PR #5, opened 2018-11-02
Merge Sprint #1 to Us Tax Court Staging Branch

codyseibert:


web-client/src/components/UsaBanner.jsx
```diff
@@ -0,0 +1,81 @@
...
+              <img src={usFlag} alt="U.S. flag" />
+              <p>An official website of the United States government&nbsp;</p>
+              <button
+                className="usa-accordion-button usa-banner-button"
+                aria-expanded={showDetails}
+                aria-controls="gov-banner"
+                onClick={() => toggleUsaBannerDetails()}
+              >
+                <span className="usa-banner-button-text">
+                  Here&apos;s how you know&nbsp;
```
waldoj: Whenever possible, I would prefer to see entities like `&apos;` nixed in favor of the actual character that's to be represented (`’`). The exceptions to this are those that HTML requires be recorded as entities (e.g., `&gt;`), plus `&nbsp;` since the Unicode character for a non-breaking space looks like any other space and can easily be overwritten accidentally. In our era of Unicode-everywhere, HTML character entities really just create technical debt.

```suggestion
                  Here’s how you know&nbsp;
```

waldoj: (This is not a thing that needs to be changed for this pull request to be accepted! It's jjust a guideline going forward.)

web-client/src/environments/dev.js
```diff
@@ -0,0 +1,12 @@
+const API_URL = 'https://7wjn89p2z4.execute-api.us-east-1.amazonaws.com/dev/v1';
```
waldoj: What is this URL?

web-client/terraform/bin/deploy-app.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+
+ENVIRONMENT=$1
+
+BUCKET=gov.ustaxcourt.ef-cms.terraform.deploys
```
waldoj: (Placeholder to indicate that the same question I asked about this variable's value in `document-service` applies here, too.)

waldoj:
A notable wrinkle is that `terraform.tfvars`' `environment` variable is used as the basis for S3 bucket names. It can't be left as `efcms`, because Flexion's S3 buckets already have that name. The Court can go with `ef-cms`, I suppose, but this feels like something that will eventually need a less awkward solution.

waldoj:
I've been unable to deploy this to AWS. After 5 minutes of reattempts to connect to the bastion host, it gives up, reporting:

```
* module.management.aws_instance.jenkins_web: timeout - last error: Error connecting to bastion: ssh: handshake failed: ssh: unable to authenticate, attempted methods [publickey none], no supported methods remain
```

The repeated attempts to connect look like this:

```
module.management.aws_instance.jenkins_web (remote-exec): Using configured bastion host...
module.management.aws_instance.jenkins_web (remote-exec):   Host: 34.226.217.160
module.management.aws_instance.jenkins_web (remote-exec):   User: ubuntu
module.management.aws_instance.jenkins_web (remote-exec):   Password: false
module.management.aws_instance.jenkins_web (remote-exec):   Private key: true
module.management.aws_instance.jenkins_web (remote-exec):   SSH Agent: true
module.management.aws_instance.jenkins_web (remote-exec):   Checking Host Key: false
```

And yet `ssh -i ssh/id_rsa ubuntu@34.226.217.160` allows me to connect to the bastion host just fine.

Although it's the bastion that times out first, those repeated log entries _also_ appear for the Jenkins instance, so I assume that's having the same problem.

web-client/terraform/bin/deploy-app.sh
codyseibert: This is actually ok (versus the hard coded flexion domain was bad above).  All the terraform related code is currently stored in a single bucket.

web-client/src/components/UsaBanner.jsx
dubharmonic: I'd agree if this were raw HTML, though I think JSX is a bit more touchy. ESLint's recommended rules suggest escaping entities: https://github.com/yannickcr/eslint-plugin-react/blob/HEAD/docs/rules/no-unescaped-entities.md

mgwalker: I believe that eslint rule only applies to straight single and double quotes, not the curled quote that @waldoj posted in his suggestion.  Since the curly quotes aren't part of a JSX string or an escape code, they're safe to use.

sadlerw:
We've updated the PR to address the comments made above. thanks!

randyhart:
@mgwalker this PR is ready for your review. If you're ok with it but don't have permissions to merge let me know and I'll merge it. Thanks.

document-service/swagger.json
```diff
@@ -0,0 +1,159 @@
+{
+  "swagger" : "2.0",
+  "info" : {
+    "description" : "Documents API for U.S. Tax Court EF-CMS",
+    "version" : "2018-10-25T19:16:39Z",
+    "title" : "will-ef-cms",
```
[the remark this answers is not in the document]
mgwalker: Likewise to this, but since it's swagger, it seems fine to let it go for now and let @waldoj decide how to handle it when he gets back.  I mean, everything should work fine, except possibly a swagger-ui instance using the `host` field from the next comment.

document-service/terraform/template/s3.tf
```diff
@@ -0,0 +1,36 @@
+provider "aws" {
+  region = "us-east-1"
+  alias = "us-east-1"
+}
+
+provider "aws" {
```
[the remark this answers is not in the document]
mgwalker: Just flagging this as an unanswered question for the future.

document-service/update-swagger.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+# This script parses the serverless.yml file to determine the name
+# of the apigateway resource that was deployed, and then uses that
+# to fetch the apigateway ID from aws.  Using that ID, this script is then
+# able to download the swagger.json from the deployed aws apigateway.
+
+# requires python
+# install jq
```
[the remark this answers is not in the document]
mgwalker: Also flagging this one

mgwalker:
Looks sensible to me.  Most of Waldo's comments/questions are addressed, and I like moving a lot of domain info out into env vars.  Flagged a couple of Waldo's questions that weren't answered, but I think this is good to merge to keep everyone rolling.  👍

waldoj:
Thanks so much for taking care of this in my absence, folks!

PR #11, opened 2018-11-19
Support for Configurable Sonarqube project keys

codyseibert:


waldoj:
I merged without review, since this is going into the pile of sprint 2 code that I'm reviewing now. (That is, it's going to be reviewed along with everything else, and there's no need to do that preemptively.)

PR #16, opened 2018-11-20
Addressing some PR comments

codyseibert:
Addressing some PR comments, including
- update readme to have `npm run install:dynamodb`
- update readme to contain top level overview of how to run project in two terminals
- clearing the error alert if the user is successfully logged in


waldoj:
Thanks!

PR #20, opened 2018-11-28
API README editing pass

waldoj:
The API README was in a random-seeming order, and needed a general editorial pass. It's definitely not right yet, but it's improved.


waldoj:
Hannah, my goal here is to _start_ to cleanup the API `README`. It's definitely not done. I still don't follow some bits. But I figure it's best to iterate.

PR #22, opened 2018-12-11
Specify directory in which to run Jenkins setup

waldoj:
It won't work in any other directory, because Terraform can't find the site data.


waldoj:
@hannahkane, the problem turned out to be that I accidentally hadn't given you write-level access. I set up a team for 18F, added us to it, and gave us both write-level access.

PR #25, opened 2019-01-03
Reduce API Deploy Zip Size

codyseibert:
- fixes the issue related to the serverless deploy package being too large... this was because we were including a lot of development dependencies in the zip.

### DIFF
--- a/Dockerfile.deploy
+++ b/Dockerfile.deploy
@@ -1,0 +1,45 @@
+########
+# Shellcheck Section
+FROM koalaman/shellcheck-alpine as shellcheck
+
+FROM node:8-alpine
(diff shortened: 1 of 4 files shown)


waldoj:
For the record, this is an addendum to [the sprint 4 pull request](https://github.com/ustaxcourt/ef-cms/pull/24/), after attempting to deploy it to staging yielded this error: `Unzipped size must be smaller than 262144000 bytes`.

It's impractical to test this prior to accepting the pull request (since the error is in the process of deploying the `staging` branch to the staging environment on AWS), so after a quick code review, I will accept the pull request to see if it works.

PR #32, opened 2019-01-11
Delete code review

waldoj:
This has been moved to `staging`, so that it can travel with the work in progress, so it should be removed from here, to avoid having to keep them in sync.


waldoj:
@hannahkane, this is just deleting the file that we already copied to `staging`.

PR #42, opened 2019-02-11
Document OWASP ZAP review process

waldoj:
Sprint 7 incorporates Docker-based OWASP ZAP testing. Flexion is running it on their end to ensure that there are no errors. This could be gated as a part of the CD pipeline, but that would only work if there were _no_ errors and, in fact, there are several errors that we consider acceptable. We may modify this process down the line (e.g., configure OWASP ZAP to ignore those errors), but for now, the process is for them to run OWASP ZAP before filing an end-of-sprint PR and direct the output to `docs/`, which we then review to confirm that there are no errors that we have not previously said are acceptable (in the code review process, on a PR).


waldoj:
It's not clear that this is the process that we'll use — I'm creating this PR to get in writing the process that we're _considering_ using.

PR #54, opened 2019-03-28
Add a "merge to master" issue template

waldoj:
This adds a new template for GitHub issues, one that’s intended for proposing to move code from `staging` into `master`. It's a checklist of everything that needs to be provided in order to approve that pull request.

This would be better as a pull-request template, but GitHub doesn't support multiple pull-request templates like they do for issue templates, so we’ll start like this. Presumably GitHub will add that feature before we use this functionality many times.

### DIFF
--- a/.github/ISSUE_TEMPLATE/merge-to-master.md
+++ b/.github/ISSUE_TEMPLATE/merge-to-master.md
@@ -1,0 +1,23 @@
+---
+name: Merge to Master
+about: A proposal to merge from staging to master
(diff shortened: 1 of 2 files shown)


waldoj:
@JessicaMarine, I've tagged you on this PR, but this is different than past pull requests — it's about changing how our GitHub repository works, in a small way. I'm happy to explain further, if that would be helpful!

waldoj:
And I've also added a file explaining how I think this process ought to work.

PR #55, opened 2019-03-29
Add Sonarcloud badges to the README

waldoj:
To support our planned process for merging to `master`, here I put a trio of build status badges — for each subproject — on the README.

(Build status badges are little graphics used to show the real-time status of software in a repository. How good is the security? Is it working at this moment? Etc.)

These are only from Sonarcloud, and I've selected the three status badges that I think best represent the risks/concerns that people may have around the project. It remains to add Jenkins badges, to show whether the software is currently building successfully, but that requires improvements to Jenkins.


waldoj:
Looking at the code really isn't helpful here! [Here you can see what it'll look like to end-users](https://github.com/ustaxcourt/ef-cms/blob/d808f936c435d55ae5069f4e57e6180adc220d81/README.md) who are visiting this repository.

PR #93, opened 2019-07-24
Work can now be found in staging and master

waldoj:
Previously, it was necessary to specify that that `staging` holds all work, and `master` would show no activity. Now that we've merged work from `staging` to `master`, and intend to do so routinely, that's no longer the case. But now the absence of any badges for `master` is weird (there's no build status, etc.), so I've replaced the old disclaimer with a new disclaimer.


waldoj:
I'd written "all work can be in the `staging` branch" because, unusually, the `master` branch contained nothing. A reasonable person would have thought there the repository was empty. Since we move work over from `staging` to `master` after each sprint, now we don't need that disclaimer anymore. While it's true that all work can be seen in `staging`, it can also be seen in `master`, where people expect it. :)

PR #95, opened 2019-08-07
Merge sprint 19 to master

waldoj:
# Release Notes

## Summary
This provides incremental user-facing improvements and some structural changes around the build process, dependencies, and architecture (see "Notes" for details on the latter).

## New Features
- additional functionality around service
- orphaned document uploads
- general improvements to the practitioner interface
- a new flow for petitions clerks for paper-filed petitions
- a print view for case records
- improvements to the "file a document" form

## Bug Fixes
There are no serious bug fixes.

## Breaking Changes
There are no breaking changes. (Until work is in production, "breaking" isn't a concept that exists.)

## Sprints Encompassed
[Sprint 19](https://github.com/ustaxcourt/ef-cms/pull/94).

## Notes
Flexion is rearchitecting to get away from the heavy reliance on third-party plugins and, problematically, forks of those plugins. (Those forks have largely been created to make security changes that the maintainers were unwilling to make.) They have eliminated several so far, and intend to continue down that path. They're also moving to something closer to a microservices architecture. These changes are in line with AWS' prescribed approach for Lambda. This should be the final release that uses 3 `package.json` files — sprint 20 is reducing those to just 1, which reduces some risk around dependencies.

# Checklist

- [x] There are standardized release notes
- [x] All dependencies are pinned
- [x] No dependencies are deprecated
- [x] Major events are logged
- [x] Output from vulnerability scan with OWASP ZAP is included
- [x] Dependency analysis (with e.g. Code Climate, Snyk, NPM Audit) reveals no non-trivial vulnerabilities that would be deployed to production, and has the output included
- [x] Code security analysis (with e.g. Code Climate, Bandit, or Sonarcloud) reveals no non-trivial vulnerabilities, and has the output included
- [x] All volatile data storage is on redundant infrastructure
- [x] Periodic snapshots of all volatile data storage are configured
- [x] There is monitoring, alerting a specific person, for both downtime and error/performance problems
- [x] There is a system/network diagram
- [x] There is a README badge for dependency analysis
- [x] There is a README badge for static code analysis
- [x] A “beta” label is prominently featured on every page


waldoj:
Sonarcloud

Here is the Sonarcloud scan. The two vulnerabilities in `EF CMS UI` are simply two uses of JavaScript `alert`s, which are acceptable and do not pose a security risk in these uses.

waldoj:
Here's the Snyk scan.

Snyk Overview

The medium-level vulnerability in `web-client/package.json` is about reverse tabnabbing in Quill:

Snyk Details

This vulnerability was identified after we accepted this work into `staging` but before starting this PR to `master`. We are using the "Snow" theme, so we suffer from this vulnerability.

However, it is my recommendation that this not be an obstacle to moving this code into `master`, because the security exploit isn't present in this PR, but instead in code that is already in `master`. Accepting this PR neither makes the problem better or worse. (This, of course, highlights the need for a process to fix security problems in `master`, but that's out of the scope of this process.)

PR #97, opened 2019-08-19
Update template to reflect its use

waldoj:
Now that we've used this template a couple of times, it's clear that some elements are going to be present in every PR. This incorporates those elements.


waldoj:
@JessicaMarine, when we open a pull request to merge work from `staging` to `master`, this is what appears by default, instead of requiring that everything be written by scratch each time. It's basically a form to be completed, and I've added more form fields.

waldoj:
Just a quick reminder about this, @JessicaMarine — this is the template that comprises the body of each pull request from staging to master. I've been manually using this template for the past two pull requests, and it's saved a bunch of time.

PR #112, opened 2019-11-13
Add missing IAM policy

waldoj:
Without giving Cognito the `UpdateUserPool` permission, deployment fails.


waldoj:
@JessicaMarine, this is a patch for a permission missing from the IAM policy. This is what was blocking the deployment of sprint 26. I'm fixing it here for posterity.

PR #127, opened 2020-01-21
Add Snyk badge to the merge-to-master template

waldoj:
Including a Snyk badge means that it can be determined at a glance whether there are any dependencies that have known security holes.


waldoj:
@JessicaMarine, this is a tiny change — it's a status badge from Snyk at the top of each PR to master. I've been including it for the past few sprints, but here I formalize that by adding it to the template that we use.

PR #133, opened 2020-02-03
Document how work is deployed to test environment

waldoj:
This replaces the merge-to-master document with a doc about overall branch management, which now includes documentation about how work is moved into the test environment.


waldoj:
@JessicaMarine, would you look at this proposed change and consider merging it? It's an update to our documentation about how work moves through environments (Flexion -> staging -> test -> production).

PR #135, opened 2020-02-04
Merge sprint 32 to Test

waldoj:
This will deploy the work done in sprint 32 to Test. This deployed to Staging seamlessly, and there is no reason to suspect that it will not deploy to Test just as simply.

Note that the deploy to Test will begin as soon as this pull request is merged, so it will be necessary to ensure that this update will not disrupt any testing taking place. (Whether the deploy goes well or goes badly, there _should_ be no disruption, but pending further testing to confirm that, it would be best to anticipate disruptions.)


waldoj:
@JessicaMarine, I'm going to leave it to you to merge this PR, to establish the best practice that the Court should decide when it is a good time to update Test (based on what actual use the environment is getting at any time). I know you're not actually using the environment right now, but good habits take practice.

PR #176, opened 2020-03-19
Describe need for larger Circle instances

waldoj:
This could save folks a lot of frustration.


waldoj:
This is a minor documentation update, @JessicaMarine.

PR #180, opened 2020-03-20
Update troubleshooting document

waldoj:
- Add a new section about dealing with permissions problems
- Update the formatting a little


waldoj:
Sorry about all these little PRs, @JessicaMarine. :) I'm making a series of changes to the documentation, wanting to document everything that's in my head, and the "right" way to do this is to have one class of change per pull request.

PR #184, opened 2020-03-27
Clean up README, link to "How We Work"

waldoj:
Add a link to the new "How We Work" document, and also do a little textual cleanup.


waldoj:
My apologies, @JessicaMarine, I should have included this as a part of the PR when "How We Work" was added.

PR #195, opened 2020-04-14
Add parameter-checking, force JSON

waldoj:
This confirms that the provided parameters exist and, in the case of the password, are conformant. This also ensures that JSON is returned, instead of relying on the user's account-level configuration.


waldoj:
Sorry to foist this review of a technical PR on you, @JessicaMarine. This just adds some niceties to the script that creates new users.

PR #202, opened 2020-04-27
Add variable verification to environment destruction

waldoj:
Make sure that the necessary variables are set before proceeding.


waldoj:
@JessicaMarine, this is just a little sanity-check before we destroy environments, to make sure that the correct information has been provided to the script before it starts running.

PR #211, opened 2020-05-26
Merging sprint 39 into test.

adunkman:
Creating to identify the changes between `staging` and `test`. This should ideally be sprint 39’s changeset only, but appears there are additional changes. 

This is a work in progress; will likely create a pull request to staging to resolve the merge issues.


adunkman:
Summary: There are commits on `test` which remove the deployment steps that: 

- Create Cognito users, court users, and IRS user
- Run smoke tests

These commits are not present in `staging`, and cause merge conflicts between the two branches.  It’s also not desired to have these users in the Court’s environments.

adunkman:
This is ultimately a problem with mixing environment configuration into code, rather than keeping it either as (a) environmental variables or (b) in other infrastructure configuration. 

I think the short-term fix is to duct-tape it — add a switch to CircleCI config which runs these scripts only in certain environments. That way, at least, our code is consistent and we can see the differences in environments in the repository. 

The long-term fix is to solve this wider issue of environment-specific configuration.

waldoj:
We had to remove the user creation steps (and the smoke tests that depend on those users' existence) to deploy to the `migration` environment. (It would be a significant security problem to allow anybody to log into an environment creating live data.) I do not remember making these changes in `test`, and I no longer have access to Trello, Slack, etc., where I might be able to see why that was done.

adunkman:
Going to re-mark this PR as a draft; it shouldn’t be merged until we can address the pre-sprint 39 differences between `test` and `staging`. I’ve addressed them in https://github.com/ustaxcourt/ef-cms/pull/212 — after that PR is merged, this should be a quick and easy merge.

kfoley-18F:
@JessicaMarine - Merge the work done in sprint 39 to the Test environment. Note that this should not be merged until such time as you're willing to potentially interrupt any user testing that's underway.

PR #217, opened 2020-06-01
Fix user permissions across branches, including IRS.

adunkman:
Following up to #214, this PR once again tweaks the user deployment configuration, because: 

1. There’s a missing space which doesn’t cause an issue in zsh (locally) but does in bash (in CI) (fixed in 2625e136b519611d3b244f1670c23a659a07addf).
2. We have [further guidance](https://docs.google.com/spreadsheets/d/1ZE9Ip2M8jaRGiChwOarFy44tqx0V5B2LxDqJi1lXYeg/edit#gid=0) from @JessicaMarine on users: 

    | | Staging | Test | Production | IRS Test | Migration |
    |-|-------|------|------------|-------|------|
    | Generic users | ✓ |  |  | ✓ |  |
    | Court users (CSV) | ✓ | ✓ |  |  |  |
    | IRS Superuser (temp) |  |  |  | ✓ |  |
    | Mike (admin) |  |  |  |  | ✓ |
    | Final court users (TBD) |  |  | ✓ |  |  |
    | Final IRS Superuser |  |  | ✓ |  |  |

Once again I’d like to highlight that this is a short-term patch until we have a longer-term user management strategy in place. I hope to tackle that shortly, as it’s likely to amplify pain over time.

Two notes:

- Because the smoke tests require **Generic users**, they can only be run in `staging` and `IRS Test` environments.
- I additionally picked up some merge commits in this PR so we can fast-forward merge in the future.


adunkman:
@JessicaMarine I recommend the court accept these changes into `staging`, they reflect the user account spreadsheet referenced above (as created by you). 

(I also slacked you about this + a few upcoming PRs).

PR #258, opened 2020-06-12
5158 DOD - Remove authorizer from public route

rachelschneiderman:


rachelschneiderman:
Oops, meant to submit this as a PR to develop.

PR #289, opened 2020-07-10
Remove no-minify flag from npx parcel command

julialeague:
Our deployment is currently failing due to reaching a size limit on a lambda function: https://app.circleci.com/pipelines/github/ustaxcourt/ef-cms/299/workflows/342dc4ea-5c58-4805-a164-5af9ec99cc1d/jobs/3139

Flexion ran into this issue too and resolved to remove the `--no-minify` flag to move past the issue - there may be work in the future to switch to deploying via S3 instead of direct upload. @ericsorenson


julialeague:
@JessicaMarine , I recommend you accept this PR to the staging environment. It should allow us to move past our failing deployment. Thanks!

PR #303, opened 2020-07-21
Update docker-to-ecr.sh.

adunkman:
A few issues when running to create the new production account build:

- Two commands are missing their region definition and as such failed.
- The bash script continues executing on failed commands. Using `set -e` terminates on failed commands instead.
- aws ecr get-login is deprecated and removed from AWS CLI 2, [in favor of](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/ecr/get-login-password.html) `get-login-password` and piping to docker login.

Notes:

- The `AWS` username is [just how you do it](https://docs.aws.amazon.com/AmazonECR/latest/userguide/Registries.html), not configured or environment-dependent.

### DIFF
--- a/docker-to-ecr.sh
+++ b/docker-to-ecr.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
+
 # get aws account id if it does not exist in env var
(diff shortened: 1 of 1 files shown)


adunkman:
Tagging @sutt0n for :eyes: as well due to the code area.

julialeague (approved):
Looks good!

PR #323, opened 2020-07-29
Add iam:CreateServiceLinkedRole to CircleCI user.

adunkman:
This is used the first time a custom domain is provisioned for API Gateway to grant permission for CircleCI to create a Cloudfront distribution in an AWS-owned account as described in [AWS documentation](https://docs.aws.amazon.com/apigateway/latest/developerguide/how-to-custom-domains.html#edge-optimized-custom-domain-names). 

The non-production AWS account hasn’t experienced this issue because the CircleCI user wasn’t a limited permission account the first time a custom domain was provisioned. 

I believe this is a reasonable security risk (giving the automated user additional permissions) because it is only usable by AWS services themselves.

- [The permission](https://docs.aws.amazon.com/IAM/latest/APIReference/API_CreateServiceLinkedRole.html) has very specific use cases (emphasis added):

  > Creates an IAM role that is linked to a specific AWS service. … To attach a policy to this service-linked role, **you must make the request using the AWS service** that depends on this role.

- A bit more information about [service-linked roles](https://docs.aws.amazon.com/IAM/latest/UserGuide/using-service-linked-roles.html) (emphasis added):

  > A service-linked role is a unique type of IAM role that is linked directly to an AWS service. Service-linked roles are predefined by the service and include all the permissions that the service requires to call other AWS services on your behalf. … The linked service defines the permissions of its service-linked roles, and unless defined otherwise, **only that service can assume the roles**. 

Tagging @mmarcotte for review, and also /cc @ericsorenson, @sutt0n, and @codyseibert since this will be needed for a deploy to succeed in a new AWS account. There are also likely to be additional deploy fixes, coming from [our most recent failed prod build](https://app.circleci.com/pipelines/github/ustaxcourt/ef-cms/330/workflows/daa80d03-d118-49e8-9c5c-f9952ed755e7/jobs/3490).

### DIFF
--- a/iam/terraform/account-specific/main/circle-ci.tf
+++ b/iam/terraform/account-specific/main/circle-ci.tf
@@ -187,6 +187,7 @@ resource "aws_iam_policy" "circle_ci_policy" {
         "iam:ListPolicyVersions",
         "iam:ListInstanceProfilesForRole",
         "iam:AddRoleToInstanceProfile",
+        "iam:CreateServiceLinkedRole",
         "iam:ListAttachedRolePolicies"
       ],
       "Resource": [


codyseibert (approved):

mmarcotte (approved):
thanks for shoring this up!

PR #328, opened 2020-07-31
Clear Migration Environment Scripts

ericsorenson:
Please change the password for the migrator@example.com account before loading data.


web-api/setup-migration-user.sh
```diff
@@ -0,0 +1,149 @@
...
+
+  session=$(echo "${response}" | jq -r ".Session")
+
+  if [ "$session" != "null" ]; then
+    aws cognito-idp admin-respond-to-auth-challenge \
+      --user-pool-id  "${USER_POOL_ID}" \
+      --client-id "${CLIENT_ID}" \
+      --region "${REGION}" \
+      --challenge-name NEW_PASSWORD_REQUIRED \
+      --challenge-responses 'NEW_PASSWORD="Testing1234$",'USERNAME="${email}" \
```
mmarcotte: Just trying to understand how this works. I see that the `ustcadmin` account creates an `adminToken`, which in turn is used to create the migrator@example.com account. I just don't quite get how this account could not be used to gain access to the system.

clear-env-mig.sh
```diff
@@ -0,0 +1,36 @@
...
+if [[ "$?" == "1" ]]; then
+  echo "Terraform was not found on your path. Please install terraform."
+  exit 1
+fi
+
+./web-api/clear-elasticsearch-index.sh $ENV
+./web-api/setup-elasticsearch-index.sh $ENV
+
+pushd web-api
+node clear-dynamodb-table.js $ENV
```
mmarcotte, line 34: @ericsorenson -- another question, as this clears out the DynamoDB table, would this also create orphaned Cognito users that would have been associated with the user records in DynamoDB? If that's correct, is there any way to handle them?

mmarcotte (changes requested):
I'm gunshy about the `setup-migration-user.sh` script after the recent security event, and @mark-meyer already has a python script at the ready to create users. So, can you please remove that script and simply include the `clear-env-mig.sh` script?

mmarcotte (approved):
Thanks Eric!!!

PR #330, opened 2020-08-05
Enable logging for Elasticsearch in Terraform Config

mmarcotte:
In order to handle issues should they arise in Elasticsearch (eg, migration building the indexes that power the UI), this branch should turn on logging based on the [Terraform Docs for the aws_elasticsearch_domain resource](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/elasticsearch_domain#log_publishing_options). 

Need someone to look over the ARN because it may be attempting to be too prescriptive with the log group:
```
/aws/aes/debug_${var.environment}
```

### DIFF
--- a/web-api/terraform/template/elasticsearch.tf
+++ b/web-api/terraform/template/elasticsearch.tf
@@ -15,4 +15,10 @@ resource "aws_elasticsearch_domain" "efcms-search" {
   snapshot_options {
     automated_snapshot_start_hour = 23
   }
+
+  log_publishing_options = {
+    enabled                  = "true"
+    cloudwatch_log_group_arn = "arn:aws:logs:${var.aws_region}:${data.aws_caller_identity.current.account_id}:log-group:/aws/aes/debug_${var.environment}"
+    log_type                 = "ES_APPLICATION_LOGS"
+  }
 }
--- a/web-api/terraform/template/elasticsearch.tf
+++ b/web-api/terraform/template/elasticsearch.tf
@@ -1,3 +1,31 @@
+resource "aws_cloudwatch_log_group" "elasticsearch_application_logs" {
+  name = "/aws/aes/debug_${var.environment}"
+}
+
+resource "aws_cloudwatch_log_resource_policy" "allow_elasticsearch_to_write_logs" {
+  policy_name = "allow_elasticsearch_to_write_logs"
+
+  policy_document = <<CONFIG
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Principal": {
+        "Service": "es.amazonaws.com"
+      },
+      "Action": [
+        "logs:PutLogEvents",
+        "logs:PutLogEventsBatch",
+        "logs:CreateLogStream"
+      ],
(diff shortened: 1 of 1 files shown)


web-api/terraform/template/elasticsearch.tf
```diff
@@ -15,4 +15,10 @@ resource "aws_elasticsearch_domain" "efcms-search" {
   snapshot_options {
     automated_snapshot_start_hour = 23
   }
+
+  log_publishing_options = {
+    enabled                  = "true"
+    cloudwatch_log_group_arn = "arn:aws:logs:${var.aws_region}:${data.aws_caller_identity.current.account_id}:log-group:/aws/aes/debug_${var.environment}"
```
adunkman: Re: your comment about this ARN and being too prescriptive — an easy way to make sure we’re using the right ARN would be to create the log group and reference it: 

```hcl
resource "aws_cloudwatch_log_group" "elasticsearch_application_logs" {
  name = "/aws/aes/debug_${var.environment}"
}

resource "aws_elasticsearch_domain" "efcms-search" { 
  …
  log_publishing_options {
    …
    cloudwatch_log_group_arn = aws_cloudwatch_log_group.elasticsearch_application_logs.arn
  }
}
```

web-api/terraform/template/elasticsearch.tf
```diff
@@ -15,4 +15,10 @@ resource "aws_elasticsearch_domain" "efcms-search" {
   snapshot_options {
     automated_snapshot_start_hour = 23
   }
+
+  log_publishing_options = {
```
adunkman: No need to use `=` here — I’m actually surprised this compiled, but blocks like this can be referenced by name without assigning them with equals (same as `snapshot_options`, `ebs_options`, and `cluster_config` above).

adunkman (commented):
A few comments below — I think we’re also missing a policy which grants permission for Elasticsearch to write logs, based on the [Log Publishing to CloudWatch Logs](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/elasticsearch_domain#log-publishing-to-cloudwatch-logs) example in the Terraform documentation: 

```hcl
resource "aws_cloudwatch_log_resource_policy" "allow_elasticsearch_to_write_logs" {
  policy_name = "allow_elasticsearch_to_write_logs"

  policy_document = <<CONFIG
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Principal": {
        "Service": "es.amazonaws.com"
      },
      "Action": [
        "logs:PutLogEvents",
        "logs:PutLogEventsBatch",
        "logs:CreateLogStream"
      ],
      "Resource": "arn:aws:logs:*"
    }
  ]
}
CONFIG
}
```

mmarcotte:
@adunkman -- thank you. Can't believe I missed that example usage, but everything makes so much more sense now. Policy added, log group resource added, statement should be fixed and optional enabled boolean dropped as it served no real purpose.

adunkman (approved):
Looks great! 

It’s so beautiful!

julialeague (approved):
LGTM!

PR #343, opened 2020-08-11
Clean up tripping points when doing local PR review.

adunkman:
This was my first time doing local pull request review, and a few things tripped me up! 

@ericsorenson, can you double-check me on these two points: 

- Mock logins require their full email address, suffixed by `@example.com`, rather than what was documented in the README. 
- The public API was not exposed from docker (port `5000`), which prevented the public UI from loading.

@julialeague, @mmarcotte, and @kfoley-18F can you double-check me on: 

- The code review checklist was outdated, specifically: 
  - We don’t run `docker system prune` or delete the repo and fresh-clone it.
  - We don’t test in IE 11.
  - We use the device emulator in Chrome rather than a real mobile device or other hardware emulator.


docs/CODE_REVIEW.md
```diff
@@ -87,17 +87,16 @@ We wouldn’t want anyone being mean to us because of an oversight, mistake, or
 
 We use this list when performing a code review to ensure that all tasks have been completed.
 
-- [ ] begin by deleting any existing copy of the Git repository, running `docker system prune` to remove any cached artifacts from any prior code review
-- [ ] clone a fresh copy of the Git repository (`git clone https://github.com/ustaxcourt/ef-cms.git`), fetch the pull request and name it for the sprint (e.g., `git fetch origin pull/204/head:sprint-43`), and then switch to that branch (e.g. `git checkout sprint-43`)
+- [ ] Fetch the pull request for the sprint (e.g., `git fetch origin pull/204/head:sprint-43`), and then switch to that branch (e.g. `git checkout sprint-43`)
```
julialeague: I actually do run `docker system prune`, mostly because if I don't do that, I might have a lot of storage space taken up by older, unused image builds.

julialeague: While we're here, should we change the explicit pull # references with something like `pull/{PR #}/head:sprint-{sprint #}` ?

docker-run.sh
```diff
@@ -6,6 +6,7 @@ docker run -it \
   -p 1234:1234 \
   -p 5678:5678 \
   -p 4000:4000 \
+  -p 5000:5000 \
```
julialeague, line 9: I thought I was able to access the public UI via http://localhost:5678, but maybe I'm mistaken.

adunkman: It appeared as a loading indicator indefinitely, since the backend service couldn’t be reached. Confirmed with Eric that this port should be accessible for the service — if you find that you’re able to run this locally without it, there might be something fishy up!

docs/CODE_REVIEW.md
adunkman: Added a discretionary step in a2ed1d4: 

> If desired, run `docker system prune` to remove any unused docker images from previous code review.

adunkman: Sure thing! Updated in a2ed1d4.

adunkman:
I believe this is ready for re-review — confirmed with @ericsorenson yesterday that this PR looks good!

docker-run.sh
julialeague: ah ok, I do get the load page. good to know!

julialeague (approved):

mmarcotte (approved):
Looks great, thank you!

kfoley-18F:
@adunkman @mmarcotte 

Re: 
- The code review checklist was outdated, specifically:
   - We don’t run docker system prune or delete the repo and fresh-clone it.
   - We don’t test in IE 11.
   - We use the device emulator in Chrome rather than a real mobile device or other hardware emulator.

For a while, Docker needed literally all of my laptop resources to run Dawson locally for testing, so IE11 testing was just hard-to-impossible. With the lambda refactoring, Docker usage was able to decrease for local operations recently. 
With Docker usage shrunk down again, it's possible to run IE 11 in VirtualBox for testing like: https://github.com/18F/handbook/issues/1709 and https://gsa-tts.slack.com/archives/C039MHHF8/p1594140502302800?thread_ts=1594137533.299400&cid=C039MHHF8

However.... it's a lot of setup for not much gain to operate on an every-week basis. 
I think it's worth considering adding a ticket for IE11 testing and spending some time on it prior to soft launch. 
And/or... add code to try to autodetect browser in the HTTP request tags, and add a banner that recommends Chrome or Firefox instead. There are enough free browser options and IE is historically really costly to support because it's often not as compliant in how it displays things and requires more workarounds. 

Also, device emulator in Chrome was how I have been doing mobile testing for PRs. It's probably worth doing some serious mobile testing against an actual physical env (test? staging?) sometime prior to soft launch too. 

I have needed to docker prune once, and have needed to delete repo and fresh clone twice to be able to review PRs in the past, usually when changes have been pushed and I'd already made a branch for the PR and there were env-specific file versioning issues. I'd love to update the github branch labeling instructions so that the PR branch is a published branch at origin and not just local repo, and so it follows the gitflow strategy better. 

My $0.02....

kfoley-18F (approved):
👍

PR #345, opened 2020-08-11
Add `dynamodb:UpdateGlobalTable` for CircleCI

mmarcotte:
In order to clear the migration environment's DynamoDB table more efficiently, we opted to use `Delete Table` via the Console, which took 1 minute rather than 175 minutes deleting Documents 25 x 25. 

However, on the next deploy in order to use Terraform to recreate the deleted table we ran into the following error:

```
ProcessGlobalTables: Error: AccessDeniedException: User: arn:aws:iam::************:user/CircleCI is not authorized to perform: dynamodb:UpdateGlobalTable on resource: arn:aws:dynamodb:us-west-1:************:table/efcms-mig
```

So we are having to add the `dynamodb:UpdateGlobalTable` permission to the policy for the CircleCI user in order to restore the environment.

### DIFF
--- a/iam/terraform/account-specific/main/circle-ci.tf
+++ b/iam/terraform/account-specific/main/circle-ci.tf
@@ -166,7 +166,8 @@ resource "aws_iam_policy" "circle_ci_policy" {
         "dynamodb:DescribeStream",
         "dynamodb:GetRecords",
         "dynamodb:GetShardIterator",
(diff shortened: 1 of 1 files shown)


adunkman (approved):
👍 ! This user already has permission to `CreateGlobalTable`, it makes perfect sense that it should be able to update what it has created. :shipit:

PR #301, opened 2020-07-20
Simplify account-specific deploy steps with Terraform.

adunkman:
This change brings 7 additional resources under Terraform configuration which were previously manually created on the AWS Account-specific level, removing their manual documentation from the [Environment setup documentation](https://github.com/ustaxcourt/ef-cms/blob/f189a24cf04e8d1b1da87d29e9431a0cfa4f0f28/docs/environments/setup.md).

- I wouldn’t have been able to get past "Setup a role & policy for accessing the Dynamsoft ZIP file" without copying over from the existing non-production AWS account — this pull request adds that configuration to Terraform so we configure it the same next time around.

⚠️ When this change is merged, existing resources will need to be manually imported into Terraform’s state file so it doesn’t attempt to create duplicate resources when the account-specific manual deployment step is run:

```bash
export TF_VAR_dns_domain=${EFCMS_DOMAIN}

terraform import aws_route53_zone.primary [HOSTED_ZONE_ID]
terraform import aws_iam_user.circle_ci [USER_NAME]
terraform import aws_iam_user_policy_attachment.circle_ci_policy_attachment [USER_NAME]/[CI_POLICY_ARN]
terraform import aws_s3_bucket.dynamsoft [DYNAMSOFT_BUCKET_NAME]
terraform import aws_iam_role.dynamsoft_s3_download_role dynamsoft_s3_download_role
terraform import aws_iam_policy.access_dynamsoft_s3_bucket [DYNAMSOFT_POLICY_ARN]
terraform import aws_iam_role_policy_attachment.allow_dynamsoft_role_access_to_dynamsoft_s3_bucket dynamsoft_s3_download_role/[DYNAMSOFT_POLICY_ARN]
```

It’s easiest to retrieve these values using the AWS console, as they were created manually and as such may vary slightly:

- `HOSTED_ZONE_ID`: Retrieve from the Hosted Zone in Route 53.
- `USER_NAME`: Retrieve from the Circle CI user in IAM.
- `CI_POLICY_ARN`: Retrieve from the ARN from the circle_ci_policy in IAM.
- `DYNAMSOFT_BUCKET_NAME`: Retrieve the bucket name from S3 which contains the Dynamsoft archive.
- `DYNAMSOFT_POLICY_ARN`: The policy created which grants access to the Dynamsoft S3 bucket.

<details>
<summary>On first run, it’s recommended to avoid <code>-auto-approve</code> and confirm Terraform’s execution plan will perform the desired steps before approving.</summary><br>

[...]

PR #360, opened 2020-08-17
Move even more account information into Terraform

mmarcotte:
This change brings one additional resource under Terraform control that was previously manually created on the AWS Account-specific level

⚠️ When this change is merged, existing resources will need to be manually imported into Terraform's state file so it doesn't attempt to create duplicate resources when the account-specific manaul deployment step is run.

To import this resource:

```bash
cd iam/terraform/account-specific/main
export ZONE_NAME=${ZONE_NAME}
export TF_VAR_dns_domain=${EFCMS_DOMAIN}
../bin/deploy-init.sh
terraform import aws_iam_instance_profile.dynamsoft_instance_profile dynamsoft_s3_download_role
```

On first run, it's recommended to not use `-auto-approve` to confirm that Terraform is tracking the resources correctly and will perform the desired steps."


adunkman (approved):
Looks great! Sorry for missing this in #301!

PR #373, opened 2020-08-25
Add zone_name to environment destruction.

adunkman:
These need to be specified for the command to run.


mmarcotte (approved):
Thanks @adunkman 

demo

PR #376, opened 2020-08-25
Update documentation for IRS_SUPERUSER_EMAIL

adunkman:
My bad, this should have been included in #363!


adunkman:
I chose to use `STG`, `TEST`, and `PROD` to be the same as the other documentation items, but for the IRS environment we’ll need to configure `IRS_SUPERUSER_EMAIL_IRS`

mmarcotte (approved):
Thanks @adunkman 

All Good

PR #383, opened 2020-08-28
remove color check from irs user script

mmarcotte:
In order to reset the IRS Super User account information, we needed to adjust the script that creates that account. Simply needed to remove some old code that was checking for the release color.

### DIFF
--- a/web-api/setup-irs-user.sh
+++ b/web-api/setup-irs-user.sh
@@ -20,8 +20,6 @@
 ENV=$1
 REGION="us-east-1"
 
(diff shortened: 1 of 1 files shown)


adunkman (approved):
Looks great — I’m so happy this was such a straightforward change. Closes [setup-irs-user.sh script isn't working](https://trello.com/c/dI7LNcbX/751-setup-irs-usersh-script-isnt-working), if I’m not mistaken (or, should that card remain open until a regression test is added?).

Either way,  👍 🚀 !

PR #385, opened 2020-09-01
Add new system architecture diagram.

adunkman:
Updated to what is currently deployed, and is now a online-editable format (with source and directions). Definitely not as pretty as the other diagram.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,27 @@ digraph {
 ```
 </details>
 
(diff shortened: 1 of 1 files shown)


julialeague (approved):
GraphViz is cool! Thanks for doing this!

mmarcotte (approved):
Okay, that's pretty cool. 💯

PR #387, opened 2020-09-02
Add configuration for developer ES admin access, new ES log ingestion instance

julialeague:
This change adds a new Elasticsearch instance for our app logs that mimics the configuration of the other instance. We can play around with that configuration if needed!

We also add an IAM user group called "developers", and an ES admin access policy that is attached to the developers group.

Question to @mmarcotte - do you think we should add [`aws_iam_group_membership` config](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_group_membership) for the developers group?

Ran a successful `terraform plan` locally with the `iam_groups.tf` config, but I'm having trouble doing the same for the ES instance. See comment below for details. 

**Update**: `terraform plan` has now been run successfully for both config additions!


julialeague:
I'm having trouble testing the ES logs addition locally - getting errors when running `web-api/terraform/bin/deploy-app.sh` with the `apply` step commented out:

```
Error: Error in function call
  on ../template/layers.tf line 16, in resource "aws_s3_bucket_object" "clamav_layer_object":
  16:   etag   = "${filemd5("../../runtimes/clamav/clamav_lambda_layer.zip")}"
Call to function "filemd5" failed: no file exists at
../../runtimes/clamav/clamav_lambda_layer.zip.
Error: Error in function call
  on ../template/layers.tf line 31, in resource "aws_s3_bucket_object" "puppeteer_layer_object":
  31:   etag   = "${filemd5("../../runtimes/puppeteer/puppeteer_lambda_layer.zip")}"
Call to function "filemd5" failed: no file exists at
../../runtimes/puppeteer/puppeteer_lambda_layer.zip.
```

julialeague:
Update on testing the ES instance: I managed to run `terraform plan` for the ES instance with some local and unrelated tweaks, and it shows a successful creation of the instance!

iam/terraform/account-specific/main/iam_groups.tf
```diff
@@ -0,0 +1,26 @@
+resource "aws_iam_group" "developers" {
+  name = "developers"
```
adunkman: Maybe this could be something like `log_viewers`? Just thinking of something that would say more about what the group does.

web-api/terraform/template/elasticsearch.tf
```diff
@@ -49,3 +49,27 @@ resource "aws_elasticsearch_domain" "efcms-search" {
...
   }
 }
+
+resource "aws_elasticsearch_domain" "efcms-app-logs" {
+  domain_name           = "efcms-app-logs-${var.environment}"
+  elasticsearch_version = "7.4"
+
+  cluster_config {
+    instance_type = "t2.small.elasticsearch"
+    instance_count = var.es_instance_count == "" ? "1" : var.es_instance_count
```
adunkman: I’m not sure Terraform allows variables to have blank values — does it? 🤔

web-api/terraform/template/elasticsearch.tf
```diff
@@ -49,3 +49,27 @@ resource "aws_elasticsearch_domain" "efcms-search" {
     log_type                 = "ES_APPLICATION_LOGS"
   }
 }
+
+resource "aws_elasticsearch_domain" "efcms-app-logs" {
+  domain_name           = "efcms-app-logs-${var.environment}"
```
adunkman: Maybe `efcms-logs-${var.environment}` instead? This name seems to indicate that the index only contains logs from `app.${var.dns_domain}` — but I think we’re planning on dropping all the logs in here (correct me if I’m wrong!).

web-api/terraform/template/elasticsearch.tf
```diff
@@ -49,3 +49,27 @@ resource "aws_elasticsearch_domain" "efcms-search" {
...
+    ebs_enabled = true
+    volume_size = 10
+  }
+
+  snapshot_options {
+    automated_snapshot_start_hour = 23
+  }
+
+  log_publishing_options {
+    cloudwatch_log_group_arn = aws_cloudwatch_log_group.elasticsearch_application_logs.arn
```
adunkman: We’ll probably want a separate log group for these logs — I’m worried we might get into an infinite loop if we intermingle the application’s elasticsearch logs with the logging logs (new log stream item -> elasticsearch -> log stream -> new log stream item -> elasticsearch…).

… does that make sense? I used the word "log" so many times :joy:

web-api/terraform/template/elasticsearch.tf
```diff
@@ -49,3 +49,27 @@ resource "aws_elasticsearch_domain" "efcms-search" {
...
+    log_type                 = "ES_APPLICATION_LOGS"
+  }
+}
```
adunkman: I think we’ll also want to require HTTPS:

```hcl
resource "aws_elasticsearch_domain" "efcms-app-logs" {

  …

  domain_endpoint_options {
    enforce_https = true
    tls_security_policy = "Policy-Min-TLS-1-2-2019-07"
  }
}
```

adunkman (commented):
Dropped in some comments — if I recall, Kibana doesn’t work with IAM-based authentication, but I might be wrong (I haven’t set this up before). Maybe @mark-meyer knows more?

web-api/terraform/template/elasticsearch.tf
julialeague: I copied this from the [ES search instance config](https://github.com/ustaxcourt/ef-cms/blob/b2016aa10a6e0789d1cb4e9f55819e792c5e4819/web-api/terraform/template/elasticsearch.tf#L29-L36), so I guess it does - it looks like we don't set [this variable](https://github.com/ustaxcourt/ef-cms/blob/a40d775c0f6e090e2d5b8c8c5d0e76dd03614ff3/web-api/terraform/bin/deploy-app.sh#L39) right now in CircleCI. We probably don't want to share this instance_count var between all ES config, though - seems like there could be different needs between them. I'll add a separate var for the app logs ES config.

julialeague: Found this Terraform doc about [conditional expressions](https://www.terraform.io/docs/configuration/expressions.html#conditional-expressions)

web-api/terraform/template/elasticsearch.tf
julialeague: I wondered about this too - it seems potentially confusing if nothing else! I'll separate them out.

web-api/terraform/template/elasticsearch.tf
julialeague: Nice! Terraform rocks

web-api/terraform/template/elasticsearch.tf
julialeague: good point, I agree that the use of "app" sounds restrictive. Will change!

sutt0n: what about `logs.${var.zone_name}`?

web-api/terraform/template/elasticsearch.tf
sutt0n: you could probably just say `instance_count = var.es_instance_count` here and in the `variables.tf`, declare it as:

```tf
variable "es_instance_count" {
  type = string
  default = null
}
```

iam/terraform/account-specific/main/cognito.tf
```diff
@@ -0,0 +1,39 @@
+resource "aws_cognito_user_pool" "log_viewers" {
+  name = "log_viewers"
+  password_policy {
+    minimum_length = 8
+  }
```
adunkman, line 10: We should probably specify at least as strict a policy as [the application’s policy](https://github.com/ustaxcourt/ef-cms/blob/7cab0b95e9f90ec2666e79b65c9a17fc9cb57962/web-api/terraform/template/cognito.tf#L65-L71), if not more strict.

iam/terraform/account-specific/main/elasticsearch.tf
```diff
@@ -0,0 +1,143 @@
...
+  name = "/aws/aes/kibana"
+}
+
+resource "aws_elasticsearch_domain" "efcms-logs" {
+  domain_name           = "info"
+  elasticsearch_version = "7.4"
+
+  cluster_config {
+    instance_type = "t2.small.elasticsearch"
+    instance_count = var.es_logs_instance_count == "" ? "1" : var.es_logs_instance_count
```
adunkman: This has a default value of 1 in the variable block, so it will never be blank.

web-api/terraform/template/streams.tf
```diff
@@ -25,4 +25,10 @@ resource "aws_lambda_event_source_mapping" "streams_mapping" {
   event_source_arn  = aws_dynamodb_table.efcms-east.stream_arn
   function_name     = aws_lambda_function.zip_streams.arn
   starting_position = "TRIM_HORIZON"
+
+  # The below config was added due to a potential Terraform bug
+  # see https://github.com/terraform-providers/terraform-provider-aws/issues/14522
+  maximum_batching_window_in_seconds = 0
+  maximum_record_age_in_seconds      = 604800
+  maximum_retry_attempts             = 10000
```
adunkman: Hmm… is this an accidental add? Or, what was this added to fix? The [commit that added it](https://github.com/ustaxcourt/ef-cms/pull/387/commits/bb22c05a8247c5f97af693b7028e419744356f68) seems to also introduce another change which has been reverted.

iam/terraform/account-specific/main/lambda-es-role.tf
```diff
@@ -0,0 +1,45 @@
...
+    },
+    {
+      "Effect": "Allow",
+      "Action": "es:ESHttpPost",
+      "Resource": "arn:aws:es:*:*:*"
+    }
+  ]
+}
+EOF
+}
```
adunkman: These two resources seem unused, unless I’m missing where they’re used. It seems to be allowing lambdas to write cloudwatch logs and also post things to any elasticsearch instance?

adunkman (commented):
Looking pretty good! A few questions — I’ll follow-up with a commit to address the straightforward changes given that Julia’s OOO, but I’m not sure we can move forward without understanding the `maximum_batching_window_in_seconds` change.

iam/terraform/account-specific/main/lambda-es-role.tf
julialeague: You were right that we didn't need them previously, but now that we have configuration for log subscriptions, this role/policy will allow the LogsToElasticsearch_info lambda to publish logs to Cloudwatch.

web-api/terraform/template/streams.tf
julialeague: I will remove this with the caveat that it's possible we run into the issue again when creating a fresh environment. Not sure about that, though.

iam/terraform/account-specific/main/lambda-es-role.tf
julialeague: I removed the ES permission, I don't think that is needed: https://github.com/ustaxcourt/ef-cms/pull/387/commits/288e626fd9ab0db16564ee52df044a3806bf5827

julialeague:
This PR now includes the configuration for the lambda function and log subscription groups that are needed to allow this new ES instance to ingest specific log groups from Cloudwatch. I ran the account-specific `deploy-app.sh` script locally, and the resources have been created successfully for all environments in our non-production AWS account.

I added a new environment variable to the [Terraform configuration](https://github.com/ustaxcourt/ef-cms/pull/387/commits/2ccee77432d3bb501dececba6a95ab70e9b68406#diff-99c36220f7685cf83a9bddb0d9509dc3) and [`deploy-app.sh`](https://github.com/ustaxcourt/ef-cms/pull/387/commits/04d8cf618a9f578d92c761654754eeb5cadb7d44) to allow us to specify deployment environments - this enables the log group subscription configuration that will add a subscription for each environment's corresponding log group.

iam/terraform/account-specific/bin/deploy-app.sh
```diff
@@ -22,4 +22,4 @@ fi
 set -eo pipefail
 
 terraform init -backend=true -backend-config=bucket="${BUCKET}" -backend-config=key="${KEY}" -backend-config=dynamodb_table="${LOCK_TABLE}" -backend-config=region="${REGION}"
-TF_VAR_my_s3_state_bucket="${BUCKET}" TF_VAR_my_s3_state_key="${KEY}" terraform apply -auto-approve -var "dns_domain=${EFCMS_DOMAIN}"
+TF_VAR_my_s3_state_bucket="${BUCKET}" TF_VAR_my_s3_state_key="${KEY}" terraform apply -auto-approve -var "dns_domain=${EFCMS_DOMAIN}" -var "es_logs_instance_count=${ES_LOGS_INSTANCE_COUNT}" -var "environments=${ENVIRONMENTS}"
```
mmarcotte: Appears this might be a good opportunity for a little tidying

```
TF_VAR_my_s3_state_bucket="${BUCKET}" 
TF_VAR_my_s3_state_key="${KEY}" 
TF_VAR_dns_domain="${EFCMS_DOMAIN}"
TF_VAR_es_logs_instance_count="${ES_LOGS_INSTANCE_COUNT}"
TF_VAR_environments="${ENVIRONMENTS}"
terraform apply -auto-approve
```

iam/terraform/account-specific/main/lambda-logs-to-elasticsearch.tf
```diff
@@ -0,0 +1,49 @@
...
+
+resource "aws_lambda_permission" "allow_cloudwatch" {
+  statement_id  = "AllowExecutionFromCloudWatch"
+  action        = "lambda:InvokeFunction"
+  function_name = aws_lambda_function.logs_to_es.function_name
+  principal     = "logs.amazonaws.com"
+}
+
+resource "aws_cloudwatch_log_subscription_filter" "api_filter" {
+  count = length(var.environments)
```
mmarcotte, line 34: cool use of count 😀

julialeague: I love count now that I know about it!

aws/lambdas/LogsToElasticSearch_info/index.js
```diff
@@ -0,0 +1,272 @@
...
+/* in AWS Console when setting up a log subscription to our Elasticsearch Kibana domain.
+/* We need to host this lambda in order to manage it via Terraform.
+*/
+
+// v1.1.2
+var https = require('https');
+var zlib = require('zlib');
+var crypto = require('crypto');
+
+const ENV = process.env;
```
julialeague: This is passed in via the lambda_function Terraform config - it's the only thing that I changed about the lambda function that's auto-generated in AWS Console.

web-api/terraform/template/streams.tf
julialeague: removed https://github.com/ustaxcourt/ef-cms/pull/387/commits/2ccee77432d3bb501dececba6a95ab70e9b68406#diff-cc38121f0a101dec01859ae2f231227f

aws/lambdas/LogsToElasticSearch_info/index.js
```diff
@@ -0,0 +1,272 @@
...
+
+var endpoint = ENV.es_endpoint;
```
adunkman: [non-blocking comment, take it or leave it] 

Below we’re referencing `process.env` directly, but here it’s assigned to `ENV` and accessed there. Typically in nodejs, I’d expect it to be referenced with `process.env.ES_ENDPOINT`.

iam/terraform/account-specific/main/lambda-logs-to-elasticsearch.tf
```diff
@@ -0,0 +1,49 @@
+resource "aws_lambda_function" "logs_to_es" {
+  filename = "../../../../aws/lambdas/LogsToElasticSearch_info.zip"
```
adunkman: I don’t see where this zip file is generated — do we need an `archive_file` resource?

adunkman: Ooh, I see that it’s committed — I would love to use `archive_file` to generate it instead and add the `zip` to `.gitignore` so we only have one copy of the code in the repo (the other lambdas use `archive_file` already to generate zips — ripe for inspiration!).

adunkman (commented):
Looking really good! I think just generating the zip file is the only thing I would say should probably make it into this PR — my other comment shouldn’t hold up this PR.

iam/terraform/account-specific/main/lambda-logs-to-elasticsearch.tf
julialeague: Done! https://github.com/ustaxcourt/ef-cms/pull/387/commits/b6d2d97e80083e6e5196a332b1b35eaf76fd7955
https://github.com/ustaxcourt/ef-cms/pull/387/commits/6ed2048c0aabb444b7c5ad30260894293886f169
https://github.com/ustaxcourt/ef-cms/pull/387/commits/ae8b6b68eff79543291f1bb63cebcb3188873e95

aws/lambdas/LogsToElasticSearch_info/index.js
julialeague: Thanks, this doesn't seem necessary - changed! https://github.com/ustaxcourt/ef-cms/pull/387/commits/b6d2d97e80083e6e5196a332b1b35eaf76fd7955

adunkman (approved):
🚀 🎸 🎉 👍 🥇 💯

PR #395, opened 2020-09-04
IRS fix: cache keys from more than one user pool in authorizer. 

adunkman:
This fixes an error seen in the IRS environment:

```
TypeError: Expected "jwk" to be an Object
    at t (/var/task/cognito-authorizer.js:172:109)
    at i (/var/task/cognito-authorizer.js:524:512)
    at Runtime.exports.handler (/var/task/cognito-authorizer.js:524:969)
    at Runtime.handleOnce (/var/runtime/Runtime.js:66:25)
```

The issue is that the lambda caches the keys retrieved from the user pool endpoint, and since the IRS is using two user pools simultaneously, the second request fails as it is not signed by the first request’s user pool keys.

The fix is to allow the lambda to cache keys from more than one user pool, and look up the appropriate pool’s authorized keys (or fetch them) when authorizing.

Animated GIF of a stealthy woman sipping a piña colada

### DIFF
--- a/web-api/terraform/template/lambdas/cognito-authorizer.js
+++ b/web-api/terraform/template/lambdas/cognito-authorizer.js
@@ -22,8 +22,15 @@ const generatePolicy = (principalId, effect, resource) => {
   return authResponse;
 };
 
-const verify = (methodArn, token, keys, kid, cb) => {
+const verify = (methodArn, token, keys, kid, iss, cb) => {
   const k = keys.keys.find(k => k.kid === kid);
+
+  if (!k) {
+    throw new Error(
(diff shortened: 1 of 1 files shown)


adunkman:
A note here as well, we’ve added a Trello card to [introduce test coverage for the cognito authorizer lambda](https://trello.com/c/1wEU2tOw/779-cover-the-cognito-authorizer-lambda-with-unit-tests) so we can get a regression test in place.

mmarcotte (approved):
Thumbs up 

This was a fun one; thanks @adunkman . 

@codyseibert I know you're going to like the resolution.

PR #422, opened 2020-09-23
change order of token validation to match order of use

mark-meyer:
This changes the order of token validation to match order of use to ensure the used token is the one that is validated.


mmarcotte (approved):
major kudos to @mark-meyer

PR #424, opened 2020-09-24
ensure queryStringParameters is available

mmarcotte:
#422 introduced an `typeError` that `queryStringParameters` was not set upon a `POST` request.

```
{ “errorType”: “TypeError”, “errorMessage”: “Cannot read property ‘token’ of undefined”, “stack”: [ “TypeError: Cannot read property ‘token’ of undefined”, ” at Runtime.exports.handler (/var/task/cognito-authorizer.js:65266:35)“, ” at Runtime.handleOnce (/var/runtime/Runtime.js:66:25)” ] }
```

Found in `cognito_authorizer_lambda_{env}`

Quick fix to ensure that it is properly set before proceeding.


mark-meyer (approved):
Looks good to me.

Issue #317, opened 2020-07-28
Application monitoring


vickimcfadden:
As the Court product owner and lead software engineer, I need to understand site performance, so that I know when it is not operating within an acceptable range and can make necessary changes. 

Acceptance criteria:

- [x] Investigate what is available out-of-the-box from AWS CloudWatch(i.e. uptime, ping times, logs being generated, database calls per day, calls across gateways, CPU/Memory load levels, HTTP response codes/error logging, number of failed/successful log-ins)
- [X] Investigate what is available with HoneyBadger. *Decision made not to use in production environments but keep enabled in other environments*
- [X] Determine if there is any necessary work to be done prior to installing AWS monitoring **yes, see below**
- [x] 18F provide a cost estimate to the Court for additional elasticsearch instance **$30.55 / month, see below**
- [x] Terraform configuration @julialeague 
- [x] Ongoing work: Configure Kibana dashboard(s) to track errors, exceptions, and metrics. Approved to proceed. @julialeague 
- [ ] ~18F + Court performs a logging audit (number of log-ins, number of documents QC'd, number of documents signed, number of documents completed, did X Court user view this document, email bounces).~ Mike/Laura/Andrew to kick off Nov 6, tracking progress in https://github.com/ustaxcourt/ef-cms/issues/474.
- [ ] ~Court consider what metrics it wants tracked in Kibana dashboard (i.e. number of log-ins, number of documents QC'd, number of signatures, did X user view this document, etc). @JessicaMarine to talk to Fig & Loretta~ (Moved to https://github.com/ustaxcourt/ef-cms/issues/371)
- [x] Talk with Flexion about elasticsearch at upcoming design meeting, who creates, if 18F creates, does Flexion want a mirror image with terraform? **18F creates and Flexion can ingest data if they want to.**  @mmarcotte to follow up with Flexion
- [x] Add a user story to the Flexion backlog that describes the work to be done, as appropriate. @adunkman @mmarcotte to develop a story for log redaction of security tokens.
- [ ] ~Figure out how to archive information. Elasticsearch will only hold data for 1-2 weeks.~ (Moved to https://github.com/ustaxcourt/ef-cms/issues/413)
- [ ] ~Do not index log items that start with: `REPORT` or `END`~ (Moved to https://github.com/ustaxcourt/ef-cms/issues/474).

Related to #137 and #339.

adunkman:
Pulling in from #233: 

> Notifications are sent when:
[...]

PR #433, opened 2020-09-30
Increase elasticsearch resources for mig & test

mmarcotte:
In order to have better elasticsearch functionality in the environments where we are housing a significant amount of data, this change increases the number of nodes from 1 to 3 for the `test` and `migration` environment. and from 2 to 3 for the `prod` environment ahead of testing next week.


adunkman (approved):
Especially interested in which metrics respond in more positive ways after this change — let’s keep an eye out so we can set overloaded alarms in #317.

PR #435, opened 2020-09-30
Add Feature Request template

mmarcotte:
Flexion recently created two issue templates. After production launch, we will probably be removing their USTC Story template, and going with a version of this one for the general public to report their issues. We are going to adopt some version of the Flexion template for Bugs.


adunkman (approved):
The sweet sweet smell of fresh issue templates. Looks great! 👍

PR #441, opened 2020-10-01
Disassociate `log_group` from `_type`

mmarcotte:
@mark-meyer found that `_type` is a field that must be unique in Kibana. The magic lambda from AWS links it with the log_group. So, in order to index multiple log groups, we are setting a static string to `_type`. 

We will have to figure out a way to separate logs by environment later, but for now i think we can search off of `log_group`


mark-meyer:
Yes, this is how we dealt with the error related to multiple `_type` not being allows in Elasticsearch post v6.0.

mark-meyer (approved):

PR #446, opened 2020-10-02
Fix kibana searching

mmarcotte:
The recently introduced `LogsToElasticSearch_info` lambda did not have privileges to write to Cloudwatch. This enables that policy so it can function properly and hopefully begin to populate Kibana.


adunkman (approved):
Looking good to me — looks like we might have been struck from the console auto-magically creating things for us again. 

In the spirit of Fat Bear Week 2020…

A cartoon bear getting buried in hearts

Issue #430, opened 2020-09-29
Identify and scope out and implement v1 API.


adunkman:
As an API consumer like the IRS, so that I can maintain uptime on services which use Dawson API, I need to be able to use a stable API with limited changes.

### Notes

- We should retain our ability to change our data structures internally while providing a stable API for consumers.

adunkman:
Current usage of fields from the IRS:

| IRS field name | EF-CMS field |
|---------------|--------------|
tc_case_name                | case.caseCaption
tc_docket_number_suffix     | case.docketNumberSuffix
tc_case_status              | case.status
tc_case_type                | case.caseType
tc_served_date              | document.servedAt
tc_filed_by                 | document.filedBy
tc_filed_date               | document.filingDate
tc_trial_location           | case.trialLocation
tc_preferred_trial_city     | case.preferredTrialCity
tc_practitioner_name        | practitioners.name
tc_practitioner_address     | practitioners.contact
tc_served_parties           | document.servedParties
tc_contact_primary          | case.contactPrimary
tc_street_address           | case.contactPrimary
tc_city                     | case.contactPrimary
tc_state                    | case.contactPrimary
tc_zipcode                  | case.contactPrimary
tc_initial_case_title       | case.initialTitle
tc_sort_docket_number       | case.sortableDocketNumber
tc_lead_case_docket_number  | case.leadCaseId
tc_party_type               | case.partyType
tc_filing_type              | case.filingType
tc_contact_secondary        | case.contactSecondary
tc_entity_name              | case.entityName
tc_practitioner_bar_number  | practitioners.barNumber
tc_practitioner_phone       | practitioners.contact
tc_practitioner_email       | practitioners.email
tc_practitioner_svc_ind     | practitioners.serviceIndicator
tc_trial_date               | case.noticeOfTrialDate
tc_respondent_name          | respondents.name
tc_respondent_address       | respondents.contact
tc_respondent_bar_number    | respondents.barNumber
tc_respondent_phone         | respondents.contact
tc_respondent_email         | respondents.email
tc_respondent_svc_indicator | respondents.serviceIndicator
tc_order_for_amend_petition | case.orderForAmendedPetition
tc_order_for_ap_filing_fee  | case.orderForAmendedPetitionAndFilingFee
tc_order_for_filing_fee     | case.orderForFilingFee
tc_order_for_ods            | case.orderForOds
tc_order_for_ratification   | case.orderForOds
[...]

PR #447, opened 2020-10-02
Add v1 api specification.

adunkman:
Regarding #430, this PR introduces a OpenAPI specification for the API which doesn’t yet exist. This is the two endpoints used by the IRS, with the properties they mentioned in use, leaving off those called out in #430. 

[View the rendered documentation](https://app.swaggerhub.com/apis-docs/adunkman/dawson/1).

Before merging, I’d like to run it by folks all around the team + at Flexion, and then check with the IRS team that this will be sufficient.

### DIFF
--- a/docs/api/v1.yaml
+++ b/docs/api/v1.yaml
@@ -143,24 +143,24 @@ components:
     Contact:
       type: object
       properties:
+        name:
+          type: string
+        phone:
+          type: string
+        email:
+          type: string
         address1:
           type: string
-          example: 111 South West St.
         address2:
           type: string
-          example: null
         address3:
           type: string
-          example: null
         city:
           type: string
-          example: Orlando
         state:
           type: string
-          example: AL
         postalCode:
           type: string
-          example: "12345"
     DocketEntry:
       type: object
       properties:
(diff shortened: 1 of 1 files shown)


adunkman:
Updated based on our conversation yesterday in 26904d4; I believe this is good to get a trial implementation going.

docs/api/v1.yaml
```diff
@@ -0,0 +1,222 @@
...
+          schema:
+            type: string
+      responses:
+        '200':
+          description: case object
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Case'
+        '401':
```
mmarcotte, line 42: might at a 404 if case isn't found?

mmarcotte (commented):
This is beautiful. Helps me organize my thoughts so well. Thank you for the effort. One quick q?

docs/api/v1.yaml
adunkman: Good catch — it should, definitely. Updated in cb3f6da!

mmarcotte (approved):
So organized
So organized. Thank you!!

adunkman:
Communicated this with the IRS today, they don’t expect any changes will be needed.

PR #448, opened 2020-10-05
Fix kibana searching

mmarcotte:
Continue work to get Kibana indexing logs.


adunkman (approved):
Thanks for digging into this, it’s so exciting to see this in Kibana! 👍

PR #449, opened 2020-10-05
add statuspage env var to circle

mmarcotte:
Inform CircleCI that to pass in the `STATUSPAGE_DNS_RECORD` environment variable. This way `deploy-app.sh` can make use of it and pass it along to terraform


adunkman (approved):
👍 🚀 

One of Ru Paul’s drag queens posting on their phone, "status update: dead"

PR #462, opened 2020-10-14
Revert behavior change for unknown docket numbers.

adunkman:
Introduced in a2285afde8c1dccb9fdf0dede17260bc1fb90a7f, this commit reverts the change in behavior to return a default object instead of null when a docket number is not found.

By doing so, we restore the 404 behavior on the get case API endpoint — that is, unknown docket numbers return a 404 Not Found instead of 400 Bad Request (due to the default object failing case validation checks). 

This change causes test failures I’m currently looking into — it appears some tests are incompletely mocked:

- Test mocks `getCaseByDocketNumber` to return a non-existent case.
- During the course of executing the test subject, the subject calls a non-mocked `getCaseByDocketNumber`.
- This non-mocked method returns null, because the mocked case does not exist. 
- The test fails due to null references.

### DIFF
--- a/shared/src/persistence/dynamo/cases/getCaseByDocketNumber.js
+++ b/shared/src/persistence/dynamo/cases/getCaseByDocketNumber.js
@@ -24,7 +24,7 @@ exports.getCaseByDocketNumber = async ({
     applicationContext,
   });
 
-  if (!caseItems) {
+  if (caseItems.length === 0) {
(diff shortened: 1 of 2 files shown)


adunkman:
Worked around the issue for now and filed https://github.com/flexion/ef-cms/issues/6853 for the dev team to take a further look.

PR #463, opened 2020-10-15
Implement v1 API.

adunkman:
Implements the v1 API as specified in ustaxcourt/ef-cms#6659. 

- This PR adds two additional error cases to the API spec, 403 (when your credentials are authorized but you don’t have permission to access the resource, distinct from the 401 which already exists and indicates that your credentials were rejected) and 500 (💥 ).
- The two new routes wrap the existing interactors as implemented to catch all the already-implemented business logic. 
- The `web-api/v1/` folder contains all the version-locked objects (most notably, a set of marshallers), leaving the remaining objects elsewhere in the application to change as needed.
- This morning when I came in with a fresh mind, I decided to implement a workaround for ustaxcourt/ef-cms#6670 instead of solving it right away. I don’t think this needs to be held up on that bug fix. 

Remaining: 

- [x] Implement marshaller tests.
- [x] Implement tests of the lambdas. We don’t have a pattern of lambda tests in this application — I think given that the v1 API has additional requirements to not change, it would be appropriate to assert the lambda function behaviors (most notably that errors are serialized to the required format).
- [x] Report bugs for skipped tests, document them in a comment above the skipped test, and reference the test in the bug report.


adunkman:
This is ready for review! 

- There are a few skipped tests, because the underlying API endpoints do not behave in the expected way. To move the API forward, I skipped those tests so we have it documented how the API should behave, and filed bugs https://github.com/flexion/ef-cms/issues/6853 and https://github.com/flexion/ef-cms/issues/6854.
- This PR introduces lambda tests for the API endpoints, which assert behavior of the route handlers down to the objects which interact with the AWS APIs.

web-api/src/applicationContext.js
```diff
@@ -1275,7 +1266,15 @@ const gatewayMethods = {
...
-  if (appContextUser) setCurrentUser(appContextUser);
+  let user;
+
+  if (appContextUser) {
+    user = new User(appContextUser);
+  }
+
+  const getCurrentUser = () => {
+    return user;
+  };
```
adunkman, line 1251: Ooh, one more thing to mention — the `user` associated with an `applicationContext` was previously stored in a module-scoped variable. This meant that:

```js
const appContext1 = createApplicationContext(user1);
appContext1.getCurrentUser(); // returns user1

const appContext2 = createApplicationContext(user2);
appContext2.getCurrentUser(); // returns user2

// has been modified by the second function, now returns user2
appContext1.getCurrentUser(); // returns user2
```

There aren’t tests for the applicationContext, but I moved it to a function-scoped variable here so it no longer interacts with other instances.

web-api/src/v1/getCaseLambda.js
```diff
@@ -0,0 +1,27 @@
+const { genericHandler } = require('../genericHandler');
+const { marshallCase } = require('./marshallers/marshallCase');
+const { v1ApiWrapper } = require('./v1ApiWrapper');
+
+/**
+ * used for fetching a single case and returning it in v1 api format
+ *
+ * @param {object} event the AWS event object
+ * @returns {Promise<*|undefined>} the api gateway response object containing the statusCode, body, and headers
```
mmarcotte, line 10: add quick line for `options` parameter

mmarcotte (approved):
Yes Please

web-api/src/v1/getCaseLambda.js
adunkman: Oops! Added in 521b0dc.

PR #470, opened 2020-10-20
Tweaks for Dry-Run in Test Environment

mmarcotte:
This work includes the following:

- Following other patterns in the code base, if a petitioner record exists in Cognito, attempt to create the record in DynamoDB anyhow. We are doing this because we are creating the Cognito records ahead of time. This function currently catches instances where this isn't the case, and then attempts to create that record.
- Disable Smoketests for `test` and `migration` environment.
- Suppress welcome email when creating Petitioner record during Migration process
- Adjust script to bulk import practitioners
- Ensures proper env variables are configured when performing the switch-colors step locally


adunkman (approved):
Looks good!

.circleci/config.yml
```diff
@@ -951,9 +951,7 @@ workflows:
             branches:
               only:
                 - irs
-                - migration
```
mmarcotte, line 954: don't want to perform smoketests in `migration` or `test` environment that would contaminate migration data

shared/src/persistence/dynamo/users/createMigratedPetitionerUser.js
```diff
@@ -36,6 +36,7 @@ exports.createMigratedPetitionerUser = async ({ applicationContext, user }) => {
     const response = await applicationContext
       .getCognito()
       .adminCreateUser({
+        MessageAction: 'SUPPRESS',
```
mmarcotte, line 39: don't want to send emails during migration for petitioners who haven't already been created

shared/src/persistence/dynamo/users/createMigratedPetitionerUser.js
```diff
@@ -65,6 +66,34 @@ exports.createMigratedPetitionerUser = async ({ applicationContext, user }) => {
       userId: response.User.Username,
     });
   } catch (err) {
+    // the user already exists in Cognito
     applicationContext.logger.error(err);
+    const response = await applicationContext
```
mmarcotte, line 71: reusing existing pattern for trying to get the user from Cognito if creating the user in Cognito fails. Then proceed with creating the DynamoDB record...

shared/src/persistence/dynamo/users/createPractitionerUser.js
```diff
@@ -52,8 +52,13 @@ exports.createUserRecords = async ({ applicationContext, user, userId }) => {
 
 exports.createPractitionerUser = async ({ applicationContext, user }) => {
   let userId = applicationContext.getUniqueId();
+  const practitionerRoleTypes = [
```
mmarcotte, line 55: this borrows work from https://github.com/flexion/ef-cms/pull/6870

web-client/switch-colors.sh
```diff
@@ -1,5 +1,11 @@
 #!/bin/bash
 
+[ -z "${ENV}" ] && echo "You must have ENV set in your environment" && exit 1
```
mmarcotte, line 3: for when this is run locally, it's helpful to short circuit if the environment isn't properly configured

.circleci/config.yml
adunkman: Partial revert of https://github.com/ustaxcourt/ef-cms/pull/482, linking in for future reference.

shared/src/persistence/dynamo/users/createMigratedPetitionerUser.test.js
```diff
@@ -46,6 +57,7 @@ describe('createMigratedPetitionerUser', () => {
     expect(
       applicationContext.getCognito().adminCreateUser,
     ).toHaveBeenCalledWith({
+      MessageAction: 'SUPPRESS',
```
adunkman, line 60: 👏

adunkman (approved):
👍 Thanks for the test updates!

Issue #481, opened 2020-10-23
Improve documentation and procedure for deploying to production


mmarcotte:
Acceptance Criteria:

- [ ] Tech lead has instructions that are clear and tested to perform a Blue/Green deployment to Production environment

PR #482, opened 2020-10-23
Ensure that blue/green migration process is automated on Court environments

mmarcotte:
This PR ensures that the `migrate`, `smoketests`, and the `switch-colors` steps are performed by CircleCI for the following Court environments:

* Staging (already working)
* IRS
* Test
* Migration

This will speed up the process of moving fixes through the Court's environments and reduce the chance for errors performing these steps manually.

The process is manual for production; however, it can still be improved upon. That is [tracked here](https://github.com/ustaxcourt/ef-cms/issues/481).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -923,9 +923,9 @@ workflows:
             branches:
               only:
                 - irs
+                - migration
                 - staging
                 - test
-                - migration
                 - master
                 - experimental1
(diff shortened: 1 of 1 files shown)


adunkman (approved):
Looks good! Glad we’re ensuring these are run — I didn’t realize they were being skipped. 

Boo

PR #492, opened 2020-10-26
Use SES to send email on behalf of Cognito; Tidy noreply@ emails

### LINKED ISSUE #460: BUG: Mail from domain does not match sending domain
Following up from https://github.com/ustaxcourt/ef-cms/commit/39ca8a61b37fadabdb0c2f34e7eb920831baafc4, a part of https://github.com/flexion/ef-cms/pull/6328 to fix https://github.com/flexion/ef-cms/issues/4834 which came over in https://github.com/ustaxcourt/ef-cms/pull/390, the mail sending domain is still configured incorrectly in two places: 

- https://github.com/ustaxcourt/ef-cms/blob/7cab0b95e9f90ec2666e79b65c9a17fc9cb57962/web-api/terraform/template/cognito-trigger.tf#L36
- https://github.com/ustaxcourt/ef-cms/blob/bed4468129b3fd6a4396fab62b3324a63edd9dfb/web-api/terraform/template/cron.tf#L14

### LINKED ISSUE #6668: (6) Creating orders (Merged with #532, this is a dup)
As an internal court user, so that I can prompt parties in a case to take action, I need the ability to issue an order.

As an internal court user, it is my job to ensure that parties take all necessary action in a case and in a timely manner. To do this, I issue court orders to parties in a case which require them to perform a certian action by a certain period of time. Because the judges are very particular about the type of language they use in an order, I need to create orders using the verbiage my judge prefers. 

Today, my judge asked me to issue an order for petitioner in a bankruptcy case, requesting that they file additional paperwork by Feb 5, 2019. To create the order, I prefer to utilize a template that auto-populates basic case information into the header and/or footer, because it is tedious to enter this information every time I have to create an order. Once the order is generated with the auto-populated header and footer information, I copy the text from a word processing program I use to store my judge's prefered order verbiage and then I paste it into the template the system has generated for me. The order verbiage contains instructions to the petitioner to submit the bankruptcy documents, along with the date by which the judge would like the documents submitted.

Once I've completed drafting the order, I apply the judges signature and serve it to the petitoner, who recevies electronic notificaction through the system that there's been action in the case. The petitioner can then log into the system and view the order. 

A few days later, my judge asks me if I issued the order, but I cannot remember, so I log into the system and see that I issued the order.

Roles: 

## Tasks
- [ ] Discuss what is missing from the epic. Esp. in light of only looking at happy path.
- [ ] Answer and document all open questions
- [ ] Plan user research needed to complete the epic
- [ ] Conduct any additional user research needed to complete the epic
- [ ] Create or update system and UX flows and diagrams (USTC e-file process, domain model, etc)
- [ ] Determine approach for form error handling
- [ ]  Plan and conduct usability testing

## Definition of Done 
- [ ] Add UX flow docs to code repository
- [ ] Ensure Figma is archived as pages for each sprint; if we move away from sprints, maybe at the end of each epic?

mmarcotte:
To complete ustaxcourt/ef-cms#6666, in conjunction with ustaxcourt/ef-cms#6674 and https://github.com/flexion/ef-cms/issues/6824, this PR aims to establish SES as the sender of email and inform Cognito of this change. It includes documentation in docs/architecture/email.md for manually verifying an email in order to permit Cognito to use the ARN of the Verified Email address. It also cleans up the noreply@ email to use the `EFCMS_DOMAIN` (removing the errant `mail.` prefix.)

This also fixes ustaxcourt/ef-cms#6668.

### DIFF
--- a/docs/architecture/email.md
+++ b/docs/architecture/email.md
@@ -44,3 +44,56 @@ Notes:
 2. EF-CMS uses Amazon SES Easy DKIM, [which uses a 1024-bit key](https://docs.aws.amazon.com/ses/latest/DeveloperGuide/send-email-authentication-dkim-easy.html).
 3. Amazon SES does not support DNSSEC on its Easy DKIM signing domains.
 4. Encrypting emails with S/MIME would reduce the effectiveness of email notifications, as some recipients would not be able to decrypt the emails. No sensitive information is sent in email notifications.
+
+### Verifying SES Email
+
+At present this is a manual process as a human needs to click on the link in the email in order to verify ownership. We verify ownership so that we can use that verified email's ARN in [our Cognito configuration](../../web-api/terraform/template/cognito.tf). This lets us send more email than Cognito's very low quotas allow. 
+
+Here are the steps required to verify a new email:
+
+1. Need to create an s3 bucket (e.g. `mail-verification.example.com`)
+2. Need to grant SES permission to write to it. Replace `BUCKET-NAME` and `AWSACCCOUNTID` in the following policy, and apply it to the newly created bucket:
+
+    ```json
+    {
+        "Version": "2012-10-17",
+        "Statement": [
+            {
+                "Sid": "AllowSESPuts",
+                "Effect": "Allow",
+                "Principal": {
+                    "Service": "ses.amazonaws.com"
+                },
+                "Action": "s3:PutObject",
+                "Resource": "arn:aws:s3:::BUCKET-NAME/*",
+                "Condition": {
+                    "StringEquals": {
+                        "aws:Referer": "AWSACCOUNTID"
+                    }
+                }
+            }
+        ]
+    }
+    ```
+
+3. In SES, create a rule set: (e.g., `confirm_email_helper`)
+4. Need to create a Rule in that has the following attributes: (Replace `BUCKET-NAME`)
+
+   ```json
+   {
+      "Name": "Confirm Email Rule",
+      "Enabled": true,
+      "Actions": [
+        {
+          "S3Action": {
+            "BucketName": "BUCKET-NAME"
+          }
+        }
+      ]
+    }
+    ```
+
+5. Need to make the Rule Set active
+6. Use SES to send a verification email to that account.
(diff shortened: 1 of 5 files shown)


julialeague (approved):
Looks good to me - I'm not clear on how emails will be sent out going forward, but that's beyond the scope here.

docs/architecture/email.md
```diff
@@ -44,3 +44,56 @@ Notes:
...
+          }
+        }
+      ]
+    }
+    ```
+
+5. Need to make the Rule Set active
+6. Use SES to send a verification email to that account.
+7. Check the S3 bucket. It should have an item in there with a link to confirm the Email Address. Follow the link, and the email address should be verified.
+8. When complete, delete the S3 Bucket, SES Rule, and SES Rule Set. Future deploys should rely on the verified email.
```
adunkman, line 99: I’m a bit confused here — will this store emails in S3 from this point forward? This looks to be nearly directly the example from *[How can I use Amazon SES to receive inbound emails, and then store those emails on Amazon S3?](https://aws.amazon.com/premiumsupport/knowledge-center/ses-receive-inbound-emails/)*. 

Is this the desired functionality? :scratching-head:

adunkman: Maybe we should grab a few minutes to pair. I’m definitely not an expert on SES.

mmarcotte: Yes, this is the desired functionality. And good find, I did use that as a guide for setting this up.

This is what lets you verify the email address without actually requiring an inbox or to forward the user. 

And in fact, I think we're going to use this functionality by setting practitioner and petitioner emails to `md5(${originalEmail})` + `@mig.ef-cms.ustaxcourt.gov` during the next dry-run. This will help us:

- verify that no emails are sent during the migration.
- verify that the welcome emails are all sent
- verify that service emails are being sent properly.

... all by checking the S3 bucket.

adunkman: Gotcha! I thought "This is what lets you verify the email address without actually requiring an inbox or to forward the user." meant you were intending to use this S3 "inbox" as a way to verify an email address, rather than an S3 destination — I can’t believe this is a desired and intended process from AWS, but… /shruggie.

👍 🎉

PR #493, opened 2020-10-26
Increase the volume size for Kibana

mmarcotte:
To assist with #317 -- we need increased resources to index all of the items that we are attempting to index in Elasticsearch to monitor in Kibana. This increases the ebs from 10GB to 30GB. With 3 nodes, that should be enough (for now).


julialeague (approved):
🥳

adunkman (approved):
Looks great! 

Pumpkin dancing

adunkman:
This no longer represents the resource requirements for info ES cluster; we’ll need to do a follow-up. Andrew and Mike to pair on a next implementation.

PR #498, opened 2020-10-27
Disable auto court users on `test` env

mmarcotte:
Simple one line fix to prevent CircleCI from creating the test Court users in the `test` environment. This environment is now housing real Court users, and we don't want to add confusion.


adunkman (approved):
Thumbs up, but extra cheesy

julialeague (approved):

PR #514, opened 2020-11-04
Configure JSON logging with Winston.

adunkman:
This PR gets us most of the way to ustaxcourt/ef-cms#6678 by implementing JSON logging and turning off debug log statements.

- This PR introduces a new `debug` log level which is visible when running the application locally. This PR also sets the log level to `info` in production, and it can be tweaked ad-hoc in the AWS console by modifying the `LOG_LEVEL` environment variable if we want to turn up logging temporarily.

- In production, the logger outputs JSON strings with request IDs attached. There are three request IDs that are AWS-assigned, and all three are output on log messages: 
  1. The API gateway request id, served as a response header on all responses. 
  2. The application load balancer request id (also known as the trace ID), set as a request header on all requests. 
  3. The lambda invocation request id, shown in cloudwatch logs for the lambda invocation.

  By specifying all three, we can search Kibana for any request ID we have and find the related log entries without having to know what kind of request ID it is. 

Thanks for taking the lead on this for so long, @julialeague!


shared/src/utilities/createLogger.js
```diff
@@ -0,0 +1,43 @@
...
+const {
+  createLogger: createWinstonLogger,
+  format,
+  transports,
+} = require('winston');
+const { cloneDeep, unset } = require('lodash');
+
+const redact = format(logEntry => {
+  const copy = cloneDeep(logEntry);
+  ['token', 'headers.authorization'].forEach(k => unset(copy, k));
```
adunkman, line 14: This redacts `token` and `headers.authorization` from any logged object, accomplishing https://github.com/flexion/ef-cms/issues/6408 for in-application logging. 

https://github.com/flexion/ef-cms/issues/6408 should still address any logging happening outside of the application.

web-api/src/lambdaWrapper.test.js
```diff
@@ -100,4 +103,65 @@ describe('lambdaWrapper', () => {
       'ERROR: we do not support this return type',
     );
   });
+
+  describe('request IDs', () => {
+    it('passes request IDs to event if set', async () => {
+      // set by aws-serverless-express.eventContext()
```
julialeague, line 109: thank you for this explanation 👍

julialeague (approved):
This looks great! Looking forward to seeing this over in Kibana now.

shared/src/utilities/createLogger.js
mmarcotte: @adunkman -- So, in other words, https://github.com/flexion/ef-cms/issues/6408 would still seek to remove instances where `console.log(...)` or some other logging could be exposing information. It would ensure that we use the `applicationContext.logger` for those instances...?

web-api/src/lambdaWrapper.js
```diff
@@ -5,6 +5,12 @@ export const lambdaWrapper = lambda => {
       path: req.path,
       pathParameters: req.params,
       queryStringParameters: req.query,
+      requestId: {
+        apiGateway: (((req.apiGateway || {}).event || {}).requestContext || {})
+          .requestId,
+        applicationLoadBalancer: req.headers['x-amzn-trace-id'],
+        lambda: ((req.apiGateway || {}).context || {}).awsRequestId,
+      },
```
mmarcotte, line 13: just, wow.

mmarcotte (approved):
Bravo

Bravo @adunkman and @julialeague . This is going to be instrumental for monitoring and troubleshooting throughout the life of the application. You've built a solid foundation. Thank you!

shared/src/utilities/createLogger.js
julialeague: That seems apropos - any potentially sensitive output from shell scripts and Honeybadger being the low-hanging fruit.

PR #526, opened 2020-11-10
Update LogsToElasticSearch_info.

adunkman:
Note for review: this may be easiest to step through commit-by-commit, as the refactoring commits introduced what appears to be a large changeset. It’s more understandable if you look one change at a time. 😄 

This PR: 

- ♻️ Refactors the `LogsToElasticSearch_info` function:
  - To follow `eslint` JavaScript standards defined elsewhere in the project.
  - To use `async` and `await` instead of callbacks, bringing it up-to-date with the rest of the code base.
  - To abstract the decompression and parsing logic further for greater readability.
- 🤫 Only sends JSON logs to Kibana. Any log entry which fails JSON parsing is ignored, which includes log entries created by AWS with metadata about lambda invocations (desired). 
- 🧹 Introduces an Index Template to our Elasticsearch cluster which will store all log data passed, but indexes only the known log keys. This allows our cluster to accept all log entries without having memory issues or rejecting entries, even those which may have more than 1,000 keys. 

Caveats:

- 🤔 The Index Template needs to be applied to the Elasticsearch cluster after it is created manually. There is [a promising Terraform Elasticsearch provider](https://registry.terraform.io/providers/phillbaker/elasticsearch/latest/docs) which may allow us to manage this through Terraform — however, it requires Terraform 0.13. I suggest we backlog a ticket to convert this (and any other) Elasticsearch logging resource to Terraform at a later date.

Status:

- This change has already been applied to the Court’s non-production AWS account manually — so you can see the result of this change in the Court’s Kibana interface.

### DIFF
--- a/aws/lambdas/LogsToElasticSearch_info/index-template.md
+++ b/aws/lambdas/LogsToElasticSearch_info/index-template.md
@@ -1,0 +1,52 @@
+# Elasticsearch Index Template
+
+When parsing log entries into an index, Elasticsearch will attempt to create a new index with dynamic field mappings if it does not already exist (the normal behavior). We can control how Elasticsearch performs that operation with an Index Template.
+
+Because log entries may contain data which is of an unknown shape and size, it is desirable to [limit the dynamic field mappings](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping.html#mapping-limit-settings) that occur. Say more!
+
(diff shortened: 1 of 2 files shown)


mmarcotte (approved):
Thank you so much for this. Appreciate the doc on the index template. I'll add an item to the backlog for doing this when we move to Terraform 0.13.

clap clap clap

Issue #496, opened 2020-10-26
BUG: log_viewers cognito pool recreated on account-specific terraform run


adunkman:
The `log_viewers` cognito pool is modified every time the account-specific terraform is run, showing this difference: 

```
  # aws_cognito_identity_pool.log_viewers will be updated in-place
  ~ resource "aws_cognito_identity_pool" "log_viewers" {
        allow_unauthenticated_identities = false
        arn                              = "arn:aws:cognito-identity:us-east-1:350455059537:identitypool/us-east-1:55e8d40b-2580-4f6f-a710-a4e074270133"
        id                               = "us-east-1:55e8d40b-2580-4f6f-a710-a4e074270133"
        identity_pool_name               = "kibana dashboard identity pool"
        openid_connect_provider_arns     = []
        saml_provider_arns               = []
        supported_login_providers        = {}
        tags                             = {}

      - cognito_identity_providers {
          - client_id               = "641p536mtf742ev6s24q9604d" -> null
          - provider_name           = "cognito-idp.us-east-1.amazonaws.com/us-east-1_qJr5ZVWNT" -> null
          - server_side_token_check = true -> null
        }
    }
```

It appears that we need to specify the configuration for the `cognito_identity_providers` block in terraform. This looks like configuration that’s auto-created when applying the pool to Kibana.

adunkman:
Discovered this is actually an issue causing intermittent lockouts of Kibana — the cognito identity providers are being automagically applied by AWS Elasticsearch and then un-applied by terraform. Pulling in a fix to #532!

adunkman:
Fixed in #532.

PR #532, opened 2020-11-12
Adjust account-specific terraform to manually-applied changes.

adunkman:
These tweaks have been made to update our logging infrastructure; apply them to terraform to bring things back in line. This change goes hand-in-hand with an update to our [Court-specific environment documentation](https://github.com/ustaxcourt/ef-cms/wiki/AWS-environment-configuration) to document these values.

### DIFF
--- a/iam/terraform/account-specific/bin/deploy-app.sh
+++ b/iam/terraform/account-specific/bin/deploy-app.sh
@@ -10,6 +10,16 @@ if [ -z "$ES_LOGS_INSTANCE_COUNT" ]; then
   exit 1
 fi
 
+if [ -z "$ES_LOGS_INSTANCE_TYPE" ]; then
+  echo "Please export the ES_LOGS_INSTANCE_TYPE variable in your shell"
+  exit 1
+fi
+
+if [ -z "$ES_LOGS_EBS_VOLUME_SIZE_GB" ]; then
+  echo "Please export the ES_LOGS_EBS_VOLUME_SIZE_GB variable in your shell"
+  exit 1
+fi
+
 if [ -z "$COGNITO_SUFFIX" ]; then
   echo "Please export the COGNITO_SUFFIX variable in your shell"
   exit 1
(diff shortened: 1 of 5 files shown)


mmarcotte (approved):
super

Thanks for this Andrew!

adunkman:
Added one more commit to fix #496: 

- When deploying a domain, AWS Elasticsearch creates a Cognito user pool client and its mapped identity provider, configured to point at the domain endpoint. 
- This creates an issue when Terraform runs against the Cognito identity pool, since it sees no such mapped identity provider and removes it. (Reported at #496). This breaks authentication for Kibana when the run happens. 
- Any time the domain is deployed — on changes to its volumes, instances, configuration, etc — it re-creates the user pool client and mapped identity provider if it doesn’t exist. This fixes authentication. 

…

FIGHT! 

- I initially attempted to bring these resources under Terraform control, which would allow us to manage them ourselves as we do all other AWS resources. This creates a cycle:
  - To configure the user pool client, you need the endpoint for Elasticsearch.
  - To determine the endpoint for Elasticsearch, you need the user pool client.
- My attempted fix was to use a known domain for the Elasticsearch endpoint (`logs.ef-cms.ustaxcourt.gov`) to break the cycle. This worked great! Except…
  - Even when Elasticsearch is configured to use a custom domain (https://github.com/hashicorp/terraform-provider-aws/issues/16059), it still forces the Cognito user pool client and its mapped identity provider to point to the AWS-managed domain. The one that we can’t know until after the Elasticsearch domain is created. 

There is no permanent fix without a change in AWS behavior (https://github.com/hashicorp/terraform-provider-aws/issues/5557) —  https://github.com/ustaxcourt/ef-cms/pull/532/commits/c853d8bcfcb2eeafe15624cd073b411696f88d0d implements a workaround (to have Terraform ignore changes to `cognito_identity_providers`, sidestepping the fight).

adunkman:
Since this PR is still open, tacked on one more commit to fix a bash scripting bug which we discovered right before our load test yesterday — the IF-statement conditional is reversed for determining if `LOG_GROUP_ENVIRONMENTS` is set, which ensured the default value was always used. Fixed in https://github.com/ustaxcourt/ef-cms/pull/532/commits/de5aa717196d42560e5e5c71e7d1536bf3ea4a8a.

mmarcotte (approved):

Issue #553, opened 2020-11-17
Create a toggle for maintenance mode


mmarcotte:
Without requiring a deployment, we need to devise and implement a switch that would turn the application into read-only maintenance mode. The application would not be able to receive writes to DynamoDB or Elasticsearch.

vickimcfadden:
@adunkman to add ticket to Flexion's backlog.

adunkman:
Reported as https://github.com/flexion/ef-cms/issues/7095 — I believe that handles this ticket and will close this issue, but feel free to reopen @mmarcotte if you wanted to track it here as well or if this represents any outstanding work.

Issue #415, opened 2020-09-21
Devise and implement backup and restore strategy for DynamoDB


mmarcotte:
In the case of a catastrophic event at AWS or security incident, we need to have a mechanism to backup a snapshot of the DynamoDB to a secure location outside of the AWS Account. Will confer with @michael-mcvicker about the best place for this location, perhaps in an on-prem resource.

adunkman:
May involve piggybacking on some recurring cron jobs, or using dynamo’s point-in-time backups. @mark-meyer and @mmarcotte to sync!

mark-meyer:
The DynamoDB tables used for Dawson have [point-in-time recovery](https://aws.amazon.com/dynamodb/pitr/) enabled, which maintains backups for 35 days. It also maintains a 35 day backup for any table that has been deleted. Restoring from the backup is easily done with either the aws cli or the console. 

However, it is not clear what steps will need to be taken. You cannot restore a backup to an existing table; instead you need to restore to a new table and then somehow direct the application to use this table. 

**A possible recovery plan could look like:**

1. Place the app in a maintenance or read-only mode to prevent new data from being written to a possibly corrupt table.
2. Restore from PITR backup to a new DynamoDB table.
3. Re-index elasticsearch so it is in sync with the recovered DynamoDB table.
4. Redirect app to use new dynamoDB table.
5. Delete corrupt table.
6. Exit maintenance mode.

I don't know if there is currently a way to perform items 1 or 4 from this list.

vickimcfadden:
Test blue/green deploy to see if it works and document workflow. @mark-meyer 
@mmarcotte to share documentation on how to run script.

mmarcotte:
Made a PR with improved documentation for how the blue green deployments currently work as well as some notes about how this could inform this ticket. Also logged https://github.com/ustaxcourt/ef-cms/pull/554 to add a switch to put the application into maintenance mode.

mark-meyer:
**Findings for experimenting in exp1**

Both point-in-time and recovery from a deleted dynamoDB table work with the current configuration of ef-cms. With a fully-populated dynamoDB table the recovery into a new table from a backup takes about 70 minutes.

[...]

PR #554, opened 2020-11-17
Improved blue green documentation

mmarcotte:
As a step to informing #415 , this seeks to add clarity to the Blue / Green deployment process. 

We could consider the approach of

1. Restoring from a DynamoDB backup to a table with the name `efcms-ENV-backup`. 
2. Using `backup` as the `source-table-version` with a `migrate` = `true` deployment
3. Then the deployment should use the source table (restored backup) as a basis to repopulate the destination table, which would reindex the information into Elasticsearch.

It's further conceivable that the application could be running on `beta` while restoring to `alpha`. However, we would lose any writes to the system. It would seem that we would need a control to turn the system into maintenance / read-only mode. https://github.com/ustaxcourt/ef-cms/issues/553


mark-meyer (approved):
Thanks. This seems clear to me.

Issue #541, opened 2020-11-13
Improve monitoring and logging with lessons learned from load test.


adunkman:
Creating this issue to track improvements to monitoring and logging from our run of #138: 

- [x] No errors were logged — we suspect an error in the implementation of #502. — #556
- [x] It would have been nice to search and view request headers. — #556
- [x] It was hard to judge how widespread `400` issues may be without a breakdown of HTTP response codes for all API responses. — #556
- [x] Tweak Index Template for logging format changes. #611
- [x] We could not see failed authorization attempts. #572

adunkman:
Pinging @mark-meyer @julialeague @mmarcotte for your contributions to the task list above from things you noticed/noted — feel free to edit! 

I’m dropping this in the "In Progress" column with my name on it as I’m rolling the remaining task from #474 into here as well (since it was something we observed yesterday).

mark-meyer:
@mmarcotte and I were debugging some of the dynamoDB migrations last night. One issue with the current log setup is that we need to be thorough in avoiding uncaught exceptions. The `migration-segments` lambda has places where the lambdas are allowed to fail (by design so the SQS items return to the queue), but since the errors are not caught, there is no explicitly logged json, preventing them from showing up in Kibana. I wonder if there is any way to detect uncaught errors in the logging lambda, or if we just need to be a bit more disciplined in catching and logging errors.

adunkman:
> No errors were logged — we suspect an error in the implementation of #502.

Confirmed that anything passed to `applicationContext.logger.error()` was logged, but the resulting JSON was not very helpful. If there were errors that occurred during our load test that were logged this way, we would have seen log entries with `level: 'error'` but `message: null` — which we didn’t see. However, there are (of course) improvements to be made — #556 implements uncaught exception, uncaught promise rejection logging, and better error formatting (message and call stack now included).

>  - It would have been nice to search and view request headers.
>  - It was hard to judge how widespread 400 issues may be without a breakdown of HTTP response codes for all API responses.

#556 changes how requests are logged, moving them up in the execution stack to the request level. This allows us to ensure that for every request, the outgoing response code, overall response time, and incoming request data are logged. We’ll now be able to see every request which hits the API, even those which do not match routes, and their response codes in the logs.

adunkman:
In meeting with @lauraGgit just now, we determined that we’ll need to index the user email and role for auditing purposes (and, of course, usefulness purposes!). 

Will include in the index tweaks.

PR #556, opened 2020-11-17
Logging improvements based on load testing.

adunkman:
This PR gets us most of the way to #541:

- Improves error formatting for errors logged as instances of Error (`applicationContext.logger.error(new Error())`) so their messages and stack traces appear in the logs.
- Adds log handlers for uncaught exceptions and uncaught promise rejections, so they appear in logs as well.
- Always logs incoming requests and their headers.
- Always logs outgoing response statusCodes and the overall response time in milliseconds (required changing where the logger was instantiated, moving it to ExpressJS middleware instead of in the applicationContext (which still has a reference to it)). We’ll now be able to see every request which hits the API, even those which do not match routes, and their response codes in the logs.

Will drop a few inline comments below, too.

### DIFF
--- a/shared/src/utilities/createLogger.js
+++ b/shared/src/utilities/createLogger.js
@@ -1,7 +1,8 @@
+const colorize = require('logform/colorize');
 const combine = require('logform/combine');
+const errors = require('logform/errors');
 const json = require('logform/json');
 const printf = require('logform/printf');
-const timestamp = require('logform/timestamp');
 const {
   config,
   createLogger: createWinstonLogger,
@@ -12,34 +13,41 @@ const { cloneDeep, unset } = require('lodash');
 
 const redact = format(logEntry => {
   const copy = cloneDeep(logEntry);
-  ['token', 'headers.authorization'].forEach(k => unset(copy, k));
+  ['user.token', 'request.headers.authorization'].forEach(k => unset(copy, k));
   return copy;
 });
 
-const nonProductionFormatters = [
-  combine(
-    timestamp(),
-    printf(
-      info =>
-        `${info.timestamp} ${info.level}: ${info.message} ${JSON.stringify(
-          info,
-          null,
-          2,
-        )}`,
-    ),
-  ),
-];
+const console = () => new transports.Console();
 
-exports.createLogger = defaultMeta =>
-  createWinstonLogger({
-    defaultMeta,
-    format: combine(
-      redact(),
-      ...(process.env.NODE_ENV === 'production'
-        ? [json()]
-        : nonProductionFormatters),
-    ),
+exports.createLogger = (transport = console()) => {
+  const options = {
+    defaultMeta: {},
     level: process.env.LOG_LEVEL || 'debug',
     levels: config.syslog.levels,
-    transports: [new transports.Console()],
-  });
+    transports: [transport],
+  };
+
+  const formatters = [errors({ stack: true }), redact()];
+
+  if (process.env.NODE_ENV === 'production') {
+    options.format = combine(...formatters, json());
+  } else {
+    options.format = combine(
+      ...formatters,
+      colorize(),
+      printf(info => {
+        const metadata = Object.assign({}, info, {
+          level: undefined,
+          message: undefined,
+        });
+
+        const stringified = JSON.stringify(metadata, null, 2);
+        const lines = stringified === '{}' ? [] : stringified.split('\n');
+
+        return [`${info.level}:\t${info.message}`, ...lines].join('\n  ');
+      }),
+    );
+  }
+
+  return createWinstonLogger(options);
+};
--- a/web-api/src/applicationContext.js
+++ b/web-api/src/applicationContext.js
@@ -1242,7 +1242,7 @@ const gatewayMethods = {
   zipDocuments,
 };
 
-module.exports = (appContextUser, requestId) => {
+module.exports = (appContextUser, logger = createLogger()) => {
   let user;
 
   if (appContextUser) {
@@ -1253,13 +1253,13 @@ module.exports = (appContextUser, requestId) => {
     return user;
   };
 
-  const logger = createLogger({
-    environment: {
-      color: environment.currentColor,
-      stage: environment.stage,
-    },
-    requestId,
-  });
+  if (process.env.NODE_ENV === 'production') {
+    const authenticated = user && Object.keys(user).length;
+    logger.defaultMeta = logger.defaultMeta || {};
+    logger.defaultMeta.user = authenticated
+      ? user
+      : { role: 'unauthenticated' };
+  }
 
   return {
     barNumberGenerator,
--- a/web-api/src/documents/addCoversheetLambda.js
+++ b/web-api/src/documents/addCoversheetLambda.js
@@ -15,5 +15,5 @@ exports.addCoversheetLambda = event =>
         ...event.pathParameters,
       });
     },
-    { logEvent: true, logResults: false },
+    { logResults: false },
   );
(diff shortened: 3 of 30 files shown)


shared/src/utilities/createLogger.js
```diff
@@ -12,34 +13,41 @@ const { cloneDeep, unset } = require('lodash');
 
 const redact = format(logEntry => {
   const copy = cloneDeep(logEntry);
-  ['token', 'headers.authorization'].forEach(k => unset(copy, k));
+  ['user.token', 'request.headers.authorization'].forEach(k => unset(copy, k));
```
adunkman, line 16: These are now logged at these locations in a standardized way — it’ll enable us to index the user and request properties in the Index Template (not changed in this PR, figured it would require some tweaking).

shared/src/utilities/createLogger.js
```diff
@@ -12,34 +13,41 @@ const { cloneDeep, unset } = require('lodash');
...
+  const options = {
+    defaultMeta: {},
     level: process.env.LOG_LEVEL || 'debug',
     levels: config.syslog.levels,
-    transports: [new transports.Console()],
-  });
+    transports: [transport],
+  };
+
+  const formatters = [errors({ stack: true }), redact()];
```
adunkman, line 30: This `errors({ stack: true})` does the magic for converting instances of `Error` into log messages nicely.

shared/src/utilities/createLogger.js
```diff
@@ -12,34 +13,41 @@ const { cloneDeep, unset } = require('lodash');
...
+        const metadata = Object.assign({}, info, {
+          level: undefined,
+          message: undefined,
+        });
+
+        const stringified = JSON.stringify(metadata, null, 2);
+        const lines = stringified === '{}' ? [] : stringified.split('\n');
+
+        return [`${info.level}:\t${info.message}`, ...lines].join('\n  ');
+      }),
```
adunkman, line 48: This function handles the log format in development only.

web-api/src/applicationContext.js
```diff
@@ -1253,13 +1253,13 @@ module.exports = (appContextUser, requestId) => {
...
-    },
-    requestId,
-  });
+  if (process.env.NODE_ENV === 'production') {
+    const authenticated = user && Object.keys(user).length;
+    logger.defaultMeta = logger.defaultMeta || {};
+    logger.defaultMeta.user = authenticated
+      ? user
+      : { role: 'unauthenticated' };
+  }
```
adunkman, line 1262: Instead of logging `'Public User'` for the user, this PR changes it so the object structure is the same in the logs — so we can filter and graph user roles consistently. In the logs, public users will now appear as `user.role == 'unauthenticated'` alongside `practitioner`, `petitioner`, etc.

web-api/src/documents/addCoversheetLambda.js
```diff
@@ -15,5 +15,5 @@ exports.addCoversheetLambda = event =>
         ...event.pathParameters,
       });
     },
-    { logEvent: true, logResults: false },
+    { logResults: false },
```
adunkman, line 18: `logEvent` was set to true by default, and always set to true in overrides. I removed it both in references and from the lambdas. 

This diff is the cleanest, but in other lambdas below required ESLint reformatting (so don’t get thrown off by the larger changeset).

web-api/src/documents/getDocumentDownloadUrlLambda.js
```diff
@@ -9,22 +7,11 @@ const { getUserFromAuthHeader } = require('../middleware/apiGatewayHelper');
...
-      await applicationContext.notifyHoneybadger(e);
-      throw e;
-    }
+  genericHandler(event, async ({ applicationContext }) => {
+    return await applicationContext
+      .getUseCases()
+      .getDownloadPolicyUrlInteractor({
+        applicationContext,
+        ...event.pathParameters,
+      });
```
adunkman, line 16: There were two lambdas which logged using the same pattern as the `genericHandler` — so they were logging twice, and notifying honeybadger twice. I removed the duplication.

web-api/src/genericHandler.js
```diff
@@ -45,41 +45,26 @@ exports.genericHandler = (event, cb, options = {}) => {
...
-      logUserLabel = 'User',
-      skipFiltering,
-    } = options;
+      createApplicationContext(user, event.logger);
 
-    try {
-      if (logEvent && applicationContext) {
-        applicationContext.logger.info(logEventLabel, event);
-      }
+    delete event.logger;
```
adunkman, line 50: This is a little bit hacky (okay, pretty darn hacky). I needed to pass the logger instance, which has request-specific details already attached to it, to the `applicationContext` — there is only one object which goes through this path, the `event`. If we were to add another object, we’d need to edit the signature of _every lambda_ which seemed like too much. 

Happy for other opinions/suggestions on a path forward!

web-api/src/genericHandler.js
```diff
@@ -45,41 +45,26 @@ exports.genericHandler = (event, cb, options = {}) => {
...
       const results = await cb({ applicationContext, user });
 
-      const returnResults = !skipFiltering
-        ? exports.dataSecurityFilter(results, {
-            applicationContext,
-          })
-        : results;
+      const returnResults = exports.dataSecurityFilter(results, {
+        applicationContext,
+      });
```
adunkman, line 62: The `dataSecurityFilter` was never skipped. I removed the ability to skip it.

web-api/src/genericHandler.test.js
```diff
@@ -8,7 +8,7 @@ const token =
 
 const MOCK_EVENT = {
   headers: {
-    Authorization: `Bearer ${token}`,
+    authorization: `Bearer ${token}`,
```
adunkman, line 11: This was a test bug — the headers are lowercased.

web-api/src/lambdaWrapper.js
```diff
@@ -5,24 +5,16 @@ export const lambdaWrapper = lambda => {
       path: req.path,
       pathParameters: req.params,
       queryStringParameters: req.query,
-      requestId: {
-        apiGateway: (((req.apiGateway || {}).event || {}).requestContext || {})
-          .requestId,
-        applicationLoadBalancer: req.headers['x-amzn-trace-id'],
-        lambda: ((req.apiGateway || {}).context || {}).awsRequestId,
-      },
```
adunkman, line 13: This parsing is handled over in the ExpressJS middleware now.

web-api/src/lambdaWrapper.js
```diff
@@ -5,24 +5,16 @@ export const lambdaWrapper = lambda => {
...
 
-    if (process.env.USTC_ENV === 'dev') {
-      console.log(`${req.method}: ${event.path}`);
-    }
-
     const response = await lambda({
       ...event,
       body: JSON.stringify(req.body),
+      logger: req.locals.logger,
     });
```
adunkman, line 16: Here‘s where the logger is passed into the `event` so it can be referenced from `applicationContext`.

web-api/src/public-api/casePublicSearchLambda.js
```diff
@@ -15,8 +15,5 @@ exports.casePublicSearchLambda = event =>
         ...event.queryStringParameters,
       });
     },
-    {
-      isPublicUser: true,
-      user: {},
-    },
+    { user: {} },
```
adunkman, line 18: `isPublicUser` was used for logging purposes only. Now, the logger checks if the user is `{}`, which is the same check other parts of the application use to determine if it is a public user.

mmarcotte (approved):

PR #574, opened 2020-11-24
Improve logging for pdf processing lambda

mmarcotte:
The process that currently scrapes legacy PDFs into files and saves to DDB is challenging! The Lambdas often fail, even when fully allocated with resources. I've had to manually increase this amount in previous runs in order for it to finish. Some still prove to max out the resources that Lambdas were equipped to do.

For any that fail, we need to be able to better identify where they fail in the logs. So, I've included the `docketNumber` and the uuid for the document so we can more easily track down how big they are, whether or not they exist _somewhere_ (perhaps they didn't get synced over?), and process them manually.


adunkman (commented):
Looks good! I think you’ll have a few test failures — looks like the specific error messages are asserted on in `parseLegacyDocumentsInteractor.test.js` & they’ll need an update.

julialeague (approved):

mmarcotte:
Believe it or not @adunkman the test still passed! The error assertions were a mere substring. So I expanded on the substring to include the information I was looking for. And I added a test for when the scraping of the PDFs happen. 

Will want to get this in before we process the 3.7M PDFs waiting to be processed in the `mig` environment.

adunkman (approved):
🎉 Looks great — I kicked off a rebuild of `build-client-integration` just now, appears to have failed due to a networking issue, so I don’t have any concerns about this code. 

🚢 🚢 :shipit: 🚢 🚢

PR #584, opened 2020-11-27
Add statuspage.io embed to the Public App

mmarcotte:
In order to better communicate to the public (and Court users before they attempt to login), this PR aims to add a JavaScript embed to the component responsible for serving the Public App of the site. It should not and does not embed this script in the Private App, which would expose the application and sensitive information to a third party.

We should consider whether or not we want to make this dynamic (so we can add more in the future if desirable). This could also play into using Analytics.


julialeague (approved):
LGTM! This checks off one of my boxes for issues that I noted during resiliency testing, so thank you!

PR #604, opened 2020-12-02
ustc-581: increase sqs visibility and retention

mmarcotte:
A recent deploy failed to the migration environment because we had adjusted the settings for the SQS queue in the console to have a retention time of 14 days. Unfortunately, this cleared out the queue. And the Lambda has been reconfigured to have a timeout time of 15 minutes. Terraform was [unable to update in place](https://app.circleci.com/pipelines/github/ustaxcourt/ef-cms/1353/workflows/d5bb12a2-a3ea-41a0-b3e3-62b47b4920f8/jobs/13120) the queue because its visibility timeout needs to be greater than the Lambda timeout. This fixes that, and helps keep our queue lasting longer.

### DIFF
--- a/web-api/terraform/template/sqs.tf
+++ b/web-api/terraform/template/sqs.tf
@@ -1,6 +1,6 @@
 resource "aws_sqs_queue" "migrate_legacy_documents_queue" {
   name                       = "migrate_legacy_documents_queue_${var.environment}"
-  visibility_timeout_seconds = "70"
+  visibility_timeout_seconds = "930"
 
   redrive_policy = jsonencode({
     deadLetterTargetArn = aws_sqs_queue.migrate_legacy_documents_dl_queue.arn
--- a/web-api/terraform/template/sqs.tf
+++ b/web-api/terraform/template/sqs.tf
@@ -1,6 +1,7 @@
 resource "aws_sqs_queue" "migrate_legacy_documents_queue" {
   name                       = "migrate_legacy_documents_queue_${var.environment}"
-  visibility_timeout_seconds = "930"
+  visibility_timeout_seconds = 930
+  message_retention_seconds  = 1209600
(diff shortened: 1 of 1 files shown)


adunkman (approved):
👍 Looks good! 

I had to look this up, so posting here in case others are in the same boat: 

> [**Amazon SQS visibility timeout**](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html)
> 
> When a consumer receives and processes a message from a queue, the message remains in the queue. Amazon SQS doesn't automatically delete the message. Because Amazon SQS is a distributed system, there's no guarantee that the consumer actually receives the message (for example, due to a connectivity issue, or due to an issue in the consumer application). Thus, the consumer must delete the message from the queue after receiving and processing it.
> 
> Immediately after a message is received, it remains in the queue. To prevent other consumers from processing the message again, Amazon SQS sets a visibility timeout, a period of time during which Amazon SQS prevents other consumers from receiving and processing the message.

One thing I didn’t see — does the Lambda also need timeout adjustment, since this is adjusting the queue settings on how much time it will give the Lambda but not the timeout that the Lambda will use? 

It wasn’t immediately apparent to me where that configuration lives, but this change is good to go regardless.

mmarcotte:
> One thing I didn’t see — does the Lambda also need timeout adjustment, since this is adjusting the queue settings on how much time it will give the Lambda but not the timeout that the Lambda will use?

https://github.com/ustaxcourt/ef-cms/blob/staging/web-api/terraform/template/legacy-document-migration.tf

We actually adjusted that in #574. And because I adjusted the SQS settings on this environment via the console, Terraform went about adjusting it, and that's why it complained about the discrepancy. 

Thanks!

Issue #616, opened 2020-12-04
Add checkbox for Admissions Clerk to bypass docket entry creation when editing Practitioner contact info


mmarcotte:
**Is your feature request related to a problem? Please describe.**
Sometimes due to data entry or missing information, we may need to adjust the contact information for a Practitioner without creating and serving documents. This would be in exceptional circumstances. The default should always be to generate the docket entry.

**Describe the solution you'd like**
Add a checkbox (and safeguard) to the page where the Admissions Clerk edits a Practitioner's contact information to use the flag that #591 introduced in the subsequent API call. 

**Describe alternatives you've considered**
We can make the change via an API call directly. It would just be nice to have via the UI.

JessicaMarine:
Don't think this is necessary.  If it becomes necessary, can add a story to Flexion's backlog at that time.

Issue #591, opened 2020-12-01
IRS Practitioners need to have addresses updated


mmarcotte:
None of the IRS practitioners who had an Address3 field received have this information in the system. We will need to figure out the best way to add this information in without generating documents or sending service emails.

*Acceptance Criteria*
- [ ] Update practitioner's addresses with their correct contact information without creating docket entries

mmarcotte:
When working on testing this in Migration I encountered this blocker. While I can update the practitioner address record, the logic to update each case is tricky.

Right now the logic only looks for open cases or closed cases within the last 6 months to update. My belief is that this logic is to figure out whether or not to generate and serve the change of address document. However, would we still want to update the address for the practitioner on that case? Or is it okay to leave a stale record there with old address information?

PR #617, opened 2020-12-04
Add optional bypass  for Docket Entry creation when modifying Practitioner contact info

mmarcotte:
As #591  requires, we need to be able to adjust some practitioner's contact information without generating a new document, serving that document, and Docket Entry. The solution here is only in the API call, and there is no UI change for this. (Have logged https://github.com/ustaxcourt/ef-cms/issues/616 to consider this in the future)

The solution introduces an optional `bypassDocketEntry` property to the JSON that gets passed when we update a Practitioner. If it is not set or false-y, then the document will get created, served, and added to the docket record. If it's truth-y, the document will not get created, but the address will get updated for the user and the lookup records for each case they are associated with.

The main work was simply moving all of the logic that generates, serves, and adds to the record into its own function that gets called if the bypass toggle is false.


shared/src/business/useCases/users/generateChangeOfAddress.js
```diff
@@ -53,173 +53,41 @@ exports.generateChangeOfAddress = async ({
...
-      const irsPractitioner = caseEntity.irsPractitioners.find(
-        practitioner => practitioner.userId === user.userId,
-      );
-      if (irsPractitioner) {
-        oldData = clone(irsPractitioner.contact);
-        irsPractitioner.contact = contactInfo;
-        irsPractitioner.name = practitionerName;
+      if (!practitionerObject) {
+        throw new Error(
+          `Could not find user|${user.userId} barNumber: ${user.barNumber} on ${docketNumber}`,
```
adunkman, line 65: Thank you for this!

adunkman (approved):
This looks great!

Issue #621, opened 2020-12-07
Support pagination for Elasticsearch queries that yield greater than 10,000 hits


mmarcotte:
When working on #591 , I noticed that we may need to improve pagination logic for our Elasticsearch queries.

We cannot use the `from` parameter to paginate Elasticsearch queries that have more than 10,000 hits ([documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/paginate-search-results.html)). We will need to "use the [search_after parameter](https://www.elastic.co/guide/en/elasticsearch/reference/current/paginate-search-results.html#search-after) with a point in time (PIT)." 

At the moment I don't suspect this is a problem for the operation getCasesByUserId, as the most in the DB right now is about 6,000 for practitioners. However, it's quite possible that some day a practitioner could surpass that 10,000 threshold. We would need a better way to get ALL of the results to perform operations like change of address.

mmarcotte:
Use this: https://www.elastic.co/guide/en/elasticsearch/reference/current/scroll-api.html

PR #622, opened 2020-12-07
Move logic from ES Query to function that generates the docket entry.

mmarcotte:
As I was testing out the previous fix for updating practitioners in the Migration environment, I encountered an issue where the code would only update practitioner's contact information on the first 5,000 cases that were open or closed in the last six months. Some IRS practitioners are associated with more than 5,000 cases, and we still need all cases updated with correct information. However, we only want to create docket entries for open cases or cases that have closed within the last 6 months. This work moves that logic from the query of users associated with a case to the function call that creates the docket entry.

Furthermore, this fixes a bug where Admissions Clerks would be unable to update a practitioner who is associated with many cases. It moves the handler for the API call to the `/async/` Lambda that has a 15 minute timeout, and it utilizes the websocket messaging to inform the user of the progress just as it does when a user updates their own address. In doing so, I had to introduce new "admin" versions of these messages and sequences since their behavior was different. (Practitioners would get transported to the dashboard after updating their address whereas an Admissions Clerk would want to remain on the current page.)

There is one area of concern, and that is the current max size of the Elasticsearch query. In order to paginate beyond the first 10,000, we need to adopt a more involved pagination strategy, and so I logged https://github.com/ustaxcourt/ef-cms/issues/621 to track that. For now, the most cases a practitioner is associated with is around 6,000.


shared/src/business/useCases/users/generateChangeOfAddress.test.js
```diff
@@ -720,4 +722,115 @@ describe('generateChangeOfAddress', () => {
...
+    ).toHaveBeenCalled();
+    expect(docketEntryForNoticeOfChangeOfAddress).toBeDefined();
+    expect(cases).toMatchObject([
+      expect.objectContaining({ docketNumber: MOCK_CASE.docketNumber }),
+    ]);
+    // We still want to update our case!
+    expect(
+      applicationContext.getPersistenceGateway().updateCase,
+    ).toHaveBeenCalled();
+  });
```
julialeague, line 854: Can you add a test for the `isOpen` case too?

shared/src/business/useCases/users/generateChangeOfAddress.test.js
```diff
@@ -720,4 +722,115 @@ describe('generateChangeOfAddress', () => {
...
+    });
+
+    const docketEntryForNoticeOfChangeOfAddress = cases[0].docketEntries.find(
+      entry => entry.documentTitle.includes('Notice of Change'),
+    );
+
+    expect(
+      applicationContext.getDocumentGenerators().changeOfAddress,
+    ).toHaveBeenCalled();
+    expect(docketEntryForNoticeOfChangeOfAddress).toBeDefined();
```
julialeague, line 843: should this check if the `saveWorkItemForNonPaper` function was called also?

julialeague (approved):
Outside of those couple test-related comments, looks great to me!

 santa kitties approved!

shared/src/business/useCases/users/generateChangeOfAddress.test.js
mmarcotte: Well, I think all of the other tests all run on open cases as `MOCK_CASE` comes in with `CASE_STATUS_TYPES.new` And those cases wouldn't have a `closedDate`. So effectively the isOpen logic should be tested.

shared/src/business/useCases/users/generateChangeOfAddress.test.js
mmarcotte: Again the default `MOCK_CASE` has everyone on the case be `SI_ELECTRONIC`. So, that function would not have been called. For good measure, though, I have made the case in this test have the secondary contact who requires paper service and reintroduced this expect.

Now you have me wondering about that name of that function. It's literally doing the opposite... saving a work item for a case that requires paper service. I've logged a refactoring ticket to discuss further.

PR #633, opened 2020-12-10
Improve Logging visibility for Blue/Green Migrations

mmarcotte:
In order to have a better understanding of how a blue/green migration is performing, this PR introduces the `NODE_ENV` environmental variable to the Lambdas that perform the migration. This way anything logged using the Application logger will get indexed into Kibana. In doing so, I replaced the `console.log(...)` statements to use the Application logger and introduced a couple more logs.

### DIFF
--- a/web-api/migration-terraform/main/migration-segments.tf
+++ b/web-api/migration-terraform/main/migration-segments.tf
@@ -18,10 +18,11 @@ resource "aws_lambda_function" "migration_segments_lambda" {
 
   environment {
     variables = {
-      SOURCE_TABLE       = var.source_table
-      ENVIRONMENT        = var.environment
-      SEGMENTS_QUEUE_URL = aws_sqs_queue.migration_segments_queue.id
       DESTINATION_TABLE  = var.destination_table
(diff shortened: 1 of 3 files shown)


web-api/migration-terraform/main/migration-segments.tf
```diff
@@ -18,10 +18,11 @@ resource "aws_lambda_function" "migration_segments_lambda" {
 
   environment {
     variables = {
-      SOURCE_TABLE       = var.source_table
+      DESTINATION_TABLE  = var.destination_table
       ENVIRONMENT        = var.environment
+      NODE_ENV           = "production"
       SEGMENTS_QUEUE_URL = aws_sqs_queue.migration_segments_queue.id
-      DESTINATION_TABLE  = var.destination_table
+      SOURCE_TABLE       = var.source_table
```
mmarcotte, line 25: Just alphabetized for housekeeping while I was there.

adunkman (approved):
Looking great!

julialeague (approved):
Looks good!

PR #639, opened 2020-12-10
Fix wonky react sequence for updating a practitioner

mmarcotte:
Due to the move to the async handler, the sequences triggered by websockets were competing with the sequences performing the API call. This moves the navigation to `practitioner-detail` and the success messaging to the sequences triggered by websockets and pulls the practitioner information out of state.


web-client/src/presenter/actions/navigateToPractitionerDetailAction.js
```diff
@@ -6,8 +8,13 @@
  * @param {object} providers.props the cerebral props object
  * @returns {Promise} async action
  */
-export const navigateToPractitionerDetailAction = async ({ props, router }) => {
-  const { barNumber } = props;
+export const navigateToPractitionerDetailAction = async ({
+  get,
```
julialeague, line 13: Can you add doc for this parameter?

julialeague (approved):
Looks good other than the tiny doc issue!

PR #644, opened 2020-12-11
Add stage to client

mmarcotte:
In order to manage various UI/UX functionality in the React App, this introduces the `CLIENT_STAGE` environment variable that we can specify in CircleCI. Since the Court uses contexts in CircleCI, we can specify this exclusively for `prod`. Otherwise, the app will default to 'local' if it is not set. If we need to introduce more stages (perhaps Test env could be a "training" stage with some helper tips), we'll figure out a way to support that by making use of this variable.


matthopson (approved):

adunkman (approved):
Future Us will thank you for the documentation update! 🙏 

You and me, both

matthopson:
I was just thinking, I would believe that David Rose is Johnny Rose's actual son... turns out....

PR #646, opened 2020-12-11
Ensure CLIENT_STAGE env var is available

mmarcotte:
#644 helped get us there, but `CLIENT_STAGE` wasn't available to the docker container.


adunkman (approved):
Oops! 👍

Issue #581, opened 2020-11-27
Remove race condition when scraping PDFs.


mmarcotte:
@mark-meyer found a race condition in the process for scraping PDFs. The script that attempts to do that tries to load and save an entire case. Meanwhile other operations are running concurrently working with the same case, and they overwrite the previously saved information for the case in Dynamo, and the saved record gets lost. This process could also explain why we saw a high volume of writes in Dynamo when performing this operation. 

Upon further examination, these writes may be entirely unnecessary. All they are doing is saving a system generated UUID after creating a file with that UUID in the same bucket where the pdfs are stored. We already have a UUID, and we could re-use that for the scraped contents if we simply stored it in another location. It would be more efficient and less confusing as the UUID for the docketEntryId would match up exactly for the PDF and the file containing the scraped contents. 

One final thing is to remove files if there is a failure because previous attempts that encountered the DDB throttling probably created a bunch of files as they scraped and saved the contents, but did not clean them up upon DDB failure.

Tasks
- [x] Make sure we have long lasting retention on the SQS queue
- [ ] Restage the queue in the migration environment
- [ ] Create bucket for Document Contents in Terraform (or use a .txt extension in the same bucket)
- [ ] Allow the save to storage allow for mimetime specification (right now it forces PDF)
- [ ] Copy (or refactor) Lambda that saves the Document to the Documents bucket, to save the Document Contents to its new location
- [ ] Rather than write the case back to DynamoDB, attempt to index the document on behalf of the case in ES.
- [ ] refactor application to remove `documentContentsID` and look at the right spot for ☝️

mmarcotte:
Need to run on Migration Environment and Prod

mark-meyer:
From the code perspective it looks like we would be getting rid of `documentContentsId`. Any place that currently uses `documentContentsId` would instead use `docketEntryId` and simply reference a different S3 bucket. 

The question is, can we entirely avoid writing to DDB when we scrape the PDFs?  Are there places in the code where ef-cms needs to know *if* the document contents currently exist without looking in S3? For example, there is a code in `/shared/src/business/useCases/processStreamRecordsInteractor.js` that tests for this:

```
if (fullDocketEntry.documentContentsId) {
        // TODO: for performance, we should not re-index doc contents if we do not have to (use a contents hash?)
        try {
            ...
```

[...]

PR #652, opened 2020-12-14
Fix PDF Scraping and improve logging

mmarcotte:
To help finish off #581 , we needed to validate the docketEntry before attempting to save it to persistence. 

Additionally, to help troubleshoot any issues that arise, this PR improves the logging and functionality for the Lambda that performs the scraping.

- introduces adds `NODE_ENV` to the Lambda so logs are properly formatted for Kibana
- uses `applicationContext.logger.error` to format failures so they can be observed in Kibana
- subscribes to the log group for the scraping (this has been run)
- adds logging for when the process succeeded
- does not attempt retries if 
  1. the item could not be found in S3 or 
  1. the docketEntryId doesn't exist


shared/src/business/useCases/migration/parseLegacyDocumentsInteractor.js
```diff
@@ -75,10 +76,17 @@ exports.parseLegacyDocumentsInteractor = async ({
 
   foundDocketEntry.documentContentsId = documentContentsId;
 
+  const validatedDocketEntry = new DocketEntry(foundDocketEntry, {
+    applicationContext,
+  }).validate();
```
adunkman, line 81: Oops, did I cause a bug before? I didn’t see how this could not pass validation (since we’re just adding the contents) but please educate me if I made a mistake! 🐑

web-api/terraform/template/lambdas/legacy-documents-migration.js
```diff
@@ -11,11 +11,28 @@ exports.handler = async event => {
...
+    // try again if error does not have one of the following
+    if (
+      !(
+        err.message.includes('Docket entry document not found in S3.') ||
+        err.message.includes('Docket entry not found.')
+      )
+    ) {
+      throw err;
+    }
+  }
```
adunkman, line 35: Nice!

adunkman (approved):
Looks great!

shared/src/business/useCases/migration/parseLegacyDocumentsInteractor.js
mmarcotte: I don't think you made a mistake, but I don't think you had a good way to test it. And so I learned something to trying to test this.

The persistence gateway was throwing an error that the DocketEntry hadn't been validated, and so it put the brakes on the process before updating the DB. All I had to do was validate it first, and then the save was successful. In looking closer at that validation process, I'm a little curious about [what other effects that has on the entity](https://github.com/ustaxcourt/ef-cms/blob/staging/shared/src/business/entities/DocketEntry.js#L133).

PR #668, opened 2020-12-16
Increase size of EBS size for Test, Mig, and Prod

mmarcotte:
With pdfs scraped, we need more space in our cluster.


adunkman (approved):
Confirmed these nodes are `m5.large` with a maximum EBS volume size of `512`.

Issue #669, opened 2020-12-17
Only index Orders and Opinions into ES


mmarcotte:
In order to improve Advanced Document Search performance as well as reindexing times and resources, add logic to ONLY index Orders and Opinions. We will still scrape the contents of all documents, but let's only index what we need (for now).

PR #670, opened 2020-12-17
Elasticsearch optimization: only index contents of Orders and Opinions

mmarcotte:
To help satisfy #669 , this additional check only indexes the `documentContents` of Orders and Opinions into the `efcms-docket-entry` index. Other docket entries will still be indexed, but their `documentContents` will be ignored as they are not used. One test was changed and two introduced to test this functionality.


adunkman (approved):
Code looks good to me! 👍 Thanks for the tests!

PR #671, opened 2020-12-17
7343 - cherry picking commits to remove documentContents

codyseibert:
- removes documentContents from the _source array so that the full text of the PDF isn't returned to the front end (which we think is causing the UI / API endpoint to crash because it's over 6mb)
- filters out certain contact information from the api response that might include sensitive addresses and contact info of practitioners and contacts.


mmarcotte (approved):
Awesome, thanks for helping us get to the finish line
racecar

Issue #393, opened 2020-09-03
Court: Ability to see "bounced" service emails


JessicaMarine:
As a court user, so that I can tell if a party did not receive their service email, I need the ability to see if any service emails were bounced back to the server as undeliverable.

## Acceptance Criteria
* Court user can see any service emails that were returned/bounced back to the server as undeliverable. 

## Tasks

- [x] Test in exp1
- [x] Merge in #678 
- [x] Verify bounced email `bounced-service-emails@ustaxcourt.gov` in SES
- [x] Specify Environment Variable
- [x] Deploy to Prod

vickimcfadden:
Docket section needs to have access to this bounce service so they can re-serve or make changes to contact info.

adunkman:
Docket needs the ability to monitor bounce-back emails.

adunkman:
Mike has a crude way to see failed email addresses, by parsing S3 email records — but it’s hard to see which email messages themselves bounced.

vickimcfadden:
Can see them in S3 bucket now. How do we want to see emails bounce? What would docket do with this? What data do they need?Kicking the tires: 1) Might be able to set up an email address that a list of individuals can monitor. 2) product design to show email bounce-back within Dawson. etc etc

vickimcfadden:
Have emails bounce to an email address (possibly in MS365).

mmarcotte:
dawson.bounced.email@ustaxcourt.gov

mmarcotte:
Looks like we already have an email account to use: `bounced-service-emails@ustaxcourt.gov` Will be sending email there.

mark-meyer:
SES allows bounced email notifications [to be sent to an SNS topic](https://docs.aws.amazon.com/ses/latest/DeveloperGuide/configure-sns-notifications.html). You can subscribe to SNS topics in many ways, including via email, which will result in bounce notification sending an email like this to whomever is subscribed to the SNS stream:

`
[...]

PR #678, opened 2020-12-18
Add support for specifying bounced service emails

mmarcotte:
To get #393 working, this introduces a new environmental variable to which we will send bounced emails.

Added tasks to the original ticket to complete the ticket.


shared/src/dispatchers/ses/sendBulkTemplatedEmail.js
```diff
@@ -38,6 +38,8 @@ exports.sendBulkTemplatedEmail = async ({
         },
         ReplacementTemplateData: JSON.stringify(destination.templateData),
       })),
+      ReturnPath:
```
julialeague, line 41: Why do we require the `BOUNCED_EMAIL_RECIPIENT` for deployment if there's a default?

julialeague (commented):
Outside of my question (which may just be a misunderstanding), this looks good!

julialeague (approved):

shared/src/dispatchers/ses/sendBulkTemplatedEmail.js
mmarcotte: Good call, I just removed the requirement.

PR #694, opened 2020-12-21
Add multi-field mapping on messages.

adunkman:
This indexes log messages in two ways: 

- It leaves the existing full-text index. This gives us the ability to search the log messages with free-text search.
- It adds a keyword index, which gives us the ability to aggregate based on the log message. For error dashboards, we can use the error message to group similar errors together. 

This change has already been applied to the staging AWS account.


mmarcotte (approved):
Cool!

cool

Issue #697, opened 2020-12-22
Error preventing Dynamsoft in Production


mmarcotte:
When trying to load a page that uses Dynamsoft, I'm presented with this error:

```
getScannerInterface.js:74 Refused to load the script 'https://dynamsoft-lib.dawson.ustaxcourt.gov/dynamic-web-twain-sdk-14.3.1/dynamsoft.webtwain.initiate.js' because it violates the following Content Security Policy directive: "script-src 'self' 'unsafe-inline' https://dynamsoft-lib.stg.ef-cms.ustaxcourt.gov https://lynmjtcq5px1.statuspage.io resource://pdf.js". Note that 'script-src-elem' was not explicitly set, so 'script-src' is used as a fallback.
```

Looks like we need to add it to the CSP

Issue #696, opened 2020-12-22
Disable extra Dynamsoft EC2 instances


mmarcotte:
We are currently only using the on on Staging, but we are running ec2 instances effectively doing nothing for:

- irs
- mig
- test
- dev
- exp1

adunkman:
These may be account-level resources instead of environment-specific.

PR #698, opened 2020-12-22
Add Dynamsoft Production URL to CSP

mmarcotte:
In order to resolve the Content Security Policy directive error, and solve #697, this introduces `dynamsoft-lib.dawson.ustaxcourt.gov` to the CSP. Hopefully this helps resolve our issues using the scanning software in production.


adunkman (commented):
Retroactive 👍 — I see you’ve also filed #696. These seem to be account-level (or even less than account-level) resources, so we might want to move them there longer-term.

PR #703, opened 2020-12-23
increase es resources for production

mmarcotte:
To help over-prepare for launch, we are increasing the resources of our ES cluster.


adunkman (approved):
Biggie jiving

PR #704, opened 2020-12-23
Increase async lambda’s memory to 5GB.

### LINKED ISSUE #661: Check to see how larger PDFs perform
We have some real large PDFs. Let's get a basis for how they perform with DAWSON.

- Large PDFs are ones that are between 1-2GB in size.

Understand how a file of this size performs when: 

- Serving the file (a coversheet is attached)
- Downloading a trial session (it is included in a ZIP archive)

If it does not perform, explore adjusting resources available to the lambda functions used.

adunkman:
Closes #661.

Creating zip archives with trial sessions requires enough memory to hold all documents in memory, with the largest of documents being 2.14gb.


mmarcotte (approved):
brain

PR #718, opened 2020-12-29
consider record.isFileAttached when adding link

mark-meyer:
Consider `record.isFileAttached` when calculating if a docket entry should be linked. 
Add `record.isFileAttached` missing in test fixtures.


mmarcotte (approved):

mmarcotte:
Thanks @mark-meyer !

i can see

PR #759, opened 2021-01-12
Tune down es one step for production

mmarcotte:
Based on observed usage, our ES cluster is over-provisioned. We do not need the firepower that we are currently paying for. This tunes it down one step to `xlarge` from `2xlarge`.


adunkman (approved):
Looks great! Happy that we have visibility into our usage patterns enough to see when we’re over-prepared.  👍

PR #777, opened 2021-01-15
Increase signal-to-noise ratio on system health alarms.

### LINKED ISSUE #712: Fix Dynamsoft Health Check for Prod
Currently the Dynamsoft check is [hardcoded for the Staging environment](https://github.com/ustaxcourt/ef-cms/blob/staging/shared/src/business/useCases/health/getHealthCheckInteractor.js#L58). We need to make this work for the Production environment that uses its own set of URLs. Additionally, these are frequently returning an unhealthy status. We should investigate why we receive so many 503 errors.

Tasks: 

- [x] Point production application to the production EC2 instance. (#771)
- [ ] Understand why staging Dynamsoft EC2 instance reports 503s.

adunkman:
Closes #712.

We have seen instability in our alarms for the system health endpoint due to the Dynamsoft EC2 host failing — there were a few contributing factors, but since we do not have user reports of failure, the remaining issues appear to be intermittent network issues. Continuing onward with this theory until we receive user reported issues. 

This PR applies two changes: 

1. **Specify regions to deploy the Health Check.** This results in a significant [overall reduction of health pings](https://github.com/ustaxcourt/ef-cms/issues/712#issuecomment-760954580). 
    - The default regions are:
      1. US East (N. Virginia)
      2. US West (N. California)
      3. US West (Oregon)
      4. EU (Ireland)
      5. Asia Pacific (Singapore)
      6. Asia Pacific (Sydney)
      7. Asia Pacific (Tokyo)
      8. South America (São Paulo)
    - The new regions are (minimum 3 required):
      1. US East (N. Virginia)
      2. US West (N. California)
      3. US West (Oregon)
2. **Increases the failure threshold to 3.** The alarm will now be triggered after one minute and 30 seconds of observed downtime in any of the above regions. 

Confirmed the changes on `exp1`.

### DIFF
--- a/web-client/terraform/main/main.tf
+++ b/web-client/terraform/main/main.tf
@@ -170,6 +170,7 @@ resource "aws_route53_health_check" "public_ui_health_check" {
   resource_path     = "/"
   failure_threshold = "2"
   request_interval  = "30"
+  regions           = [ "us-east-1", "us-west-1", "us-west-2" ] # Minimum of three regions required
 }
 
 resource "aws_cloudwatch_metric_alarm" "ui_health_check" {
@@ -198,6 +199,7 @@ resource "aws_route53_health_check" "ui_health_check" {
   resource_path     = "/"
   failure_threshold = "2"
   request_interval  = "30"
+  regions           = [ "us-east-1", "us-west-1", "us-west-2" ] # Minimum of three regions required
 }
 
 resource "aws_cloudwatch_metric_alarm" "status_health_check" {
@@ -228,4 +230,5 @@ resource "aws_route53_health_check" "status_health_check" {
   request_interval   = "30"
   invert_healthcheck = true
   search_string      = "false" # Search for any JSON property returning "false"
+  regions            = [ "us-east-1", "us-west-1", "us-west-2" ] # Minimum of three regions required
 }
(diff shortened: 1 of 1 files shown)


julialeague (approved):

web-client/terraform/main/main.tf
```diff
@@ -168,8 +168,9 @@ resource "aws_route53_health_check" "public_ui_health_check" {
   port              = 443
   type              = "HTTPS"
   resource_path     = "/"
-  failure_threshold = "2"
+  failure_threshold = "3"
   request_interval  = "30"
+  regions           = [ "us-east-1", "us-west-1", "us-west-2" ] # Minimum of three regions required
```
mmarcotte, line 173: Why not continue using the default set of regions that AWS uses? They're global... Just curious.

adunkman: Thanks for the question! It’s very confusing, and I just learned this! By overriding these regions, we reduce health checking traffic by 62%. 

When configuring Route53 health checks, you’re actually configuring health check workers _per AWS datacenter_. I assume this is so you can regionally react based on latency and scale per-data center. 

The result is that when you use the default regions, you’re actually configuring workers in **8 data centers** which results in approximately 15x higher traffic than configured here. I measured this observed traffic at 1,779 requests per hour per environment.

By overriding to custom regions here, we configure workers in **3 data centers** instead. This results in observed traffic of 671 requests per hour per environment. 

You can follow along with my learning on [the original issue](https://github.com/ustaxcourt/ef-cms/issues/712#issuecomment-760258118), including the graphs used to calculate the observed traffic.

mmarcotte (approved):

PR #800, opened 2021-01-22
Introduce some Elasticsearch Tooling

mmarcotte:
To help consolidate code, and open up tooling to more environments, this PR moves some code from the `test-elasticsearch.js` script into `web-api/elasticsearch/client.js`. This lets us create a client that we can use to execute queries on our deployed ES clusters. This will be handy for tuning, testing, and making better use of ES. 

The scripts introduced rely on two arguments to inform the script where to go looking. The first is the environment (e.g, `mig`), and the second is the version (`alpha` or `beta`):

```
node ./web-api/elasticsearch/get-users.js mig beta
```

This will help configure the correct client to use when performing searches. You need to have IAM privileges to read from the cluster for these scripts to work.


mark-meyer (approved):
At some point we might consider moving some of these reporting scripts into admin-scripts and organize other similar utilities we have in other places.

shared/admin-tools/elasticsearch/health-migration.js
```diff
@@ -0,0 +1,74 @@
...
+
+  const totals = {
+    alpha: 0,
+    beta: 0,
+  };
+
+  console.log(`## ${environmentName} Index Summary`);
+  console.log('+----+----+----+----+');
+  console.log('| Index | Alpha | Beta | Difference |');
+  console.log('+----+----+----+----+');
```
kkoskelin: Cool.  In a future version, you could even experiment with this: https://developer.mozilla.org/en-US/docs/Web/API/Console/table

kkoskelin (approved):

shared/admin-tools/elasticsearch/health-migration.js
mmarcotte: neat-o. done!

kkoskelin (approved):

Issue #773, opened 2021-01-14
Investigate irregularities concerning E-Service to IRS


mmarcotte:
It looks like DAWSON has tried to serve multiple copies of the same document:

- Docket No: 7118-19
- Docket Entry: 30
- Docket Code: MISCL

appeared to be served four times (12/29) with Document Ids:

- 9c21f708-8d76-4fbc-93d0-d4fd1fd6c2cb
- 1d7065a2-cf11-4fc5-adf0-b21d3de6c370
- 1a1a8637-6f2d-413b-9f7c-2ad7a4502f60
- 1d7065a2-cf11-4fc5-adf0-b21d3de6c370

Second and fourth appear to have identical Document Ids but came in at different times.   All four creating errors on our end that do not return the pdf document.

Similar with respect to:

- Docket No: 11559-20
- Docket Entry: 5
- Docket Code: DISC

We appeared to receive this at least two times (12/28) with Document Ids:

- 41613810-06d8-414a-aacb-efe8553a98f5
-  7a2f3c37-4a86-4371-a4c6-df966bb2b8e2

PR #808, opened 2021-01-25
Only send service email after validation

mmarcotte:
In order to address #773  -- due to a bug, which has since been addressed, sometimes filing a docket entry generates an error with validation. However, the system was still serving that document before validation. This adjusts that order of operations so it only sends after validation has occurred.

It also includes a bit of refactoring to simplify the logic.


shared/src/business/useCases/docketEntry/fileDocketEntryInteractor.test.js
```diff
@@ -268,4 +268,34 @@ describe('fileDocketEntryInteractor', () => {
         .deleteCaseTrialSortMappingRecords,
     ).toBeCalled();
   });
+
+  it('does not send the service email if an error occurs while updating the case', async () => {
+    applicationContext
+      .getPersistenceGateway()
+      .updateCase.mockRejectedValue(new Error('bad!'));
```
adunkman, line 275: Bad! :joy:

adunkman (approved):
Looks good to me — thanks for the refactoring as well, looks great!

Issue #682, opened 2020-12-18
Export database into different account — AWS Glue


adunkman:
Goal: Enable Flexion to test in a production-like data environment, including the contents of documents. 

TODO: 

- [x] Consider whether we want to create a stream so other environment's data is always up-to-date. 
- [x] Determine approach: should we migrate from DynamoDB to DynamoDB, or use the migration scripts and endpoints? 
- [x] Identify data that should not be migrated or redacted (@JessicaMarine is working on this)
- [x] Perform the migration
- [x] Document how to do it again in the future
- [x] Sync PDFs
- [ ] Create a script to assume the identity of another user in the system

adunkman:
May make sense to migrate _all_ PDFs with contents. Prioritized after launch.

adunkman:
#793 is related.

mark-meyer:
Mike and I talked about the first todo:  (`Consider whether we want to create a stream so other environment's data is always up-to-date`) and realized we probably don't want to do this. Day-to-day use of the migration environment may include writing/deleting/updating rows which could cause validation issues if that data no longer aligns with data coming from the production DB. This will mean periodically syncing the migration environment as needed.

PR #815, opened 2021-01-28
Admin User Scripts for Glue Project

mmarcotte:
In order to finish up #682 , these scripts help developers who have the required AWS privileges lookup and assume the identity of certain users in the designated environment. There are two scripts, which have their own documentation for usage, that allow users with access to perform ES queries as well as certain Cognito API calls. These scripts are included in `package.json`.

```bash
$ npm run admin:lookup-user docketClerk "Remington"
$ npm run admin:become-user uuiduuid-uuid-uuid-uuiduuiduuid
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -152,6 +152,7 @@
   },
   "scripts": {
     "admin:lookup-user": "node shared/admin-tools/glue/lookup-user.js",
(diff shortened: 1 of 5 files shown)


adunkman (approved):
Looks great to me!

PR #821, opened 2021-02-02
Mmeyer elastic mappings

mark-meyer:
This changes the mappings of many of the Elasticsearch indexes to move fields that should not be analyzed as full-text,(which causes both on original input and query text to be tokenization), in preference of `keyword` fields which are stored whole and intended for exact match searches. 

Currently, fields that are UUIDs are stored as text fields. This means they are entered in the Elasticsearch index as a bunch of tokens. For example, a field like `docket-entry|5bd2f4eb-e08a-41e4-8d18-13b9ffd4514c` will be stored in the index as seven  independent tokens: `['docket', 'entry', '5bd2f4eb', 'e08a', 41e4', '8d18', '13b9ffd4514c']`. Searching for this with a `match` query will find the original document, but it is expensive (it requires seven term queries under the hood) and can result in results that do not exactly match the uuid -- for example if only a couple tokens match. Changing the default operator to `and` reduces the chances of matching incorrect documents (it will require all tokens, but does not necessarily in the same order), but it does not improve the cost.

By changing these to `keyword` fields we will only add a singe token token to the index — the exact match:  `docket-entry|5bd2f4eb-e08a-41e4-8d18-13b9ffd4514c`. This has a side-effect of not breaking matches like `{"match": {"pk.S": "docket-entry|"}}` because `docket` and `entry` are no longer independent terms in the index. It seems like this can be avoided by matching on `entityName` when/if this kind of match is needed.

This PR changes a few other fields to keyword fields such as event codes, and case statuses. Like UUIDs, we will never need to match individual tokens from these fields, i.e. we *never* want a search for case status `General Docket - At Issue (Ready for Trial)` by searching for the token `general` and we don't want `General Docket - Not at Issue` to ever match. In other words we are never asking for a score — "how well does the query match?"; we are always asking for a yes/no — "does this query match?"

A couple of the fields are less clear, for example: `preferredTrialCity`. This changes `preferredTrialCity` to a keyword because the exact values are enforced by the Case schema. However this may be incorrect if we expect user input such as `"reno, nv"` to match the value `"Reno, Nevada"`. 

In addition to the changes to the mapping types, this removes many matches in queries that appear unnecessary such as matching for `match: { 'sk.S': 'case|' }` in the `efcms-case` index. It appears *all* cases in the index match this query. One place this is a little unclear is the `efcms-work-item` index. Queries such as `'pk.S':  "section|docket"` appear to always have an `sk.S` of the form `'work-item|'` so this match has been removed. Reviewers should take a critical look at these.

### DIFF
--- a/shared/src/persistence/elasticsearch/advancedDocumentSearch.js
+++ b/shared/src/persistence/elasticsearch/advancedDocumentSearch.js
@@ -46,16 +46,8 @@ exports.advancedDocumentSearch = async ({
   const docketEntryQueryParams = [
     {
       bool: {
-        must_not: [
-          {
-            term: { 'isStricken.BOOL': true },
-          },
-        ],
-        should: documentEventCodes.map(eventCode => ({
-          term: {
-            'eventCode.S': eventCode,
-          },
-        })),
+        must: [{ terms: { 'eventCode.S': documentEventCodes } }],
+        must_not: [{ term: { 'isStricken.BOOL': true } }],
       },
     },
   ];
--- a/shared/src/persistence/elasticsearch/advancedDocumentSearch.js
+++ b/shared/src/persistence/elasticsearch/advancedDocumentSearch.js
@@ -52,7 +52,7 @@ exports.advancedDocumentSearch = async ({
           },
         ],
         should: documentEventCodes.map(eventCode => ({
-          match: {
+          term: {
             'eventCode.S': eventCode,
           },
         })),
@@ -92,7 +92,7 @@ exports.advancedDocumentSearch = async ({
 
   if (docketNumber) {
     caseQueryParams.has_parent.query.bool.must = {
-      match: { 'docketNumber.S': { operator: 'and', query: docketNumber } },
+      term: { 'docketNumber.S': docketNumber },
     };
   } else if (caseTitleOrPetitioner) {
     caseQueryParams.has_parent.query.bool.must = {
@@ -142,12 +142,7 @@ exports.advancedDocumentSearch = async ({
 
   if (opinionType) {
     docketEntryQueryParams.push({
-      match: {
-        'documentType.S': {
-          operator: 'and',
-          query: opinionType,
-        },
-      },
+      term: { 'documentType.S': opinionType },
     });
   }
 
@@ -186,8 +181,7 @@ exports.advancedDocumentSearch = async ({
       query: {
         bool: {
           must: [
-            { match: { 'pk.S': 'case|' } },
-            { match: { 'sk.S': 'docket-entry|' } },
+            { term: { 'entityName.S': 'DocketEntry' } },
             {
               exists: {
                 field: 'servedAt',
--- a/shared/src/persistence/elasticsearch/advancedDocumentSearch.test.js
+++ b/shared/src/persistence/elasticsearch/advancedDocumentSearch.test.js
@@ -37,8 +37,7 @@ describe('advancedDocumentSearch', () => {
   });
 
   const orderQueryParams = [
-    { match: { 'pk.S': 'case|' } },
-    { match: { 'sk.S': 'docket-entry|' } },
+    { term: { 'entityName.S': 'DocketEntry' } },
     {
       exists: {
         field: 'servedAt',
@@ -53,12 +52,12 @@ describe('advancedDocumentSearch', () => {
         ],
         should: [
           {
-            match: {
+            term: {
               'eventCode.S': orderEventCodes[0],
             },
           },
           {
-            match: {
+            term: {
               'eventCode.S': orderEventCodes[1],
             },
           },
@@ -68,8 +67,7 @@ describe('advancedDocumentSearch', () => {
   ];
 
   const opinionQueryParams = [
-    { match: { 'pk.S': 'case|' } },
-    { match: { 'sk.S': 'docket-entry|' } },
+    { term: { 'entityName.S': 'DocketEntry' } },
     {
       exists: {
         field: 'servedAt',
@@ -84,12 +82,12 @@ describe('advancedDocumentSearch', () => {
         ],
         should: [
           {
-            match: {
+            term: {
               'eventCode.S': opinionEventCodes[0],
             },
           },
           {
-            match: {
+            term: {
               'eventCode.S': opinionEventCodes[1],
             },
           },
@@ -133,8 +131,8 @@ describe('advancedDocumentSearch', () => {
 
     if (docketNumber) {
       query.bool.must = {
-        match: {
-          'docketNumber.S': { operator: 'and', query: docketNumber },
+        term: {
+          'docketNumber.S': docketNumber,
         },
       };
     }
@@ -257,15 +255,12 @@ describe('advancedDocumentSearch', () => {
       ...orderQueryParams,
       getCaseMappingQueryParams(), // match all parents
       {
-        match: {
-          'documentType.S': {
-            operator: 'and',
-            query: 'Summary Opinion',
-          },
+        term: {
+          'documentType.S': 'Summary Opinion',
         },
       },
     ];
-    expectation[4].has_parent.query.bool.must_not = [
+    expectation[3].has_parent.query.bool.must_not = [
       { term: { 'isSealed.BOOL': true } },
     ];
 
@@ -286,15 +281,12 @@ describe('advancedDocumentSearch', () => {
       ...orderQueryParams,
       getCaseMappingQueryParams(), // match all parents
       {
-        match: {
-          'documentType.S': {
-            operator: 'and',
-            query: 'Summary Opinion',
-          },
+        term: {
+          'documentType.S': 'Summary Opinion',
         },
       },
     ];
-    expectation[4].has_parent.query.bool.must_not = [
+    expectation[3].has_parent.query.bool.must_not = [
       { term: { 'isSealed.BOOL': true } },
     ];
     expect(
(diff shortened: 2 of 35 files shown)


kkoskelin:
Last November I'd done some work related to this which you can see here: https://github.com/flexion/ef-cms/pull/7071/
Most of the index and search syntax changes got rolled back because we hadn't wanted to take the time to reindex our environment at the time -- but the principles within are sound.

shared/src/persistence/elasticsearch/advancedDocumentSearch.js
```diff
@@ -52,7 +52,7 @@ exports.advancedDocumentSearch = async ({
           },
         ],
         should: documentEventCodes.map(eventCode => ({
-          match: {
+          term: {
             'eventCode.S': eventCode,
           },
         })),
```
kkoskelin: I think this entire `should` block could be written as a `must > terms` block as seen here: https://github.com/flexion/ef-cms/pull/7071/files#diff-b6ca2376effc775062456ff2b69149507d5b5e7b3c96e185c7568e9a0457d0ccR24
It would obviate the need for calling `.map` with an anonymous function, but could instead do 
```
terms: {  'eventCode.S': documentEventCodes }
```

shared/src/persistence/elasticsearch/advancedDocumentSearch.test.js
```diff
@@ -53,12 +52,12 @@ describe('advancedDocumentSearch', () => {
         ],
         should: [
```
kkoskelin: This `should[]` array can be replaced with a similar construction as mentioned earlier: 
```
must: [ 
 {
   terms: { columnName: termArray}
 }
]
```

shared/src/persistence/elasticsearch/advancedDocumentSearch.test.js
```diff
@@ -84,12 +82,12 @@ describe('advancedDocumentSearch', () => {
...
```
kkoskelin: use must/terms here also

shared/src/persistence/elasticsearch/getIndexedCasesForUser.js
```diff
@@ -32,16 +32,14 @@ exports.getIndexedCasesForUser = async ({
...
-            { match: { 'sk.S': 'case|' } },
-            { match: { 'gsi1pk.S': 'user-case|' } },
             {
               bool: {
                 should: statuses.map(statusStr => ({
-                  match: {
+                  term: {
                     'status.S': statusStr,
                   },
                 })),
```
kkoskelin: Replace with a `terms` construction here also.

shared/src/persistence/elasticsearch/messages/getSectionInboxMessages.js
```diff
@@ -7,12 +7,10 @@ exports.getSectionInboxMessages = async ({ applicationContext, section }) => {
...
           must: [
             {
-              match: {
-                'toSection.S': { operator: 'and', query: section },
-              },
+              term: { 'toSection.S': section },
             },
             {
-              match: { 'isRepliedTo.BOOL': false },
+              term: { 'isRepliedTo.BOOL': false },
```
kkoskelin, line 13: excellent

web-api/elasticsearch/efcms-case-deadline-mappings.js
```diff
@@ -4,7 +4,7 @@ module.exports = {
       type: 'text',
     },
     'caseDeadlineId.S': {
-      type: 'text',
+      type: 'keyword',
```
kkoskelin, line 7: Consider introducing `index: false` to all columns that you expect to _never_ appear in the query predicates, but only returned in the `_source` arrays of queries.  This should also speed up indexing.

kkoskelin (approved):
Some nitpicks about constructing slightly more efficient queries, but overall, pretty good!
I like the idea of setting `index: false` for all columns that won't appear in query predicates.
A task for another day is probably to audit all query scripts, collecting all columns that are in query predicates vs `_source` array and see if there are any other index mapping columns that could be updated or removed entirely.
A great start to this effort!

web-api/elasticsearch/efcms-case-deadline-mappings.js
mark-meyer: @kkoskelin quick question about using `index:false`...it looks like [`elasticsearch-index-settings.js`](https://github.com/ustaxcourt/ef-cms/blob/staging/web-api/elasticsearch/elasticsearch-index-settings.js#L56) sets `dynamic: false` on all indexes which should prevent indexing any fields not explicitly set in the mapping. Is there any advantage to introducing `index:false` vs removing these fields from the mapping?

kkoskelin: @mark-meyer I am far from an elasticsearch expert -- so please take this all with a few grains of salt.
I think we may need some columns specified here so that their values are stored, and can then be retrieves using the `_sources` array, even if they're not part of the actual search predicates.
This is partly why I suggested a sort of "audit":
For any given ES index:
1) compile a list of columns mentioned in the `_sources`
2) ...and any columns and in the query predicates
3) any columns that appear ONLY in `_sources` should probably have `index: false` unless we think there's a chance that column will appear in a future query predicate.
4) any columns that appear in the index-mappings, but NOT in the `_sources` nor query predicates should be removed from the index-mappings.

I think.  I wouldn't mind folks checking my work on the above logic.  😄

shared/src/persistence/elasticsearch/advancedDocumentSearch.js
mark-meyer: Thanks @kkoskelin that's a great suggestion. I've made these changes where that pattern occurs.

shared/src/persistence/elasticsearch/getCaseInventoryReport.js
```diff
@@ -38,16 +38,12 @@ exports.getCaseInventoryReport = async ({
       from,
       query: {
         bool: {
-          must: [
+          must: [],
```
mmarcotte, line 41: Just curious, do we need a 

```
must: [],
```

if we have a `must_not: [ ...` ?

shared/src/persistence/elasticsearch/getCasesByUserId.js
```diff
@@ -21,18 +21,15 @@ exports.getCasesByUserId = async ({ applicationContext, userId }) => {
...
-                  ],
-                  query: `"${userId}"`,
-                },
+                term: { 'privatePractitioners.L.M.userId.S': `${userId}` },
+              },
+              {
+                term: { 'irsPractitioners.L.M.userId.S': `${userId}` },
+              },
+              {
+                term: { 'userId.S': `${userId}` },
```
mmarcotte, line 32: I think we can take replace ``${userId}`` with `userId` in these

shared/src/persistence/elasticsearch/getIndexedCasesForUser.js
```diff
@@ -32,19 +32,13 @@ exports.getIndexedCasesForUser = async ({
         bool: {
           must: [
             {
-              match: {
-                'pk.S': { operator: 'and', query: `user|${userId}` },
+              term: {
+                'pk.S': `user|${userId}`,
```
mmarcotte, line 36: Perhaps this should be done in a follow up ticket, but i would love to see the PK and SKs dropped in favor of the UUIDs being their own prop. I realize this would affect the processStreamsInteractor since that what does the indexing.

shared/src/persistence/elasticsearch/getCaseInventoryReport.js
mark-meyer: @mmarcotte, agreed that this is a little confusing. Items are selectively pushed into that array further down in the function. `searchParameters.body.query.bool.must.push...`.

web-api/elasticsearch/efcms-case-deadline-mappings.js
mark-meyer: @kkoskelin I think these mappings only effect which terms get put in the inverted index. By default, the entire document will be saved to `_source` regardless of what we put in the mappings unless we explicitly exclude them with something like:

 ```
_source: {
    'excludes': [
      'some_field'
    ]
  }
```

On the other hand, no fields will be included in the inverted index (i.e. be searchable) unless they are included in the mapping above. So if we put nothing in the mapping, we would have the complete document saved in `_source`, but no searchable fields. 

We can see this in action with the seed data when the app is running locally. The `efcms-user-mappings.js` does not include the `email` field, but requesting all users with `http://localhost:9200/efcms-user/_search` shows emails like `"petitionsclerk@example.com"` in the `_source` object of the returned results. However, we can't search by email, because it's not in the mappings (and hence not in the index). 

That's not to say I think an audit is a bad idea — it makes sense to only index the things you are searching for. I just think the mechanism of `index: false` isn't needed.

kkoskelin: @mark-meyer I'll take your word for it -- so much of what I've learned is by reading and re-reading documentation and wishing we had an ES expert at our disposal.

shared/src/persistence/elasticsearch/getCaseInventoryReport.js
kkoskelin: I am uncertain -- though it seems likely that any query ought to have some "positive" terms, either using `must` or `should`, I think?
Regarding composition of queries, I've considered something like this: https://blog.logrocket.com/elasticsearch-query-body-builder-node-js/

kkoskelin (approved):
Once again if we're passing tests, at this point I gotta give it a thumb. 👍

PR #846, opened 2021-02-10
[hotfix] 7751: BUG: Unable to add trial transcript to docket record

sutt0n:
This bug fixes the need for `documentContents` to be returned in the `getCaseInteractor`, and fetch it _only_ where needed to reduce the size returned from API requests.

Draft TODO: 
- [x] Create migration to remove nested `draftOrderState.draftOrderState`
- [x] Fix `draftOrderState` nesting itself when modifying a draft order


shared/src/business/useCases/document/getDocumentContentsForDocketEntryInteractor.js
```diff
@@ -0,0 +1,32 @@
...
+ * @param {object} providers the providers object
+ * @param {object} providers.applicationContext the application context
+ * @param {object} providers.documentContentsId document contents id
+ * @returns {string} url for the generated document on the storage client
+ */
+exports.getDocumentContentsForDocketEntryInteractor = async ({
+  applicationContext,
+  documentContentsId,
+}) => {
+  try {
```
rachaelparris, line 25: Should there be some permissions checks in here to verify that the user is allowed to see this content?

mrinsin: Good idea, there probably should be!

rachaelparris: Should there be more permission checks here, the same as in `getCaseInteractor`? The way this is written now, anyone with petitioner credentials and a docketEntryId could get the full text of a document on a case they are not associated with. Maybe that's not a concern, just thought I should throw it out there.

rachaelparris (approved):

web-client/src/presenter/actions/getDocumentContentsAction.js
```diff
@@ -0,0 +1,34 @@
...
+  applicationContext,
+  props,
+}) => {
+  const { caseDetail, docketEntryIdToEdit } = props;
+
+  if (docketEntryIdToEdit) {
+    const { documentContentsId } = caseDetail.docketEntries.find(
+      d => d.docketEntryId === docketEntryIdToEdit,
+    );
+
```
mmarcotte, line 19: Should this action throw an error if 
a) it doesn't receive a docketEntryIdToEdit
b) it can't find docketEntryIdToEdit in caseDetail.docketEntries

mmarcotte (commented):
Looks great, just the one nit-picky question.

web-client/src/presenter/actions/getDocumentContentsAction.js
sutt0n: That flow _shouldn't_ get hit; furthermore, there are other actions that do this and don't throw an error for the same reason. Here's a few, for example:

- completeWorkItemForDocumentSigningAction.js
- getDocumentEditUrlAsPathAction.js
- setDocumentToEditAction.js

mmarcotte: I agree that logically shouldn't happen. However, if it does, it will fail pretty quietly. No need to fix now, but it's something to consider for future actions that we write.

mmarcotte (approved):

PR #860, opened 2021-02-16
7787 HOTFIX: Unable to QC document after changing document type from external to internal

rachelschneiderman:


web-client/src/presenter/actions/FileDocument/refreshExternalDocumentTitleFromEventCodeAction.test.js
```diff
@@ -122,4 +122,52 @@ describe('refreshExternalDocumentTitleFromEventCodeAction', () => {
...
+      {
+        modules: { presenter },
+        state: {
+          form: internalFilingDocument,
+        },
+      },
+    );
+
+    expect(result.state.form.documentTitle).toEqual(
+      internalFilingDocument.documentTitle,
```
mmarcotte: Hmm, I think the runAction might be mutating internalFilingDocument. This test should probably be matching a string and not the object. When I run this locally, they are both getting set to `[Miscellaneous]`

web-client/src/presenter/actions/FileDocument/refreshExternalDocumentTitleFromEventCodeAction.test.js
```diff
@@ -101,25 +89,59 @@ describe('refreshExternalDocumentTitleFromEventCodeAction', () => {
...
+        modules: { presenter },
+        state: {
+          form: {
+            category: 'Miscellaneous',
+            eventCode: 'MISC',
+          },
+        },
+      },
     );
+
```
mmarcotte, line 144: Hmm these last two tests are the same, but have different names. I think we should test for the secondary document.

mmarcotte (changes requested):

web-client/src/presenter/actions/FileDocument/refreshExternalDocumentTitleFromEventCodeAction.test.js
```diff
@@ -101,25 +89,59 @@ describe('refreshExternalDocumentTitleFromEventCodeAction', () => {
...
+    expect(result.state.form.documentTitle).toEqual('[Miscellaneous]');
   });
```
mmarcotte: These last two tests are the same, but have different names. 

I think we should test for the secondary document.

rachelschneiderman: You are right. I will update that now. Apologies!

mmarcotte (approved):
Looks great, many thanks!

yipee

Issue #874, opened 2021-02-19
Retry Failed Service Email Attempts


mmarcotte:
AC:

- As a DAWSON Administrator, in order to ensure that service email is sent to the required parties, I need the system to reattempt to send any failed service emails
- As a DAWSON Administrator, in order to inform the Court, I need to know when a service email cannot be sent after multiple retries
  - logger.error(...)   
  - Possible Email to Clerk of the Court?

NOTES:
```
const response = await SES.sendBulkTemplatedEmail(params).promise();
```
This bit of code returns the response from the AWS call. It could include successes or failures. Failures most likely are where SES did not even attempt to send the email and we should try again, removing the ones that were successful. 

Here's an example with mixed results:
```
{
  "Status": "Failed",
  "Error": "Unknown error occurred. Please try again later"
},
{
  "Status": "Failed",
  "Error": "Unknown error occurred. Please try again later"
},
{
  "Status": "Success",
  "MessageId": "01000177833178a1-473348fc-4011-4ab9-83a8-2a37613af68e-000000"
},
{
  "Status": "Failed",
  "Error": "Unknown error occurred. Please try again later"
},
{
  "Status": "Failed",
  "Error": "Unknown error occurred. Please try again later"
},
{
  "Status": "Failed",
  "Error": "Unknown error occurred. Please try again later"
}
```

adunkman:
Clearing my assignment here — @mmarcotte got to this first and was a WIP branch at [`874-mm-retry-failed-service-emails`](https://github.com/ustaxcourt/ef-cms/compare/874-mm-retry-failed-service-emails), deploying to `exp1` to test.

PR #881, opened 2021-02-24
Retry Failed Service Emails

mmarcotte:
In order to satisfy #874, this PR ensures that we re-attempt any failures reported by SES. It sets an arbitrary number of retries to 6, giving us a total of 7 attempts to send service email. Should a message be un-sendable after that many attempts, we log an error the logs. 

This PR also introduces a couple of new helper functions that facilitate the reattempts. This is advisory practice for handling failures with AWS to avoid throttling.

- `sleep` simply takes the number of milliseconds as a parameter, and returns a Promise that resolves itself after the specified number of milliseconds
- `backOff` calls this new `sleep` function with a calculated number of milliseconds based on the number of times we have retried. The calculation is `(2 ^ retryCount) * 100ms`.

#878 attempts to go a step further and put the application / SES in an unhealthy status if we get repeated failures sending.

### DIFF
--- a/shared/src/dispatchers/ses/sendBulkTemplatedEmail.js
+++ b/shared/src/dispatchers/ses/sendBulkTemplatedEmail.js
@@ -1,4 +1,6 @@
-const MAX_SES_RETRIES = 10;
+const { backOff } = require('../../tools/helpers');
+
+const MAX_SES_RETRIES = 6;
 
 /**
  * calls SES.sendBulkTemplatedEmail
@@ -85,6 +87,9 @@ exports.sendWithRetry = async ({
       .join(',');
     throw `Could not complete service to ${failures}`;
   }
+
+  // exponential back-off
+  await backOff(retryCount);
 
   await exports.sendWithRetry({
     applicationContext,
--- a/shared/src/dispatchers/ses/sendBulkTemplatedEmail.js
+++ b/shared/src/dispatchers/ses/sendBulkTemplatedEmail.js
@@ -27,8 +27,6 @@ exports.sendBulkTemplatedEmail = async ({
   destinations,
   templateName,
 }) => {
-  const SES = applicationContext.getEmailClient();
-
   try {
     const params = {
       DefaultTemplateData: JSON.stringify(defaultTemplateData),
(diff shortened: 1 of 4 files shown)


shared/src/dispatchers/ses/sendBulkTemplatedEmail.test.js
```diff
@@ -72,4 +72,158 @@ describe('sendBulkTemplatedEmail', () => {
...
+          }),
+        },
+      ],
+    });
+    expect(
+      applicationContext.getEmailClient().sendBulkTemplatedEmail,
+    ).toHaveBeenCalledTimes(2);
+  });
+
+  it('should retry a failed mailing 10 times, and then throw an error', async () => {
```
matthopson: Should this be "7 times" ? - Also, what do you think about storing the `MAX_SES_RETRIES` on applicationContext, so we can mock it out / pull it from constants?

shared/src/dispatchers/ses/sendBulkTemplatedEmail.js
```diff
@@ -44,12 +46,57 @@ exports.sendBulkTemplatedEmail = async ({
...
 
-    applicationContext.logger.info('Bulk Email Params', params);
+    await exports.sendWithRetry({ applicationContext, params });
+  } catch (err) {
+    applicationContext.logger.error(`Error sending email: ${err}`, err);
+  }
+};
+
+/**
+ * Sends the email via SES with a MAX_SES_RETRIES = 10;
```
matthopson: Currently set to `7` above. Added a comment below about maybe sticking this on the constants.

shared/src/dispatchers/ses/sendBulkTemplatedEmail.js
```diff
@@ -44,12 +46,57 @@ exports.sendBulkTemplatedEmail = async ({
...
+
+  if (retryCount >= MAX_SES_RETRIES) {
+    const failures = needToRetry
+      .map(dest => dest.Destination.ToAddresses[0])
+      .join(',');
+    throw `Could not complete service to ${failures}`;
   }
+
+  // exponential back-off
+  await backOff(retryCount);
```
codyseibert, line 91: there is a package called `promise-retry` I have used in the past which has built-in expo backoff which might help reduce the extra logic, but it's probably not critical to use

adunkman: I’ve used the same package in the past as well and have a similar conviction — that is, probably not critical to use, listen to your heart! 🎵

adunkman (approved):
Looks good to me — could use a bit of 🧹 around the constant values and a potential package to use for exponential backoff, but I would consider none of it to be blocking for the pull request. 

Thanks for the test!

shared/src/dispatchers/ses/sendBulkTemplatedEmail.test.js
mmarcotte: thanks! it had been 10. I reduced it after adding the exponential backoff due to the anticipation of timing out a Lambda if un-sendable

shared/src/dispatchers/ses/sendBulkTemplatedEmail.js
```diff
@@ -44,12 +46,57 @@ exports.sendBulkTemplatedEmail = async ({
...
+  applicationContext.logger.info('Bulk Email Params', params);
+  const response = await SES.sendBulkTemplatedEmail(params).promise();
+  applicationContext.logger.info('Bulk Email Response', response);
 
-    const response = await SES.sendBulkTemplatedEmail(params).promise();
+  // parse response from AWS
+  const needToRetry = response.Status.map((attempt, index) => {
+    // AWS returns 'Success' and helpful identifier upon successful send
+    return attempt.Status !== 'Success' ? params.Destinations[index] : false;
+  }).filter(row => !!row);
```
kkoskelin: not even a _nit_ but an observation I made in a recent commit -- you can use `.filter(Boolean)` which gives you only the truthy entries, which is kind of nice.  no change necessary.

shared/src/dispatchers/ses/sendBulkTemplatedEmail.test.js
```diff
@@ -72,4 +72,158 @@ describe('sendBulkTemplatedEmail', () => {
...
+            welcomeMessage: 'Welcome to Flavortown',
+            whoAmI: 'The Sauce Boss',
+          },
+        },
+      ],
+      templateName: 'case_served',
+    });
+    expect(
+      applicationContext.getEmailClient().sendBulkTemplatedEmail,
+    ).toHaveBeenCalledTimes(7);
```
kkoskelin: let's use the constant here, too.

kkoskelin (approved):

kkoskelin (approved):

PR #917, opened 2021-03-15
Latest to migration

rosiehsmith:
Includes
* 7827
* 7772


rosiehsmith:
Will probably merge from PR instead

PR #920, opened 2021-03-17
Reduce WAFv2 regex pattern sets.

adunkman:
This configures the regex pattern sets per-environment instead of per-color, resulting in a 1/2 reduction of resources. There is a [hard cap of 10 pattern sets per account per region](https://docs.aws.amazon.com/waf/latest/developerguide/limits.html) for… reasons!?… and this ensures we can support up to 10 environments per account.


adunkman:
This is now ready for review, deployed to `dev` and it completed successfully. Since we were at our AWS account limit, I had to accomplish through two deploys (first deploy deleted two records but couldn’t create the new single replacement because the count of records hadn’t been updated yet with the removed records).

I don’t expect this to be a problem going forward — but if it does pop up, just restart the build.

mmarcotte (approved):

PR #927, opened 2021-03-18
Restrict USWDS to patch-level versions of ~2.10

kkoskelin:
Upgrading to USWDS version `^2.11.x` introduces unexpected changes to icons, so we're keeping it as `~2.10` for now.
Symptom of upgrading to 2.11 seen below:
Screen Shot 2021-03-18 at 10 23 16 AM


kkoskelin:
Because the court does not run `npm update`, this can be managed through the pull-request pipeline coming from flexion/ef-cms. Closing.

PR #939, opened 2021-03-22
Refactor for applicationContext

mmarcotte:
Looks like during recent refactor for pulling applicationContext out of providers object left some missing.

### DIFF
--- a/web-client/src/presenter/actions/getUsersInSectionAction.js
+++ b/web-client/src/presenter/actions/getUsersInSectionAction.js
@@ -23,8 +23,7 @@ export const getUsersInSectionAction = ({ section }) =>
     }
     const users = await applicationContext
       .getUseCases()
-      .getUsersInSectionInteractor({
-        applicationContext,
+      .getUsersInSectionInteractor(applicationContext, {
         section: sectionToGet,
       });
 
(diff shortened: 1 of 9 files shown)


web-client/src/presenter/actions/getUsersInSectionAction.js
```diff
@@ -23,8 +23,7 @@ export const getUsersInSectionAction = ({ section }) =>
     }
     const users = await applicationContext
       .getUseCases()
-      .getUsersInSectionInteractor({
-        applicationContext,
+      .getUsersInSectionInteractor(applicationContext, {
```
sutt0n, line 26: this should be a proxy, shouldn't have to refactor this

web-client/src/presenter/actions/getUsersInSelectedSectionAction.js
```diff
@@ -17,8 +17,7 @@ export const getUsersInSelectedSectionAction = async ({
   }
   const users = await applicationContext
     .getUseCases()
-    .getUsersInSectionInteractor({
-      applicationContext,
+    .getUsersInSectionInteractor(applicationContext, {
```
sutt0n, line 20: this should be a proxy, shouldn't have to modify this.

PR #954, opened 2021-03-24
Add WAFv2 firewall rule to block banned IP addresses.

### LINKED ISSUE #728: Ban an IP address
Acceptance Criteria:

As a sysadmin, so that I can keep bad actors from impacting the performance of the site, I need to quickly ban IP addresses from making requests to the API Gateway. 

Notes:

- Should be a bit of Terraform or aws CLI to be able to ban (or unban) an IP address with one command. 
- Possibly make it easy automate banning an IP in case as we add more triggers to catch suspicious or malicious activity.
- Need to be able to quickly identify an IP address in Kibana,
- IP Address is easily visible in Kibana logs by `request.headers.x-forwarded-for`

References:
https://aws.amazon.com/premiumsupport/knowledge-center/api-gateway-resource-policy-access/

adunkman:
Closes #728.

This PR introduces a new firewall rule to block requests from specific IP addresses. There are example IPs deployed by default to demonstrate how to use the rules (they’re IPs in reserved "example" space, so they will never match incoming requests).

Additionally, this PR adds a bash script to quickly ban an IP address for an environment, and documentation showing how to see blocked requests and manually access the banned IPs through the AWS Console.

Deployed and tested in the `dev` environment.

### DIFF
--- a/docs/operations/ip-blocking.md
+++ b/docs/operations/ip-blocking.md
@@ -1,0 +1,62 @@
+# Blocking IP addresses
+
+This describes a mechanism to block access to EF-CMS, which can be used when an IP address is causing unexpected or unusual traffic. This is an augmentation to other, automated rate limiting and security measures which may also block access, and at best will be used sparingly.
+
+## How to block an IPv4 or IPv6 address
+
+Blocking is done through [AWS Web Application Firewall v2](https://docs.aws.amazon.com/waf/latest/APIReference/Welcome.html), and requests which match blocked patterns will receive `403 Forbidden` status codes when attempting to make API requests.
+
+### Using a bash script
+
+To block an IP address, run:
+
+```bash
+./web-api/ban-ip-address.sh [ENVIRONMENT] [IP ADDRESS]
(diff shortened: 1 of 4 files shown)


docs/operations/ip-blocking.md
```diff
@@ -0,0 +1,62 @@
+# Blocking IP addresses
+
+This describes a mechanism to block access to EF-CMS, which can be used when an IP address is causing unexpected or unusual traffic. This is an augmentation to other, automated rate limiting and security measures which may also block access, and at best will be used sparingly.
```
sutt0n, line 3: nit: should `DAWSON` be used instead of `EF-CMS`?

mmarcotte: Good question. I would say DAWSON is the currently deployed production environment of EF-CMS. There could be others if this repo was forked and repurposed, and so it's probably suitable to use EF-CMS as a generic reference. Curious what others think.

mmarcotte (approved):

PR #961, opened 2021-03-24
[hotfix] 8035: scale down scanned images

sutt0n:
Addresses https://github.com/flexion/ef-cms/issues/8035
Scaled down scanned images to match the coversheet. 

image


mmarcotte (approved):
honey i shrunk the kids

Looks great, thank you!

PR #969, opened 2021-03-26
8051: strict timestamps

kkoskelin:
This effort always converts YYYY-MM-DD to strict ISO before storage, and should do the same for any comparisons.  It is currently focused on just a couple of date/time fields, and probably can plan this for even more in the near future.

We push a date like "2020-01-02" through `createISODateAtStartOfDayEST()` to produce "2020-01-02T05:00:00.000Z".

* This PR also contains a very important bugfix in a `DateHandler` function that was used to decide whether or not two dates represented the same calendar day.

*Additional note* the commit history of this PR is a bit messy because it was cherry-picked from commits on a different fork/branch and had several conflicts that needed resolution.


shared/src/business/utilities/DateHandler.js
```diff
@@ -112,16 +118,15 @@ const formatNow = formatStr => {
...
-    // at least one date has a simple format, compare only year, month, and day
-    const [aSimple, bSimple] = [
-      a.match(simpleDatePattern)[0],
-      b.match(simpleDatePattern)[0],
-    ];
-    if (aSimple.localeCompare(bSimple) == 0) {
+    // at least one date has a simple format, compare only year, month, and day according to EST
+    const dayDifference = calculateDifferenceInDays(
+      createISODateString(a),
+      createISODateString(b),
```
mmarcotte, line 127: In reading this, I think I would expect these both to be:

```
createISODateAtStartOfDayEST(a), 
createISODateAtStartOfDayEST(b),
```

mmarcotte: But... not so fast because if one of these is a timestamp (and not  a date), then it would rely upon `createISODateString(...)`

matthopson: Yeah, I think the intent is to just create a full ISO date string from either a simple date or timestamp and then JUST compare the date portions from that - basically a means to normalize inputs for comparison. You ultimately end up with an ISO date string - which then gets mutated by the `calculateDifferenceInDays` method (the hours are set to 12). So you're right in that we could just create a start of day date from what's given, but it's a no-op either way.

shared/src/business/utilities/DateHandler.test.js
```diff
@@ -287,6 +329,16 @@ describe('DateHandler', () => {
...
+
+    it('should return a non-zero if two calendar-dates appear to be the same but are different according to EST', () => {
+      let result;
+      const date1 = '2001-01-02'; // i.e. Jan 2, midnight EST
+      const date2 = '2001-01-02T02:40:55.007Z'; // Jan 1, 9:40m EST
+
+      result = DateHandler.dateStringsCompared(date1, date2);
+      expect(result).not.toEqual(0);
+    });
+
```
mmarcotte, line 341: ☝️ all of these tests are the jam☝️ 
👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏

matthopson (approved):

mmarcotte (approved):
timezones are hard

Thanks for this challenging work. Love the tests you've introduced.

PR #973, opened 2021-03-29
8026 - Pending Report Fixes

codyseibert:


web-client/src/presenter/computeds/formattedCaseDetail.js
```diff
@@ -256,7 +256,15 @@ export const formattedCaseDetail = (get, applicationContext) => {
...
-    d => d.pending && applicationContext.getUtilities().isServed(d),
+    docketEntry =>
+      (docketEntry.pending &&
+        applicationContext.getUtilities().isServed(docketEntry)) ||
+      (docketEntry.pending &&
+        applicationContext
+          .getConstants()
+          .UNSERVABLE_EVENT_CODES.find(
+            unservableCode => unservableCode === docketEntry.eventCode,
+          )),
```
mmarcotte: Just curious, could we not make use of the logic introduced in this PR in `shared/src/business/entities/DocketEntry.js`

codyseibert: good catch, we will fix that

TomElliottFlexion (commented):
Test

PR #979, opened 2021-03-29
7839: refactor all contacts into petitioners array

rachaelparris:
A blue/green migration will need to be run with these changes.


web-api/migration-terraform/main/lambdas/migration-segments.js
```diff
@@ -1,5 +1,11 @@
...
 const createApplicationContext = require('../../../src/applicationContext');
+const {
+  migrateItems: migration0024,
+} = require('./migrations/0024-docket-entry-docket-number-served-parties-code');
+const {
+  migrateItems: migration0025,
+} = require('./migrations/0025-add-contacts-to-petitioners-array');
 
 const {
   migrateItems: migration0024,
```
mmarcotte, line 11: @rachelschneiderman -- FYI, you'll need to remove one of these; I got it going manually for tonight.

rachelschneiderman: Thank you! I'll do that now so we don't forget.

PR #991, opened 2021-04-01
7839: (Refactoring) Merge contactPrimary, contactSecondary, otherFilers, and otherPetitioners into one array on Case entity

rachelschneiderman:


shared/src/business/useCases/createCaseFromPaperInteractor.js
```diff
@@ -124,6 +128,8 @@ exports.createCaseFromPaperInteractor = async (
 
   const petitionDocketEntryEntity = new DocketEntry(
     {
+      contactPrimary: caseToAdd.getContactPrimary(),
+      contactSecondary: caseToAdd.getContactSecondary(),
```
mmarcotte, line 132: No changes requested, but is this also an area where this sort of refactoring effort will need to be continued in the future?

rachaelparris: It seemed like it would be more difficult to use an array for the create case forms because only 2 petitioners are allowed in that instance, so we left it this way purposefully, unless there becomes a need to refactor further during the other stories.

shared/src/business/useCases/users/generateChangeOfAddress.test.js
```diff
@@ -73,21 +78,25 @@ describe('generateChangeOfAddress', () => {
   };
 
   beforeEach(() => {
+    mockCase = mockCaseWithPrivatePractitioner;
```
mmarcotte, line 81: is there any worry of mutating `mockCaseWithPrivatePractitioner` in one of these tests, and this wouldn't be the same thing the next time we made this call?

shared/src/business/useCases/users/generateChangeOfAddress.test.js
```diff
@@ -105,15 +114,13 @@ describe('generateChangeOfAddress', () => {
     ]);
   });
 
-  it('attempts to run a change of address when address1 changes for an irs practitioner', async () => {
+  it('should run a change of address when address1 changes for an irs practitioner', async () => {
+    mockCase = mockCaseWithIrsPractitioner;
```
mmarcotte, line 118: would seem that this isn't necessary due to the `beforeEach`?

web-api/migration-terraform/main/lambdas/migrations/0025-add-contacts-to-petitioners-array.js
```diff
@@ -0,0 +1,49 @@
...
+
+      item.petitioners = [
+        ...(item.petitioners || []),
+        ...(item.otherFilers || []),
+        ...(item.otherPetitioners || []),
+      ];
+
+      const updatedCaseRaw = new Case(item, { applicationContext })
+        .validate()
+        .toRawObject();
```
mmarcotte, line 29: This migration script is incredibly tidy and easy to follow. One thing I can't quite tell (without digging back into Case) is precisely how the primary and secondary petitions get their CONTACT_TYPES associated with them? I think that's in the constructor, right?

mmarcotte: ... seems well covered in the tests below. 😄  👍

shared/src/business/useCases/users/generateChangeOfAddress.test.js
rachelschneiderman: Good catch. It looks like some of the tests are modifying `mockCaseWithPrivatePractitioner`, we will go and update those so they modify `mockCase` instead!

web-client/src/presenter/computeds/formattedCaseDetail.test.js
```diff
@@ -1,3 +1,4 @@
+import { CONTACT_TYPES } from '../../../../shared/src/business/entities/EntityConstants';
 import { MOCK_CASE } from '../../../../shared/src/test/mockCase';
 import { SERVICE_INDICATOR_TYPES } from '../../../../shared/src/business/entities/EntityConstants';
```
mmarcotte: minor fix: 
```
'../../../../shared/src/business/entities/EntityConstants.js' imported multiple times
```

mmarcotte (approved):

PR #995, opened 2021-04-01
8083 - Find Mis-Associated petitioners script

codyseibert:
`node find-petitioners-missing-cases.js mig alpha https://search-efcms-search-mig-alpha-dwffrub5hv5f4w4vlxpt4v65ni.us-east-1.es.amazonaws.com` is an example of how you could run this again mig alpha.

arg1 = env
arg2 = version
arg3 = ES endpoint


matthopson (commented):
I assume this is probably going to be short-lived, but do you think it's worth adding a test for it for good measure? I think it's pretty straight-forward, otherwise.

rachelschneiderman (approved):

mmarcotte (approved):

PR #1001, opened 2021-04-02
8019 Bug - Trial Session List / Filter

matthopson:
From #8019 comments:

## The Problem
The root of this issue is, when updating a note on a hearing, the persistence method is adding a `gsi1pk: trial-session-catalog` - and should not be. This is allowing the hearing to be duplicated in the trial session list. In cases where a given duplicate session (it's not _actually_ a duplicate - it's a separate distinct case hearing record that is coming back because of that gs1ipk) falls on the same trial week as another session for a given filtered judge, that record hangs around - which is also odd, but I believe it's due to how the item is being indexed in the React template. I suspect React isn't being told, based on its key, to REMOVE that element from the UI so it hangs around.

## Steps to Reproduce
*Note, this only works if the later steps that create this condition haven't been met for the given trial sessions - it would also work to just create two trial sessions in the same week with two different judges.

- Find two trial sessions that fall on the same week with two different judges (Judge A and Judge B).
- Go to a case and add trial session from Judge A
- View trial sessions, filter by Judge A and note that there are no unintended duplicates
- Filter by Judge B and note that no records from Judge A show
- Taking it a step further, you could look in dynamo at the mapping record generated by adding the case to the hearing and note there is NO gsi1pk for this record.
- Go back to the case and update the note on the hearing and save
- Go to trial sessions and repeat steps from above, and note there will be a duplicate when filtering by Judge A
- Filter by Judge B and note that the session from Judge A is now present.
- Looking at that same mapping record in dynamo will now show a gsi1pk with `trial-session-catalog` (and manually deleting this property will revert the condition back to before updating the note on the trial session - no duplicates, and no mis-filtered items).

## The Fix
The fix here is to update the persistence method to prevent it from adding that `gsi1pk` and create a migration that updates any case hearing record (`pk: case|${docketNumber`, `sk: hearing|${trialSessionId}`) with a gsi1pk of `trial-session-catalog` to no longer have that gsi1pk.

## Deploy Notes
- Requires Blue/Green Migration


web-api/migration-terraform/main/lambdas/migrations/0026-hearing-remove-gsi1pk.test.js
```diff
@@ -0,0 +1,91 @@
+const { migrateItems } = require('./0026-hearing-remove-gsi1pk.js');
+
+describe('migrateItems', () => {
+  let mockDocketEntry;
```
rachelschneiderman: World's largest nit: I think this variable is more like `mockHearing`?

rachelschneiderman (approved):

web-api/migration-terraform/main/lambdas/migrations/0026-hearing-remove-gsi1pk.test.js
matthopson: Thanks! Copy/Pasta!

mmarcotte (approved):

PR #1003, opened 2021-04-05
Devex 547

rosiehsmith:
* combined docket-record, work-queue, and ustc-table into one class to reduce redundant css
* replaced all references to docket-record and work-queue with ustc-table instead
* https://trello.com/c/wtBJsoIr/547-css-cleanup-tablesscss


rosiehsmith:
closing this one - PR https://github.com/ustaxcourt/ef-cms/pull/1006 into test instead of staging since that follows the git flow experiment - refactoring goes in the same flow as bugs

PR #1006, opened 2021-04-05
Devex 547

rosiehsmith:
- combined docket-record, work-queue, and ustc-table into one class to reduce redundant css
- replaced all references to docket-record and work-queue with ustc-table instead
- https://trello.com/c/wtBJsoIr/547-css-cleanup-tablesscss


mmarcotte (approved):

mrinsin (approved):

rachelschneiderman (approved):

web-client/src/views/DocketRecord/DocketRecord.jsx
```diff
@@ -104,7 +104,7 @@ export const DocketRecord = connect(
                     <td>
                       <span
                         className={classNames(
-                          entry.isStricken && 'stricken-docket-record',
+                          entry.isStricken && 'stricken-ustc-table',
```
rachaelparris: This doesn't look right - but maybe it is. Just thought I'd note it!

rosiehsmith: Good catch! I think that's not right.

mmarcotte (approved):

PR #1014, opened 2021-04-05
Upgrade node modules and Terraform

ericsorenson:


web-api/terraform/api/api-public.tf
```diff
@@ -23,6 +23,8 @@ resource "aws_lambda_function" "api_public_lambda" {
 resource "aws_api_gateway_rest_api" "gateway_for_api_public" {
   name = "gateway_api_public_${var.environment}_${var.current_color}"
 
+  minimum_compression_size = "1"
+
```
mmarcotte, line 27: Is this part of the gzip refactoring story?

ericsorenson: Yes!

rachelschneiderman (approved):

web-api/terraform/api/api-public.tf
mmarcotte: cool, and it's harmless to include here? is there more work for that devex task?

ericsorenson: It's harmless, I'm currently debugging to find the reason why the setting hasn't impacted the browser.

mmarcotte: Thanks. Sorry that another PR snuck through and is now conflicting with your node modules update. 😢

kkoskelin (approved):
Our workflow seems to have introduced some of my accessibility-related changes into this PR (Tabs.jsx) but hopefully if we can merge this soon, there will not be merge conflicts with later efforts which also alter Tabs.jsx

PR #1016, opened 2021-04-06
Bug 8090: Entered and Served Stamp not displayed on Stipulated Decision

mrinsin:


rachaelparris (approved):

shared/src/business/useCases/courtIssuedDocument/addServedStampToDocument.test.js
```diff
@@ -12,7 +12,9 @@ describe('addServedStampToDocument', () => {
   let drawTextMock;
```
rachelschneiderman, line 12: I know from experience that writing tests for these issues is incredibly difficult. Was there any way to cover this bug with a test?

.circleci/config.yml
```diff
@@ -876,23 +876,6 @@ build-and-deploy-with-context-defaults: &build-and-deploy-with-context-defaults
 
 workflows:
   version: 2
-  scheduled-testing:
```
rachelschneiderman, line 879: Was this removed on purpose?

rachaelparris: It looks like this happened because we branched off of staging. @mmarcotte can you weigh in on if this commit belongs in test or not?

shared/src/business/useCases/courtIssuedDocument/addServedStampToDocument.test.js
mrinsin: Great question, I am not sure if this bug could have been covered by a test :( however the test suite did fail after we incorporated our bug fix and we updated it to reflect that change.

rachelschneiderman: Ok, that works! Thanks!

.circleci/config.yml
matthopson: fwiw, I've seen this frequently. It might be less confusing if, when merging anything to staging, staging is immediately back-merged to test.

mmarcotte: my bad. i did a bit of merging yesterday, and I missed this after updating the scheduled testing. should be up to date now.

rachelschneiderman (approved):

mmarcotte (approved):

PR #1017, opened 2021-04-06
8091 - Fixing Invalid Service Indicator Failure

codyseibert:
This PR addresses a bug related to trying to serve an order for a case when the case has no paper/electronic service parties.  The underlying issue was that the `DocketEntry.servedPartiesCode` was getting set to an invalid issue when the case has no electronic or paper parties causing entity validation to fail.  We added two fixes in this change:

- cause the fileAndServeCourtIssuedDocumentInteractor to validate earlier before any database changes so that if something goes wrong, it doesn't leave the data in a half-finished state
- reset the contactPrimarylserviceIndicator back to electronic or paper if the counsel is removed from the case

We need to provide scripts to figure out the following:
- did anyone else try to serve orders with cases in a bad state
- are there any other cases out there with no privatePractitioners and a contactPrimary.serviceIndicator === 'None'

## Deploy Notes
Needs blue/green migration


matthopson (approved):

shared/src/business/entities/cases/Case.js
```diff
@@ -1741,6 +1745,18 @@ Case.prototype.removeConsolidation = function () {
   return this;
 };
 
+/**
+ * checks all the practitioners on the case to see if there is a privatePractitioner associated with the userId
+ *
+ * @param {String} userId the id of the user
+ * @returns {boolean} if the userId has a privatePractitioner associated with them
+ */
+Case.prototype.isUserIdRepresentedByPrivatePractitioner = function (userId) {
```
kkoskelin, line 1754: Just to keep coverage up and intent clear, we should probably have some tests around the new functions on the `Case` entity and `DocketEntry` too.

web-api/migration-terraform/main/lambdas/migrations/0027-contact-primary-secondary-service-preference.js
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        new Case(
+          { ...caseRecord, itemToModify },
+          {
+            applicationContext,
+          },
+        ).validate();
+
+        applicationContext.logger.info(
```
matthopson, line 128: @mmarcotte Just wanted to note these are the items to look for in the logs. I've added flags on whether contactPrimary or contactSecondary was updated for a little more info. Let me know if anything else would be helpful here.

web-api/migration-terraform/main/lambdas/migration-segments.js
```diff
@@ -3,6 +3,9 @@ const createApplicationContext = require('../../../src/applicationContext');
 const {
   migrateItems: migration0025,
 } = require('./migrations/0025-docket-entry-received-at-strict-timestamp');
+const {
+  migrateItems: migration0027,
```
matthopson, line 10: Named this 0027, assuming it might get merged alongside the other PR (https://github.com/ustaxcourt/ef-cms/pull/1001) with 0026, but we can address this differently as needed if that's not the case.

kkoskelin (approved):

shared/src/business/entities/cases/Case.js
matthopson: Good catch, I made a mental note to go back and check that, but got wrapped up in this migration.

kkoskelin (approved):

mmarcotte (approved):

PR #1024, opened 2021-04-07
DevEx 737: Non ISO compliant dates

TomElliottFlexion:
- Updated preparedAndFormattedDateISO function in dateHandler utility to avoid passing dates through moment twice, avoiding the deprecation warning that is triggered when passing an invalid object/string into moment.
- Added console error to dateHandler when moment returns an `Invalid` object, to avoid this being silently swallowed. 
- Requesting feedback on _need_ for console error. Is recorded ~dozen times in tests. 

- Updated test data to pass in invalid dates less frequently, since this is already covered in the dateHandler tests.
- Updated dateHandler tests for the new functionality.

### DIFF
--- a/shared/src/business/utilities/DateHandler.js
+++ b/shared/src/business/utilities/DateHandler.js
@@ -38,10 +38,11 @@ const prepareDateFromString = (dateString, inputFormat) => {
     dateString = createISODateString();
   }
   const dateFromMoment = moment.tz(dateString, inputFormat, USTC_TZ);
-  console.log('moment.tz result', dateFromMoment);
-  if (dateFromMoment.toString().includes('Invalid')) {
+  // If an invalid string is passed in, we're able to check the result and log so error is
+  // not silently swallowed
+  if (dateString !== null && dateFromMoment.toString().includes('Invalid')) {
     console.error('Invalid date', {
-      message: `moment cannot recognize string '${dateString}'`,
+      message: `moment cannot recognize dateString: [${dateString}] dateFromMoment: [${dateFromMoment}]`,
     });
   }
 
@@ -300,9 +301,11 @@ const computeDate = ({ day, month, year }) => {
   if (!PATTERNS.YYYYMMDD.test(dateToParse)) {
     return dateToParse;
   }
-  const preparedDateISO = prepareDateFromString(dateToParse, FORMATS.YYYYMMDD);
-  const yyyymmdd = formatDateString(preparedDateISO, FORMATS.YYYYMMDD);
-  return yyyymmdd;
+  const preparedAndFormattedDateISO = prepareDateFromString(
+    dateToParse,
+    FORMATS.YYYYMMDD,
+  ).format(FORMATS.YYYYMMDD);
+  return preparedAndFormattedDateISO;
 };
 
 module.exports = {
--- a/shared/src/business/utilities/DateHandler.js
+++ b/shared/src/business/utilities/DateHandler.js
@@ -40,12 +40,9 @@ const prepareDateFromString = (dateString, inputFormat) => {
   const dateFromMoment = moment.tz(dateString, inputFormat, USTC_TZ);
   // If an invalid string is passed in, we're able to check the result and log so error is
   // not silently swallowed
-  if (dateString !== null && !dateFromMoment.isValid()) {
-    console.error('Invalid date', {
-      message: `moment cannot recognize dateString: [${dateString}] dateFromMoment: [${dateFromMoment}]`,
-    });
+  if (dateString != null && !dateFromMoment.isValid()) {
+    console.error('Invalid date', dateString);
   }
-
   return dateFromMoment;
 };
 
--- a/shared/src/business/utilities/DateHandler.js
+++ b/shared/src/business/utilities/DateHandler.js
@@ -37,7 +37,15 @@ const prepareDateFromString = (dateString, inputFormat) => {
   if (dateString === undefined) {
     dateString = createISODateString();
   }
-  return moment.tz(dateString, inputFormat, USTC_TZ);
+  const dateFromMoment = moment.tz(dateString, inputFormat, USTC_TZ);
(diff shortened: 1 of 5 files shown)


shared/src/business/utilities/DateHandler.js
```diff
@@ -37,7 +37,13 @@ const prepareDateFromString = (dateString, inputFormat) => {
   if (dateString === undefined) {
     dateString = createISODateString();
   }
-  return moment.tz(dateString, inputFormat, USTC_TZ);
+  const dateFromMoment = moment.tz(dateString, inputFormat, USTC_TZ);
+  // If an invalid string is passed in, we're able to check the result and log so error is
+  // not silently swallowed
+  if (dateString != null && !dateFromMoment.isValid()) {
+    console.error('Invalid date', dateString);
```
matthopson, line 44: How is this error being thrown? Or you're saying it is logged for archeological purposes?

TomElliottFlexion: Simply logged for archeological purposes, with a stack trace to pinpoint the issue since it can be a little convoluted

dsalls: Yeah, bigger question was how/whether we wanted to capture something in the logs that we were receiving a string that can't possibly be a date.

web-client/integration-tests/journey/docketClerkAddsTranscriptDocketEntryFromOrder.js
```diff
@@ -55,10 +56,18 @@ export const docketClerkAddsTranscriptDocketEntryFromOrder = (
       value: 'Anything',
     });
 
+    // suppressing moment deprecation warnings temporarily as moment returns an
+    // invalid object when called with a single day or month value
+    moment.suppressDeprecationWarnings = true;
```
matthopson, line 61: Why couldn't we use a two-digit value in the test instead of removing the warning?

web-client/integration-tests/journey/docketClerkCreatesARemoteTrialSession.js
```diff
@@ -69,13 +69,12 @@ export const docketClerkCreatesARemoteTrialSession = (test, overrides = {}) => {
 
     expect(test.getState('validationErrors')).toEqual({
       startDate: errorMessages.startDate[1],
-      term: errorMessages.term,
```
matthopson, line 72: Could you explain this? Shouldn't the intent be to create a validation error and prove that it is displayed?

web-client/integration-tests/journey/docketClerkCreatesATrialSession.js
```diff
@@ -89,13 +89,12 @@ export const docketClerkCreatesATrialSession = (test, overrides = {}) => {
 
     expect(test.getState('validationErrors')).toEqual({
       startDate: errorMessages.startDate[1],
-      term: errorMessages.term,
```
matthopson, line 92: Same comment as above - I think we want to create an invalid term so we can prove a validation message is displayed.

shared/src/business/utilities/DateHandler.js
kkoskelin: 1) I don't like invoking "null" almost ever. (I personally wish it did not exist in JS).  Is there any evidence that it's ever being provided as the value of the first argument?
2) if you want to simply note the issue but not create big traces, I think perhaps `applicationContext.logger.warn(<string>)` would be a better choice.

web-client/integration-tests/journey/docketClerkAddsTranscriptDocketEntryFromOrder.js
kkoskelin: Agreed, to keep things clean, frankly I would prefer that no other tests or files ever need know that `moment` even exists.

TomElliottFlexion: The `updateCourtIssuedDocketEntryFormValueSequence` only updates on form value at a time, and regenerates the document title each time it is called. Calling the sequence with only one day/month value results in a `undefined-month-undefined` or `day-undefined-undefined` string being passed into moment.

kkoskelin: The timestamp business is a can of worms, and this function is in the middle of it.  The mmmm-yy-dd format that's being constructed/deconstructed is really more of a presenter-related validation function IMO and doesn't belong in DateHandler IMO.  If it were pulled out, it would also solve the issue of moment ever seeing `undefined`.

matthopson: So does this seem maybe like an opportunity for refactoring in the sequence? Not 100% sure what that looks like, but I'd prefer that over disabling a deprecation warning if we can help it.

shared/src/business/utilities/DateHandler.js
matthopson: Kris, I also considered that, but we _currently_ do not pass in applicationContext here 😬

TomElliottFlexion: We do pass null in several times, as several functions in dateHandler itself return null. The only important change is the one on lines 301 to 305, which prevents the invalid `moment` object being passed _back_ into moment in `formatDateString`, eliminating a lot of deprecation warnings. The rest is an attempt to log when a bad string is passed into moment, because it happens often and is being swallowed silently.

web-client/integration-tests/journey/docketClerkCreatesARemoteTrialSession.js
TomElliottFlexion: The thought was that since the `term` calculation is already being tested in `computeTrialSessionFormDataAction`, we could avoid passing a 13th month into moment, which returns an invalid object. In retrospect that seems backward.

PR #1045, opened 2021-04-09
8088 - Two Paged Notice of Trial Bug

codyseibert:
The bug was due to larger case captions causing extra lines to push the content down into two pages.  This issue will still happen with very large case captions, but this fix should address 95% of the average cases.

- reduced the margin bottom on 2 elements to leave more space
- renamed Password to Passcode `per Jessica's request`
- changed `is` to `are` to fix a grammar issue
- added a test to verify this PDF is 1 line 
- fixing an issue with this documentGenerators test not working

Screen Shot 2021-04-09 at 11 29 26 AM

### DIFF
--- a/shared/src/business/utilities/htmlGenerator/index.scss
+++ b/shared/src/business/utilities/htmlGenerator/index.scss
@@ -560,6 +560,18 @@ th {
   }
 }
 
+.notice-of-trial-title {
+  margin-bottom: 0px;
+}
+
+#notice-of-trial-pdf .info-box {
+  margin-bottom: 20px;
+}
+
+#notice-body > p {
+  margin-top: 0px;
+}
+
 @font-face {
   font-family: 'nimbus_roman';
   font-style: normal;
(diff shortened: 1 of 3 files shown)


shared/src/business/utilities/htmlGenerator/index.scss
```diff
@@ -560,6 +560,18 @@ th {
...
+.notice-of-trial-title {
+  margin-bottom: 0px;
+}
+
+#notice-of-trial-pdf .info-box {
+  margin-bottom: 20px;
+}
+
+#notice-body > p {
+  margin-top: 0px;
```
kkoskelin, line 572: _nit_: zero-values in CSS require no units (i.e. can be written as `margin-bottom: 0`).  I thought we had a linter rule that complains about this.  No biggie.

kkoskelin (approved):

mmarcotte (approved):

PR #1048, opened 2021-04-09
8105: semantic markup, aria adjustments

kkoskelin:
using td instead of th for accurate semantic markup; moving role and aria attributes to correct elements


rachelschneiderman (approved):

web-client/src/views/AdvancedSearch/DocumentSearchResults.jsx
```diff
@@ -39,8 +39,8 @@ export const DocumentSearchResults = connect(
             <table className="usa-table search-results ustc-table responsive-table">
               <thead>
                 <tr>
-                  <th aria-hidden="true" className="small-column"></th>
-                  <th aria-hidden="true" className="small-column"></th>
+                  <td aria-hidden="true" className="small-column"></td>
```
codyseibert, line 42: is this related to aria why it needs to become a td?

kkoskelin: Explanation found here, I just went with it: https://github.com/flexion/ef-cms/issues/8105

mmarcotte (approved):

adunkman:
> moving role and aria attributes to correct elements

Thank you for this! I noticed this issue yesterday while doing PO review of one of your other fixes, and it’s delightful to see it fixed today. 💖 💖 💖

PR #1062, opened 2021-04-13
8111 - removing email from contact primary

codyseibert:
This PR addresses all the cases which currently have a contactPrimary.email that matches one of the email's in the practitioners array.  This does not address the cases where a practitioner might have been deleted leaving the old email on the contactPrimary.email.  Matt and I think it would be better to do these data migrations in separate migration steps.


codyseibert:
quick note.  Depending on when the contact refactoring makes it in to test / staging, I may need to refactor this depending on if this gets merged before or after

web-api/migration-terraform/main/lambdas/migrations/0028-contact-primary-email.js
```diff
@@ -0,0 +1,84 @@
...
+
+      if (
+        privatePractitioners.find(
+          practitioner =>
+            practitioner.email === itemToModify.contactPrimary.email,
+        )
+      ) {
+        delete itemToModify.contactPrimary.email;
+        delete itemToModify.contactPrimary.serviceIndicator;
+
```
mmarcotte, line 65: because we are deleting their email and service indicator, could we please log out this mutation?

I think all we would need to record are the: 

```
{
  pk,
  contactPrimary.email,
  contactPrimary.serviceIndicator
}
```

web-api/migration-terraform/main/lambdas/migrations/0028-contact-primary-email.js
```diff
@@ -0,0 +1,84 @@
...
+          ...itemToModify,
+          privatePractitioners: [...caseRecord.privatePractitioners],
+        });
+
+        itemToModify.contactPrimary = {
+          ...item.contactPrimary,
+          email: null,
+          serviceIndicator: contactPrimary.serviceIndicator,
+        };
+      }
```
mmarcotte, line 76: Also, it would seem that we don't modify the item if we don't make it into this if block. So we could just push the item into `itemsAfter`?

mmarcotte (changes requested):

mmarcotte (approved):

PR #1070, opened 2021-04-14
8078 - More explicit error logging, catch missing s3 files

matthopson:
The issue, as reported, isn't necessarily solvable with this PR, as there is not enough information to diagnose - however, a few issues were identified in this archiving process that may be contributors - or will at least help produce information for diagnosing this and future problems as they arise.

- Ensures that removed cases do not make it into the batch.
- Ensures files being archived actually exist on s3 before proceeding.
- Ensures the process fails early when there's an error while archiving (bubbling the error to the pipe) - otherwise the error wasn't being displayed until the socket timed out (so it would just appear to hang on x%).

### DIFF
--- a/shared/src/business/useCases/trialSessions/batchDownloadTrialSessionInteractor.js
+++ b/shared/src/business/useCases/trialSessions/batchDownloadTrialSessionInteractor.js
@@ -56,7 +56,11 @@ const batchDownloadTrialSessionInteractor = async (
     .replace(/,/g, '');
 
   sessionCases = sessionCases
-    .filter(caseToFilter => caseToFilter.status !== CASE_STATUS_TYPES.closed)
+    .filter(
+      caseToFilter =>
+        caseToFilter.status !== CASE_STATUS_TYPES.closed &&
+        !caseToFilter.removedFromTrial,
+    )
     .map(caseToBatch => {
       const caseTitle = Case.getCaseTitle(caseToBatch.caseCaption);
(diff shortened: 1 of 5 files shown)


rachelschneiderman (approved):

shared/src/business/useCases/trialSessions/batchDownloadTrialSessionInteractor.js
```diff
@@ -162,6 +180,17 @@ const batchDownloadTrialSessionInteractor = async (
     });
   };
 
+  const onError = error => {
+    applicationContext.getNotificationGateway().sendNotificationToUser({
```
mmarcotte, line 184: should we also log this to the error logs? Otherwise it's just the user who will see this error message?

matthopson: Yeah, I actually removed that by accident...
https://github.com/ustaxcourt/ef-cms/pull/1070/files#diff-3e2d7a71712966189525afdc55445ca74523cc3da42c233a36eebf1f62756fe2L73

matthopson: (well, technically I removed it on purpose to test something, but forgot to put it back)

kkoskelin (approved):

mmarcotte (approved):

PR #1071, opened 2021-04-15
8116 - removing duplicate contactId from secondary when present

codyseibert:
This was to fix an issue with a practitioner showing as associated to both petitioners on the case for some cases (see 101-21 or 103-21 on migration and add a practitioner to the primary and you'll see it also say it's attached to the secondary).  The reason for this was that both contactPrimary.contactId and contactSecondary.contactId were matching which was the root of the bug.  This data duplication was happening when the user clicked the "Use same name & address" checkbox in the petition QC page which was copying over the contactId from the primary and putting it into the secondary.  This also includes a migration to generate a new UUID for contactSecondary.contactId if it happens to match the primary contactId.


codyseibert:
Don't merge this, I'm fixing a bug

kkoskelin (approved):

mmarcotte (approved):

PR #1072, opened 2021-04-15
8108: track previously-focused element

kkoskelin:
when FocusLock is removed from DOM, place focus back on the element which last had focus prior to FocusLock
attempt to set focus on first focusable child on render


mmarcotte (approved):
this is cool!

ericsorenson (approved):

PR #1074, opened 2021-04-15
7466 - adding logic to support firmName updates

codyseibert:
Displaying the Firm Name input box for privatePractitioners on the edit contact information page.  This required adding in some logic to the proxy & interaction to handle updates to only the firmName by itself so that it didn't go and generate changeOfAddress pdfs.

Screen Shot 2021-04-15 at 5 16 28 PM

### DIFF
--- a/shared/src/business/useCases/users/updateUserContactInformationInteractor.test.js
+++ b/shared/src/business/useCases/users/updateUserContactInformationInteractor.test.js
@@ -48,6 +48,7 @@ describe('updateUserContactInformationInteractor', () => {
       birthYear: '1902',
       employer: EMPLOYER_OPTIONS[2],
       entityName: irsPractitionerEntityName,
+      firmName: 'broken',
       firstName: 'Roy',
       lastName: 'Rogers',
       originalBarState: 'OR',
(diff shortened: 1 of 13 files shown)


codyseibert:
Any clue why this only ran 3 checks?  It didn't run integration tests or anything else.

shared/src/business/useCases/users/updateUserContactInformationInteractor.test.js
```diff
@@ -304,4 +305,56 @@ describe('updateUserContactInformationInteractor', () => {
...
+    await updateUserContactInformationInteractor(applicationContext, {
+      contactInfo,
+      firmName: mockUser.firmName,
+      userId: 'f7d90c05-f6cd-442c-a168-202db587f16f',
+    });
+
+    expect(
+      applicationContext.getPersistenceGateway().updateUser,
+    ).not.toBeCalled();
+  });
```
mmarcotte: Do my eyes deceive me, or is this a duplicate test?

mmarcotte (changes requested):
Looks like one dupe test in there.

shared/src/business/useCases/users/updateUserContactInformationInteractor.test.js
codyseibert: I think so, fixing now.

mmarcotte (approved):

rachaelparris (approved):

PR #1110, opened 2021-04-23
Webpack for Client bundles

mmarcotte:


web-client/src/views/CreateOrder/TextEditor.jsx
```diff
@@ -2,12 +2,15 @@
...
 import { QuillDeltaToHtmlConverter } from 'quill-delta-to-html';
 import React, { Suspense, useEffect, useRef } from 'react';
+import reactQuill from 'react-quill';
 
 const inlineStylesFontSizes = {};
 const fontSizes = ['10px', '12px', '14px', '16px', '18px', '20px'];
 
 const ReactQuill = React.lazy(async () => {
-  const reactQuill = await import('react-quill');
+  // TODO: make this import lazy again
```
mmarcotte: Just checking. Is there a refactoring backlog item for this, or is the @TODO standing on its own?

mmarcotte (commented):
You're a badass, @ericsorenson

web-client/src/views/CreateOrder/TextEditor.jsx
ericsorenson: Just forgot to remove that TODO.

PR #1111, opened 2021-04-23
Devex 932: make docketNumber required on DocketEntry records

mmarcotte:


web-api/migration-terraform/main/lambdas/migration-segments.js
```diff
@@ -30,12 +33,14 @@ const sqs = new AWS.SQS({ region: 'us-east-1' });
...
+  applicationContext.logger.error('about to run migration 0028');
   items = await migration0028(items, documentClient);
-  applicationContext.logger.info('about to run migration 0029');
+  applicationContext.logger.error('about to run migration 0029');
   items = await migration0029(items, documentClient);
+  applicationContext.logger.error('about to run migration 0030');
+  items = await migration0030(items, documentClient);
 
-  applicationContext.logger.info('about to run validation migration');
+  applicationContext.logger.error('about to run validation migration');
```
mmarcotte: Curious why the switch to `.error` instead of `.info` on these?

Also, I think it might be time to switch these to `.debug`. We haven't been making much use of them once they make it to the Court. 🤔

mmarcotte (commented):
Looks good, one quick change. Also, looks like you based this branch off of the Progress Bar work. Will have to rely on that work getting approved first unless you are able to rebase.

web-api/migration-terraform/main/lambdas/migration-segments.js
rachaelparris: Whoops, that was just for me to see some logs locally. I can switch them to debug!

PR #1112, opened 2021-04-23
8156 - Bug - Trial Session Download Logging

mmarcotte:
https://github.com/ustaxcourt/ef-cms/pull/1104

> We've discovered that when testing this on very large trial sessions with cases with lots of docket entries, this process is a bit slow. One option was to add some kind of indicator that this is being done, but after chatting with @klohman about it, it seemed like doing this step on every download was impacting users far more than the few that will fail.
> 
> There was already better error handling / logging in a previous issue related to downloads, the intent of these code changes on top of that was for diagnosing when cases fail due to unreachable documents - something that did happen, but should not happen that often.
> 
> To both provide a better user experience and preserve the diagnostic value in checking the files first that logging was left in place in this PR, but wrapped in a conditional that can be passed via a REST client on an offending trial session to generate better logs in the system.


mmarcotte (commented):
approve 👍

PR #1114, opened 2021-04-23
8101 progress a11y

mmarcotte:
https://github.com/ustaxcourt/ef-cms/pull/1101
> throttle status updates to every 3 seconds
> js-driven focus on progress indicator assists a11y users


web-client/src/presenter/computeds/fileUploadStatusHelper.test.js
```diff
@@ -112,11 +120,68 @@ describe('fileUploadStatusHelper', () => {
...
+        state: {
+          fileUploadProgress: {
+            isUploading: true,
+            noThrottle: false,
+            timeRemaining: 30,
+          },
+        },
+      });
+
+      expect(result2.statusMessage).toEqual(result1.statusMessage);
```
mmarcotte, line 156: While certainly efficient, I think in case this test ever changes in the future, I'd just put 'Preparing Upload' in this expect. Usually better to explicitly state what we are expecting something to be.

web-client/src/presenter/computeds/fileUploadStatusHelper.test.js
```diff
@@ -112,11 +120,68 @@ describe('fileUploadStatusHelper', () => {
...
+        state: {
+          fileUploadProgress: {
+            isUploading: true,
+            noThrottle: false,
+            timeRemaining: 30,
+          },
+        },
+      });
+
+      expect(result2.statusMessage).not.toEqual(result1.statusMessage);
```
mmarcotte, line 184: Same here, unless I'm missing something obvious. (Sorry if I am.)

web-client/src/presenter/computeds/fileUploadStatusHelper.test.js
kkoskelin: If the test case text changes, I change it just on line 144.  As for the way this `expect` statement is written, I feel it clearly conveys that "this should equal the previous thing".

web-client/src/presenter/computeds/fileUploadStatusHelper.test.js
kkoskelin: here, in my mind it says "this should NOT equal the previous thing".

kkoskelin (approved):

PR #1125, opened 2021-04-26
8199 - Live Migrations

codyseibert:
The migration lambda that is run when in the middle of a blue / green deploy which migrates LIVE data from source -> destination table was never fully implemented.  This PR runs the exact same migration scripts used in the segment migrations so that if someone updates data during a blue / green deploy, those updates will get migrated over as well in the new format.

### DIFF
--- a/web-api/migration-terraform/main/lambdas/migration.js
+++ b/web-api/migration-terraform/main/lambdas/migration.js
@@ -47,5 +47,9 @@ exports.getFilteredGlobalEvents = getFilteredGlobalEvents;
 exports.processItems = processItems;
 exports.handler = async event => {
   const items = getFilteredGlobalEvents(event);
-  await processItems({ documentClient: docClient, items, migrations });
+  await processItems({
+    documentClient: docClient,
+    items,
+    migrateRecords: migrations,
+  });
 };
--- a/web-api/migration-terraform/main/lambdas/migration.js
+++ b/web-api/migration-terraform/main/lambdas/migration.js
@@ -15,7 +15,6 @@ const documentClient = new AWS.DynamoDB.DocumentClient({
 
 const processItems = async items => {
   const promises = [];
-  await processItems();
 
   items = migrateRecords({ documentClient, items });
 
--- a/web-api/migration-terraform/main/lambdas/migration.js
+++ b/web-api/migration-terraform/main/lambdas/migration.js
@@ -1,6 +1,6 @@
 const AWS = require('aws-sdk');
 const { get } = require('lodash');
-const { migrateRecords } = require('./migration-segments');
+const { migrateRecords: migrations } = require('./migration-segments');
 
 const dynamodb = new AWS.DynamoDB({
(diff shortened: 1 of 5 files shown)


matthopson (approved):

rachelschneiderman (approved):

web-api/migration-terraform/main/lambdas/migration.js
```diff
@@ -1,41 +1,55 @@
...
-      .promise();
-    promises.push(result);
+    promises.push(
+      documentClient
+        .put({
+          Item: item,
+          TableName: process.env.DESTINATION_TABLE,
+        })
+        .promise(),
+    );
```
mmarcotte, line 29: What happens if one of these fails due to throttling? Would it just error out, and then the item would still be in the DDB Stream to be retried?

web-api/migration-terraform/main/lambdas/migration.js
```diff
@@ -1,41 +1,55 @@
...
+  return Records.filter(record => {
     // to prevent global tables writing extra data
     const NEW_TIME_KEY = 'dynamodb.NewImage.aws:rep:updatetime.N';
     const OLD_TIME_KEY = 'dynamodb.OldImage.aws:rep:updatetime.N';
     const newTime = get(record, NEW_TIME_KEY);
     const oldTime = get(record, OLD_TIME_KEY);
     return newTime && newTime !== oldTime;
   }).map(item => AWS.DynamoDB.Converter.unmarshall(item.dynamodb.NewImage));
-  await processItems(items);
+};
```
mmarcotte, line 44: Would these writes need to happen _after_ the migration is complete? 

In other words, is there risk in this scenario happening?

- a b/g migration is kicked off
- a user makes an update to a Case DDB record
- this lambda picks it up and writes it to the destination table _before_ the migration-segments Lambda does
- the migration-segments.js Lambda overwrites the newer record with stale information.

rachelschneiderman: ^ Fantastic question. I would love to learn more about this if y'all have a zoom conversation about it!

web-api/migration-terraform/main/lambdas/migration.js
codyseibert: From what I know, Kinesis doesn't move onto the next batch of events unless it processes everything successfully.  It will keep retrying the batch.  We have it set to retry to 30 `maximum_retry_attempts        = 30`.  Additionally, the dynamodb driver is set to   `maxRetries: 10,`, so I don't think throttling would be an issue.

web-api/migration-terraform/main/lambdas/migration.js
codyseibert: I addressed this issue by adding a conditional write on the segments-migration script.  If the record already exists in the destination table (since we always plan to keep the destination table empty before a migration), it will skip migrating the record since a user more than likely did a manual update on a record

PR #1133, opened 2021-04-28
8188: caption spaces problem

kkoskelin:
test


mmarcotte:
Please re-open with the label `bugfix`

PR #1142, opened 2021-04-28
8188: fixing caption space bug

kkoskelin:
unrelated: log more detail if duplicate docket numbers are found

### DIFF
--- a/shared/src/business/entities/cases/Case.test.js
+++ b/shared/src/business/entities/cases/Case.test.js
@@ -1652,6 +1652,40 @@ describe('Case entity', () => {
       });
       expect(caseCaption).toEqual('Bob Smith, Petitioner');
     });
+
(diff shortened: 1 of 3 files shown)


shared/src/business/entities/cases/Case.test.js
```diff
@@ -1639,6 +1639,19 @@ describe('Case entity', () => {
...
+        petitioners: [
+          {
+            contactType: CONTACT_TYPES.primary,
+            name: '    Bob Smith                 ',
+          },
+        ],
+      });
+      expect(caseCaption).toEqual('Bob Smith, Petitioner');
+    });
   });
```
matthopson, line 1690: Should we test secondary, and titles here to ensure branch coverage?

matthopson (approved):
Small nit re: test coverage

mmarcotte (approved):

PR #1147, opened 2021-04-29
7552: sort trial session cases for report according to docket number

kkoskelin:
Refactored to permit separate testing of case filtering and sorting;
Changed persistence to immediately fetch all case info before composing the result rather than awaiting each call.


shared/src/business/useCases/trialSessions/generateTrialCalendarPdfInteractor.test.js
```diff
@@ -109,6 +114,25 @@ describe('generateTrialCalendarPdfInteractor', () => {
     );
   });
 
+  describe('format cases', () => {
+    it('should filter out cases not removed from trial', () => {
```
matthopson: SHould that read "should filter out cases removed from trial" ?

matthopson (approved):

mmarcotte (approved):

PR #1182, opened 2021-05-05
V2 api add firm name

mmarcotte:
In order to include more information about a case, we have received a request to include `firmName` for a Practitioner in the Cases endpoint in the V2 API. This code adds Firm Name to the Practitioner Schema. The information was already readily available. This also updates the tests for the getCaseLambda as the firmName was specified in the mock data.

docs/api/v2.yaml is also updated to reflect the addition.

### DIFF
--- a/docs/api/v2.yaml
+++ b/docs/api/v2.yaml
@@ -271,6 +271,7 @@ components:
           example: privatePractitioner1@example.com
         firmName:
           type: string
+          example: Some Firm
         name:
           type: string
           example: Test private practitioner1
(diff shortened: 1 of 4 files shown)


kkoskelin (approved):
Presuming that because this is an addition to the fields available and won't break anything that was already expected that this particular `v2` modification is completely safe to make.

mmarcotte:
@kkoskelin -- good observation. Yes, that's partially my thinking. The rest is because the primary consumer of this API (the IRS) requested this change _without_ a version increase. Seemed like a straightforward addition.

PR #1190, opened 2021-05-07
7945: show progress indicator

kkoskelin:
prevents double-submit


rachelschneiderman (approved):
Is this worth a sequence test?

kkoskelin:
> Is this worth a sequence test?

@rachelschneiderman I'm open to suggestions on the best way to write that test.  During the execution of the sequence, a progress flag is set in the state, but then it gets unset as we exit.

I could maybe mock the `showSequenceProgressDecorator` and verify that it gets called?

edit: I tried to mock the decorator, but couldn't get the test to properly declare the sequence. The juice might not be worth the squeeze. 🍋  🥃

rachelschneiderman:
Thank you for trying! I agree, if it's proving difficult, abandon :)

mmarcotte (approved):

PR #1191, opened 2021-05-07
8081: search on pk and sk

kkoskelin:
ensure the records we are retrieving are user records and not user/case mappings

### DIFF
--- a/shared/src/persistence/elasticsearch/getPractitionersByName.js
+++ b/shared/src/persistence/elasticsearch/getPractitionersByName.js
@@ -1,7 +1,7 @@
 const {
   MAX_SEARCH_CLIENT_RESULTS,
-  ROLES,
 } = require('../../business/entities/EntityConstants');
+const { IS_PRACTITIONER } = require('./helpers/searchClauses');
 const { search } = require('./searchClient');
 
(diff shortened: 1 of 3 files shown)


rachaelparris (approved):

rachaelparris:
Is there any type of test that could be written to cover this?

kkoskelin:
> Is there any type of test that could be written to cover this?

@rachaelparris I am open to suggestions on how to better test persistence methods connected to ES.  In the past there have been tests which basically say, "Make sure the search client gets JSON that looks like [this]" but it's not clear to me whether those tests actually have any value.

rachaelparris:
> > Is there any type of test that could be written to cover this?
> 
> @rachaelparris I am open to suggestions on how to better test persistence methods connected to ES. In the past there have been tests which basically say, "Make sure the search client gets JSON that looks like [this]" but it's not clear to me whether those tests actually have any value.

Maybe an integration test that performs a practitioner search that would have previously returned duplicates, and expect that only a single result is returned for each practitioner? There may even be an existing one you could just expand on to ensure there aren't duplicates.

mmarcotte (approved):

PR #1197, opened 2021-05-10
7599: remove coversheets if necessary

kkoskelin:
Added interactor to remove coversheet from a document which previously had one.  This was a use-case which had not yet been covered.

Before, when docket entry required a coversheet:
Screen Shot 2021-05-10 at 12 09 30 PM
After, when entry was changed to event code MISC which requires none (and coversheet is removed):
Screen Shot 2021-05-10 at 12 10 27 PM

### DIFF
--- a/shared/src/business/useCaseHelper/coverSheets/removeCoversheet.js
+++ b/shared/src/business/useCaseHelper/coverSheets/removeCoversheet.js
@@ -64,11 +64,13 @@ exports.removeCoversheet = async (
     throw err;
   }
 
-  const { numberOfPages, pdfData: newPdfData } =
-    await exports.removeCoverFromPdf({
-      applicationContext,
-      pdfData,
-    });
+  const {
+    numberOfPages,
+    pdfData: newPdfData,
+  } = await exports.removeCoverFromPdf({
+    applicationContext,
+    pdfData,
+  });
 
   docketEntryEntity.setAsProcessingStatusAsCompleted();
   docketEntryEntity.setNumberOfPages(numberOfPages);
--- a/shared/src/business/useCaseHelper/coverSheets/removeCoversheet.js
+++ b/shared/src/business/useCaseHelper/coverSheets/removeCoversheet.js
@@ -1,4 +1,4 @@
-const { Case } = require('../entities/cases/Case');
+const { Case } = require('../../entities/cases/Case');
 
 /**
  * a helper function which removes a coversheet from a pdf and returns the new pdf data
@@ -24,7 +24,7 @@ exports.removeCoverFromPdf = async ({ applicationContext, pdfData }) => {
 };
 
 /**
- * removeCoversheetInteractor
+ * removeCoversheet
  *
  * @param {object} applicationContext the application context
  * @param {object} providers the providers object
@@ -32,7 +32,7 @@ exports.removeCoverFromPdf = async ({ applicationContext, pdfData }) => {
  * @param {string} providers.docketNumber the docket number of the case
  * @returns {Promise<*>} updated docket entry entity
  */
-exports.removeCoversheetInteractor = async (
+exports.removeCoversheet = async (
   applicationContext,
   { docketEntryId, docketNumber },
 ) => {
@@ -64,13 +64,11 @@ exports.removeCoversheetInteractor = async (
     throw err;
   }
 
-  const {
-    numberOfPages,
-    pdfData: newPdfData,
-  } = await exports.removeCoverFromPdf({
-    applicationContext,
-    pdfData,
-  });
+  const { numberOfPages, pdfData: newPdfData } =
+    await exports.removeCoverFromPdf({
+      applicationContext,
+      pdfData,
+    });
 
   docketEntryEntity.setAsProcessingStatusAsCompleted();
   docketEntryEntity.setNumberOfPages(numberOfPages);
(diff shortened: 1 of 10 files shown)


TomElliottFlexion (approved):

shared/src/business/useCases/removeCoversheetInteractor.js
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * removeCoversheetInteractor
+ *
+ * @param {object} applicationContext the application context
+ * @param {object} providers the providers object
+ * @param {string} providers.docketEntryId the docket entry id
+ * @param {string} providers.docketNumber the docket number of the case
+ * @returns {Promise<*>} updated docket entry entity
+ */
+exports.removeCoversheetInteractor = async (
```
rachaelparris: You _could_ make this a use case helper instead of an interactor. I realize it's just a name change, but I think it makes it more clear that it could be called from multiple places and isn't a standalone use case.

matthopson (approved):

shared/src/business/useCases/removeCoversheetInteractor.js
ericsorenson: Excellent point @rachaelparris !

shared/src/business/useCases/docketEntry/updateDocketEntryMetaInteractor.js
```diff
@@ -122,12 +122,14 @@ exports.updateDocketEntryMetaInteractor = async (
   const entryRequiresCoverSheet = COURT_ISSUED_EVENT_CODES_REQUIRING_COVERSHEET.includes(
     editableFields.eventCode,
   );
-  const originalEntryDoesNotRequireCoversheet = !COURT_ISSUED_EVENT_CODES_REQUIRING_COVERSHEET.includes(
+  const originalEntryRequiresCoversheet = COURT_ISSUED_EVENT_CODES_REQUIRING_COVERSHEET.includes(
```
ericsorenson, line 125: 🎉

shared/src/business/useCases/docketEntry/updateDocketEntryMetaInteractor.js
```diff
@@ -182,6 +184,21 @@ exports.updateDocketEntryMetaInteractor = async (
...
+  } else if (shouldRemoveExistingCoverSheet) {
+    await applicationContext.getPersistenceGateway().updateDocketEntry({
+      applicationContext,
+      docketEntryId: docketEntryEntity.docketEntryId,
+      docketNumber,
+      document: docketEntryEntity.validate(),
+    });
+    const updatedDocketEntry = await applicationContext
+      .getUseCases()
+      .removeCoversheetInteractor(applicationContext, {
```
ericsorenson: We're trying to move away from interactors calling other interactors.

rachaelparris (approved):

matthopson (approved):

TomElliottFlexion (approved):

shared/src/business/useCaseHelper/coverSheets/removeCoversheet.js
```diff
@@ -0,0 +1,94 @@
...
+ * @param {object} options the providers object
+ * @param {object} options.applicationContext the application context
+ * @param {object} options.pdfData the original document pdf data
+ * @returns {object} the new pdf with coversheet removed
+ */
+exports.removeCoverFromPdf = async ({ applicationContext, pdfData }) => {
+  const { PDFDocument } = await applicationContext.getPdfLib();
+  const pdfDoc = await PDFDocument.load(pdfData);
+
+  pdfDoc.removePage(0);
```
mmarcotte, line 15: I wish there was some way to know this was a cover page and not just a page in the document. 🤔

shared/src/business/useCaseHelper/coverSheets/removeCoversheet.js
```diff
@@ -0,0 +1,94 @@
...
+  docketEntryEntity.setNumberOfPages(numberOfPages);
+
+  const updatedDocketEntryEntity = docketEntryEntity.validate();
+
+  await applicationContext.getPersistenceGateway().updateDocketEntry({
+    applicationContext,
+    docketEntryId,
+    docketNumber: caseEntity.docketNumber,
+    document: updatedDocketEntryEntity.toRawObject(),
+  });
```
mmarcotte, line 85: Doesn't it make more sense to have the updateDocketEntry validate before it saves to the database, like inside of `updateDocketEntry(...)`? It would seem like we would want to have this happen every time, no? Save us a line of code (and brainpower to remember to do it) every time we updateDocketEntry.

kkoskelin: It certainly would be a good thing to do this.  Perhaps in a use-case helper that performs both validation and then calls persistence.  @ericsorenson possibly another thing to consider in our devex task?  Meanwhile, @mmarcotte do you have objections to this being merged, otherwise?

shared/src/business/useCaseHelper/coverSheets/removeCoversheet.js
kkoskelin: So do I!  Maybe a new story can cover this, storing metadata about documents so we know for certain.

mmarcotte (approved):

shared/src/business/useCaseHelper/coverSheets/removeCoversheet.js
mmarcotte: added to trello :)

PR #1205, opened 2021-05-12
8207 - Bug - Disable / Re-Enable DDB Stream During Migration Deploy

matthopson:
During a migration deploy, this disables the dynamo stream that triggers indexing on ES until after the migration has been completed. Re-enabling the stream allows ES to catch up on the writes to the new table.

Closed https://github.com/ustaxcourt/ef-cms/pull/1203 due to weirdness with the PR


.circleci/config.yml
```diff
@@ -724,13 +735,25 @@ jobs:
           name: Wait for SQS Queue to Empty
           command: |
             if [ "$MIGRATE_FLAG" == "true" ]; then
+              no_output_timeout: 5h
```
kkoskelin, line 738: I know I'm late to the party, but -- I'd like to learn more about this line, what it's doing, and why.  I don't object, just want to learn. :)

mmarcotte: since the migration can take upwards of an hour with a production-sized data set and the current wait until complete script doesn't output anything, this increases the timeout time from 10m (default) to a range that would accommodate a prod-sized blue/green migration.

i think the final version should probably bring this down to 2h _or_ have the script output status with how many messages it received. could help provide a status update.

PR #1211, opened 2021-05-13
7913: progress indicator for private practitioners

kkoskelin:
prevent progress bar from showing if practitioner is updating *only* their firm name.
prevent "NaN%" from being displayed due to division-by-zero, instead show "0%" if this should ever be calculated.
Do not close & reopen sockets, which is one possible cause of socket messages being dropped and not processed by the application.

### DIFF
--- a/shared/src/business/useCases/users/generateChangeOfAddress.js
+++ b/shared/src/business/useCases/users/generateChangeOfAddress.js
@@ -53,15 +53,17 @@ exports.generateChangeOfAddress = async ({
     });
 
   let completedCases = 0;
-  await applicationContext.getNotificationGateway().sendNotificationToUser({
-    applicationContext,
-    message: {
-      action: `${websocketMessagePrefix}_contact_update_progress`,
-      completedCases,
-      totalCases: docketNumbers.length,
-    },
-    userId: requestUserId || user.userId,
-  });
+  if (docketNumbers.length > 0) {
+    await applicationContext.getNotificationGateway().sendNotificationToUser({
+      applicationContext,
+      message: {
+        action: `${websocketMessagePrefix}_contact_update_progress`,
+        completedCases,
+        totalCases: docketNumbers.length,
+      },
+      userId: requestUserId || user.userId,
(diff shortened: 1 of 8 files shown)


matthopson (approved):

shared/src/business/useCases/users/generateChangeOfAddress.js
```diff
@@ -53,15 +53,17 @@ exports.generateChangeOfAddress = async ({
...
+    await applicationContext.getNotificationGateway().sendNotificationToUser({
+      applicationContext,
+      message: {
+        action: `${websocketMessagePrefix}_contact_update_progress`,
+        completedCases,
+        totalCases: docketNumbers.length,
+      },
+      userId: requestUserId || user.userId,
+    });
+  }
```
mmarcotte: Looking at this and the code, I think it would be simpler to have this block just return an empty array `[]` if `docketNumbers.length === 0` since the rest of the code also hinges on there being docketNumbers.

web-client/src/providers/socket.js
```diff
@@ -22,34 +22,34 @@ export const socketProvider = ({ socketRouter }) => {
...
+            // the socket needs to be open for a short period or it could miss the first message
+            setTimeout(() => {
+              resolve();
+            }, 300);
+          };
+        } catch (e) {
+          if (applicationContext) {
+            applicationContext.logger.error(
+              'Failed to establish WebSocket connection',
+              e,
```
mmarcotte: Just because we have seen [some other strange behavior passing the Error object along](https://github.com/ustaxcourt/ef-cms/pull/1171/files#diff-f5f2fff53bc2682640ec3a00b96baf967e5c45c6e2c428f8ec06936f0f3edd63R276), can you encapsulate `e` in an object

```
applicationContext.logger.error(
  'Failed to establish WebSocket connection',
  { e },
);
```

mmarcotte (changes requested):

mmarcotte (approved):

PR #1213, opened 2021-05-14
8297 - Bug - File Document Without Served Petition

matthopson:
This checks for a served petition before showing the File a Document button for external parties, as well as prevents filing of a document on a case without a served petition at the interactor level.


matthopson:
Closing per discussion with PO.

PR #1233, opened 2021-05-24
8303 - Update Case Type With Statistics - Review Fixes

matthopson:
Addresses some feedback from review of 8303

### DIFF
--- a/shared/src/business/useCases/trialSessions/updateTrialSessionInteractor.js
+++ b/shared/src/business/useCases/trialSessions/updateTrialSessionInteractor.js
@@ -147,6 +147,18 @@ exports.updateTrialSessionInteractor = async (
           caseToUpdate: caseEntity,
         });
       }
+
+      const matchingHearing = caseToUpdate.hearings.find(
+        hearing =>
+          hearing.trialSessionId == newTrialSessionEntity.trialSessionId,
+      );
+      if (matchingHearing) {
+        applicationContext.getPersistenceGateway().updateCaseHearing({
+          applicationContext,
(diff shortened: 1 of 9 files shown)


kkoskelin (approved):

shared/src/business/useCases/trialSessions/updateTrialSessionInteractor.js
```diff
@@ -147,6 +147,18 @@ exports.updateTrialSessionInteractor = async (
...
+
+      const matchingHearing = caseToUpdate.hearings.find(
+        hearing =>
+          hearing.trialSessionId == newTrialSessionEntity.trialSessionId,
+      );
+      if (matchingHearing) {
+        applicationContext.getPersistenceGateway().updateCaseHearing({
+          applicationContext,
+          docketNumber: calendaredCase.docketNumber,
+          hearingToUpdate: newTrialSessionEntity.validate().toRawObject(),
```
mmarcotte, line 159: I still don't quite grasp why it makes more sense to `validate` and `toRawObject` something before/as we call this method rather than from within `updateCaseHearing`. It seems like something we would do every time, and so why code it every time rather than once? @kkoskelin I know you mentioned something about this pattern yesterday during the refactoring meeting.

matthopson: @mmarcotte I would just point out that is from another PR. Those top 3 commits are not part of my PR. Maybe test out of sync?

mmarcotte (approved):

PR #1239, opened 2021-05-25
7450: remove event listener

kkoskelin:
a listener was preventing `touchmove` scrolling on ipad (possibly others).  Original intent was to prevent the background from scrolling, but this is currently deemed a harmless price to pay so that these "tall" modals can scroll as desired for ipad users.
future strategy: prefer non-modal interfaces when content has any chance of over-running the height of the viewport.

### DIFF
--- a/web-client/src/providers/socket.js
+++ b/web-client/src/providers/socket.js
@@ -28,7 +28,7 @@ export const socketProvider = ({ socketRouter }) => {
           socket = createWebSocketClient(token);
           socket.onmessage = socketRouter(app);
           socket.onerror = error => {
-            applicationContext.logger.error('Websocket error detected', error);
+            console.error(error);
             return reject(error);
           };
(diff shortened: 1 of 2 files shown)


matthopson (approved):

web-client/src/providers/socket.js
```diff
@@ -28,7 +28,7 @@ export const socketProvider = ({ socketRouter }) => {
           socket = createWebSocketClient(token);
           socket.onmessage = socketRouter(app);
           socket.onerror = error => {
-            applicationContext.logger.error('Websocket error detected', error);
+            console.error(error);
```
rachelschneiderman, line 31: Were these changes intentional?

kkoskelin: These are odd merge artifacts from another PR.

PR #1247, opened 2021-05-26
7518 - Bug - Keep draft selected while on case

matthopson:
When navigating within a case, this will keep the last selected draft document the default when returning to the Drafts tab. When leaving the case and coming back, it will default to the first document in the list.


rachaelparris (approved):

web-client/integration-tests/petitionsClerkViewsDraftDocuments.test.js
```diff
@@ -0,0 +1,107 @@
...
+      primaryTab: 'drafts',
+    });
+
+    expect(
+      test.getState(
+        'currentViewMetadata.caseDetail.caseDetailInternalTabs.drafts',
+      ),
+    ).toEqual(true);
+
+    // this gets fired when the component is mounted, so it is being explicitly called here
```
kkoskelin, line 50: 🌮

kkoskelin (approved):

mmarcotte (approved):

PR #1266, opened 2021-05-27
7943: No edit link appears on Correspondence document when viewed from message inbox preview

rachelschneiderman:
Before:
image

After: 
image

### DIFF
--- a/web-client/src/presenter/computeds/messageDocumentHelper.test.js
+++ b/web-client/src/presenter/computeds/messageDocumentHelper.test.js
@@ -4,41 +4,41 @@ import { messageDocumentHelper as messageDocumentHeperComputed } from './message
 import { runCompute } from 'cerebral/test';
 import { withAppContextDecorator } from '../../withAppContext';
 
-const { USER_ROLES } = applicationContext.getConstants();
-
-const messageDocumentHelper = withAppContextDecorator(
-  messageDocumentHeperComputed,
-  applicationContext,
-);
-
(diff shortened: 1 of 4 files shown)


web-client/src/presenter/computeds/messageDocumentHelper.test.js
```diff
@@ -4,41 +4,41 @@ import { messageDocumentHelper as messageDocumentHeperComputed } from './message
...
-const getBaseState = user => {
-  return {
-    permissions: getUserPermissions(user),
+  const getBaseState = user => {
+    return {
+      permissions: getUserPermissions(user),
+    };
+  };
+
+  const docketClerkUser = {
```
rachaelparris, line 21: Nit: I know you just moved these lines, but they can be imported from `/shared/src/test/mockUsers` rather than redefined here, if you'd like to clean this up a bit.

rachaelparris (approved):

web-client/src/presenter/computeds/messageDocumentHelper.test.js
rachaelparris: Wait, nevermind, that's only on develop. Resolving this!

rachelschneiderman: ~Thank you! Love it!!!~ Darn it....

mmarcotte (approved):

PR #1276, opened 2021-05-28
Bug 8368: refactor work item inbox records

rachaelparris:
Delete work item inbox records and instead use the main work item record (pk `work-item|{workItemId}` and sk `work-item|{workItemId}`) with appropriate filters.


rachelschneiderman (approved):

mmarcotte (approved):
clap clap clap

PR #1283, opened 2021-06-01
Bug 7946: File Upload element for "Supporting Document" is not consistent with other file inputs

rachelschneiderman:
Before:
image

After:
image


web-client/src/views/FileDocument/StateDrivenFileInput.jsx
```diff
@@ -9,6 +9,7 @@ export const StateDrivenFileInput = connect(
   {
     ariaDescribedBy: props.ariaDescribedBy,
     constants: state.constants,
+    file: props.file,
```
mmarcotte, line 12: Have a look at the linting catch below. `file` is already defined. We might want to break out that function or use a more descriptive name here for `file`.

mmarcotte (changes requested):

kkoskelin (approved):
Once `no-shadow` rules are in the clear, you betcha.

mmarcotte (approved):

PR #1293, opened 2021-06-01
Bug 8008: PDF Signature Stamp Slightly Off

rachelschneiderman:
Applying Signature (updated after PR comments, PT 2):
image

Signature Applied (updated after PR comments, PT 2):
image

### DIFF
--- a/shared/src/business/useCases/generateSignedDocumentInteractor.js
+++ b/shared/src/business/useCases/generateSignedDocumentInteractor.js
@@ -140,21 +140,21 @@ exports.generateSignedDocumentInteractor = async ({
 
   const { signatureName, signatureTitle } = sigTextData;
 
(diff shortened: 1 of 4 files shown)


mrinsin (approved):

kkoskelin:
It appears that the fonts are still unmatched (see the lower-case G).  I think the font mismatch is the primary issue... isn't it?

kkoskelin:
Please forgive the observation, but we've traded one font for another which is also a mismatch, I think?  Probably the original font seen in the "final product" is the one we want to use.  Can we find a way to use it in all places, not only in the final product but also in the canvas signature-application tool?

rachelschneiderman:
Nothing to forgive! Thanks for your feedback. I'll work on standardizing!

mmarcotte (approved):

PR #1294, opened 2021-06-01
8325 adding back legacy migration code

codyseibert:


codyseibert:
I actually see code where we catch the exception if the file is not found in s3 so that we don't retry the event... I'm going to close this.

PR #1303, opened 2021-06-02
8384 - Ensure old mapping records are removed before creating new ones

matthopson:
When creating `eligible-for-trial-case-catalog` mapping records, ensure any currently exist records are removed. There can be only one! (well, two, actually because there is a separate hybrid record that is created).


rachelschneiderman (approved):

shared/src/persistence/dynamo/cases/createCaseTrialSortMappingRecords.js
```diff
@@ -15,6 +18,27 @@ exports.createCaseTrialSortMappingRecords = async ({
...
+    applicationContext,
+  });
+
+  if (oldSortRecords.length) {
+    await deleteCaseTrialSortMappingRecords({
+      applicationContext,
+      docketNumber,
+    });
+  }
+
```
mmarcotte, line 41: This is starting to look exactly like `updateCaseTrialSortMappingRecords.js` Do you think it would make enough sense to just consolidate these two into one?

matthopson: Yeah, I think you're right. Maybe only use `create...`?

mmarcotte (approved):

PR #1338, opened 2021-06-07
Bug 8379: prevent users from creating docket entries for same draft documents

rachaelparris:


matthopson (approved):

mmarcotte (approved):
Thanks for the refactoring in the tests, too.

PR #1346, opened 2021-06-08
Bug 8176: use consistent valid format for practitioner original bar state

rachaelparris:


shared/src/utilities/JoiValidationDecorator.js
```diff
@@ -162,6 +162,7 @@ exports.joiValidationDecorator = function (
     });
 
     if (error) {
+      console.log('entity with error-------------', this);
```
kkoskelin, line 165: Probably want to remove this?

kkoskelin (approved):

shared/src/utilities/JoiValidationDecorator.js
rachaelparris: I left it there intentionally for troubleshooting errors in migrations. It will only print out when `validateForMigration` is called and if there is an error.

web-api/migration-terraform/main/lambdas/migrations/0002-original-bar-state.js
```diff
@@ -0,0 +1,47 @@
...
+      item.pk.startsWith('user|') &&
+      item.sk.startsWith('user|') &&
+      item.role &&
+      item.role.includes('Practitioner')
+    ) {
+      if (
+        item.originalBarState &&
+        !validStateAbbreviations.includes(item.originalBarState)
+      ) {
+        const stateAbbreviation = invert(US_STATES)[item.originalBarState];
```
rachelschneiderman, line 32: Alt Text

rachelschneiderman (approved):

mmarcotte (approved):

PR #1376, opened 2021-06-15
7512: set focus on header after initial application load

kkoskelin:
When the application is loading for the very first time, the first value for `currentPage` is `Interstitial`, and the second value is the view which represents the user's dashboard.
This code will prevent us from programatically setting the keyboard / voiceover focus on the main content's header until at least the second "main view change".  Thus on initial load, voiceover and keyboard users will still step into the application by first encountering the skip-navigation link at the beginning of the page.  Thereafter, they will be directed to the content headers.

### DIFF
--- a/web-client/src/views/AppComponent.jsx
+++ b/web-client/src/views/AppComponent.jsx
@@ -165,6 +165,8 @@ const pages = {
   WorkQueue,
 };
 
+let initialPageLoaded = false;
+
 /**
  * Root application component
  */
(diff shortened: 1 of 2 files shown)


rachaelparris (approved):

adunkman (commented):
Nice solution! Looks good to me, will leave the formal green check approval to one of your teammates. 😄 

Thanks for working through all the alternatives along the way, and for suggesting https://github.com/uswds/uswds/issues/4231 to make it better for the next person stepping into your shoes!

mmarcotte (approved):

Issue #1223, opened 2021-05-20
New item


mmarcotte:
In the docs of the `ef-cms` repo, we would like to document key decisions that have been made for the application. For example, DynamoDB versus other datastores, Terraform versus Serverless, etc. 

As a new employee to the Tax Court, vendor team, or as an open source contributor, so that I can understand the trade-offs and context of decisions made, I want to be able to read documentation to bring me up to speed.

TODO: Will add more details in here.

Potential starting list: 

- [x] Deployment with Terraform over using Serverless 
    - ustaxcourt/ef-cms#6831
- [x] Using one Lambda and Express as a router over multi-lambda deployment (may be included in the above)
    - ustaxcourt/ef-cms#6831
- [x] Choice of data stores (DynamoDB, Elasticsearch, S3)
    - ustaxcourt/ef-cms-flexion#1377 
- [x] Use of infrastructure-as-a-service over platform-as-a-service
- [ ] Blue-green migrations

adunkman:
I’d like to move forward with adopting [Architecture Decision Records](https://adr.github.io/) — it’s a pretty lightweight process, effectively just adopting a markdown template. 

https://engineering.18f.gov/architecture-reviews/ maintains a list of repos which have example ADR documentation, which are useful to get an idea of what it looks like in practice.

mmarcotte:
@adunkman and @mmarcotte might have the original proposal from Flexion that may offer hints to these decisions. @JessicaMarine to help connect the dots with the proposal.

adunkman:
Next up is getting the blue/green migration approach documented

mmarcotte:
Created these tickets on Flexion's board to help continue and wrap up the conversations.
- https://github.com/flexion/ef-cms/issues/8816
- https://github.com/flexion/ef-cms/issues/8815
- https://github.com/flexion/ef-cms/issues/8814

mmarcotte:
Possibly an interesting onboarding pairing activity with new hire Jim Lerza. Good opportunity to show Jim where the documentation for decision records are in the repo and learn why we are where we are. 🍪 🐦🐦

PR #1377, opened 2021-06-15
Document the decision to use DynamoDB.

adunkman:
As a part of #1223, this pull request introduces documentation around the decision to use DynamoDB. The purpose of this documentation is to:

- Create a pattern of documentation for major architectural choices, to help remind future team members (Future Us, or others) of the context and thoughts around how major choices were made, the mitigations we took for risks, and make revisiting or changing the decision explicit and intentional.
- Assist in onboarding new technical employees, like the DevSecOps Engineer (current open for candidates), to reduce churn on questions like "why did we do this?" which may be easily misunderstood or misinterpreted as a change proposal instead of information gathering.

To be clear, the intention is not to revisit these choices or perform analysis or postmortem reviews. 

This is my best take at recovering what is available to me — it will likely benefit from some folks who were in the room taking a look at it and filling in any missing information.


adunkman:
We discussed in standup this morning and vocally 👍’d this — if anything was missed, we’re happy to do a follow-up PR.

PR #1383, opened 2021-06-17
Bug 8429: messages occasionally not getting indexed into ES

rachaelparris:
I have not been able to reproduce this bug, but we still believe this is a solid fix for the instances where the situation we've outlined in the bug _could_ occur. 

We were able to reduce the number of queries by only querying dynamo if the message being indexed has `isRepliedTo = false`. A message will never go from `isRepliedTo = true` to `isRepliedTo = false`, so we don't have to worry about checking for up-to-date data if `isRepliedTo` is already `true`.


rosiehsmith (approved):

shared/src/business/useCases/processStreamRecordsInteractor.js
```diff
@@ -329,11 +397,23 @@ exports.processStreamRecordsInteractor = async (
       useTempBucket: false,
     });
 
+  const getMessage = ({
```
rachelschneiderman: Does this method need to be marked as async?

rachelschneiderman (approved):

shared/src/business/useCases/processStreamRecordsInteractor.js
rachaelparris: I don't believe so. I copied from `getCase` and `getDocument` above. Those seem to work correctly without `async`.

shared/src/business/useCases/processStreamRecordsInteractor.js
```diff
@@ -248,10 +256,72 @@ const processWorkItemEntries = async ({
...
+              S: messageNewImage.pk.S,
+            },
+            sk: {
+              S: messageNewImage.sk.S,
+            },
+          },
+          NewImage: marshalledMessage,
+        },
+        eventName: 'MODIFY',
+      };
```
mmarcotte: Just a gut check here. Do we even need to do _anything_ if the fetched DDB record has `isRepliedTo === true`. It would seem we would just wait until that event made it through Kinesis (or maybe it already has somehow?)

shared/src/business/useCases/processStreamRecordsInteractor.js
```diff
@@ -294,10 +363,9 @@ const processRemoveEntries = async ({ applicationContext, removeRecords }) => {
   });
 
   if (failedRecords.length > 0) {
-    applicationContext.logger.error(
-      'the records that failed to delete',
+    applicationContext.logger.error('the records that failed to delete', {
```
mmarcotte, line 366: We seem to be re-using this throughout this file. Adding something to Trello backlog to address.

shared/src/business/useCases/processStreamRecordsInteractor.js
```diff
@@ -381,6 +462,18 @@ exports.processStreamRecordsInteractor = async (
...
     );
 
+    await processMessageEntries({
+      applicationContext,
+      messageRecords,
+      utils,
+    }).catch(err => {
+      applicationContext.logger.error(
+        "failed to process message records',",
+        err,
```
mmarcotte: can you encapsulate the error into an object and clean up the log message just a bit 🙏

shared/src/business/useCases/processStreamRecordsInteractor.js
rachaelparris: I can't think of any reason why we have to. Would you like for me to make that change?

mmarcotte: Yes, please. However, in the future should we still have messages sticking around, we may opt for this method that covers that base thoroughly.

PR #1412, opened 2021-06-22
Fix v1 and v2 API for Petitioner info

mmarcotte:
A recent update migrated `CONTACT_TYPES.primary` and `CONTACT_TYPES.primary` to consolidate to `CONTACT_TYPES.petitioner` for all served cases. This caused `getContactPrimary` and `getContactSecondary` to fail on served cases and thus not passed along to the versioned APIs. This fix uses the status of the case to determine what to report for contactPrimary and contactSecondary.

It also tidies up some test data to be more diligent with `.primary`, `.secondary`, and `.petitioner` in the appropriate places (whether or not the case is new or not).


shared/src/business/entities/contacts/ContactFactory.test.js
```diff
@@ -125,7 +125,7 @@ describe('ContactFactory', () => {
     expect(caseExternal.getFormattedValidationErrors()).toEqual(null);
   });
 
-  it('passes validation when primary contact is defined and everything else is valid on a served case', () => {
+  it('passes validation when petitioner exists and everything else is valid on a served case', () => {
```
rachaelparris: I think ContactFactory is only used when a case is in `New` status now, so this whole test is irrelevant and can probably be removed.

rachaelparris (approved):
Looks great!

shared/src/business/entities/contacts/ContactFactory.test.js
mmarcotte: I had a sneaking suspicion. Thanks!

shared/src/business/entities/cases/Case.js
```diff
@@ -1589,16 +1589,43 @@ Case.prototype.setAdditionalNameOnPetitioners = function (rawCase) {
...
 
+/**
+ * Retrieves the contact from the case based on whatever contactType is specified
+ *
+ * @param {Object} providers the providers object
+ * @param {String} providers.contactType the type of contact we are looking for (primary or secondary)
+ * @param {Object} providers.rawCase the raw case
+ * @returns {(Object|undefined)} the contact object on the case
+ */
+const getContactPrimaryOrSecondary = function ({ contactType, rawCase }) {
```
rosiehsmith, line 1600: 👍

rosiehsmith (approved):

PR #1427, opened 2021-06-24
Bug 8501: View link on practitioners not working for some users 

rachaelparris:
This bug was occurring when a case was created by a private practitioner with a party type that had a contact primary and secondary, then when the case was QCed by a petitionsclerk the party type was changed to one with only a contact primary. The contact secondary was not removed from the practitioner's representing array, so the practitioner modal view was attempting to look up the name of a petitioner that was no longer on the case.


sutt0n (approved):

TomElliottFlexion (approved):

matthopson:
Reopening per Mike's request.

mmarcotte (approved):

PR #1450, opened 2021-07-06
8538 BUG: Indexing starts prematurely on Blue/Green Deploy

mrinsin:


rachaelparris (approved):

web-api/disable-deploying-dynamo-stream-trigger.sh
```diff
@@ -7,3 +7,4 @@
 
 UUID=$(aws lambda list-event-source-mappings --function-name "arn:aws:lambda:us-east-1:${AWS_ACCOUNT_ID}:function:streams_${ENV}_${DEPLOYING_COLOR}" --region us-east-1 | jq -r ".EventSourceMappings[0].UUID")
 aws lambda update-event-source-mapping --uuid "${UUID}" --region us-east-1 --no-enabled
+sleep 60
```
mmarcotte: What happens if this process takes longer than 60 seconds?

It would seem that we should wait X time (30-60 seconds or so), and then check to see if the status is Disabled or Disabling. If Disabled proceed, if Disabling wait X time and try again.

Also could this please get a newline added? 🙏

kkoskelin: Presumably if the process takes longer, then records would *still* get indexed... although perhaps not as many.  I (personally) do not know of a way to query the trigger to see if it's truly disabled, nor whether a query of that kind would tell us useful things.  I certainly think that there's little harm done in this initial 60second sleep, but we could double it, triple it, whatever.

mrinsin: To add on to what Kris said, when we were debugging this issue we noticed that even when the status was returned Disabled, as opposed to Disabling, the trigger did not truly seemed to be disabled as records were still getting indexed, which is why we chose to add a sleep. Only once we added the sleep did we see no records getting indexed!

mrinsin: However I can still increase the time, Mike, to a number of your choosing!

mmarcotte: I think it's a [pretty straightforward API call](https://docs.aws.amazon.com/lambda/latest/dg/functions-states.html) to get the status. 

```
aws lambda get-function-configuration --function-name my-function
```

mmarcotte: I think it's a [pretty straightforward API call](https://docs.aws.amazon.com/cli/latest/reference/lambda/get-event-source-mapping.html) to get the state of the trigger.

```
aws lambda get-event-source-mapping \
    --uuid "a1b2c3d4-5678-90ab-cdef-11111EXAMPLE"
```

That should let us know whether or not it's `Disabled`, `Enabled`, `Disabling`. or `Enabling` via the `State` key.

mmarcotte (approved):

PR #1453, opened 2021-07-06
8554 BUG: Unable to change filing party on entry

mrinsin:


shared/src/business/entities/EntityValidationConstants.js
```diff
@@ -114,7 +114,11 @@ const DOCKET_ENTRY_VALIDATION_RULE_KEYS = {
     .description(
       'The party who filed the document, either the petitioner or respondent on the case.',
     ),
-  filers: joi.array().items(JoiValidationConstants.UUID).optional(),
+  filers: joi.when('servedAt', {
+    is: joi.exist().not(null),
+    otherwise: joi.array().items(JoiValidationConstants.UUID).optional(),
+    then: joi.array().required(),
```
rachaelparris, line 120: Should it be required? Is there a filers array for court-issued docs?

mrinsin: It didn't seem to be required originally and when I looked into the constructor for this entity I realized why - it is always defaulted to an empty array when a value has not been passed in!

rachaelparris (approved):

mmarcotte (approved):

PR #1473, opened 2021-07-12
8536 fallback dynamodb region from staging

codyseibert:
Adding fallback logic which wraps our getDocumentClient call which is used anywhere in our application we attempt to fetch or store data into dynamodb.  This will try to make a request to the main region the API is deployed on, but fall back if the Resource does not exist (we manually deleted the table) or if AWS returns a 503 error (which we think only happens if their service is having issues).

### DIFF
--- a/shared/src/utilities/getDynamoEndpoints.js
+++ b/shared/src/utilities/getDynamoEndpoints.js
@@ -2,12 +2,17 @@ const AWS = require('aws-sdk');
 const { DynamoDB } = AWS;
 
 /**
- * deleteByGsi
+ * getDynamoEndpoints
  *
  * @param {object} providers the providers object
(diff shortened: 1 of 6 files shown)


mrinsin (approved):

matthopson (approved):
D5zH9SyxCKdC62Ei6Uwo5CyLmWyrWSsLNte9qLuRBaHNo4M62yTvdhKqeTZqpjr1ZLg3QmzZYsEiFEDVN1SDF2x69rmycCQpDQUs8GavpqM7ugdXiLA44QDRDY7JpN1U6oFuEu

kkoskelin (approved):

mmarcotte (approved):

PR #1482, opened 2021-07-14
8576: BUG remove unneeded async decorators

kkoskelin:
action had 'async' decoration.  Removed it, and suddenly sequence worked correctly.

tried a variety of things, but if the leading action in the 'ignore' path returned a promise, the second action (which had also been async) would not fire and the jsx would throw an error.


rachaelparris (approved):
Weird.

mmarcotte (approved):

PR #1536, opened 2021-07-29
8670 - remove records result in empty objects trying to be processed

codyseibert:
There is an issue with live migrations failing on REMOVE events on the dynamodb stream.  This issue was created due to an earlier refactoring where we upgraded to dynamodb 2019.  The current code was taking an empty object {} and trying to run it through the migrations scripts which would cause it to fail with `item.pk of undefined`.

### DIFF
--- a/web-api/migration-terraform/main/lambdas/migration.js
+++ b/web-api/migration-terraform/main/lambdas/migration.js
@@ -37,6 +37,13 @@ const getFilteredGlobalEvents = event => {
   );
 };
 
+const getRemoveEvents = event => {
+  const { Records } = event;
+  return Records.filter(item => item.eventName === 'REMOVE').map(item =>
+    AWS.DynamoDB.Converter.unmarshall(item.dynamodb.OldImage),
+  );
+};
+
 exports.getFilteredGlobalEvents = getFilteredGlobalEvents;
 exports.processItems = processItems;
 exports.handler = async event => {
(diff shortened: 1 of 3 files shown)


web-api/migration-terraform/main/lambdas/migration.js
```diff
@@ -32,7 +32,7 @@ const processItems = async ({ documentClient, items, migrateRecords }) => {
 
 const getFilteredGlobalEvents = event => {
   const { Records } = event;
-  return Records.map(item =>
+  return Records.filter(item => item.eventName !== 'REMOVE').map(item =>
```
adunkman, line 35: Confirming my understanding here — this would mean that records which were deleted _while a migration is running_ are not removed from the destination table, leaving orphaned records?

adunkman (commented):
Question for you!

web-api/migration-terraform/main/lambdas/migration.js
codyseibert: That sounds correct, I can try to work on a solution for this missing functionality.

adunkman: Thanks, Cody!

adunkman (approved):
This looks great — thanks for addressing the fix for missing items!

adunkman:
Will merge w/ #1542 upon [coordination](https://ustaxcourt.slack.com/archives/C015SEE9CUU/p1627917596015900).

adunkman:
I’ve run the environment-specific deploy step in the `test` environment prior to merging here, and will flag to Mike to run across the other environments as we merge.

PR #1542, opened 2021-07-30
BUG 8693: Checkboxes on Section QC Page are not working

rosiehsmith:
This is happening due to a bug in USWDS.
Awaiting a release from them which will include a fix for this 🎉 (so we don't have to ensure that the classes `usa-radio__label` and `usa-checkbox__label` are always last in class name lists)

image
image

### DIFF
--- a/web-client/src/views/FileDocument/SupportingDocumentInclusionsForm.jsx
+++ b/web-client/src/views/FileDocument/SupportingDocumentInclusionsForm.jsx
@@ -92,7 +92,7 @@ export const SupportingDocumentInclusionsForm = connect(
                 }}
               />
               <label
-                className="usa-checkbox__label inline-block supporting-document-certificate-of-service"
+                className="inline-block supporting-document-certificate-of-service usa-checkbox__label"
                 htmlFor={`${type}-certificateOfService`}
               >
                 Certificate Of Service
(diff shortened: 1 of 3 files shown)


adunkman:
Leaving myself a note for when I forget what I’m doing later — this is awaiting review by a Flexion engineer & needs coordination with confirmation in the `test` environment by @ttlenard.

rachaelparris (approved):

adunkman (commented):
Related USWDS issue — https://github.com/uswds/uswds/issues/4251. Thanks for keeping your eye on both the immediate fix and the longer-term fix! 

I would expect a fix for a like this (prevents a user from using the application) to contain a regression test asserting that this issue doesn’t present itself again — is there a pattern of regression testing that y’all have been using?

rosiehsmith:
The checkboxes don't have anything to assert against with since they still contained the `checked` class and everything in the DOM seems to match what we would expect, so a visual test would be the only thing to capture the bug (to grab a snapshot of how it looks). We don't have a pattern for that yet, and since this is something that is coming from an external library (with a fix incoming!) this doesn't seem like it warrants a new pattern?

adunkman (approved):
Thanks for the additional context! I would agree. 

Confirming timing on the merge, but I’m 👍 on the PR!

rosiehsmith:
Thanks so much @adunkman!

PR #1548, opened 2021-08-05
Reverts 7542 - Preventing Service If Document Processing Fails

matthopson:
While investigating a bug (#8717), it was determined that changes to the flow where coversheets are generated was data that did not reflect the currently processing state of the given document - which was causing an inaccurate cover sheet data being used. Further investigation showed other areas where a similar bug would be possible. Reverting work done in 7542 corrects this.


TomElliottFlexion (approved):

adunkman (approved):
This looks great — and many thanks to @TomElliottFlexion for keeping the initial pull requests and commits tidy so this was easy to review! 

- https://github.com/flexion/ef-cms/pull/8592
- https://github.com/flexion/ef-cms/pull/8632
- https://github.com/flexion/ef-cms/pull/8637

PR #1552, opened 2021-08-09
8506 - refresh migrate flag after running setup blue green if needed

codyseibert:
We added automation to turn on the migrate true flag, but circle was keeping a reference to an old cache value which was causing the terraform API to run thinking the migrate flag was false even though the script prior turned it on.

same issue with destination table

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -379,6 +379,9 @@ jobs:
               -e "ENV=${ENV}" \
               --rm efcms /bin/sh -c "./setup-for-blue-green-migration.sh"
       - run:
+          name: Setup Migrate Flag
+          command: echo "export MIGRATE_FLAG=$(./scripts/get-migrate-flag.sh $ENV)" >> $BASH_ENV
+      - run:
           no_output_timeout: 20m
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -378,6 +378,12 @@ jobs:
               -e "ZONE_NAME=${ZONE_NAME}" \
               -e "ENV=${ENV}" \
               --rm efcms /bin/sh -c "./setup-for-blue-green-migration.sh"
+      - run:
+          name: Setup Migrate Flag
+          command: echo "export MIGRATE_FLAG=$(./scripts/get-migrate-flag.sh $ENV)" >> $BASH_ENV
+      - run:
+          name: Setup Destination Table
+          command: echo "export DESTINATION_TABLE=$(./scripts/get-destination-table.sh $ENV)" >> $BASH_ENV
```
adunkman, line 386: Thanks for the fix! I see why this is needed.

I’m a bit worried that we’ll remove these in the future by accident, but that’s just speculation at this point.

adunkman (approved):

PR #1553, opened 2021-08-09
turning on 8648

codyseibert:
this feature flag needs to be off on production, but we need to turn it on for the demo tomorrow and redeploy


codyseibert:
this PR will have to wait until https://github.com/ustaxcourt/ef-cms/pull/1554 is merged to migration before it can be re-tested

PR #1586, opened 2021-08-18
8803: trial sort tags compared carefully

kkoskelin:
The previous case may not have had a preferred trial city selected.  Check for the ability to generate the trial sort tags prior to comparing them.


kkoskelin:
EOD confusion

kkoskelin:
something's definitely wrong. re-creating.

PR #1600, opened 2021-08-22
remove migrations that have been run across all Court environments

mmarcotte:


codyseibert:
I forget where we landed on this code.. should we remove these since we don't have those migrations anymore?
```
const ranMigrations = {
    ...(await hasMigrationRan('bug-0035-private-practitioner-representing.js')),
    ...(await hasMigrationRan('bug-0036-public-served-parties-code.js')),
    ...(await hasMigrationRan('0036-phone-number-format.js')),
    ...(await hasMigrationRan('devex-0037-combine-work-items.js')),
    ...(await hasMigrationRan('0038-parse-generated-orders.js')),
  };
  ```

ericsorenson (approved):

ericsorenson:
Good point @codyseibert I think we should remove those as well.

adunkman:
Yes, I think we should remove the checks for if a migration has run as well. Otherwise, I don’t believe it’s possible to create a new environment.

codyseibert (approved):

PR #1603, opened 2021-08-24
8842 - omit filers from validation when validating document type

codyseibert:
It seems that legacy docket entries have a blank filers array which means the docket entry fails validation since we expect at least 1 filer to exist, but I don't think we have a way to figure out the userId of who filed the document when we migrated these legacy docket entries.  When editing a docket entry, we don't display the filers checkboxes, so there shouldn't be a reason why we validate them on the UI.

### DIFF
--- a/web-client/src/presenter/actions/EditDocketRecordEntry/validateDocumentAction.test.js
+++ b/web-client/src/presenter/actions/EditDocketRecordEntry/validateDocumentAction.test.js
@@ -107,6 +107,31 @@ describe('validateDocumentAction', () => {
     expect(errorStub).toHaveBeenCalled();
   });
 
+  it('should not cause an error if filers is empty with document tuype', async () => {
+    applicationContext
+      .getUseCases()
+      .validateDocumentInteractor.mockReturnValue(null);
+    applicationContext
+      .getUseCases()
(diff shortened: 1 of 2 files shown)


web-client/src/presenter/actions/EditDocketRecordEntry/validateDocumentAction.test.js
```diff
@@ -107,6 +107,31 @@ describe('validateDocumentAction', () => {
     expect(errorStub).toHaveBeenCalled();
   });
 
+  it('should not cause an error if filers is empty with document tuype', async () => {
```
rachelschneiderman: Extremely small nit: typo in 'tuype'

rachelschneiderman (approved):

web-client/src/presenter/actions/EditDocketRecordEntry/validateDocumentAction.test.js
mmarcotte: will merge once this typo's fixed.

codyseibert: fixed

mmarcotte (approved):

PR #1608, opened 2021-08-24
8828 - Bug - receivedAt migration

matthopson:
This verifies the recievedAt for a case hasn't been updated from that of the petition for electronic cases due to a bug that was introduced that allowed this to happen during the QC process.

NOTE: This is going to updated case records that may have a few ms difference in time. Should we account for that? An example from a test run:
`Updating case.receivedAt from 2020-06-25T20:58:55.411Z to 2020-06-25T20:58:55.414Z to match petition.`

### DIFF
--- a/web-api/migration-terraform/main/lambdas/migrations/bug-0040-case-received-at.js
+++ b/web-api/migration-terraform/main/lambdas/migrations/bug-0040-case-received-at.js
@@ -1,0 +1,64 @@
+const createApplicationContext = require('../../../../src/applicationContext');
+const {
+  aggregateCaseItems,
+} = require('../../../../../shared/src/persistence/dynamo/helpers/aggregateCaseItems');
+const {
+  Case,
+} = require('../../../../../shared/src/business/entities/cases/Case');
+const {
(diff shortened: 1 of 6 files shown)


web-api/migration-terraform/main/lambdas/migrations/bug-0040-case-received-at.js
```diff
@@ -0,0 +1,64 @@
...
+} = require('../../../../../shared/src/business/entities/EntityConstants');
+const applicationContext = createApplicationContext({});
+
+const migrateItems = async (items, documentClient) => {
+  const itemsAfter = [];
+
+  for (const item of items) {
+    if (item.pk.startsWith('case|') && item.sk.startsWith('case|')) {
+      const fullCase = await documentClient
+        .query({
```
mmarcotte: Again, this makes me wonder about `.queryFull(...)` versus `.query(...)`

codyseibert (approved):

sutt0n (approved):

mmarcotte (approved):

PR #1616, opened 2021-08-26
Add step-by-step for running Glue job

mark-meyer:
Adds a markdown file with instructions on running Glue jobs between production and staging environments.


mmarcotte:
Thanks @mark-meyer ! I may someday move this to the docs folder, or at least add a reference to it from a broader index within the docs folder. This is great for now, though. Thanks for the clearly laid out steps.

PR #1624, opened 2021-08-30
Batch 9: Kudos

ericsorenson:
- 8880 [Opinion Search: UI Fix](https://github.com/flexion/ef-cms/issues/8880) 
- 8793 [Order Search - Automate performance testing to utilize variable data for a simulated load](https://github.com/flexion/ef-cms/issues/8793) 
- 8511 [Automate Switch Colors Step in Prod](https://github.com/flexion/ef-cms/issues/8511) 
- 7643 [Update or Add New Email: Notice of Change of Email Address - Party](https://github.com/flexion/ef-cms/issues/7643) 
- 7384 [Opinion Search: AND keyword search](https://github.com/flexion/ef-cms/issues/7384) 
- 7370 [Opinion Search: Locate document by exact keyword/phase](https://github.com/flexion/ef-cms/issues/7370)


shared/src/business/entities/cases/Case.js
```diff
@@ -1170,6 +1172,58 @@ Case.prototype.getPetitionerById = function (contactId) {
...
+  const MAX_CLOSED_DATE = calculateISODate({
+    howMuch: -6,
+    units: 'months',
+  });
+  const isRecent =
+    rawCase.closedDate &&
+    dateStringsCompared(rawCase.closedDate, MAX_CLOSED_DATE) >= 0;
+
+  return isOpen || isRecent;
+};
```
mmarcotte: As per this morning's parking lot, I think this function needs to be renamed to something like `shouldGenerateNoticesForCase` to avoid further confusion. Technically, cases closed longer than 6 months can still receive service; they just should not have automatically generated notices filed and served upon them.

codyseibert:
@mmarcotte  heads up, for the batch 9 deploy, you'll need to run an environment specific deploy, and also the deploy will probably fail on the web-api terraform run the first time you try to deploy.  There is a script called `setup-s3-triggers-file.sh` which you will need to run to copy the triggers.zip inside the deploys bucket.  This is a work around needed due to how the blue green is setup in terraform (it would be nice to figure out a better solution than to run the deploy twice in a row, I can make a opex card for it if needed.  It seems low priority since this is a one time setup on a new environment)

Additional notes for 7643, this story ended up involving adding blue / green functionality to the cognito triggers.  After the deploy and smoke tests, the switch colors script will modify the cognito pool and set the authentication trigger to point to the correct color.  This change, imo, is a high risk change because if something goes wrong with the cognito pool's triggers, users won't be able to login.  I think this change needs extra scrutiny just to be safe.  I've tested and deployed it various times and it seems to work fine and doesn't switch to the new blue / green trigger until after the reindex finishes, etc.

shared/src/business/useCases/updatePetitionerInformationInteractor.js
```diff
@@ -298,53 +196,49 @@ const updatePetitionerInformationInteractor = async (
...
+    hasPetitionerInfoChanged &&
     !updatedCaseContact.isAddressSealed &&
-    shouldGenerateNoticesForCase(caseEntity)
-  ) {
-    const partyWithPaperService = caseEntity.hasPartyWithServiceType(
-      SERVICE_INDICATOR_TYPES.SI_PAPER,
-    );
+    caseEntity.shouldGenerateNoticesForCase();
 
+  // shouldGenerateNoticesForCase(caseEntity)
```
mmarcotte: just a little cleanup

shared/src/business/useCases/updatePetitionerInformationInteractor.js
```diff
@@ -383,7 +296,7 @@ const updatePetitionerInformationInteractor = async (
 
   return {
     paperServiceParties: servedParties.paper,
-    paperServicePdfUrl: serviceResults?.url,
+    paperServicePdfUrl: serviceUrl,
     updatedCase,
   };
```
mmarcotte, line 300: this function is huge. how do we encourage smaller functions that this main interactor can reference? no changes requested in this PR, but something should probably be added to the refactoring backlog. I'd be curious what the team's thoughts are on this subject.

web-api/elasticsearch/client.js
```diff
@@ -12,24 +12,18 @@ const {
 } = require('../elasticsearch/elasticsearch-settings');
 
 const getHost = async DomainName => {
-  console.log(DomainName);
```
mmarcotte, line 15: ty!

web-api/migration-cron-terraform/bin/destroy-app.sh
```diff
@@ -0,0 +1,35 @@
...
+
+echo "checking for the dynamodb lock table..."
+aws dynamodb list-tables --output json --region "${REGION}" --query "contains(TableNames, '${LOCK_TABLE}')" | grep 'true'
+result=$?
+if [ ${result} -ne 0 ]; then
+  echo "dynamodb lock does not exist, creating"
+  sh ../bin/create-dynamodb.sh "${LOCK_TABLE}" "${REGION}"
+else
+  echo "dynamodb lock table already exists"
+fi
```
mmarcotte, line 22: adding another one of these feels like there's more room for refactoring.

shared/src/business/useCases/updatePetitionerInformationInteractor.js
rachelschneiderman: Previously, we have added items like this to the devex backlog! We recently wrapped up a task to refactor [`serveCaseToIrsInteractor`](https://trello.com/c/3btHuHyX/913-refactor-servecasetoirsinteractor).

web-api/migration-cron-terraform/bin/destroy-app.sh
rachelschneiderman: Do you mean refactoring the destroy-app scripts? Or refactoring the infrastructure?

web-client/integration-tests/journey/docketClerkQCsNoticeOfChange.js
```diff
@@ -0,0 +1,78 @@
...
+    expect(cerebralTest.getState('currentPage')).toEqual('CaseDetailInternal');
+
+    await cerebralTest.runSequence('chooseWorkQueueSequence', {
+      box: 'inbox',
+      queue: 'section',
+    });
+    const workQueueFormatted = runCompute(formattedWorkQueue, {
+      state: cerebralTest.getState(),
+    });
+    console.log('docket number', cerebralTest.docketNumber);
```
mmarcotte: just checking, is this `console.log(...)` intentional? If not, can we remove it?

web-client/integration-tests/modifyContactInfo.test.js
```diff
@@ -56,6 +62,6 @@ describe('Modify Petitioner Contact Information', () => {
   petitionerEditsCasePrimaryContactAddressAndPhone(cerebralTest);
 
   loginAs(cerebralTest, 'docketclerk@example.com');
-  docketClerkViewsNoticeOfChangeOfAddress(cerebralTest);
+  docketClerkViewsNoticeOfChangeOfAddress({ cerebralTest });
```
mmarcotte, line 65: This appears to deviate from the previous, and imo more desirable, form that `loginAs(...)` or `userUpdatesEmailAddressToOneAlreadyInUse(...)` employs.

I would have expected:

```
docketClerkViewsNoticeOfChangeOfAddress(cerebralTest);
```

or

```
docketClerkViewsNoticeOfChangeOfAddress(cerebralTest, { documentTitle: 'Example' });
```

I don't think we need to change this now, but perhaps we can talk about it in the parking lot next week? Just trying to keep things consistent.

mmarcotte (approved):

web-api/migration-cron-terraform/bin/destroy-app.sh
mmarcotte: The scripts -- they all make use of this same check to see if the lock table exists, and it would be nice to have one script to re-use for all of the destroy (and init) scripts.

rachelschneiderman: I agree! I believe there is a devex task for this ([here](https://trello.com/c/cgOYAxn7/822-dry-up-the-deploy-appsh-scripts))! I will link this comment in there as well.

PR #1638, opened 2021-09-01
7341 - bug - only use last name for judge opinion search

matthopson:
This is a temporary fix preventing middle initials which are shared between judges from returning mixed results when a judge is given in opinion search.

This _could_ still produce unexpected results in the case of judges with the same last name (or a first name that matches the last name being queried). PO intends to work on stories that will create a more unique way of identifying / associating judges with documents in the system.


kkoskelin (approved):

mmarcotte:
I think this PR needs to go to the `mig` environment. `test` does not have Opinion Search enabled.

PR #1667, opened 2021-09-21
Add script to determine the difference between es clusters on a given index

mmarcotte:
While [documenting a recent deployment](https://github.com/ustaxcourt/ef-cms/wiki/2021-09-18-%28Saturday%29-Deployment) it became important to identify the difference between the alpha and beta clusters to see what may have gone missing.


shared/admin-tools/elasticsearch/determine-difference-es-index.js
```diff
@@ -0,0 +1,107 @@
...
+      },
+    },
+    index: indexName,
+    scroll: '30s',
+    size: 5000,
+  });
+
+  responseQueue.push(response);
+
+  while (responseQueue.length) {
```
mrinsin, line 52: should some unit tests be added to cover this logic?

PR #1668, opened 2021-09-21
8952 - bug - serve documents

matthopson:
Uses case status to determine if documents are eligible for service on a case.


matthopson:
Marking as draft, pending question in bug comment: https://github.com/flexion/ef-cms/issues/8952#issuecomment-924432911

matthopson:
Failing test re: the exposed case status.

mmarcotte (approved):
Looks super clean and clear

shared/src/business/entities/cases/Case.js
```diff
@@ -2267,8 +2267,20 @@ const caseHasServedDocketEntries = rawCase => {
...
 };
 
+/**
+ * determines if the case is in a state where documents can be served
+ *
+ * @param {string} caseCaption the original case caption
+ * @returns {string} caseTitle the case caption with the postfix removed
+ */
+
+const canAllowDocumentServiceForCase = rawCase => {
```
kkoskelin, line 2279: While working on #7643 we created a function similar to this -- and I'm wondering if it should be used instead, or if its conditions are entirely different:
https://github.com/flexion/ef-cms/blob/7643-sqs-integration/shared/src/business/entities/cases/Case.js#L1193
(the above branch is part of code that is in a long-running story)
As you can see, our code also checks if case has been closed inside the last six months.  I'm not certain whether this is the way to do it, or if these require different logic.  Just wanted to put this out there.

kkoskelin:
@mmarcotte heads-up that I have attempted to update some of the failing tests in this PR, but also there are some very slight changes to the `getCaseLambda` endpoints on v1 and v2.  You can see those changes reflected in the test updates I have committed.  I know that alterations to v1 and v2 are something not to be taken lightly, and I would like to discuss this with you.  There are no properties being added or removed from a potential response, but there is now some data being returned which was previously absent.

Happy to discuss in person!

web-api/src/v1/getCaseLambda.test.js
```diff
@@ -96,16 +96,17 @@ describe('getCaseLambda (which fails if version increase is needed, DO NOT CHANG
...
-    expect(JSON.parse(response.body).userId).toBeUndefined();
+    const parsedResponse = JSON.parse(response.body);
+    expect(parsedResponse).toHaveProperty('caseCaption', expect.any(String));
+    expect(parsedResponse.assignedJudge).toBeUndefined();
+    expect(parsedResponse.contactPrimary).toMatchObject({
+      name: 'Test Petitioner',
+      state: 'TN',
+    });
+    expect(parsedResponse.noticeOfTrialDate).toBeUndefined();
+    expect(parsedResponse.status).toBe('Calendared');
```
mmarcotte: I think this part confuses me. the v1 API would only be consumed by the IRS Super User. And the IRS Super User would be associated with any case that is not in New Status.

mmarcotte: Also, we should not be changing these tests. These are what keep the API returning what we have promised to the V1 and V2 API consumers.

web-api/src/v2/getCaseLambda.test.js
```diff
@@ -119,16 +119,17 @@ describe('getCaseLambda (which fails if version increase is needed, DO NOT CHANG
...
+    expect(parsedResponse).toHaveProperty('caseCaption', expect.any(String));
+    expect(parsedResponse.assignedJudge).toBeUndefined();
+    expect(parsedResponse.contactPrimary).toMatchObject({
+      name: 'Test Petitioner',
+      state: 'TN',
+    });
+    expect(parsedResponse.status).toBe('Calendared');
+    expect(parsedResponse.trialDate).toBeUndefined();
+    expect(parsedResponse.trialLocation).toBeUndefined();
+    expect(parsedResponse.userId).toBeUndefined();
```
mmarcotte: We should not be changing these tests. These are what keep the API returning what we have promised to the V1 and V2 API consumers.

web-api/src/v1/getCaseLambda.test.js
```diff
@@ -96,16 +96,15 @@ describe('getCaseLambda (which fails if version increase is needed, DO NOT CHANG
...
-    expect(JSON.parse(response.body).userId).toBeUndefined();
+    const parsedResponse = JSON.parse(response.body);
+    expect(parsedResponse).toHaveProperty('caseCaption', expect.any(String));
+    expect(parsedResponse.assignedJudge).toBeUndefined();
+    expect(parsedResponse.noticeOfTrialDate).toBeUndefined();
+    expect(parsedResponse.trialLocation).toBeUndefined();
+    expect(parsedResponse.userId).toBeUndefined();
+    // contact and status is unavailable for user unassociated with the case
+    expect(parsedResponse.contactPrimary).toBeDefined();
+    expect(parsedResponse.status).toBeDefined();
```
mmarcotte, line 106: these should be undefined as per the earlier tests.

web-api/src/v2/getCaseLambda.test.js
```diff
@@ -119,16 +119,14 @@ describe('getCaseLambda (which fails if version increase is needed, DO NOT CHANG
...
-    expect(JSON.parse(response.body).contactPrimary).toBeUndefined();
-    expect(JSON.parse(response.body).status).toBeUndefined();
-    expect(JSON.parse(response.body).trialDate).toBeUndefined();
-    expect(JSON.parse(response.body).trialLocation).toBeUndefined();
-    expect(JSON.parse(response.body).userId).toBeUndefined();
+    const parsedResponse = JSON.parse(response.body);
+    expect(parsedResponse).toHaveProperty('caseCaption', expect.any(String));
+    expect(parsedResponse.assignedJudge).toBeUndefined();
+    expect(parsedResponse.contactPrimary).toBeDefined();
+    expect(parsedResponse.status).toBeDefined();
```
mmarcotte, line 126: these should be undefined per the earlier tests.

shared/src/business/entities/cases/Case.js
matthopson: I don't think that was the intent of this particular story. Per the PO, we want to disallow document service on a case that is in NEW status. This method is _currently_ only checking for that. I left it as a method instead of just checking its status everywhere in case this logic changed, at least tit would be abstracted. I'm wondering if I can just use the `isSttatusNew` property instead of exposing the status on Public case. That would give us what we want and we wouldn't have to make any changes to the entity, API response, or associated tests.

web-api/src/v1/getCaseLambda.test.js
adunkman: As per discussion this morning, the `status` here was unintentionally missing from this API endpoint. It is included in the API contract and was set on some cases, but not always. This change is because for this specific test case, the case status _should have been_ and now _will_ be included. 

@kkoskelin, can you explain why `contactPrimary` is included as well? We talked about it this morning, and you mentioned that it was also not set sometimes when it should have been — but I can’t follow why it would have been missing before and is now set. Is there some filter which checks on `status`? 

In any case, the comment here is incorrect and contradicts the test assertions — probably worth dropping the comment so it doesn’t become outdated again.

web-api/src/v2/getCaseLambda.test.js
adunkman: Same as above — will look to the above comment thread for response.

adunkman (commented):
Thanks for the explanation today, y’all!

web-api/src/v1/getCaseLambda.test.js
kkoskelin: @adunkman The particular case used in this test was _sealed_, and previously had no `status` value.  It means that the marshalling function attempted to obtain the `contactPrimary` as seen here:
https://github.com/ustaxcourt/ef-cms/pull/1412/files#diff-384b0330bd6db4b05ec5bf32f098eb0c01feab98b385ef5653e6dbc59faf58f8R1603

However, the "else" block executed wasn't using the correct method of obtaining the contact info because this case was *not* actually New -- it was Calendared.  The contact information was most definitely available on the PublicCase entity instance, but our code was doing a poor job of attempting to retrieve it, and thus it previously had been undefined.

shared/src/business/entities/cases/PublicCase.js
```diff
@@ -104,6 +105,12 @@ const publicCaseSchema = {
     .description('Party type of the case petitioner.'),
   petitioners: joi.array().items(PublicContact.VALIDATION_RULES).required(),
   receivedAt: JoiValidationConstants.ISO_DATE.optional(),
+  status: JoiValidationConstants.STRING.valid(
+    ...Object.values(CASE_STATUS_TYPES),
+  )
+    .optional()
+    .meta({ tags: ['Restricted'] })
```
mrinsin, line 112: Hello! Just curious to know what this `[Restricted]` tag does!

matthopson:
👍 (I'm the author - can't thumb this) - but thanks @kkoskelin

mmarcotte (approved):

web-client/src/presenter/computeds/documentViewerHelper.test.js
```diff
@@ -200,6 +203,28 @@ describe('documentViewerHelper', () => {
...
             ],
+            status: CASE_STATUS_TYPES.calendared,
+          },
+        },
+      });
+
+      expect(showUnservedPetitionWarning).toBe(false);
+    });
+
+    it('should be true if an otherwise servable document is selected but the case is new', () => {
```
mrinsin, line 214: These are some great test descriptions! - Rachel, Rosie and Mrin!

mrinsin (approved):

shared/src/business/entities/cases/PublicCase.js
kkoskelin: @matthopson can you answer this?

matthopson: This was pulled over from the other definitions of this field on Case. I'm not 100% sure, as I couldn't find where it was being used, but I thought these would be restricted fields on a blocked case - which makes this a good question to try and answer for sure.

matthopson: Eric said it was used for document generation and doesn't change anything. :)

web-api/src/v1/getCaseLambda.test.js
```diff
@@ -96,16 +96,15 @@ describe('getCaseLambda (which fails if version increase is needed, DO NOT CHANG
...
-    expect(JSON.parse(response.body).status).toBeUndefined();
-    expect(JSON.parse(response.body).trialLocation).toBeUndefined();
-    expect(JSON.parse(response.body).userId).toBeUndefined();
+    const parsedResponse = JSON.parse(response.body);
+    expect(parsedResponse).toHaveProperty('caseCaption', expect.any(String));
+    expect(parsedResponse.assignedJudge).toBeUndefined();
+    expect(parsedResponse.noticeOfTrialDate).toBeUndefined();
+    expect(parsedResponse.trialLocation).toBeUndefined();
+    expect(parsedResponse.userId).toBeUndefined();
+    // contact and status is unavailable for user unassociated with the case
```
adunkman:
```suggestion

```

adunkman (approved):
(non-blocking) Reminder that there’s a comment here which isn’t accurate, as previously mentioned.

mmarcotte (approved):

PR #1669, opened 2021-09-22
Add script to cleanup outbox partitions

mmarcotte:
This is a script we used to help clear out a hot partition that was overloaded with records. It became a blocker for migrations and let to the creation of this Critical bug. https://github.com/flexion/ef-cms/issues/8933

The problem is that records were continuing collect in one partition, and it was becoming increasingly full of information that didn't really match the usage pattern by the application. Rather than delete them permanently, we opted to archive them into their own partition based on the month they pertain to. 

Running this unblocked our migrations, and these segments and partitions were no longer overloaded. Additionally, it brought our glue job of moving production data to staging environments down from 7h 54m to 2h 16m.


mmarcotte:
Closing for a more comprehensive solution: https://github.com/ustaxcourt/ef-cms/issues/1739

PR #1673, opened 2021-09-27
De-dupe the eligible case docket numbers...

mmarcotte:
This removes any duplicate docket numbers that we receive from the query of mapping records for the given Trial Location
prefix. An underlying data issue underlying causes this problem, and we are working to fix it with [this bug report](https://github.com/ustaxcourt/ef-cms/tree/diagnose-trial-session-bug). 

We have encountered this bug numerous times before, and we have resorted to fixing the data in production. For reference, here are the previous bug reports:

* https://github.com/flexion/ef-cms/issues/8186
* https://github.com/flexion/ef-cms/issues/8594
* https://github.com/flexion/ef-cms/issues/8384

### DIFF
--- a/shared/src/persistence/dynamo/trialSessions/getEligibleCasesForTrialSession.test.js
+++ b/shared/src/persistence/dynamo/trialSessions/getEligibleCasesForTrialSession.test.js
@@ -47,4 +47,32 @@ describe('getEligibleCasesForTrialSession', () => {
       },
     ]);
   });
+
+  it('should remove any duplicate cases it receives the list of cases it receives from the mapping records query', async () => {
+    client.query = jest.fn().mockReturnValue([
+      {
+        docketNumber: MOCK_CASE.docketNumber,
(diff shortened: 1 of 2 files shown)


shared/src/persistence/dynamo/trialSessions/getEligibleCasesForTrialSession.test.js
```diff
@@ -47,4 +47,32 @@ describe('getEligibleCasesForTrialSession', () => {
       },
     ]);
   });
+
+  it('should remove any duplicate cases it receives the list of cases it receives from the mapping records query', async () => {
```
matthopson: Test description seems a little mal-formed

mmarcotte: Suppressing my urge to respond with a Ralph Wiggum gif. Thank you!

matthopson (approved):

PR #1674, opened 2021-09-28
Hide status from public case

mmarcotte:
In order to keep information secure as per Court requirements, we are keeping `status` hidden from Public endpoints. Additionally, we are removing it from the v1 and v2 APIs as this information is internal and should only be conveyed to Court users.

### DIFF
--- a/docs/api/v1.yaml
+++ b/docs/api/v1.yaml
@@ -164,9 +164,6 @@ components:
         sortableDocketNumber:
           type: integer
           example: 20000384
-        status:
-          type: string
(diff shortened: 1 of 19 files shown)


kkoskelin:
(I understand that this is a draft PR)
If these changes align with the desires of the PO and the Court:
1. there should be either an issue or a bug filed on the board with a complete explanation, groomed and prioritized by the PO.
2. The bug number should be linked here in the PR title or description.
3. The changes contained herein should all have complete test coverage of altered code.

PR #1677, opened 2021-09-28
8952 - Decorate Case Status

matthopson:
This adds a little function in getCaseInteractor to compute the canAllowDocumentService field without sending case status information over the network for PublicCase objects.

As a side note, we will make use of this decorator for other fields / future stories.

### DIFF
--- a/shared/src/business/entities/cases/Case.canAllowDocumentServiceForCase.test.js
+++ b/shared/src/business/entities/cases/Case.canAllowDocumentServiceForCase.test.js
@@ -4,13 +4,31 @@ const { CASE_STATUS_TYPES } = require('../EntityConstants');
 const { MOCK_CASE } = require('../../../test/mockCase');
 
 describe('canAllowDocumentServiceForCase', () => {
+  it('returns true if the rawCase.canAllowDocumentService value is defined and true', () => {
+    expect(
+      canAllowDocumentServiceForCase({
+        ...MOCK_CASE,
+        canAllowDocumentService: true,
+      }),
+    ).toEqual(true);
+  });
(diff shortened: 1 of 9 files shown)


kkoskelin:
Sorry to add-on after the fact, but can we also refactor `Case.getContactPrimaryOrSecondary` so that it does not attempt to inspect the status of the case to obtain petitioner data?
I would propose that it tries one method after another, stopping at the first attempt which yields a defined result.  LMK if you wish to discuss.

matthopson:
@kkoskelin is this change directly impacting this story? I think there are a few other places where we want to do similar refactors where this might fit.

kkoskelin:
@matthopson it does affect this story insofar as the contactPrimary was expected to be "defined" and is now "undefined", because we were relying upon the value of the case status to accurately fetch the case contacts.

mmarcotte:
@kkoskelin -- I'm not certain that it does. `contactPrimary` had previously been expected to be undefined, as it [currently is in staging](https://github.com/ustaxcourt/ef-cms/blob/staging/web-api/src/v2/getCaseLambda.test.js#L127). This PR reverts that change that the previous attempt had made. 

If we want to expose `contactPrimary` to the `v1` and `v2` API users who are not associated with a case, I suppose that would be a bug or user story that we would want to get written and prioritized. As of right now, this is expected behavior.

kkoskelin:
Yes, the change makes it go back to the way it was.
What I'm saying is that "the way it was" only behaved that way because we had poorly-written code that could be improved (and lacked test coverage).  But prioritize it however you see fit.

Issue #1695, opened 2021-09-30
BUG: Edit Petitioner button is missing for Closed Cases > 6 months


mmarcotte:
**Describe the Bug**

The Edit Petitioner button is now hidden for Cases that are closed for greater than 6 months

**Business Impact/Reason for Severity**

Docket should always be able to edit the contact information for a petitioner for Open and Closed cases. 

**In which environment did you see this bug?**

Prod / Test

**Who were you logged in as?**

Docket Clerk

**What were you doing when you discovered this bug? (Using the application, demoing, smoke tests, testing other functionality, etc.)**

**To Reproduce**

1. Login as a Docket Clerk
2. Navigate to a case that has been closed greater than 6 months ago
3. Go to Case Details and Parties
4. Observe that you cannot edit the petitioner information

**Expected Behavior**

You should be able to edit the petitioner, and no notices should be generated for cases closed more than 6 months ago

**Actual Behavior**

The button is missing.

**Screenshots**

![bug-closed-cases-petitioner](https://user-images.githubusercontent.com/5023502/135524253-cbe83c6a-cbbf-41d5-ac72-6d01807cb5e1.png)

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Cause of Bug, If Known**

**Process for Logging a Bug:**
* Complete the above information
* Add a severity tag (Critical, High Severity, Medium Severity or Low Severity). See below for priority definition. 

**Severity Definition:**
* Critical Defect
Blocks entire system's or module’s functionality
No workarounds available
Testing cannot proceed further without bug being fixed.

* High-severity Defect
Affects key functionality of an application
There's a workaround, but not obvious or easy
App behaves in a way that is strongly different from the one stated in the requirements

* Medium-severity Defect
A minor function does not behave in a way stated in the requirements.
Workaround is available and easy

* Low-severity Defect
Mostly related to an application’s UI
Doesn't need a workaround, because it doesn't impact functionality

## Definition of Done (Updated 4-14-21)
**Product Owner**
 - [ ]  Bug fix has been validated in the Court's test environment

**Engineering**
 - [ ] Automated test scripts have been written
 - [ ] Field level and page level validation errors (front-end and server-side) integrated and functioning
 - [ ] Verify that language for docket record for internal users and external users is identical
 - [ ] New screens have been added to pa11y scripts
[...]

PR #1703, opened 2021-10-01
USTC-1695: Fix button for editing Petitioner Info for Cases closed longer than 6 months

mmarcotte:
Solves: https://github.com/ustaxcourt/ef-cms/issues/1695

Docket clerks reported that you could no longer edit the Petitioner contact for a Case that had been closed longer than 6 months. It appears that work for https://github.com/flexion/ef-cms/issues/8952 broke this as we became reliant on whether or not you could perform service on a closed case via https://github.com/ustaxcourt/ef-cms/pull/1668

However, once we made the button available, it also exposed a bug that we were still performing service on cases that were closed greater than six months. This logic needed to be performed at the Lambda level and not in the client.


mmarcotte:
Moving to another branch and incorporating 1706 changes

Issue #1706, opened 2021-10-01
BUG: Parties cannot file documents on cases closed for more than 6 months


mmarcotte:
**Describe the Bug**

It appears that https://github.com/flexion/ef-cms/issues/8952 has introduced another bug. Parties to a case can no longer file and serve documents on a case if the case has been closed for more than 6 months. 

**Business Impact/Reason for Severity**

It's vital for parties to a case to be able to file documents on a case even if the case has been closed longer than 6 months

**In which environment did you see this bug?**

`test`

**Who were you logged in as?**

Users:

* 4ebb8e80-4fea-46ea-9401-904b03660849 (IRS Practitioner)
* 5c0573d4-1af0-4afa-b6a0-287988c80a27 (Private Practitioner)

on case 16715-19

**What were you doing when you discovered this bug? (Using the application, demoing, smoke tests, testing other functionality, etc.)**

Check to see if this was an issue

**To Reproduce**

1. Become user `4ebb8e80-4fea-46ea-9401-904b03660849`
2. Navigate to 16715-19
3. See notice that you cannot file a document until the petition has been served

**Expected Behavior**

You should be able to file a document.

**Actual Behavior**

You receive a notice 

> You will be able to file documents after the Petition is processed.

And you cannot file documents

**Screenshots**

<img width="1807" alt="Screen Shot 2021-10-01 at 5 08 13 PM" src="https://user-images.githubusercontent.com/5023502/135686355-0d9f68ec-1d76-4e10-829a-608b395c4e8e.png">

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Cause of Bug, If Known**

**Process for Logging a Bug:**
* Complete the above information
* Add a severity tag (Critical, High Severity, Medium Severity or Low Severity). See below for priority definition. 

**Severity Definition:**
* Critical Defect
Blocks entire system's or module’s functionality
No workarounds available
Testing cannot proceed further without bug being fixed.

* High-severity Defect
Affects key functionality of an application
There's a workaround, but not obvious or easy
App behaves in a way that is strongly different from the one stated in the requirements

* Medium-severity Defect
A minor function does not behave in a way stated in the requirements.
Workaround is available and easy

* Low-severity Defect
Mostly related to an application’s UI
Doesn't need a workaround, because it doesn't impact functionality

## Definition of Done (Updated 4-14-21)
**Product Owner**
 - [ ]  Bug fix has been validated in the Court's test environment

**Engineering**
[...]

PR #1718, opened 2021-10-05
USTC-1706: Enable filing and serving on Closed cases

mmarcotte:
Solves: https://github.com/ustaxcourt/ef-cms/issues/1706

Continuing to get everything straightened out after managing some service bugs, this bugfix attempts to solve a bug that prevents parties to a case from filing and serving documents. We are introducing a different decorator / Case helper function that https://github.com/flexion/ef-cms/issues/7643 is also using.

At its core, it's changing the logic for `canAllowDocumentServiceForCase` to whether or not the `status` is `new`. And it's moving the previous logic to a different helper function  called `shouldGenerateNoticesForCase`


shared/src/business/entities/cases/Case.js
```diff
@@ -2285,6 +2285,20 @@ const canAllowDocumentServiceForCase = rawCase => {
...
+};
+
+/**
+ * determines whether or not to file automatically generated notices for notice of change of address
+ *
+ * @param {Object} rawCase the court case
+ * @returns {Boolean} whether or not we should automatically generate notices for a change of address
+ */
+const shouldGenerateNoticesForCase = rawCase => {
+  if (typeof rawCase.shouldGenerateNotices !== 'undefined') {
```
kkoskelin, line 2298: I think I would prefer `if (rawCase.shouldGenerateNotices !== undefined) {` as checking `typeof` is probably an unnecessary step.

PR #1725, opened 2021-10-08
ustc: Reintroduce smoke tests into migration environment

mmarcotte:
When we stood up the `migration` environment, we had no way for the `smoketests` to make use of the strong password in place for the test users. Now we've been able to encapsulate the strong password in an environment-specific environment var `DEFAULT_ACCOUNT_PASS_MIG`.


codyseibert (approved):
seems good to me!

Issue #1731, opened 2021-10-13
Increase the resources for the Production Elasticsearch Cluster


mmarcotte:
Changing `m5.xlarge` to `r5.xlarge`. Also performing the same change to `mig` and `test` so we can have realistic tests for performance.

PR #1735, opened 2021-10-13
ustc-1731: change instance types for elasticsearch cluster nodes

mmarcotte:
Completes: https://github.com/ustaxcourt/ef-cms/issues/1731

Following some recent production activity where the JVM Memory Pressure became too high, we adjusted the instance type in the console to `r5.xlarge`, which is optimized for memory. They have been holding up very well so far, and so we are committing this change. 

Additionally, we're increasing the resources for `test` and `migration` so we can conduct some realistic performance tests with the updated configuration.

### DIFF
--- a/scripts/get-es-instance-type.sh
+++ b/scripts/get-es-instance-type.sh
@@ -29,15 +29,15 @@ elif [[ $BRANCH == 'irs' ]] ; then
 elif [[ $BRANCH == 'staging' ]] ; then
   echo "t2.medium.elasticsearch"
 elif [[ $BRANCH == 'test' ]] ; then
-  echo "m5.large.elasticsearch"
+  echo "r5.xlarge.elasticsearch"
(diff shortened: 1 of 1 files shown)


kkoskelin (approved):
Fair warning: @codyseibert has committed some devex work on the Flexion fork which will need to be accounted for once merged -- variables have been extracted to instance-specific shell scripts, FYI.

mmarcotte:
> Fair warning: @codyseibert has committed some devex work on the Flexion fork which will need to be accounted for once merged -- variables have been extracted to instance-specific shell scripts, FYI.

Thanks @kkoskelin! Yeah, I'm really excited for those changes! Hopefully it will be straightforward enough to incorporate.

mmarcotte:
Screen Shot 2021-10-18 at 10 14 25 AM

Cluster continues to perform well with garbage collection after one week of this change being in place manually.

PR #1738, opened 2021-10-13
USTC-1732: turn on slow query logging for ES

mmarcotte:
Completes: https://github.com/ustaxcourt/ef-cms/issues/1732

Enable additional logging in order to troubleshoot slow queries that may be causing undue JVM Memory Pressure.


kkoskelin (commented):
Probably okay as far as I can see, but I defer to Cody on this one. ;)

Issue #1739, opened 2021-10-13
Refactor WorkItem Section Outboxes


mmarcotte:
In order to get migrations unblocked with Production Sized data sets, we needed to move records off of partitions as they had become overloaded. We created and executed [this script](https://github.com/ustaxcourt/ef-cms/pull/1669) to help clear out a hot partition that was overloaded with records. It became a blocker for migrations and led to the creation of [this Critical bug](https://github.com/flexion/ef-cms/issues/8933).

The problem is that records were (and still are) continuing collect in one partition, and it was becoming increasingly full of information that didn't really match the usage pattern by the application. Rather than delete them permanently, we opted to archive them into their own partition based on the month they pertain to. 

Running this unblocked our migrations, and these segments and partitions were no longer overloaded. Additionally, it brought our glue job of moving production data to staging environments down from 7h 54m to 2h 16m. 

Now we need work to complete the transition to the new partition structure. We will continue to create the records that are used by the UI, but we will add a time-to-live attribute so them so they expire after a period of 8 days so they do not collect. And we want these persistence methods to create the archive records just in case we want to go back and find them based on the time they occurred.

PR #1740, opened 2021-10-13
Refactor Section Outbox Work Item Partition Usage

mmarcotte:
Solves: https://github.com/ustaxcourt/ef-cms/issues/1739

Need to run the script in order to move all of the records to the correct partition in order to clean up the existing partitions. And then future records should all expire properly. We can make use of the archive data with future stories.

This refactor continues to store records on the original primary partition of `section-outbox|${section}`, but it adds a `ttl` so these partitions don't collect data forever and become overloaded.

And it also stores a copy of the record in a partition based on the month `section-outbox|${section}|{YYYY-MM}` of the `sk`. This way we can query for records if we are attempting to go back in time further than 7 days. This functionality isn't written yet, but we didn't want to lose any data in case we wanted to do this at a later date.

Additionally, it makes use of the handy `deleteByGsi(...)` persistence helper function so that we can easily remove all of the records associated with a Work Item when we need to delete it.


shared/src/persistence/dynamo/workitems/createSectionOutboxRecords.js
```diff
@@ -0,0 +1,104 @@
...
+      Item,
+      applicationContext,
+      section,
+    }),
+    await createSectionOutboxArchiveRecord({
+      Item,
+      applicationContext,
+      section,
+    }),
+  ]);
```
kkoskelin: you can actually
1. remove the `async` decorator from the function definition
2. remove all `await` decorators
3. directly return the value from `Promise.all(...)`

shared/src/persistence/dynamo/workitems/deleteWorkItem.js
```diff
@@ -1,10 +1,4 @@
...
 
-exports.deleteWorkItem = ({ applicationContext, workItem }) =>
-  client.delete({
-    applicationContext,
-    key: {
-      pk: `case|${workItem.docketNumber}`,
-      sk: `work-item|${workItem.workItemId}`,
-    },
-  });
+exports.deleteWorkItem = async ({ applicationContext, workItem }) =>
```
kkoskelin: `async` decorator can be removed

shared/src/persistence/dynamo/workitems/createSectionOutboxRecords.js
```diff
@@ -0,0 +1,104 @@
...
+  workItem,
+}) => {
+  const Item = {
+    ...workItem,
+    gsi1pk: `work-item|${workItem.workItemId}`,
+    sk: workItem.completedAt ? workItem.completedAt : workItem.updatedAt,
+  };
+
+  await Promise.all([
+    await createSectionOutboxRecentRecord({
```
codyseibert: technically you don't need these inner await calls.

shared/src/persistence/dynamo/workitems/createSectionOutboxRecords.js
```diff
@@ -0,0 +1,104 @@
...
+ *
+ * @param {Object} providers The providers object
+ * @param {Object} providers.applicationContext The application context
+ * @param {String} section docket or petitions
+ * @param {WorkItem} workItem The instantiated Work Item to create a record
+ * @returns {Promise} resolves upon completion of persistence request
+ */
+const createSectionOutboxArchiveRecord = async ({
+  applicationContext,
+  Item,
```
codyseibert, line 30: I find it strange that this is `Item` and not lower case i.  Most properties on our JS objects are all camalCase, but maybe there are a few places we just need to go back and refactor.

mmarcotte: Yeah, put takes `Item` for the dynamoDB call.

kkoskelin (approved):

mark-meyer (approved):

PR #1743, opened 2021-10-14
Bug 9009: Items unreliably getting removed from ES after removing an IRS Practitioner from Case

rosiehsmith:
This PR will remove the `efcms-user-case` index, therefore a blue/green migration is required when this is merged to deployed environments.

We would expect that the reindex step will probably have to be manually approved as we also stop indexing case/practitioner mapping records into `efcms-user` as part of this PR so the difference between source/destination indices will not be 0.


mrinsin:
The fix for this bug assumes that getIndexedCasesForUser is no longer called anywhere in the code. However, 7643 was the story in which getIndexedCasesForUser was deleted, therefore the PR for this fix will fail the build until batch 9 (ustaxcourt#1624) has been merged into ustaxcourt:staging.

rosiehsmith:
to ensure indices match during `reindex` circle step, @mmarcotte likely to write a script to delete records from `efcms-user` where the pk and sk do NOT begin with `user|`

for reference: https://flexion.slack.com/archives/CD4PNKEPK/p1634311035254400
thank you!

mrinsin (approved):
I looked for all the files relevant to this bug fix and ignored the other, lgtm!

codyseibert (approved):

TomElliottFlexion (approved):

mmarcotte:
Looks good other than the `package-lock.json` change that would cause conflict confusion down the road. We shouldn't be committing this file for bugfixes.

mmarcotte (approved):

PR #1754, opened 2021-10-20
Add script to remove deprecated index and cleanup user index

mmarcotte:
```
 node shared/admin-tools/elasticsearch/flexion-9009-cleanup-elasticsearch.js [env]
```


mmarcotte:
not needed as this work has completed on all environments.

PR #1755, opened 2021-10-21
9074: Added docketNumberWithSuffix to EligibleCase entity

TomElliottFlexion:
Screen Shot 2021-10-21 at 10 56 11 AM


rachelschneiderman (commented):
Is there a test that could be added that would have caught this bug? Maybe adding an expect to an integration test?

TomElliottFlexion:
Not sure if this _should_ be covered with an integration test. This was a matter of exposing a new attribute on the data structure

kkoskelin (approved):

rachelschneiderman:
> Not sure if this _should_ be covered with an integration test. This was a matter of exposing a new attribute on the data structure

Generally when I pick up a bug, I try and write a failing test first that covers the bug. IMO this should be covered if it can be by some type of test that would have failed before the solution was explored.

kkoskelin (approved):

rachelschneiderman (approved):
Looks great!!!

TomElliottFlexion:
> Not sure if this _should_ be covered with an integration test. This was a matter of exposing a new attribute on the data structure

Upon further discussion, this is _exactly_ the kind of thing that should be covered by an integration test

PR #1760, opened 2021-10-21
9080 Ping Pong Keep Alive

codyseibert:
- Adding the default handler to the api gateway websocket so we can read incoming messages from the client
- Adding logic to have the client send a ping even every minute to the server to keep the connection alive

### DIFF
--- a/web-api/src/notifications/defaultLambda.js
+++ b/web-api/src/notifications/defaultLambda.js
@@ -1,0 +1,19 @@
+const { genericHandler } = require('../genericHandler');
+
+/**
+ * save the information about a new websocket connection
+ *
+ * @param {object} event the AWS event object
+ * @returns {Promise<*|undefined>} the api gateway response object containing the statusCode, body, and headers
+ */
+exports.defaultLambda = event =>
+  genericHandler(
+    event,
+    () => {
+      // as of right now, this default lambda only exists so that we can send PING messages from the UI and keep
+      // the websocket connection to api gateway alive.  API Gateway has a 10 min timeout, so the socket needs traffic (either in / out)
+      // at least once before that 10 minutes to have API Gateway not kill the connection.
+      return null;
+    },
+    { bypassMaintenanceCheck: true },
+  );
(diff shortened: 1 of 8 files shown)


codyseibert:
This was verified on a flexion experimental2 environment.  I allowed the UI to sit for 15 minutes, then set a calendar session.  No spinner of doom occured.

web-api/src/notifications/defaultLambda.js
```diff
@@ -0,0 +1,19 @@
+const { genericHandler } = require('../genericHandler');
+
+/**
+ * save the information about a new websocket connection
+ *
+ * @param {object} event the AWS event object
+ * @returns {Promise<*|undefined>} the api gateway response object containing the statusCode, body, and headers
+ */
+exports.defaultLambda = event =>
```
rachelschneiderman, line 9: Would it be accurate to call this the heartbeatLambda? Just trying to think how we could name this to help inform future devs :)

web-client/src/providers/socket.js
```diff
@@ -37,6 +41,10 @@ export const socketProvider = ({ socketRouter }) => {
             setTimeout(() => {
               resolve();
             }, 300);
+
+            pingInterval = setInterval(() => {
```
rachelschneiderman, line 45: Is there a way to add a test for this so it's captured we expect heartbeat messages to be sent every so often?

codyseibert: I was having trouble figuring out the best way to test this.  Want to pair on this part?

web-api/src/notifications/defaultLambda.js
codyseibert: so the reason I named it default is because on the API gateway, you can setup 3 types of listeners ($connect, $disconnect, $default).  So I was trying to have our 3 web socket lambda map one-to-one with the naming convention of api gateway v2.

codyseibert: Screen Shot 2021-10-22 at 9 26 18 AM

here is what I'm talking about, there are only 3 routes possible.

rachelschneiderman: I gotcha! Makes sense! Thanks!

rachelschneiderman (approved):

PR #1770, opened 2021-10-25
9074 bug docket number suffix mk. 2

TomElliottFlexion:
Same as https://github.com/ustaxcourt/ef-cms/pull/1755


mmarcotte (approved):
Thanks Tom!

PR #1793, opened 2021-11-10
Bugfix 9123: sanitize pdf

kkoskelin:
Detect forms in PDF documents which contain form fields which may interfere with placement of signatures and other text elements during service.
Flatten form fields so they may no longer be interactive and allow Dawson-related modifications to appear as designed.


shared/src/business/useCases/pdf/sanitizePdfInteractor.js
```diff
@@ -0,0 +1,37 @@
...
+
+    await applicationContext.getPersistenceGateway().saveDocumentFromLambda({
+      applicationContext,
+      document: pdfBytes,
+      key,
+      useTempBucket: false,
+    });
+  }
+
+  return true;
```
rachelschneiderman: ~Nit: Is this value used for anything?~

Just saw this is still a draft. Apologies!

kkoskelin: This was copied to behave similarly to validatePdfInteractor.  Essentially if everything succeeds, we return `true`, but otherwise an error may be thrown.  Perhaps it should come out of both? I'm open to that, too.

shared/src/business/useCases/filePetitionInteractor.test.js
```diff
@@ -1,8 +1,17 @@
...
+  uploadDocumentAndMakeSafeInteractor,
+} = require('./uploadDocumentAndMakeSafeInteractor');
 const { applicationContext } = require('../test/createTestApplicationContext');
 const { filePetitionInteractor } = require('./filePetitionInteractor');
 const { ROLES } = require('../entities/EntityConstants');
 
 beforeAll(() => {
+  applicationContext
+    .getUseCases()
+    .uploadDocumentAndMakeSafeInteractor.mockImplementation(
```
rachelschneiderman: I see this is done in a couple of tests, could this be moved to `createTestApplicationContext`?

shared/src/business/useCases/pdf/sanitizePdfInteractor.js
rachelschneiderman: I think I would prefer it to be removed if it's not being consumed anywhere!

shared/src/business/useCases/pdf/sanitizePdfInteractor.js
```diff
@@ -0,0 +1,37 @@
+/**
+ * sanitizePdfInteractor
+ *
+ * @param {object} applicationContext the application context
+ * @param {object} providers the providers object
+ * @param {string} providers.key the key of the document to sanitize
+ * @returns {boolean} true if successfully sanitized
+ */
+exports.sanitizePdfInteractor = async (applicationContext, { key }) => {
```
rosiehsmith, line 8: now that i'm looking at it again with fresh(er) eyes, is "sanitize" still an accurate term for what we're doing in this interactor?

kkoskelin: The original "sanitize" code that once existed was used to attempt to remove all pdf form interactivity as well as disable script tags.  
Here's jsdoc about what this function (`flatten()`) intends to do:
https://github.com/Hopding/pdf-lib/blob/b8a44bd24b74f4f32456e9809dc4d2d9dc9bf176/src/api/form/PDFForm.ts#L517
I think that although this function is currently only flattening the forms, it has the potential to maybe do other things as well in the future.

kkoskelin: Screen Shot 2021-11-10 at 10 02 42 PM
Here's a screenshot of a PDF I uploaded to my local environment that's using `flatten()`.  It was a PDF with an interactive form which I *also* applied my "pdf signature" to.  all those things were flattened and cannot be interacted with or altered from within the pdf viewer.

rosiehsmith: seems like it's more accurately doing a flatten then! i'm fine with leaving it as is, just wanted to note.

rachelschneiderman (approved):

rosiehsmith (approved):

mmarcotte (approved):

PR #1813, opened 2021-11-29
8949 - Sporadic Logout Bug

codyseibert:
**DON'T MERGE - NEEDS MORE REWORK**

This bug fix involves two main changes.

First, we no longer store the token / user in localStorage.  We originally decided to never store the refresh token on localStorage due to security concerns.   Because of this, if a user closes their browser and open it back up at a later time, the application will continue to work (for up to 1 hour) using that stored access token, but it will eventually expire because that client's browser does NOT have the refresh token.  We talked to @mmarcotte and @cholly75; they both mentioned that forcing the user to re-login when the browser closes sounds like a better solution in regards to security, so now we don't store any tokens in localStorage (except for when doing local development).

Second, when a user opens a new tab, it broadcasts a 'requestToken' event on the BroadcastChannel.  Any browser that has a refresh token stored in memory will respond with a 'receiveToken' event so that the children tab can also have a reference to the refresh token.  This allows the user to open a new tabs and, at a later time, close the original parent tab without running into issues an hour later because now ALL tab instance have the refresh token in memory.  Each browser tab will refresh their access token periodically.


web-client/src/presenter/actions/receiveRefreshTokenAction.test.js
```diff
@@ -0,0 +1,25 @@
...
+        presenter,
+      },
+      props: {
+        refreshToken: 'someThing',
+      },
+      state: {
+        refreshToken: 'old-value',
+      },
+    });
+    expect(result.state.refreshToken).toBe('someThing');
```
halprin: A simple improvement is to variable-ize `'someThing'`.  So, you say `const newRefreshToken = 'someThing';` which allows you to do `refreshToken: newRefreshToken` in `props` and then `expect(result.state.refreshToken).toBe(newRefreshToken);`

web-client/src/presenter/actions/setUserAction.test.js
```diff
@@ -32,14 +32,5 @@ describe('setUserAction', () => {
...
     );
-    expect(
-      applicationContext.getUseCases().setItemInteractor.mock.calls.length,
-    ).toEqual(1);
-    expect(
-      applicationContext.getUseCases().setItemInteractor.mock.calls[0][1],
-    ).toMatchObject({
-      key: 'user',
-      value: user,
-    });
```
halprin: The name of this test is `stores the user onto local storage when running local` with the operative part being `running local`.  So, should we either rename this test to not state we are running locally or should we set the `process.env.IS_LOCAL` to `true` (and clear it afterward) and re-instate this deleted code?

package-lock.json
```diff
@@ -2082,9 +2082,9 @@
           }
         },
         "node-fetch": {
-          "version": "2.6.5",
```
halprin: Are you meaning to commit these changes to this file?  I thought that the only time this is updated is when we explicitly want to update the dependencies.

web-client/src/presenter/actions/startRefreshIntervalAction.test.js
```diff
@@ -33,7 +33,7 @@ describe('startRefreshIntervalAction', () => {
     ).toHaveBeenCalled();
     expect(
       applicationContext.getUseCases().setItemInteractor,
-    ).toHaveBeenCalled();
+    ).not.toHaveBeenCalled();
```
halprin, line 39: Should we add a test with `process.env.IS_LOCAL` set to `true` and therefore have an expect for having `setItemInteractor` to have been called?

halprin (commented):
All my comments should be taken with a grain of salt since I'm new and don't know all of the expectations.  🙂

web-client/src/app.jsx
```diff
@@ -154,18 +152,36 @@ const app = {
...
+      applicationContext.setCurrentUserToken(token);
+    }
 
     presenter.state.cognitoLoginUrl = applicationContext.getCognitoLoginUrl();
 
     presenter.state.constants = applicationContext.getConstants();
 
+    // we must do all this logic before we start the router or render react
+    let startRefreshSequence = false;
+    const isNewTabOpened = !window.location.href.includes('?code');
```
rachelschneiderman, line 170: How do we know this means a new tab was opened? I have never seen this before!

web-client/src/presenter/actions/setTokenAction.test.js
```diff
@@ -55,24 +55,4 @@ describe('setTokenAction,', () => {
       mockToken,
     );
   });
-
-  it('should call applicationContext.getUseCases().setItemInteractor with key token and the value of props.token', async () => {
```
rachelschneiderman, line 59: Instead of removing this, could we setup process.env.IS_LOCAL for this test so that code is still covered?

web-client/src/app.jsx
codyseibert: I'm just assuming.   There are only two ways to get into our private application.  Either you logged in via cognito (which means you were redirected to our application to `/log-in?code=ABC`, or you tried to load the application via a bookmark, new tab, or link.   Maybe I should change the variable name?

codyseibert:
We decided it would be safer to have these changes deploy via our normal refactoring / feature workflow due to the high risk of these changes.

PR #1819, opened 2021-12-02
ustc-1817: use paginated query for getting docket numbers for a given user

mmarcotte:
Some practitioners are on thousands of cases, and all of this information cannot be returned in one DDB query.


mrinsin:
Hey! Just wondering if the test failure on this build were expected/fake failures, since the PR was merged with a broken build? TY in advance!

halprin:
Hmmm, seems to be an intermittent failure.  I created [this DevEx card](https://trello.com/c/rkqiJ4tB) to address it.

PR #1828, opened 2021-12-06
Seal cases in Lower Environments as they are sealed in Production

mmarcotte:
This work aims to let the Production environment immediately inform lower environments that a case (and someday a document) has been sealed in production. This way the lower environments can immediately seal that information so that it is not publicly available.

Solves: https://github.com/ustaxcourt/ato/issues/65

This introduces two new environment variables, which are somewhat self explanatory and documented in `docs/environments/setup.md`

| Variable Name | Description |
|----|----|
| `PROD_ENV_ACCOUNT_ID` | The account ID of the AWS account with Production Data |
| `LOWER_ENV_ACCOUNT_ID` | The account ID of the AWS account where copies of Production Data might live |

It also introduces a new dispatcher to the codebase. This dispatcher is abstracted to use AWS SNS, which sends the messages to the lower environments.

And it introduces a new lambda `seal_in_lower` that performs the sealing in the lower environments, acting as a `docketclerk`.

Documentation for how this works is detailed in `docs/operations/sealing-items-in-lower-environments.md`

sealing-items-in-lower-environments


codyseibert (approved):
It looks good from what I can see.  I guess my one question is the manual setup subscription set.  That could potentially be setup in terraform right?  What was the main reason to keep that a manual step?  I think as long as we use count=0 on environments that don't care about setting this up it should all work fine?

shared/src/business/useCases/sealCaseInteractor.test.js
```diff
@@ -8,6 +8,9 @@ describe('sealCaseInteractor', () => {
     applicationContext
       .getPersistenceGateway()
       .getCaseByDocketNumber.mockReturnValue(MOCK_CASE);
+
+    applicationContext.getNotificationGateway().sendNotificationOfSealing =
```
rachelschneiderman: Nit: can this be removed? It looks like `sendNotificationOfSealing` is retrieved from `getDispatchers` which is already mocked in `createTestApplicationContext`

shared/src/persistence/dynamo/helpers/isCurrentColorActive.js
```diff
@@ -0,0 +1,20 @@
+const client = require('../../dynamodbClientService');
+
+/**
+ * Checks to see if the environment currentColor matches that of the deploy table
+ *
+ * @param {object} applicationContext The applicationContext
+ * @returns {Promise} which resolves to whether or not the current color matches what's in the deploy table
+ */
+exports.isCurrentColorActive = async applicationContext => {
+  const { current: currentColor } = await client.get({
```
rachelschneiderman: Nit: dynamoDbClientService has a `getFromDeployTable` function, perhaps that could be used here?

web-api/src/cases/sealInLowerEnvironmentLambda.js
```diff
@@ -0,0 +1,27 @@
+const createApplicationContext = require('../applicationContext');
+
+/**
+ * used for retroactively sealing a case in a lower environment after it is sealed in the Production environment
+ *
+ * @param {object} event the AWS event object received that includes any messages from our SNS subscription
+ * @returns {Promise<*>|undefined} the response to the topic
+ */
+exports.sealInLowerEnvironmentLambda = async event => {
+  const user = { role: 'docketclerk' };
```
rachelschneiderman, line 10: Could this logic be extracted into an interactor? Also, do we need to wrap this lambda in the genericHandler?

mmarcotte:
> It looks good from what I can see. I guess my one question is the manual setup subscription set. That could potentially be setup in terraform right? What was the main reason to keep that a manual step? I think as long as we use count=0 on environments that don't care about setting this up it should all work fine?

@codyseibert -- the reason for the manual subscription is that only some environments have production data. We would only want this subscription to exist with environments that have production data. I could see adding another ENV var, but thought we might want to manually manage these subscriptions. Sometimes `dev` has Prod data, so it would be easy to subscribe and then unsubscribe.

shared/src/business/useCaseHelper/sealInLowerEnvironment.js
```diff
@@ -2,32 +2,41 @@
...
-    //     docketNumber,
-    //   });
-  } else if (docketNumber) {
-    return await applicationContext
-      .getUseCases()
-      .sealCaseInteractor(applicationContext, {
-        docketNumber,
-      });
+exports.sealInLowerEnvironment = async (applicationContext, records) => {
+  const isCurrentColorActive = await applicationContext.isCurrentColorActive(
```
codyseibert, line 9: 👊 thanks for doing this!

web-api/src/cases/sealInLowerEnvironmentLambda.js
mmarcotte: Took a stab at moving the logic. lmk if that helps with visibility and testing. Thanks for your feedback!

rawillis3:
whatever is the current system for sealing records has the effect of sealing the entire docket, even if there is a published decision in the file. case in point
https://dawson.ustaxcourt.gov/case-detail/23809-17
which includes T.C. Memo. 2020-89, issued June 17, 2020, which has since been reversed by the 11th Circuit appeals court.
probably there is something in this file that should properly be sealed, though I cannot think what, but certainly there can be no justification for suppressing a published opinion

PR #1829, opened 2021-12-07
Bug 9235 missing pdf form text

shumway-tm:
Fixes bug to capture PDF Form Fields.

### DIFF
--- a/shared/src/business/useCases/pdf/sanitizePdfInteractor.js
+++ b/shared/src/business/useCases/pdf/sanitizePdfInteractor.js
@@ -1,10 +1,10 @@
 /**
- * setPdfField helper function
+ * setPdfFields helper function
  *
  * @param {array} fields array of pdf form fields
  */
 
-exports.setPdfField = fields => {
+exports.setPdfFields = fields => {
   fields.forEach(field => {
     const fieldType = field.constructor.name;
     if (fieldType === 'PDFTextField') {
@@ -35,12 +35,11 @@ exports.sanitizePdfInteractor = async (applicationContext, { key }) => {
 
   const pdfDoc = await PDFDocument.load(pdfData, { ignoreEncryption: true });
   const form = pdfDoc.getForm();
-  const fieldCount = form.getFields().length;
   const fields = form.getFields();
 
-  applicationContext.getUseCaseHelpers().setPdfField(fields);
+  applicationContext.getUseCaseHelpers().setPdfFields(fields);
 
-  if (fieldCount > 0) {
+  if (fields.length > 0) {
     form.flatten();
     const pdfBytes = await pdfDoc.save();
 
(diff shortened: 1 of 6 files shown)


TomElliottFlexion (approved):

mmarcotte (approved):

ujahio (approved):

jimlerza (approved):

shared/src/business/useCases/pdf/sanitizePdfInteractor.test.js
```diff
@@ -45,6 +52,21 @@ describe('sanitizePdfInteractor', () => {
     ).toBe(mockKey);
   });
 
+  it('calls setPdfFields when PDF has form fields', async () => {
```
rachelschneiderman: I really like this test name! Does this mean that the call to `applicationContext.getUseCaseHelpers().setPdfFields(fields)` should be moved into the if check on line 42?

shared/src/business/useCases/pdf/sanitizePdfInteractor.js
```diff
@@ -1,10 +1,27 @@
+/**
+ * setPdfFields helper function
+ *
+ * @param {array} fields array of pdf form fields
+ */
+
+exports.setPdfFields = fields => {
```
rachelschneiderman: I love that this was added as a use case helper! 

Nit: Could it be extracted into its own file in the `useCaseHelpers` folder?

shared/src/business/useCases/pdf/sanitizePdfInteractor.test.js
TomElliottFlexion: Good call!

shared/src/business/useCases/pdf/sanitizePdfInteractor.test.js
```diff
@@ -26,6 +30,9 @@ describe('sanitizePdfInteractor', () => {
     });
 
     applicationContext.getPersistenceGateway().deleteDocumentFromS3 = jest.fn();
+
+    applicationContext.getUseCaseHelpers().setPdfFields = jest.fn();
```
rachelschneiderman: Nit: `setPdfFields` and `deleteDocumentFromS3` could be added to the `createTestApplicationContext` to follow our "mocking strategy" a little more closely

shared/src/business/useCases/pdf/sanitizePdfInteractor.js
TomElliottFlexion: Let's do it

rachelschneiderman (approved):

shared/src/business/useCases/pdf/sanitizePdfInteractor.test.js
shumway-tm: Thanks for the heads up! Fixed it!

PR #1831, opened 2021-12-08
Refactor: check to see if we have anything to delete...

mmarcotte:
before deleting them. 

Previously, if one of the `.get` calls couldn't find an item, we still proceeded forward trying to access properties of an object that was undefined. 

This refactor simply checks to see if we have anything to delete.

### DIFF
--- a/shared/src/persistence/dynamo/notifications/deleteUserConnection.js
+++ b/shared/src/persistence/dynamo/notifications/deleteUserConnection.js
@@ -17,6 +17,13 @@ exports.deleteUserConnection = async ({ applicationContext, connectionId }) => {
     applicationContext,
   });
 
+  if (!connection) {
+    // it must have expired
+    return;
+  }
+
+  const toDelete = [connection];
+
   const userConnection = await client.get({
     Key: {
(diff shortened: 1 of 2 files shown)


TomElliottFlexion (approved):

jimlerza (approved):

shared/src/persistence/dynamo/notifications/deleteUserConnection.js
```diff
@@ -17,6 +17,13 @@ exports.deleteUserConnection = async ({ applicationContext, connectionId }) => {
     applicationContext,
   });
 
+  if (!connection) {
+    // it must have expired
```
rachelschneiderman: Nit: Could this comment be added to that first test name instead?

rachelschneiderman (approved):

shared/src/persistence/dynamo/notifications/deleteUserConnection.js
mmarcotte: done!

codyseibert (approved):

Issue #1830, opened 2021-12-07
API Failover


mmarcotte:
Quick fix for now to evaluate target health for domain names in route 53 east & west.

PR #1850, opened 2021-12-16
ustc-1830: toggle evaluate_target_health to true for API domains

mmarcotte:
This toggles on the evaluate target health option for the public and private apis. This way if a region is experiencing network issues and the API endpoint is deemed unhealthy by AWS, then it will not receive API requests, and the application will failover to another healthy record with the same domain name in another availability zone.

This should cover the following subdomains:

* `api-blue.{EFCMS_DOMAIN}`
* `api-green.{EFCMS_DOMAIN}`
* `public-api-blue.{EFCMS_DOMAIN}`
* `public-api-green.{EFCMS_DOMAIN}`
* `api.{EFCMS_DOMAIN}`
* `public-api.{EFCMS_DOMAIN}`

Solves: #1830 

Screen Shot 2021-12-16 at 2 39 22 PM


codyseibert (approved):

halprin (approved):
Slick!

jimlerza (approved):

PR #1854, opened 2021-12-17
Batch 14 - Gingerbread

jamesobrooks:
BUG: If Docket adds a document to pending report, and then marks a case as high priority, the UI indicates the case is Blocked #8927
BUG: Users are getting logged out of DAWSON unexpectedly#8949
BUG: Max character count error on document title displaying erroneously #8925
Court: Activate Changes to Notice of Receipt Of Petition (8652)#8714
Court: Activate Motion to Proceed Remotely in DAWSON#9225
External: Opinion Search (Public User)#9210
External Search: Turn on/off Opinion Search for External Users in Production#9261

- NOTE: Requires `scripts/setup-external-opinion-search-flag.sh` to be run, which currently defaults the feature flag to `false`

External: Opinion Search (Private/IRS Practitioner)#9211


web-client/src/presenter/actions/startRefreshIntervalAction.js
```diff
@@ -6,31 +6,24 @@ import { state } from 'cerebral';
  * @param {object} providers the providers object
  * @param {object} providers.store the cerebral store used for setting state.workItem
  */
-export const startRefreshIntervalAction = ({
+export const startRefreshIntervalAction = async ({
   applicationContext,
```
mmarcotte: just a warning, but this function doesn't have any async operations.

mmarcotte (approved):

PR #1858, opened 2021-12-23
9036 bug

ujahio:
Bumping terraform async lambda function's memory size to 7000MB for testing in court's test environment.

7000MB was chosen as the max memory based on the results of querying the lambda's memory usage while we downloaded the extremely large case, 24155-18 .
image


mrinsin (approved):

ujahio:
@jimlerza Please take a look at this PR. This PR does have a lot of the changes from the court's staging that has not been transferred over to the court's testing environment. Our changes are specific to ```web-api/terraform/api/api-async.tf```

Thank you.

mrinsin (approved):

jimlerza (approved):

PR #1868, opened 2021-12-30
Find cases missing practitioners

jimlerza:
- Added a helper for retrieving large datasets from Elasticsearch.
- Added a script to find instances where practitioner-to-case relationships exist as UserCase entities but do not also have a corresponding PrivatePractitioner or IrsPractitioner entity. The performance is simply horrendous because the aforementioned `UserCase` entities are not presently indexed in Elasticsearch.

### DIFF
--- a/shared/admin-tools/elasticsearch/find-cases-missing-practitioners.js
+++ b/shared/admin-tools/elasticsearch/find-cases-missing-practitioners.js
@@ -1,0 +1,228 @@
+// how to run
+// node find-cases-missing-practitioners.js mig alpha
+
+const AWS = require('aws-sdk');
+const { searchAll } = require('../../../web-api/elasticsearch/searchAll');
+
+const environmentName = process.argv[2] || 'exp1';
+const version = process.argv[3] || 'alpha';
+const esClientArgs = { environmentName, version };
+
+const documentClient = new AWS.DynamoDB.DocumentClient({
+  endpoint: 'dynamodb.us-east-1.amazonaws.com',
+  region: 'us-east-1',
+});
+const dynamoTableName = `efcms-${environmentName}-${version}`;
+
+/**
+ * Get active practitioners' user pks
+ *
+ * @param {string} practitionerType either 'privatePractitioner' or 'irsPractitioner'
+ * @returns {Promise<Array<String>>} array of user pks
+ */
+const getActivePractitionerUserPks = async practitionerType => {
+  const query = {
+    bool: {
+      must: [
+        {
+          term: {
+            'role.S': {
+              value: practitionerType,
+            },
+          },
+        },
+        {
+          term: {
+            'admissionsStatus.S': {
+              value: 'Active',
+            },
+          },
+        },
+      ],
+    },
+  };
+
+  const hits = await searchAll(
+    esClientArgs,
+    'efcms-user',
+    query,
+    [{ 'pk.S': 'asc' }],
+    ['pk.S'],
+  );
+
+  // eslint-disable-next-line no-underscore-dangle
+  return hits.map(hit => hit._source.pk.S);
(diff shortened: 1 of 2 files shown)


jimlerza:
Here is the find-cases-missing-practitioners script output, as run against efcms-test-alpha on 2021-12-29:

[cases-missing-practitioners.txt](https://github.com/ustaxcourt/ef-cms/files/7793314/cases-missing-practitioners.txt)

This executed for 4 hours and 45 minutes.

codyseibert (approved):

shared/admin-tools/elasticsearch/find-cases-missing-practitioners.js
```diff
@@ -0,0 +1,228 @@
+// how to run
```
jamesobrooks, line 1: Can we add this to a readme?

jimlerza: We absolutely can!

However, I have a suspicion that, given this script took nearly 5 hours to execute, it is probably not worth keeping around once [#9323](https://github.com/flexion/ef-cms/issues/9323) is addressed.

jimlerza: I don't want to "solution" that bug, but I'd imagine the bones of this script could potentially be used as the basis for a migration that generates the missing entities and then this script could be deleted.

jamesobrooks: Oh, that makes sense.

rosiehsmith (approved):

rachelschneiderman (commented):
Would it be worth it to add tests to this code? I realize the script is probably temporary, but maybe at least `searchAll.js`

jimlerza:
Should `searchAll.js` be moved into [shared/src/persistence/elasticsearch/searchClient.js](https://github.com/ustaxcourt/ef-cms/blob/staging/shared/src/persistence/elasticsearch/searchClient.js)?

Ideally, I'd like for this to be usable in as many places as necessary.

jimlerza:
> Would it be worth it to add tests to this code? I realize the script is probably temporary, but maybe at least `searchAll.js`

Yes, definitely. At very least, tests for:

* Ensuring the `search` query is not performed if the `count` query returns 0
* Ensuring the number of results returned matches the expected count

rachelschneiderman:
> Should `searchAll.js` be moved into [shared/src/persistence/elasticsearch/searchClient.js](https://github.com/ustaxcourt/ef-cms/blob/staging/shared/src/persistence/elasticsearch/searchClient.js)?
> 
> Ideally, I'd like for this to be usable in as many places as necessary.

That seems like a good idea to me!

web-api/elasticsearch/searchAll.js
```diff
@@ -0,0 +1,56 @@
...
+ * Uses Elasticsearch's `search_after` to return all results for the provided query
+ *
+ * @param {Object} clientArgs object containing the keys 'environmentName' and 'version'
+ * @param {String} index Elasticsearch index
+ * @param {Object} query Elasticsearch query
+ * @param {Array<Object>} sort array of objects describing the fields and direction to sort by
+ * @param {Array<String>} source array of fields to return
+ * @returns {Promise<Array<Object>>} results
+ */
+const searchAll = async (clientArgs, index, query, sort, source) => {
```
codyseibert, line 16: I would recommend refactoring this to be an option object instead of an order list of arguments.  Most of the functions in our application take in either 1 or 2 arguments max using objects with named properties

halprin:
The last movement on this was a month ago and three months ago.  This has approvals.  Can we merge this or close this?

PR #1869, opened 2022-01-03
BUG 9312: Public Terminal User is no longer working 

rosiehsmith:


mrinsin (approved):

codyseibert (approved):
Awesome test.  I'm disappointed our `web-client/integration-tests-public/terminalUserAccessCaseDocuments.test.js` integration test didn't catch this, but I guess it wouldn't since it's at a lower level

jimlerza (approved):

PR #1879, opened 2022-01-05
9310 Update Case status >45 days after Answer filed

jamesobrooks:
This change will require a reindex of elasticsearch indices.


codyseibert (approved):

jimlerza (approved):

jimlerza (approved):

shared/src/persistence/elasticsearch/getReadyForTrialCases.js
```diff
@@ -39,7 +34,7 @@ exports.getReadyForTrialCases = async ({ applicationContext }) => {
               },
               {
                 range: {
-                  'servedAt.S': {
+                  'createdAt.S': {
```
jimlerza, line 36: Does this need to be added to `efcms-docket-entry-mappings.js`?

jamesobrooks: Hmm. Good catch.

jamesobrooks:
Testing this out locally it's not working. Will reopen once it's working.

shared/src/persistence/elasticsearch/getReadyForTrialCases.js
```diff
@@ -17,13 +17,65 @@ exports.getReadyForTrialCases = async ({ applicationContext }) => {
...
+                    },
+                  },
+                  score: true,
+                },
+              },
+            ],
+            must_not: [
+              {
+                term: {
+                  'isStricken.BOOL': true,
```
codyseibert, line 70: do we care if an answer is isStricken?

jamesobrooks: I guess I don't know for sure. I would think so because if it's stricken, then technically there has been no Answer filed right? I'll check with Chris though.

jamesobrooks: Confirming that Chris Holly says the stricken docket entries are to be treated as if they never happened, so we do want to exclude them.

PR #1902, opened 2022-02-14
9421 - Using Promise.all to fix issue with fetching eligible cases

codyseibert:
the old approach was looping one by one over each cases which was slow; changing to promise.all


jimlerza (approved):

rosiehsmith (approved):

rachelschneiderman (approved):
Awesome job! Would love to see if there's a possible way to wrap this with a test, but otherwise... Looks good!

PR #1905, opened 2022-02-14
9323: Adding logging to ensure bug-9323 scenario is logged

TomElliottFlexion:
Adds logging to record events similar to those in bug [9323](https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/flexion/ef-cms/9323). 

When private practitioners who already have a user-case association/record (possibly due to being a petitioner on the case) but are not yet _practitioners_ on the case are added to said case, we will now have a log to aid in remedying any resulting broken data.


halprin (approved):
🔥

jimlerza (approved):

PR #1913, opened 2022-02-17
9365 removing code to sanitize pdfs

mrinsin:


codyseibert (approved):

rosiehsmith (approved):

halprin (commented):
So, the decision was to just remove sanitization all together?

Good bye!
cry salute

jimlerza (approved):

PR #1915, opened 2022-02-18
9323: Updated faulty logging level

TomElliottFlexion:


halprin (approved):
`warn` doesn't work?  Is it not a thing?

TomElliottFlexion:
Not when using the backend `applicationContext`, we have `debug`, `error`, and `info` as options.

jimlerza (approved):

PR #1931, opened 2022-03-01
9326 - Notice of Change to Remote Proceeding

jamesobrooks:


web-client/integration-tests/trialSessionChangeTRemoteProceedingJourney.test.js
```diff
@@ -0,0 +1,298 @@
+import {
+  CASE_STATUS_TYPES,
```
codyseibert: is this file named correct? trialSessionChangeTRemoteProceedingJourney.test.js

why does it have `ChangeT` in the name?

codyseibert (changes requested):

web-client/integration-tests/trialSessionChangeTRemoteProceedingJourney.test.js
jamesobrooks: Good catch. That should be `trialSessionChangeToRemoteProceedingJourney.test.js`.

codyseibert (approved):

shared/src/business/entities/EntityConstants.js
```diff
@@ -479,6 +479,11 @@ const SYSTEM_GENERATED_DOCUMENT_TYPES = {
     documentType: 'Notice of Docket Change',
     eventCode: 'NODC',
   },
+  noticeOfChangeToRemoteProceeding: {
```
rachelschneiderman, line 482: Does this needed to be added to `courtIssuedEventCodes.json` or any of the others?

shared/src/business/useCaseHelper/pdf/savePaperServicePdf.js
```diff
@@ -0,0 +1,34 @@
+/**
+ * savePaperServicePdf helper function
+ *
+ * @param {array} applicationContext the application context
+ * @param {array} document the document to be saved
+ * @returns {Object} the service information
+ */
+
+exports.savePaperServicePdf = async ({ applicationContext, document }) => {
+  let hasPaper = document.getPages().length;
```
rachelschneiderman: Has paper is the length of an array here and then a boolean elsewhere, maybe another variable is needed?

shared/src/business/useCaseHelper/pdf/savePaperServicePdf.test.js
```diff
@@ -0,0 +1,107 @@
...
+
+    applicationContext
+      .getPersistenceGateway()
+      .getDownloadPolicyUrl.mockReturnValue({ url: mockDocumentUrl });
+
+    mockDocument.getPages.mockReturnValue({ length: 1 });
+    mockDocument.save.mockReturnValue(mockDocument);
+  });
+
+  it('should not return a url or docketEntryId when there is no paper service on the case', async () => {
```
rachelschneiderman, line 25: Nit: based on the file above it looks like it does/does not do certain actions based on whether the document passed in has pages. Is that correct? If so, can the test names be updated to reflect that behavior?

shared/src/business/useCaseHelper/trialSessions/setNoticeOfChangeToRemoteProceeding.js
```diff
@@ -0,0 +1,146 @@
...
+      });
+  }
+};
+
+/**
+ * setNoticeOfChangeToRemoteProceeding
+ *
+ * @param {object} providers the providers object
+ * @param {object} providers.applicationContext the application context
+ * @param {object} providers.caseEntity the case data
```
rachelschneiderman, line 50: Nit: a couple args are missing in these docs

shared/src/business/useCaseHelper/trialSessions/setNoticeOfChangeToRemoteProceeding.js
```diff
@@ -0,0 +1,146 @@
...
+ * @param {object} providers.currentTrialSession the old trial session data
+ * @param {object} providers.newTrialSessionEntity the new trial session data
+ * @param {object} providers.userId the user ID
+ * @returns {object} the created trial session
+ */
+exports.setNoticeOfChangeToRemoteProceeding = async (
```
rachelschneiderman, line 58: This method makes excellent use of helper methods, it's extremely easy to read and understand high level what's going on. Nice job!

shared/src/business/useCaseHelper/trialSessions/setNoticeOfChangeToRemoteProceeding.js
```diff
@@ -0,0 +1,146 @@
...
+      await applicationContext.getUseCaseHelpers().countPagesInDocument({
+        applicationContext,
+        docketEntryId:
+          noticeOfChangeToRemoteProceedingDocketEntry.docketEntryId,
+      });
+
+    caseEntity.addDocketEntry(noticeOfChangeToRemoteProceedingDocketEntry);
+    const servedParties = aggregatePartiesForService(caseEntity);
+
+    noticeOfChangeToRemoteProceedingDocketEntry.setAsServed(servedParties.all);
```
rachelschneiderman, line 137: Nit: would it be useful to set the docket entry as served after the persistence call that serves it? just in case that call fails

shared/src/business/useCaseHelper/trialSessions/setNoticeOfChangeToRemoteProceeding.test.js
```diff
@@ -0,0 +1,219 @@
...
+    closedDate: '2020-03-01T21:42:29.073Z',
+    docketNumber: '999-99',
+    status: CASE_STATUS_TYPES.closed,
+    trialDate: '2019-03-01T21:42:29.073Z',
+    trialSessionId,
+  },
+  { applicationContext },
+);
+
+describe('setNoticeOfChangeToRemoteProceeding', () => {
```
rachelschneiderman, line 50: These test cases are amazing! 👏

shared/src/business/entities/EntityConstants.js
jamesobrooks: I believe @mrinsin said that the other events that are not created manually do not go into that file. Perhaps it should really be called `courtPersonnelIssuedEventCodes.json` or all should go in that file also?

mrinsin: @rachelschneiderman it seems like other auto-generated trial related notices were also defined in `EntityConstants.js` without a counterpart in any of the document related JSONs (maybe to avoid them showing up in certain dropdowns?), so we decided to follow that pattern!

shared/src/business/useCaseHelper/trialSessions/setNoticeOfChangeToRemoteProceeding.js
mrinsin: Good idea! We saw it done this way in other places that were doing similar stuff, do you think we should change those calls too? Or maybe we need stick to the scope of this story?

shared/src/business/useCaseHelper/trialSessions/setNoticeOfChangeToRemoteProceeding.js
mrinsin: Thanks! TODO: add docs

shared/src/business/useCaseHelper/pdf/savePaperServicePdf.js
mrinsin: Will do! TY

shared/src/business/useCases/trialSessions/updateTrialSessionInteractor.js
```diff
@@ -140,6 +143,31 @@ exports.updateTrialSessionInteractor = async (
       if (
         caseToUpdate.trialSessionId === newTrialSessionEntity.trialSessionId
       ) {
+        //todo add paper service url to output
```
rachelschneiderman: Nit: can this be removed?

web-client/integration-tests/journey/docketClerkCreatesATrialSession.js
```diff
@@ -132,5 +134,6 @@ export const docketClerkCreatesATrialSession = (
     expect(lastCreatedTrialSessionId).toBeDefined();
 
     cerebralTest.lastCreatedTrialSessionId = lastCreatedTrialSessionId;
+    cerebralTest.trialSessionId = cerebralTest.lastCreatedTrialSessionId;
```
rachelschneiderman, line 137: Are both of these needed? I think they are the same value?

shared/src/business/useCases/trialSessions/updateTrialSessionInteractor.js
mrinsin: TY! TODO: remove comment

rachelschneiderman (approved):

web-client/integration-tests/journey/docketClerkCreatesATrialSession.js
mrinsin: I believe `lastCreatedTrialSessionId` is used in a separate journey file that calls this! However we need `trialSessionId` to be set in state because of yet another file we call from `web-client/integration-tests/trialSessionChangeToRemoteProceedingJourney.test.js` that expects `cerebralTest.trialSessionId` to be set. I can def make a devex to DRY up trial session related integration tests, if its needed?

shared/src/business/useCaseHelper/pdf/savePaperServicePdf.test.js
mrinsin: That is correct! TODO: update test names

shared/src/business/useCaseHelper/trialSessions/setNoticeOfChangeToRemoteProceeding.js
mrinsin: Hey actually I just remembered this is not possible to do - `serveNoticesForCase` expects certain served properties to already have been set on the docket entry, and then proceeds to send the service emails. So we MUST call `setAsServed` before calling `serveNoticesForCase`

PR #1933, opened 2022-03-01
DevEx 1043: Consolidate Running Integration Tests

halprin:
The CircleCI run of our integration tests had a large number of environment variables to modify our `npx jest` run of our integration tests.  I decided to move that into an NPM script `test:client:integration:ci`.

In addition to that, I noticed that our NPM script `test:client:integration` didn't run the same stuff (or use the same Jest configuration) as our CircleCI run.  I consolidated them so that both `test:client:integration` and the new `test:client:integration:ci` call to the same configuration.

I also removed the NPM script `test:client:integration:local` because `test:client:integration` already is set for running locally and it was using a different configuration than the CircleCI version. anyway.

A future PR will come out at some point that will hopefully expand on running the integration tests locally in a parallel fashion similar to CircleCI.  That will speed up the execution of the integration tests locally and hopefully prevent some the "JavaScript out of memory" failures you sometimes get when running all the integration tests together.


codyseibert (approved):

rachelschneiderman (approved):
Could you please make a trello card for this on the devex/opex board so it can be tracked as part of the release? Thanks!

halprin:
> Could you please make a trello card for this on the devex/opex board so it can be tracked as part of the release? Thanks!

Done, @rachelschneiderman.  Created [card 1043](https://trello.com/c/C2LBr3xP).

mrinsin (approved):

PR #1934, opened 2022-03-02
DEVEX - refactoring the code to no longer call inline functions

codyseibert:
cleaning up some of this code to no longer have a bunch of nested inline functions which were hard to read.  Also adding a small refactoring to fetch the latest version of the case right before it works on updating it instead of potentially using a 10 minutes old copy of the case.


shared/src/business/useCases/trialSessions/setNoticesForCalendaredTrialSessionInteractor.js
```diff
@@ -14,6 +14,244 @@ const { DocketEntry } = require('../../entities/DocketEntry');
 const { TrialSession } = require('../../entities/trialSessions/TrialSession');
 const { UnauthorizedError } = require('../../../errors/errors');
 
+/**
+ * serves a notice of trial session and standing pretrial document on electronic
+ * recipients and generates paper notices for those that get paper service
+ *
+ * @param {object} deconstructed function arguments
```
rachelschneiderman: Nit: seems like there's a couple args missing in the docs

rachelschneiderman (approved):

codyseibert:
adding a do not merge label because some other work was done related to this code and I want to verify that code gets merged in first

PR #1940, opened 2022-03-03
9401: Sorting of Pending Items

halprin:
Our previous bug fix didn't completely work when there were multiple pages of pending items to view.  Therefore, we deleted the sorting that we did on the frontend and now have ElasticSearch sort for us.  We put a comment in the interactor that we depend on the gateway implementation to return the sorted list.  We also improved an integration test to test the that the correct order is returned even when pagination is involved.  This required us to add a new case with over 100 pending items to our seed data.


web-client/integration-tests/helpers.js
```diff
@@ -927,3 +935,25 @@ export const updateOrderForm = async (cerebralTest, formValues) => {
     'updateAdvancedOrderSearchFormValueSequence',
   );
 };
+
+export const verifySortedRecievedAtDateOfPendingItems = pendingItems => {
```
TomElliottFlexion: Nit: received

TomElliottFlexion (approved):

web-client/integration-tests/helpers.js
halprin: Hah!  Nice catch.  This has been fixed here and in the other locations where it was misspelled.

shared/src/persistence/elasticsearch/fetchPendingItems.js
```diff
@@ -73,6 +73,18 @@ exports.fetchPendingItems = async ({
         },
       },
       size,
+      sort: [
```
rachelschneiderman, line 76: Nit: does this file need some unit test coverage?

shared/src/business/utilities/DateHandler.js
```diff
@@ -432,6 +432,17 @@ const validateDateAndCreateISO = date => {
   }
 };
 
+/**
+ * Subtracts a previous year from passed in ISO date
```
rachelschneiderman: Smallest tiniest nit: Should this say "subtracts specified amount of time" instead of "previous year" as it looks like this is configurable?

web-client/integration-tests/docketClerkViewsAPendingItem.test.js
```diff
@@ -69,22 +72,14 @@ describe('a docket clerk uploads a pending item and sees that it is pending', ()
...
 
     expect(formatted.pendingItemsDocketEntries.length).toEqual(0);
-
-    await refreshElasticsearchIndex();
-
-    await cerebralTest.runSequence('gotoPendingReportSequence');
-
-    await cerebralTest.runSequence('setPendingReportSelectedJudgeSequence', {
-      judge: 'Chief Judge',
+    await docketClerkLoadsPendingReportOnChiefJudgeSelection({
```
rachelschneiderman, line 80: 👏 nice refactor!

shared/src/persistence/elasticsearch/fetchPendingItems.js
halprin: Oooo, good find.  Yes.  Yes it should.

shumway-tm: Added a unit test to cover the sort lines!

shared/src/business/utilities/DateHandler.js
shumway-tm: Changed the wording of the comment to better reflect that the function is indeed configurable!

web-client/integration-tests/docketClerkViewsAPendingItem.test.js
shumway-tm: Thank you! It was truly a mob effort!

PR #1941, opened 2022-03-03
Merge Bug Fixes

jamesobrooks:
These bug fixes are to be included in a release candidate and merged to staging so that other work can begin to be merged to test. Pending verification of final work of 9401 after deployment.


codyseibert (approved):

halprin (approved):

shared/src/business/useCaseHelper/caseAssociation/associatePrivatePractitionerToCase.test.js
```diff
@@ -201,4 +201,59 @@ describe('associatePrivatePractitionerToCase', () => {
       serviceIndicator: SERVICE_INDICATOR_TYPES.SI_ELECTRONIC,
     });
   });
+
+  it('BUG 9323: should create log if practitioner is already associated with case but does not appear in the privatePractitioners array', async () => {
```
rosiehsmith, line 205: just curious, why are we including "BUG 9323" in the log?

shared/src/business/utilities/DateHandler.test.js
```diff
@@ -521,4 +521,16 @@ describe('DateHandler', () => {
       expect(validDate).toEqual('2001-01-01T05:00:00.000Z');
     });
   });
+
+  describe('subtractISODates)', () => {
+    it('return an ISO string of the previous year', () => {
```
jimlerza, line 526: Do we need a case that takes timezones that use DST into consideration?

Ex: 

- subtracting 1 month from 2AM on 4/13/22 (ISO date "2022-04-13T02:00:00.000-0400") should return 3AM on 3/13/22 (ISO date "2022-03-13T03:00:00.000-0400") because there is no 2AM on 3/13/22.
- subtracting 1 month from the first 2AM on 11/6/22 (ISO date "2022-11-06T02:00:00.000-0400") should return 2AM on 10/6/22 (ISO date "2022-10-06T02:00:00.000-0400")
- subtracting 1 month from the second 2AM on 11/6/22 (ISO date "2022-11-06T02:00:00.000-0500") should return....uhhh....something?

jimlerza (approved):

shared/src/business/useCaseHelper/caseAssociation/associatePrivatePractitionerToCase.test.js
halprin: Hmmm.  I think because we wanted to call attention to this bug, even in our test output.  Not a good answer, but an answer.  😅

rosiehsmith: Okay thanks! My preference is not to include specific issue numbers in the code unless it's providing us some definitive value.

halprin: That makes sense.  Are you OK to merge?  I, personally, don't think this is something to stop this merge.

rosiehsmith: Yep! just a nitpick

TomElliottFlexion: Since we are intentionally NOT fixing this bug, we felt it was best to highlight the issue as often as possible

rosiehsmith: https://github.com/ustaxcourt/ef-cms/pull/1945#issuecomment-1059418145

PR #1943, opened 2022-03-04
devex - ran into some issues trying to run these teardown scripts

codyseibert:


mrinsin (approved):

env-destroy-utils/deleteCognitoPool.js
```diff
@@ -0,0 +1,29 @@
+const { CognitoIdentityServiceProvider } = require('aws-sdk');
+const cognito = new CognitoIdentityServiceProvider({
+  region: 'us-east-1',
+});
+
+exports.deleteCognitoPool = async () => {
+  const { UserPools } = await cognito
```
halprin, line 7: Weird AWS making an uppercase field in the object.

env-destroy-utils/deleteCognitoPool.js
```diff
@@ -0,0 +1,29 @@
+const { CognitoIdentityServiceProvider } = require('aws-sdk');
+const cognito = new CognitoIdentityServiceProvider({
+  region: 'us-east-1',
+});
+
+exports.deleteCognitoPool = async () => {
```
halprin: I see in the other `env-destroy-utils` scripts that we pass in `environment` which has the region and environment name specified.  Should we do that here too  so we don't need to hardcode `us-east-1` above and `stg` below?

env-destroy-utils/destroyEnvironment.js
```diff
@@ -68,6 +65,12 @@ const teardownEnvironment = async () => {
   } catch (e) {
     console.error('Error while deleting cloudwatch logs: ', e);
   }
+
+  try {
+    await deleteCognitoPool();
```
halprin: Here's where we can pass in the appropriate `environment`.

env-destroy-utils/destroyEnvironment.js
```diff
@@ -14,19 +15,15 @@ if (!process.env.AWS_ACCESS_KEY_ID || !process.env.AWS_SECRET_ACCESS_KEY) {
 const environmentName = process.argv[2] || 'exp1';
 
 const environmentEast = {
-  accessKeyId: process.env.AWS_ACCESS_KEY_ID,
```
halprin, line 17: I presume you deleted these here because you wanted the AWS SDK to use your `~/.aws/credentials` file?  That makes sense, but there are other files in the `env-destroy-utils` directory that kind of use `accessKeyId` and `secretAccessKey`.  E.g. `getS3.js`.  That file probably will still work since `environment.secretAccessKey` will be `undefined`, but, we should remove the other parts that depend on these fields since we're removing them here.

web-api/migration-cron-terraform/bin/destroy-app.sh
```diff
@@ -41,6 +41,9 @@ export TF_VAR_migrate_flag=$MIGRATE_FLAG
 export TF_VAR_source_table=$SOURCE_TABLE
 export TF_VAR_stream_arn=$STREAM_ARN
 
+mkdir -p ./lambdas/dist
+touch ./lambdas/dist/reindex-status.js
```
halprin, line 45: Why do we want to make that empty file?

env-destroy-utils/deleteCognitoPool.js
codyseibert: cognito is always in us-east-1, we don't multi-region cognito

codyseibert: right, I shouldn't have hard coded `stg`. fixing now

env-destroy-utils/destroyEnvironment.js
codyseibert: good catch!

web-api/migration-cron-terraform/bin/destroy-app.sh
codyseibert: for some reason, terraform expects those files to exist locally for it to successfully run the terraform destroy.    More specifically this code fails saying "file does not exist"

```
// web-api/migration-cron-terraform/main/reindex-status.tf
data "archive_file" "reindex_status_zip" {
  type        = "zip"
  output_path = "${path.module}/lambdas/reindex-status.js.zip"
  source_dir  = "${path.module}/lambdas/dist/"
  excludes = ["migration.js", "migration-segments.js"]
}
```

maybe there is a way to not require the file to exist when doing a tear down?  since it's a `data` resource I guess it always needs to exist.

halprin (approved):
🔥

PR #1944, opened 2022-03-04
Restart Exp Environment Script

jamesobrooks:
Adds a script to make restarting exp environments easier when the ES domains have been deleted.


scripts/load-environment-from-secrets.sh
```diff
@@ -5,10 +5,15 @@
...
 
-REGION=us-east-1
-content=$(aws secretsmanager get-secret-value --region ${REGION} --secret-id "${ENV}_deploy" --query "SecretString" --output text)
-echo $content
-echo ${content} | jq -r 'to_entries|map("\(.key)=\"\(.value)\"")|.[]' > .env
-set -o allexport
-source .env
-set +o allexport
\ No newline at end of file
+if [ $? = 0 ]; then
```
halprin: Does this work?  I thought for numbers it needs to be `-eq`, not `=`.

halprin (commented):
Nice!

scripts/load-environment-from-secrets.sh
jamesobrooks: Huh. It seemed to work but I can double check.

scripts/restart-experimental-environment.sh
```diff
@@ -0,0 +1,48 @@
...
+npm run deploy:api ${ENV}
+
+# Setting up indices
+./web-api/setup-elasticsearch-index.sh "${ENV}"
+
+# Indexing data
+node ./web-api/reindex-dynamodb-records.js "${DESTINATION_TABLE}"
+
+# Setting up users
+node shared/admin-tools/user/setup-admin.js
```
rachelschneiderman, line 41: These scripts hit api.<EFCMS_DOMAIN> endpoints without concern for the current/deploying color. Do they need to be updated to take those into account? 

Alternatively, maybe this could be a job in the circle workflow only for experimental environments?

halprin (commented):
Oh.  One last thought I had was to add some documentation into the `docs` folder if there's already a place that talks about the exp environments.  I'd include how to invoke this script and when/why one would want to invoke this script.

scripts/load-environment-from-secrets.sh
jamesobrooks: You were right. Thanks!

scripts/load-environment-from-secrets.sh
```diff
@@ -5,10 +5,14 @@
...
-echo $content
-echo ${content} | jq -r 'to_entries|map("\(.key)=\"\(.value)\"")|.[]' > .env
-set -o allexport
-source .env
-set +o allexport
\ No newline at end of file
+if [ $? -eq 0 ]; then
+  REGION=us-east-1
+  content=$(aws secretsmanager get-secret-value --region ${REGION} --secret-id "${ENV}_deploy" --query "SecretString" --output text)
+  echo $content
```
codyseibert, line 11: do we need this echo $content?

codyseibert (approved):

rachelschneiderman (approved):

scripts/restart-experimental-environment.sh
jamesobrooks: As discussed, we will make changes to the config.yml separately to address some of this, but leave these changes as-is for manual runs.

PR #1945, opened 2022-03-04
Don't Mention Bug ID in Test it Statement

halprin:
@rosiehsmith [mentioned](https://github.com/ustaxcourt/ef-cms/pull/1941#discussion_r819733249) that we shouldn't need to mention the bug ID in our test `it` statements.  @TomElliottFlexion and @shumway-tm, what do you think?


rosiehsmith (approved):
💯

TomElliottFlexion (commented):
This was intentional, as it is a specific, known issue that we are trying to keep tabs on

TomElliottFlexion (commented):
Phrased differently: these are not tests around a bug fix, they are tests around an intentionally unfixed bug

rosiehsmith:
I don't have strong feelings. Just kind of felt like a code smell to see an issue number in a log and test for something that's an open issue that's also done? If it makes sense to keep it in for logging reference and y'all see value in it, cool 👍

PR #1952, opened 2022-03-07
Add user setup to circle

jamesobrooks:
This sets up users during the CircleCI build to avoid problems with failed smoketests.

I'm not sure why it appears that `setup-test-users.sh` is _empty_, but what I changed was permissions, not content. This was run on [this exp4 build](https://app.circleci.com/pipelines/github/flexion/ef-cms/33760/workflows/79ca3b3f-ecc7-41f9-921c-6f8e81c51891).


codyseibert:
we used to do this at one point, but it was removed for some reason by the tax court (probably back when we just had a hard coded password for these users).  At this point, I don't see an issue creating these users on each deploy since it doesn't run on prod.

.circleci/config.yml
```diff
@@ -435,6 +435,38 @@ jobs:
           name: Setup Elasticsearch Index Settings
           command: |
             ./web-api/setup-elasticsearch-index.sh $ENV
+      - run:
```
halprin, line 438: Do you think we could run these new steps in its own job after the `deploy` job?  If so, we should move these new steps to their own job after `deploy` so they can have their own [branch filter](https://support.circleci.com/hc/en-us/articles/115015953868-Filter-workflows-by-branch-).  That way you don't need any of the `if [ "${CIRCLE_BRANCH}" != "prod" ]; then` statements at all.

Specifically, you'd use an `ignore` statement that lists `prod`.  [Reference](https://circleci.com/docs/2.0/configuration-reference/#filters).

halprin (commented):
Wasn't there also the step of re-indexing the data?

rachelschneiderman (dismissed):

.circleci/config.yml
jamesobrooks: I was thinking about that, but I think it needs to happen _before_ the `Deploy - Web API - Smoke Tests` steps in the `deploy` job. Otherwise if the users don't exist, then those deploys won't be able to get the tokens they need. I could of course be mistaken.

halprin: Oh hmmm.  Then I recommend making two additional jobs.  One for the adding the users (with the branch filter), and one for the smoke tests.

But I'm OK with your current approach.  I would just prefer doing it through two additional jobs.  Your decision.

rachelschneiderman: I think it depends on the problem we are trying to solve here. If we are trying to only solve the issue with experimental environments I think seeding test users could go in a step after the deploy - I say that because the `Deploy - Web API - Smoke Tests` step talks directly to cognito and doesn't actually log into the UI so we don't have to have users seeded in dynamo like we do for the cypress smoketests that occur later.

rachelschneiderman: I don't have an opinion either way! Just wanted to clarify 😄

.circleci/config.yml
```diff
@@ -435,6 +435,38 @@ jobs:
...
+          command: |
+            if [ "${CIRCLE_BRANCH}" != "prod" ]; then
+              ./shared/admin-tools/user/setup-test-users.sh $ENV
+            else
+              echo "skipping…"
+            fi
+      - run:
+          name: Setup Judge Users File Var
+          command: |
+            if [ "${CIRCLE_BRANCH}" != "prod" ] && [ "${CIRCLE_BRANCH}" != "test" ] && [ "${CIRCLE_BRANCH}" != "migration" ]; then
```
rachelschneiderman, line 457: What about ustc:staging?

jamesobrooks: Good question. I was thinking only test and migration were prod-like, but of course staging is. Is migration?

rachelschneiderman: I believe glue jobs are also performed to staging but I could be wrong. @jimlerza is that right?

jimlerza: Glue jobs are only performed to migration and test, although I think it's worth checking staging's Cognito pool to make sure the "real" judge accounts don't already exist there.

jimlerza: Confirmed - the only judge users in staging's Cognito pool are of the `@example.com` variety, so that is an environment where we can/should import judges.

jimlerza (approved):

PR #1958, opened 2022-03-09
9093 Addressing PO Review for PDF Styles & Disabling Cypress Videos

codyseibert:
- Addressing PO Review for PDF Styles
- Disabling Cypress Videos to help improve the smoketests reliability (there is an open github issue saying the cypress videos can cause the TIMEOUT -7 errors)

### DIFF
--- a/cypress-public.json
+++ b/cypress-public.json
@@ -5,6 +5,7 @@
   "reporterOptions": {
     "toConsole": true
   },
+  "video": false,
   "fixturesFolder": "cypress-integration/fixtures",
   "integrationFolder": "cypress-integration/integration",
   "pluginsFile": "cypress-integration/plugins/index.js",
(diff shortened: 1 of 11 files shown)


TomElliottFlexion (approved):

halprin (dismissed):
Looks like you have other changes in this branch other than the video.  Also the title of the PR is a little weird.

cypress-public.json
```diff
@@ -5,6 +5,7 @@
   "reporterOptions": {
     "toConsole": true
   },
+  "video": false,
```
halprin, line 8: Another possible solution identified in [the Cypress issue](https://github.com/cypress-io/cypress/issues/17627#issuecomment-1059741645) is to do...  That way, we still (possibly) get the video on a failure.

```json
"video": true,
"videoUploadOnPasses": false,
```

halprin (commented):
Can't speak to the PDF Styling.  I kind of recommend splitting this into two separate PRs.

PR #1959, opened 2022-03-09
9093 - disable videos on cypress to make tests more reliable

codyseibert:


halprin (approved):
I mentioned about doing something different in this [other comment](https://github.com/ustaxcourt/ef-cms/pull/1958#discussion_r823055531), but we don't use Cypress Dashboard, yet we still see the issue.  So, let's do it this way in this PR.

TomElliottFlexion (approved):

PR #1965, opened 2022-03-10
1046 trigger reindex automatically

jamesobrooks:
Triggers a reindex when the ES source table does not exist at time of deploy, as with the Flexion experimental environments daily.
https://trello.com/c/ZfegULzf/1046-add-reindex-without-migration

Here's a deploy from an environment that had been shut down for the night:
https://app.circleci.com/pipelines/github/flexion/ef-cms/33911/workflows/aa29e67e-7ebf-4118-90cf-cf98e26b554b


.circleci/config.yml
```diff
@@ -940,10 +994,14 @@ workflows:
             - e2e-cypress
             - e2e-cypress-smoketests-local
             - e2e-cypress-public
-      - migrate:
+      - reindex:
```
jamesobrooks, line 1019: I'm honestly not sure this should be here. We really should never have to do this for prod unless maybe we're increasing the size of the ES cluster and that can't happen in-place (like we have to create the larger instance and then reindex all of the contents from the smaller instance).

TomElliottFlexion (approved):

PR #1968, opened 2022-03-11
Devex 859: Part 4

rachelschneiderman:


web-client/src/presenter/actions/addPenaltyInputAction.js
```diff
@@ -9,7 +9,7 @@ import { state } from 'cerebral';
  * @returns {void}
  */
 export const addPenaltyInputAction = ({ get, store }) => {
-  const penalties = get(state.modal.penalties) || [];
+  const penalties = get(state.modal.penalties);
```
jamesobrooks: Is `state.modal.penalties` guaranteed to exist? From what I can tell, if penalties is null or undefined, we'll get an error on line 13. Or does cerebral guarantee that `get()` returns something that can have its length checked?

jamesobrooks: Or maybe line 13 could be:
```js
if (penalties && penalties.length < 10) {
```

rachelschneiderman: I believe it will exist, here's what I did to reach that conclusion.

`addPenaltyAction` is called in one sequence, `addPenaltyInputSequence`. `addPenaltyInputSequence` is called as part of `CalculatePenaltiesModal`. `CalculatePenaltiesModal` is opened by `showCalculatePenaltiesModalSequence` which calls `setDefaultPenaltiesAction` which defaults `state.modal.penalties` to a list of empty strings. As you can't open the model without clicking the button that triggers that sequence, I think it's guaranteed to always exist. 

I could be wrong though!

jamesobrooks: Personally I would prefer to not treat something as guaranteed because the current sequence of events that leads to executing the code makes it true. The other changes here (attachments and hearings) seem to be safe, but this one, if someone makes a change that would make us of this function without having called `setDefaultPenaltiesAction` and doesn't write a test case that would catch it, then we could error out just because we're checking length.

jamesobrooks (approved):

PR #1969, opened 2022-03-14
DevEx - Visual PDF Testing

codyseibert:
An example of what a failure in circle would look like if you break the PDFs.  This is from a simple css change to remove some padding:

Screen Shot 2022-03-14 at 12 26 16 PM

The visual testing involves creating all of the PDFs, converting the first page to an image, and comparing that image using a pixeldif library to get the % difference of pixels from the output image and expected image.  If that % difference is > 1%, we mark the PDF as failed.  This change also involves adding two libraries to docker so we can convert the pdf to an image.  In order to generate these expected images, one must run the steps outlined in the `docs/testing.md` file.


rosiehsmith (approved):

rosiehsmith:
This is awesome!

docs/index.html
```diff
@@ -4,6 +4,7 @@
     <meta charset="UTF-8" />
     <title>Document</title>
     <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1" />
+    <link rel="icon" href="_media/favicon.ico" />
```
rosiehsmith, line 7: is this to quiet that console error that was always there? if so, thank you!

codyseibert: This is actually to show a favicon in the docs.  I wasn't aware of the console error, but hopefully it fixes it lol

rosiehsmith: ahh i see. haha that would be cool. maybe i'll look into the other thing.

PR #1971, opened 2022-03-15
Update documentation to include USTC ECR mentions

jamesobrooks:
Explicitly mention both Flexion and USTC AWS accounts in reference to publishing Docker image.

### DIFF
--- a/docs/ci-cd.md
+++ b/docs/ci-cd.md
@@ -109,7 +109,7 @@ jobs:
 
 ## Docker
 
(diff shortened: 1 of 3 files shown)


jamesobrooks:
In retrospect it definitely makes sense that we would need to upload to a separate ECR given the separate CircleCI accounts with separate AWS accounts. But I did not upload prior to merging my dependency updates to test last week.

codyseibert (approved):

TomElliottFlexion (approved):

PR #1972, opened 2022-03-16
9093 fixing regressions

codyseibert:
- fixing an issue with the header font family being the wrong font
- fixing font style on the STPO and header on it
- downgrading puppeteer-core and puppeteer to match the same versions that are on the lambda layer we use for puppeteer.


TomElliottFlexion (approved):

image-compare-pdfs.js
```diff
@@ -64,13 +64,15 @@ const pdfs = [
       const totalPixels = height * width;
       const percentDifference = (pixelDifference / totalPixels) * 100;
 
-      fs.mkdirSync(`./shared/test-output/diff`, { recursive: true });
+      console.log(`${pdf}: ${percentDifference}%`);
```
rachelschneiderman, line 67: Just want to double check... Is this log left intentionally for debugging?

PR #1974, opened 2022-03-17
OpEx 1054: Delete API Gateway Mappings Only in Production

halprin:
[OpEx Card](https://trello.com/c/aLe5aGOh).

We are only deleting the API Gateway mappings for production.

Created a new job that only deletes the API Gateway mappings.  This new job is only called in the production-only workflow `build-and-deploy-with-context`.  The workflow after `switch-colors` runs both `backup-source-dynamo-table` and `delete-api-mappings` in parallel before coming back together to run `cleanup`.  This will allow for the DynamoDB back-up and the API Gateway mapping deletion to happen concurrently.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -756,7 +756,11 @@ jobs:
       - run:
           name: Delete Source API Gateway Mappings
           command: |
-            npm run delete:api-gateway-mappings -- $EFCMS_DOMAIN $DEPLOYING_COLOR
+            if [ "${CIRCLE_BRANCH}" == "prod" ]; then
+              npm run delete:api-gateway-mappings -- "${EFCMS_DOMAIN}" "${DEPLOYING_COLOR}"
+            else
+              echo "Skipping..."
+            fi
       - run:
           name: Destroy Migration Infrastructure
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -733,7 +733,7 @@ jobs:
               npm run backup:dynamo-table -- $TABLE_NAME
             fi
 
-  cleanup:
+  delete-api-mappings:
     docker:
       - image: $AWS_ACCOUNT_ID.dkr.ecr.us-east-1.amazonaws.com/ef-cms-us-east-1:1.1.0
         aws_auth:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -756,7 +756,11 @@ jobs:
       - run:
           name: Delete Source API Gateway Mappings
           command: |
-            npm run delete:api-gateway-mappings -- $EFCMS_DOMAIN $DEPLOYING_COLOR
+            if [ "${CIRCLE_BRANCH}" == "prod" ]; then
```
rachelschneiderman: Nit: The job `backup-source-dynamo-table` is only run in the prod deploy workflow so it may be worth it to rename that to be a bit less specific and move this into that job as well? That way we could make use of the built in branch filtering Circle provides

rachelschneiderman (approved):

rachelschneiderman (commented):
Sorry Peter, I did not realize this was marked as a draft! If I did I would not have commented. My bad!

halprin:
Oh no problem, @rachelschneiderman!  I only marked it as draft because I wanted to put it into an exp environment.

.circleci/config.yml
halprin: Are you sure?  🤔  I only see `backup-source-dynamo-table` having its steps defined, stating that it requires `switch-colors` to be run first, and that it is required before `backup` runs.  I don't see it having a CircleCI branch filter.

rachelschneiderman: So my understanding of how Circle wants filtering to be specified is on workflow definitions, not job definitions. So that's why `backup-source-dynamo-table` is defined without any filtering but is **only** included in the `build-and-deploy-with-context` workflow which uses the `branches: only: prod` filter

halprin: Ah, I see!

The filter isn't specified on the workflow, but on the job level still.  It just so happens that _every_ job in the `build-and-deploy-with-context` workflow has the `build-and-deploy-with-context-defaults` settings injected into it.  It is in `build-and-deploy-with-context-defaults` that has the `prod` branch filter.

Thanks for pointing this all out.  I'll look into it!

halprin: I ended up making the deletion of the API Gateway mappings its own job that only runs in production.  I updated the description of this PR with additional detail.

TomElliottFlexion (approved):

jimlerza (approved):

PR #1977, opened 2022-03-21
9326 NOCRP

codyseibert:
Converting the notice of change to remote proceeding behind a websocket to allow for longer execution times.

In addition, we added the ability to mimic the asynchronous nature of a real API Gateway while running code locally.  This led us to discover places where our tests depended on data coming back from the async endpoints when, in reality, nothing would be coming back.

### DIFF
--- a/shared/src/business/useCases/practitioners/updatePractitionerUserInteractor.js
+++ b/shared/src/business/useCases/practitioners/updatePractitionerUserInteractor.js
@@ -174,8 +174,4 @@ exports.updatePractitionerUserInteractor = async (
     },
     userId: requestUser.userId,
   });
-
-  return new Practitioner(updatedUser, { applicationContext })
-    .validate()
-    .toRawObject();
 };
--- a/shared/src/business/useCases/practitioners/updatePractitionerUserInteractor.js
+++ b/shared/src/business/useCases/practitioners/updatePractitionerUserInteractor.js
@@ -55,7 +55,6 @@ const getUpdatedFieldNames = ({ applicationContext, oldUser, updatedUser }) => {
  * @param {object} providers the providers object
  * @param {object} providers.barNumber the barNumber of the user to update
  * @param {object} providers.user the user data
- * @returns {Promise} the promise of the createUser call
  */
 exports.updatePractitionerUserInteractor = async (
   applicationContext,
--- a/shared/src/business/useCases/trialSessions/updateTrialSessionInteractor.js
+++ b/shared/src/business/useCases/trialSessions/updateTrialSessionInteractor.js
@@ -197,9 +197,9 @@ exports.updateTrialSessionInteractor = async (
     applicationContext,
     message: {
       action: 'update_trial_session_complete',
-      docketEntryId: serviceInfo?.docketEntryId,
       hasPaper: serviceInfo?.hasPaper,
       pdfUrl,
+      trialSessionId: trialSession.trialSessionId,
     },
     userId: user.userId,
   });
--- a/shared/src/business/useCases/trialSessions/updateTrialSessionInteractor.js
+++ b/shared/src/business/useCases/trialSessions/updateTrialSessionInteractor.js
@@ -203,9 +203,4 @@ exports.updateTrialSessionInteractor = async (
     },
     userId: user.userId,
   });
-
-  return {
-    newTrialSession: newTrialSessionEntity.toRawObject(),
-    serviceInfo: pdfUrl,
-  };
 };
--- a/shared/src/business/useCases/trialSessions/updateTrialSessionInteractor.js
+++ b/shared/src/business/useCases/trialSessions/updateTrialSessionInteractor.js
@@ -130,7 +130,6 @@ exports.updateTrialSessionInteractor = async (
     const calendaredCases = currentTrialSession.caseOrder;
     const { PDFDocument } = await applicationContext.getPdfLib();
     const paperServicePdfsCombined = await PDFDocument.create();
-    let processedCases = 0;
 
     for (let calendaredCase of calendaredCases) {
       const caseToUpdate = await applicationContext
@@ -155,19 +154,6 @@ exports.updateTrialSessionInteractor = async (
             user,
           });
 
-        processedCases++;
-
-        await applicationContext
-          .getNotificationGateway()
-          .sendNotificationToUser({
-            applicationContext,
-            message: {
-              action: 'notice_generation_update_progress',
-              processedCases,
-              totalCases: calendaredCases.length,
-            },
-            userId: user.userId,
-          });
         caseEntity.updateTrialSessionInformation(newTrialSessionEntity);
 
         await applicationContext.getUseCaseHelpers().updateCaseAndAssociations({
@@ -210,7 +196,7 @@ exports.updateTrialSessionInteractor = async (
   await applicationContext.getNotificationGateway().sendNotificationToUser({
     applicationContext,
     message: {
-      action: 'notice_generation_complete',
+      action: 'update_trial_session_complete',
       docketEntryId: serviceInfo?.docketEntryId,
       hasPaper: serviceInfo?.hasPaper,
       pdfUrl,
--- a/shared/src/business/useCases/trialSessions/updateTrialSessionInteractor.js
+++ b/shared/src/business/useCases/trialSessions/updateTrialSessionInteractor.js
@@ -16,7 +16,6 @@ const { UnauthorizedError } = require('../../../errors/errors');
  * @param {object} applicationContext the application context
  * @param {object} providers the providers object
  * @param {object} providers.trialSession the trial session data
- * @returns {object} the created trial session
  */
 exports.updateTrialSessionInteractor = async (
   applicationContext,
--- a/shared/src/business/useCases/trialSessions/updateTrialSessionInteractor.js
+++ b/shared/src/business/useCases/trialSessions/updateTrialSessionInteractor.js
@@ -198,6 +198,15 @@ exports.updateTrialSessionInteractor = async (
     pdfUrl = serviceInfo.url;
   }
 
+  if (trialSession.swingSession && trialSession.swingSessionId) {
+    newTrialSessionEntity.setAsSwingSession(trialSession.swingSessionId);
+  }
+
+  await applicationContext.getPersistenceGateway().updateTrialSession({
+    applicationContext,
+    trialSessionToUpdate: newTrialSessionEntity.validate().toRawObject(),
+  });
+
   await applicationContext.getNotificationGateway().sendNotificationToUser({
     applicationContext,
     message: {
@@ -207,15 +216,6 @@ exports.updateTrialSessionInteractor = async (
       pdfUrl,
     },
     userId: user.userId,
-  });
-
-  if (trialSession.swingSession && trialSession.swingSessionId) {
-    newTrialSessionEntity.setAsSwingSession(trialSession.swingSessionId);
-  }
-
-  await applicationContext.getPersistenceGateway().updateTrialSession({
-    applicationContext,
-    trialSessionToUpdate: newTrialSessionEntity.validate().toRawObject(),
   });
 
   return {
--- a/shared/src/business/useCases/trialSessions/updateTrialSessionInteractor.js
+++ b/shared/src/business/useCases/trialSessions/updateTrialSessionInteractor.js
@@ -208,6 +208,10 @@ exports.updateTrialSessionInteractor = async (
     });
   }
 
+  if (trialSession.swingSession && trialSession.swingSessionId) {
+    newTrialSessionEntity.setAsSwingSession(trialSession.swingSessionId);
+  }
+
   await applicationContext.getPersistenceGateway().updateTrialSession({
     applicationContext,
     trialSessionToUpdate: newTrialSessionEntity.validate().toRawObject(),
--- a/shared/src/business/useCases/trialSessions/updateTrialSessionInteractor.js
+++ b/shared/src/business/useCases/trialSessions/updateTrialSessionInteractor.js
@@ -125,6 +125,7 @@ exports.updateTrialSessionInteractor = async (
   }
 
   let pdfUrl = null;
+  let serviceInfo = null;
   if (currentTrialSession.caseOrder && currentTrialSession.caseOrder.length) {
     const calendaredCases = currentTrialSession.caseOrder;
     const { PDFDocument } = await applicationContext.getPdfLib();
@@ -188,25 +189,25 @@ exports.updateTrialSessionInteractor = async (
       }
     }
 
-    const serviceInfo = await applicationContext
+    serviceInfo = await applicationContext
       .getUseCaseHelpers()
       .savePaperServicePdf({
         applicationContext,
         document: paperServicePdfsCombined,
       });
     pdfUrl = serviceInfo.url;
-
-    await applicationContext.getNotificationGateway().sendNotificationToUser({
-      applicationContext,
-      message: {
-        action: 'notice_generation_complete',
-        docketEntryId: serviceInfo.docketEntryId,
-        hasPaper: serviceInfo.hasPaper,
-        pdfUrl,
-      },
-      userId: user.userId,
-    });
   }
+
+  await applicationContext.getNotificationGateway().sendNotificationToUser({
+    applicationContext,
+    message: {
+      action: 'notice_generation_complete',
+      docketEntryId: serviceInfo?.docketEntryId,
+      hasPaper: serviceInfo?.hasPaper,
+      pdfUrl,
+    },
+    userId: user.userId,
+  });
 
   if (trialSession.swingSession && trialSession.swingSessionId) {
     newTrialSessionEntity.setAsSwingSession(trialSession.swingSessionId);
--- a/web-api/src/app.js
+++ b/web-api/src/app.js
@@ -899,7 +899,7 @@ app.get(
   );
   app.put(
     '/async/practitioners/:barNumber',
-    lambdaWrapper(updatePractitionerUserLambda),
+    lambdaWrapper(updatePractitionerUserLambda, { isAsync: true }),
   );
   app.get(
     '/practitioners/:userId/printable-case-list',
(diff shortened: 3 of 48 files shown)


shared/src/business/useCases/trialSessions/updateTrialSessionInteractor.js
```diff
@@ -188,31 +175,35 @@ exports.updateTrialSessionInteractor = async (
...
-      message: {
-        action: 'notice_generation_complete',
-        docketEntryId: serviceInfo.docketEntryId,
-        hasPaper: serviceInfo.hasPaper,
-        pdfUrl,
-      },
-      userId: user.userId,
-    });
+  if (trialSession.swingSession && trialSession.swingSessionId) {
+    newTrialSessionEntity.setAsSwingSession(trialSession.swingSessionId);
```
rachelschneiderman, line 187: Out of curiosity, why was this added?

web-api/src/lambdaWrapper.js
```diff
@@ -26,21 +28,29 @@ exports.lambdaWrapper = lambda => {
       queryStringParameters: req.query,
     };
 
+    if (shouldMimicApiGateway) {
```
rachelschneiderman: Nit: It seems like this condition is checked a few times for different reasons, is there a way they could be combined/streamlined?

web-api/src/lambdaWrapper.js
```diff
@@ -26,21 +28,29 @@ exports.lambdaWrapper = lambda => {
...
+      res.status(204).send('');
```
rachelschneiderman, line 39: Are there unit tests that can be written around these newly added lines that help document why they were needed/added?

web-client/integration-tests/utils/manuallyAddCaseToTrial.js
```diff
@@ -0,0 +1,26 @@
+// import { wait } from '../helpers';
```
rachelschneiderman: Can these comments be removed?

web-client/integration-tests/utils/manuallyAddCaseToTrial.js
```diff
@@ -0,0 +1,26 @@
...
+// import { withAppContextDecorator } from '../../src/withAppContext';
+
+export const manuallyAddCaseToTrial = cerebralTest => {
```
rachelschneiderman, line 1: It looks like there are a couple other functions that also manually add a case to a trial session:

`docketClerkManuallyAddsCaseToTrialSessionWithoutNote`
`petitionsClerkManuallyAddsCaseToTrial`
`petitionsClerkManuallyAddsCaseToTrialWithoutJudge`

Could any of those been reused?

web-client/integration-tests/petitionsClerkSetsRemoteTrialSessionCalendar.test.js
```diff
@@ -69,23 +79,12 @@ describe('petitions clerk sets a remote trial session calendar', () => {
       loginAs(cerebralTest, 'petitionsclerk@example.com');
       cerebralTest.casesReadyForTrial = [];
       petitionsClerkCreatesNewCase(cerebralTest, fakeFile, trialLocation);
-      petitionsClerkManuallyAddsCaseToTrial(cerebralTest);
+      manuallyAddCaseToTrial(cerebralTest);
     });
   });
 
-  describe('petitions clerk sets calendar for trial session', () => {
-    petitionsClerkViewsNewTrialSession(cerebralTest);
```
rachelschneiderman, line 77: Why were these test steps removed?

web-client/src/presenter/actions/CaseDetail/setDefaultUnsealDocketEntryModalStateAction.test.js
```diff
@@ -0,0 +1,30 @@
+import { applicationContextForClient as applicationContext } from '../../../../../shared/src/business/test/createTestApplicationContext';
+import { presenter } from '../../presenter-mock';
+import { runAction } from 'cerebral/test';
+import { setDefaultUnsealDocketEntryModalStateAction } from './setDefaultUnsealDocketEntryModalStateAction';
+
+describe('setDefaultUnsealDocketEntryModalStateAction', () => {
```
rosiehsmith, line 6: Thanks for adding test coverage here!

web-api/src/lambdaWrapper.js
```diff
@@ -26,21 +28,29 @@ exports.lambdaWrapper = lambda => {
...
+        ...response.headers,
+        'X-Terminal-User': isTerminalUser,
+        ...exports.headerOverride,
+      });
+    }
 
-    if (
+    if (shouldMimicApiGateway) {
+      // Do nothing
+    } else if (
```
halprin: Looks like the `else if` part of this `if` statement can be combined with the `if` statement above.

web-api/src/app.js
```diff
@@ -1011,7 +1011,10 @@ app.get(
   );
   app.get('/trial-sessions', lambdaWrapper(getTrialSessionsLambda));
   app.post('/trial-sessions', lambdaWrapper(createTrialSessionLambda));
-  app.put('/trial-sessions', lambdaWrapper(updateTrialSessionLambda));
+  app.put(
+    '/async/trial-sessions',
+    lambdaWrapper(updateTrialSessionLambda, { isAsync: true }),
```
halprin, line 1016: Now that you added this `isAsync` option, shouldn't we update all of the other async endpoints in this file to have this option too?

web-client/src/presenter/actions/TrialSession/setPrintPaperDoneUrlAction.js
```diff
@@ -0,0 +1,17 @@
+import { state } from 'cerebral';
+
+/**
+ * sets the printPaperDoneUrl onto the state used for knowing where to redirect when pressing done on the PrintPaperService component.
+ *
+ * @param {object} providers the providers object
+ * @param {object} providers.store the cerebral store used for setting the state.eligibleCases
```
rachelschneiderman: Nit: docs need updating

web-api/src/lambdaWrapper.js
```diff
@@ -10,8 +10,10 @@ exports.headerOverride = {
   'X-Content-Type-Options': 'nosniff',
 };
 
-exports.lambdaWrapper = lambda => {
+exports.lambdaWrapper = (lambda, options = {}) => {
   return async (req, res) => {
+    const shouldMimicApiGateway =
```
halprin: `shouldMimicApiGateway` seems very generic.  In which way should it mimic the real AWS API Gateway?  In this case it wants to mimic the async-ness of it.  I would call it for what it is: something like `isAsyncNotInProduction`.

.... Oh wait.  this `NODE_ENV` environment variable.  That isn't the environment we deployed to, right?  I.e. not `exp3`, `test`, or `staging`.

halprin (commented):
Not done with the review, but here are some comments thus far.

shared/src/business/useCases/trialSessions/updateTrialSessionInteractor.js
```diff
@@ -188,31 +175,35 @@ exports.updateTrialSessionInteractor = async (
...
     applicationContext,
     trialSessionToUpdate: newTrialSessionEntity.validate().toRawObject(),
   });
 
+  await applicationContext.getNotificationGateway().sendNotificationToUser({
+    applicationContext,
+    message: {
+      action: 'update_trial_session_complete',
+      hasPaper: serviceInfo?.hasPaper,
+      pdfUrl,
```
rachelschneiderman, line 200: Seems like this logic was moved from `updateTrialSessionAction` to this interactor, if it's staying can test coverage be added for this new logic?

web-client/src/presenter/actions/navigateToMaintenanceAction.test.js
```diff
@@ -0,0 +1,25 @@
...
+
+  beforeAll(() => {
+    routeStub = jest.fn();
+
+    presenter.providers.router = {
+      route: routeStub,
+    };
+  });
+
+  it('navigates to FilePetitionSuccess', async () => {
```
rachelschneiderman: Seems like this test name may have been a copy/pasta 😄

web-client/src/presenter/actions/setWaitingForResponseAction.js
```diff
@@ -9,4 +9,5 @@ import { state } from 'cerebral';
 export const setWaitingForResponseAction = ({ store }) => {
   store.increment(state.progressIndicator.waitingForResponseRequests);
   store.set(state.progressIndicator.waitingForResponse, true);
+  store.unset(state.progressIndicator.waitText);
```
rachelschneiderman: Is there test coverage for this?

web-client/src/presenter/actions/setWaitingTextAction.js
```diff
@@ -0,0 +1,13 @@
+import { state } from 'cerebral';
+
+/**
+ * sets the wait text on the spinner
+ *
+ * @param {string} waitText the string to show on the spinner
+ * @returns {object} the set waiting for response action
+ */
+export const setWaitingTextAction =
```
rachelschneiderman, line 9: Missing test coverage

web-client/src/presenter/actions/setWaitingForResponseAction.js
rachelschneiderman: I'm curious why this line was added? It seems like this action is called immediately before we set the wait text

shared/src/business/useCases/trialSessions/updateTrialSessionInteractor.js
codyseibert: We used to have the UI invoke a separate endpoint to set the trial session as a swing state.  This logic seemed like it was more of a backend concern than a front end concern, so instead this single endpoint will check to set it as a swing state or not when updating the trial session.

web-api/src/lambdaWrapper.js
codyseibert: I'm not sure it will be easy to do that.  I'm updating the code to include a lot of comment to explain the purpose of this flag.

web-client/integration-tests/utils/manuallyAddCaseToTrial.js
codyseibert: We tried using some of those helper files but they have expect statements coded directly into the code, which makes them coded for a specific usecase which fails when we try to use them.  Instead of adding more arguments to the helper function to ignore the expect statements, we went with a new function that just adds the case to the trial and doesn't expect anything.  This allows us to keep the expect statement in the test file itself instead of helpers failing because we are using them for the wrong purpose.

web-client/src/presenter/actions/setWaitingForResponseAction.js
codyseibert: so the reason we added this here is because when you do a XHR request, the spinner will show up and display the waitText if it was set from the action previously in the sequence.  So if you navigate to a new page and do a different XHR request, the text will stick around on the next time we display a spinner.  We only wanted to add this spinner text to this specific user flow and not to every spinner.

web-api/src/app.js
codyseibert: yes, but it will probably break a LOT of tests since right now every test expects every endpoint to just wait until done processing.  I personally would rather make a devex item to do this work.

web-api/src/lambdaWrapper.js
codyseibert: I renamed it to `shouldMimicApiGatewayAyncEndpoint` which I think is more descriptive I hope?  yeah, NODE_ENV will only be 'production' when it's deployed to a real AWS environment

web-api/src/lambdaWrapper.js
codyseibert: I added a lot of comments to explain the purpose of this local only flag

web-client/integration-tests/petitionsClerkSetsRemoteTrialSessionCalendar.test.js
codyseibert: Because when creating a remote trial session, the status goes to `Open`.  So we can't use that helper method because it expects the trial session to be a status of `New`

web-api/src/lambdaWrapper.js
```diff
@@ -10,8 +10,15 @@ exports.headerOverride = {
   'X-Content-Type-Options': 'nosniff',
 };
 
-exports.lambdaWrapper = lambda => {
+exports.lambdaWrapper = (lambda, options = {}) => {
   return async (req, res) => {
+    // This flag was added because when invoking async endpoints on API gateway, the api gateway will return
+    // a 204 response with no body immediately.  This was causing discrepancies between how these endpoints
+    // ran locally and how they ran when deployed to an AWS environment.  We now turn this flag on
+    // whenever we are not running in product to mimic how API gateway async endpoints work.
```
halprin:
```suggestion
    // whenever we are not deployed in AWS to mimic how API gateway async endpoints work.
```

halprin: Thanks for adding the comments.  They do a good job of explaining the why!

web-client/integration-tests/utils/manuallyAddCaseToTrial.js
halprin: Since this doesn't have any `expects`, can you call this function in the aforementioned functions so we don't repeat this logic?

rachelschneiderman: Gotcha, makes sense. Thanks!

web-client/integration-tests/petitionsClerkSetsRemoteTrialSessionCalendar.test.js
rachelschneiderman: I am concerned we lost test coverage in removing these steps, or are the lines that were removed functionality that no longer exists in the system as of implementing this story?

web-client/src/presenter/actions/setWaitingForResponseAction.js
rachelschneiderman: So is that why it's not unset in the `unsetWaitingForResponseAction`?

web-client/integration-tests/petitionsClerkSetsRemoteTrialSessionCalendar.test.js
codyseibert: I'll make a helper called `petitionsClerkViewsOpenTrialSession` which should make this make more sense

halprin (commented):
These are a little late.  Hopefully they still make sense.

web-client/src/presenter/actions/TrialSession/setPrintPaperDoneUrlAction.test.js
```diff
@@ -0,0 +1,35 @@
...
+      },
+      props: {
+        trialSessionId: 'abc',
+      },
+      state: {
+        printPaperDoneUrl: null,
+      },
+    });
+
+    expect(result.state.printPaperDoneUrl).toEqual('/trial-session-detail/abc');
```
halprin: Nit: extract `abc` into a variable that is used in running the test and in the expect.

web-client/src/presenter/actions/setWaitingForResponseAction.js
halprin: Would this be a good time to use some inline comments to explain _why_ we unset this?  I'll defer to others who understand the implications more than I.

web-client/src/presenter/actions/setWaitingTextAction.test.js
```diff
@@ -0,0 +1,20 @@
+import { presenter } from '../presenter-mock';
+import { runAction } from 'cerebral/test';
+import { setWaitingTextAction } from './setWaitingTextAction';
+
+describe('setWaitingTextAction', () => {
+  it('should set the wait text when invoked with a string', async () => {
+    const EXPECTED_WAIT_TEXT = 'This is the wait text';
```
halprin, line 7: Nice!

halprin (commented):
Another batch.

web-client/src/presenter/actions/setWaitingForResponseAction.js
codyseibert: we refactored this to be placed in the unsetWaitingForResponseAction since we think it makes more sense there during cleanup.

rachelschneiderman: I think I'd prefer a named test if possible

rachelschneiderman (approved):

halprin:
@codyseibert, fixed the test.  Tomorrow I'd like to discuss the idea of removing `setPractitionerDetailAction` from the `success` path after calling `updatePractitionerUserAction` in `submitUpdatePractitionerUserSequence.js`.  Similarly, we should discuss removing passing the `barNumber` and `practitionerDetail` in the success path in `updatePractitionerUserAction.js` since it isn't used.  And really we might want to discuss removing the return statement at the bottom of `updatePractitionerUserInteractor.js` too.

shared/src/business/useCases/practitioners/updatePractitionerUserInteractor.js
```diff
@@ -174,8 +174,4 @@ exports.updatePractitionerUserInteractor = async (
     },
     userId: requestUser.userId,
   });
-
-  return new Practitioner(updatedUser, { applicationContext })
```
rachelschneiderman, line 178: I believe I understand why this is being removed but I am concerned as this is changing currently accepted/deployed functionality. What is the reason for this change?

halprin: Talked this over in Zoom.

halprin:
I believe we're done here.  The latest changes since about this time yesterday were all associated with adding the ability to mimic the asynchronous nature of the real API Gateway while running code locally.  This led us to discover places where our tests depended on data coming back from the async endpoints when, in reality, nothing would be coming back.  This also led us to add waits until after the progress indicator would disappear to some tests.

TomElliottFlexion (approved):

rosiehsmith (approved):

PR #1985, opened 2022-03-25
Devex 859: Part 6

rachelschneiderman:
- Don't need to default trial session to empty object for batch download feature
- Don't display "Download all Cases" button when the trial session doesn't have any cases added to it:
image


rosiehsmith (approved):

web-client/src/presenter/computeds/trialSessionHeaderHelper.js
```diff
@@ -22,26 +18,26 @@ export const trialSessionHeaderHelper = (get, applicationContext) => {
...
   return {
-    isStandaloneSession,
-    nameToDisplay,
+    isStandaloneSession: applicationContext
+      .getUtilities()
+      .isStandaloneRemoteSession(formattedTrialSession.sessionScope),
+    nameToDisplay: isTrialClerk
+      ? currentUser.name
+      : formattedTrialSession.formattedJudge,
+    showBatchDownloadButton: !isEmpty(formattedTrialSession.allCases),
```
rosiehsmith, line 40: that makes a lot of sense

PR #1995, opened 2022-03-28
DevEx 1025: Unify PDF Meta Header Generation

halprin:
There were two types of PDF meta header generation.  One through a React component (in `PageMetaHeaderDocket.jsx`), and one with hardcoded HTML/CSS (in `generatePdfFromHtmlInteractor.js`).  The PDF meta header is the `Docket No: 1234-22W        Page 2 of 3` text displayed at the top of pages two, and forward, in PDFs.  This PR cleans-up this code.  Specifically...

1. Removes the hardcoded HTML/CSS from `generatePdfFromHtmlInteractor.js`.  Instead it calls the `PageMetaHeaderDocket.jsx` React component by default.

2. Removes the `overwriteHeader` parameter in `generatePdfFromHtmlInteractor.js`.  Changed it so that if the `headerHtml` parameter is specified, then it is used.  Else the default is used.  There was a concept where you could use the `overwriteHeader` in conjunction with the `headerHtml` parameter that would overwrite the `DocketNo: 1234-22W` part of the meta header but keep the `Page 2 of 3` part, but this functionality wasn't used anywhere.  So... yeah, I removed `overwriteHeader` because it isn't needed.

3. Cleaned up the many calls to `generatePdfFromHtmlInteractor.js` in `documentGenerators/`.
    - Removed the usage of `overwriteHeader` because it was deleted (see 2 above).
    - Now that we are using the `PageMetaHeaderDocket.jsx` as a default, I don't need to specify `headerHtml` in all the document generators.  Only the ones where we need to overwrite the default.

4. Updated some unit tests, especially `generatePdfFromHtmlInteractor.test.js`.  Removed some test cases because they were testing the `overwriteHeader` functionality that was removed.  Added the test case of specifying an empty string for `headerHtml`.


shared/src/business/useCases/generatePdfFromHtmlInteractor.js
```diff
@@ -31,19 +33,18 @@ exports.generatePdfFromHtmlInteractor = async (
...
-    const headerContent = overwriteHeader
-      ? `${headerHtml || ''}`
-      : ` <div style="font-size: 8px; float: right;">
-              Page <span class="pageNumber"></span>
-              of <span class="totalPages"></span>
-            </div>
-            <div style="float: left">
-              ${headerHtml || `Docket Number: ${docketNumber}`}
-            </div>`;
+    if (headerHtml === undefined) {
```
mrinsin, line 36: nit question: is this the same as `!headerHtml`? And if so, is one way of writing it better than the other?

mrinsin (approved):

mrinsin:
Assuming all the PDFs affected by this change have been generated and compared against how they looked with the old code, this looks good! Thanks for doing this!

shared/src/business/useCases/generatePdfFromHtmlInteractor.js
halprin: Really good question!  It is not the same as `!headerHtml`.  If `headerHtml` is an empty string (`''`), it is a falsy value.  That then means that `!headerHtml` would be be truthy and your custom header of nothing would be overwritten with the default header.  That's also why I added a new unit test to for this exact scenario.

mrinsin: Oooh very interesting. Thank you for explaining!

halprin:
> Assuming all the PDFs affected by this change have been generated and compared against how they looked with the old code, this looks good! Thanks for doing this!

If I'm reading @codyseibert new [PDF compare run](https://app.circleci.com/pipelines/github/flexion/ef-cms/34071/workflows/067d9a07-db8b-4058-ba78-35f1f64f642f/jobs/213473/parallel-runs/0/steps/0-106), my changes have have less than 1% difference than the golden copies.  And those differences aren't due to my changes it seems.  My percent difference is in line with the [current percentage in `test`](https://app.circleci.com/pipelines/github/ustaxcourt/ef-cms/3236/workflows/f54d7576-5f4a-459b-b6db-23d1ce454dba/jobs/32529/parallel-runs/0/steps/0-106).

So, I'm thinking it's good to go.  I'll merge shortly.

PR #2002, opened 2022-03-29
9340: Auto-Generate Order to Show Cause

rosiehsmith:
image
image


shared/src/business/useCases/serveCaseToIrs/serveCaseToIrsInteractor.test.js
```diff
@@ -628,6 +628,51 @@ describe('serveCaseToIrsInteractor', () => {
     expect(applicationContext.getUtilities().uploadToS3).toHaveBeenCalled();
   });
 
+  it('should generate an order and upload it to s3 for orderToShowCause', async () => {
```
rachelschneiderman, line 631: Nit: This is quite a few expectation statements for one test, covering a lot of different AC. I would prefer to see them broken up into a few more tests that can have more specific names which will also help document the system's expected behavior.

rachelschneiderman (approved):

PR #2021, opened 2022-04-01
Dependency updates

ujahio:
Dependency updates from the week of Mar 14, 2021


rosiehsmith:
still not sure why this has so much 9001 on it... maybe it would be worth cherry picking the specific commits from your week of updates? i think this commit starting with 4600952 brought `test` into this branch which could cause some problems possibly?

shared/src/business/test/createTestApplicationContext.js
```diff
@@ -321,6 +321,7 @@ const createTestApplicationContext = ({ user } = {}) => {
     getContactPrimary: jest.fn().mockImplementation(getContactPrimary),
     getContactSecondary: jest.fn().mockImplementation(getContactSecondary),
     getCropBox: jest.fn().mockImplementation(getCropBox),
+    getDateInFuture: jest.fn().mockImplementation(DateHandler.getDateInFuture),
```
TomElliottFlexion, line 324: 🤔

TomElliottFlexion (approved):

halprin (changes requested):
I think @rosiehsmith is right.  `court/test` is in this branch.

ujahio:
I'm going to close this PR.

As of Friday, April 1, 2022, 5:03 pm ET, all of the packages are up to date. It may have been updated in the process of updating the packages from the recently merged previous dependency updates. I don't want to corrupt the current state of staging. So i'll close.

PR #2043, opened 2022-04-05
BUG 9465 (to Test)

TomElliottFlexion:
-Handles 502s correctly, adds headers
-Retrieves smaller OutboxItem for overloaded section-outbox partition
-Story to redesign section “Processed” tab to be added to backlog with help of UX

Requires `setup-section-outbox-retrieval-days.sh` to be run (added to `setup-all-env-configuration.sh`)
Requires redeploy of API Gateway.


web-api/src/applicationContext.js
```diff
@@ -1285,7 +1289,7 @@ const environment = {
   elasticsearchEndpoint:
     process.env.ELASTICSEARCH_ENDPOINT || 'http://localhost:9200',
   masterDynamoDbEndpoint:
-    process.env.MASTER_DYNAMODB_ENDPOINT || 'dynamodb.us-east-1.amazonaws.com',
+    process.env.MASTER_DYNAMODB_ENDPOINT || 'http://localhost:8000',
```
jimlerza, line 1292: Do we need to add this to the environment variables in AWS Secrets Manager? Or is it derived and set elsewhere?

jimlerza (approved):

web-api/src/applicationContext.js
TomElliottFlexion: I believe this was just incorrectly defaulted, and should have been the same as DYNAMODB_ENDPOINT @jamesobrooks

PR #2045, opened 2022-04-06
DevEx - Speed up case detail page

codyseibert:
The case detail internal page hits a lot of endpoints (deadlines, notes, trial sessions, messages, pending emails, judge).  These requests were being done in sequential order, so this PR refactors them to be executed in parallel to speed up the load time of the case detail page.

Another issue with the case detail page was it was fetching ALL of the trial sessions on page load.  All of the trial sessions are only used for when interacting with the add to trial modal (which we already fetch all the sessions on open).  This PR includes a change to only fetch the individual cases associated with the case instead of ALL trial sessions in our system.

Before this refactoring, notice how many requests this page loads (4s total):
before-load-detail-153-20

After the refactoring (1s):
Screen Shot 2022-04-06 at 4 20 53 PM

### DIFF
--- a/web-client/src/presenter/actions/TrialSession/getTrialSessionsOnCaseAction.js
+++ b/web-client/src/presenter/actions/TrialSession/getTrialSessionsOnCaseAction.js
@@ -1,0 +1,48 @@
+import { state } from 'cerebral';
+
+/**
+ * Fetches the trial sessions using the getTrialSessions use case
+ *
+ * @param {object} providers the providers object
+ * @param {object} providers.applicationContext needed for getting the getCase use case
+ * @returns {object} contains the trial sessions returned from the use case
+ */
+export const getTrialSessionsOnCaseAction = async ({
+  applicationContext,
+  get,
+  store,
+}) => {
+  const caseDetail = get(state.caseDetail);
+
+  let trialSessions = [];
+
+  if (caseDetail && caseDetail.trialSessionId) {
+    const trialSession = await applicationContext
+      .getUseCases()
+      .getTrialSessionDetailsInteractor(applicationContext, {
+        trialSessionId: caseDetail.trialSessionId,
+      });
+
+    trialSessions.push(trialSession);
+
+    store.set(
+      state.trialSessionJudge,
+      trialSession.judge || { name: 'Unassigned' },
+    );
+  }
+
+  trialSessions = [
+    ...trialSessions,
+    ...(await Promise.all(
+      caseDetail.hearings.map(hearing =>
+        applicationContext
+          .getUseCases()
+          .getTrialSessionDetailsInteractor(applicationContext, {
+            trialSessionId: hearing.trialSessionId,
+          }),
+      ),
+    )),
+  ];
+
+  return { trialSessions };
+};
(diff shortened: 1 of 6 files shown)


jamesobrooks (approved):

TomElliottFlexion (approved):

web-client/src/presenter/sequences/gotoCaseDetailSequence.js
```diff
@@ -103,7 +102,7 @@ export const gotoCaseDetailSequence = [
         USER_ROLES.reportersOffice,
         USER_ROLES.trialClerk,
       ],
-      [parallel([gotoCaseDetailInternal, fetchUserNotificationsSequence])],
```
halprin, line 106: We don't need to call the `fetchUserNotificationsSequence` anymore?

jimlerza (approved):

web-client/src/presenter/sequences/gotoCaseDetailSequence.js
codyseibert: I can add it back, but I was instead calling the two actions directly here instead of having a sequence call another sequence

web-client/src/presenter/actions/TrialSession/getTrialSessionsOnCaseAction.js
```diff
@@ -0,0 +1,48 @@
+import { state } from 'cerebral';
+
+/**
+ * Fetches the trial sessions using the getTrialSessions use case
+ *
+ * @param {object} providers the providers object
+ * @param {object} providers.applicationContext needed for getting the getCase use case
```
rachelschneiderman, line 7: Nit: docs

web-client/src/presenter/actions/TrialSession/getTrialSessionsOnCaseAction.test.js
```diff
@@ -0,0 +1,83 @@
...
+          ],
+          trialSessionId: '123',
+        },
+      },
+    });
+
+    expect(result.output.trialSessions.length).toEqual(2);
+  });
+
+  it('should set the judge name in state based on the trial session id', async () => {
```
rachelschneiderman, line 41: These tests are awesome! Is the branch that sets judge to 'Unassigned' covered in here?

halprin:
@codyseibert and @jimlerza, I just realized this was to `staging`.  My bad.  I should have reviewed this earlier when you made the PR #2044 to `test`.  Since you made a change here, we'll definitely want to make sure we make a backmerge from `staging` into `test`.

web-client/src/presenter/actions/TrialSession/getTrialSessionsOnCaseAction.test.js
codyseibert: good catch, I missed that branch, I'll make another PR to address this.

web-client/src/presenter/actions/TrialSession/getTrialSessionsOnCaseAction.js
codyseibert: I'll make another PR to address this.

PR #2049, opened 2022-04-07
9477 bugfix: Order for Filing Fee - Incorrect URL

rosiehsmith:
image

^ Note the underscores in the URL now


rachelschneiderman (approved):
Out of curiosity, would this change have caused the visual test to fail? Totally just curious because of the configured threshold 😄

TomElliottFlexion (approved):

rosiehsmith:
@rachelschneiderman Just tried it and it does fail! The percent difference is 0.276375 which is greater than the threshold we have for failure!

halprin (approved):
💣

PR #2052, opened 2022-04-07
DevEx 1010 - Adding Shell Check as Husky Pre Commit

codyseibert:
Running shellcheck against all of our .sh scripts and also running on a pre-commit.  This change will require developers to have shellcheck installed locally.  You can do this with `brew install shellcheck`.

I tested this on exp5, a deploy went smoothly: https://app.circleci.com/pipelines/github/flexion/ef-cms/34331/workflows/4992d39c-4830-4046-87e4-60294381e011


halprin:
@jimlerza, this will require a new build Docker image being built and published.

run-local.sh
```diff
@@ -40,7 +40,7 @@ S3RVER_PID=$!
 URL=http://localhost:9000/ ./wait-until.sh
 npm run seed:s3
 
-if [ ! -z "$RESUME" ]; then
+if [ -n "$RESUME" ]; then
```
halprin: Put the curly braces around the variable here?
```suggestion
if [ -n "${RESUME}" ]; then
```

scripts/elasticsearch/backup-elasticsearch.sh
```diff
@@ -26,9 +26,9 @@
 curl -XPUT 'http://localhost:9200/_snapshot/snapshot-repository' -H "Content-Type: application/json" -d'{
    "type": "s3",
    "settings": {
-      "bucket": "'${ENV}'-es-snapshots-flexion.us",
+      "bucket": "'"${ENV}"'-es-snapshots-flexion.us",
       "region": "us-east-1",
-      "role_arn": "arn:aws:iam::'${AWS_ACCOUNT_ID}':role/es-snapshots-flexion.us"
+      "role_arn": "arn:aws:iam::'"${AWS_ACCOUNT_ID}"':role/es-snapshots-flexion.us"
```
halprin, line 31: Nice that this works.  I need to remember this instead of `\"` everwhere.

web-client/build-dist.sh
```diff
@@ -4,8 +4,8 @@ ENV=$1
 DEPLOYING_COLOR=$2
 
 [ -z "${EFCMS_DOMAIN}" ] && echo "You must have EFCMS_DOMAIN set in your environment" && exit 1
-[ -z "${ENV}" ] && echo 'You must pass ENV as argument $1' && exit 1
-[ -z "${DEPLOYING_COLOR}" ] && echo 'You must pass DEPLOYING_COLOR as argument $2' && exit 1
+[ -z "${ENV}" ] && echo "You must pass ENV as argument $1" && exit 1
+[ -z "${DEPLOYING_COLOR}" ] && echo "You must pass DEPLOYING_COLOR as argument $2" && exit 1
```
halprin: I don't think this was supposed to be converted into a double quote.  Now the error message would be something like `You must pass ENV as argument `.  It no longer tells me that I need to pass the environment as the first argument.  Same for the `DEPLOYING_COLOR`/`$2`.

web-client/terraform/bin/deploy-init.sh
```diff
@@ -24,11 +24,4 @@ fi
 # exit on any failure
 set -eo pipefail
 
-DYNAMSOFT_URL="https://dynamsoft-lib.${EFCMS_DOMAIN}"
-
-if [[ -z "${IS_DYNAMSOFT_ENABLED}" ]]
-then
-  IS_DYNAMSOFT_ENABLED="1"
-fi
-
```
halprin, line 33: Oh strange.  Am I correct that this was removed because it wasn't doing anything?  `DYNAMSOFT_URL` and `IS_DYNAMSOFT_ENABLED` wasn't exported, so it didn't have any effect on the `terraform` command.  Is that right?

iam/terraform/account-specific/main/regional-log-subscription-filters/create-missing-log-groups.sh
```diff
@@ -7,8 +7,10 @@
 
 set -e
 
-ENVIRONMENTS=$@
-if [[ -z "${ENVIRONMENTS}" ]]; then
+ENVIRONMENT_STR=$*
+IFS=' ' read -r -a ENVIRONMENTS <<< "$*"
```
halprin: Specifically not supporting space delimited environments?  Hence the `$*`?

Also, can you do this?
```suggestion
ENVIRONMENT_STR=$*
IFS=' ' read -r -a ENVIRONMENTS <<< "${ENVIRONMENT_STR}"
```

halprin (commented):
Wow, these are some great changes and making everything much more uniform!

codyseibert:
> @jimlerza, this will require a new build Docker image being built and published.

I did publish a new version for the court's lower environments AWS account.  But yes, you'll need to do it on the prod account as tag 1.2.0

web-client/build-dist.sh
codyseibert: good catch, I don't think shellcheck likes the fact I have a $1 in a single quote.  Maybe I can escape the $ symbol.

web-client/terraform/bin/deploy-init.sh
codyseibert: That sounds correct.  I didn't see it used or referenced anywhere.

iam/terraform/account-specific/main/regional-log-subscription-filters/create-missing-log-groups.sh
codyseibert: can you re-review this file?  I was trying to fix a shellcheck error but I don't think I was doing the best approach.

web-client/build-dist.sh
halprin: Or maybe just reword the error by using `first argument` and `second argument` in there somewhere.

iam/terraform/environment-specific/bin/deploy-app.sh
```diff
@@ -5,6 +5,7 @@ ENVIRONMENT=$1
 # Getting the environment-specific deployment settings and injecting them into the shell environment
 export ENV=${ENVIRONMENT}
 pushd ../../../../
+# shellcheck disable=SC1091
```
halprin, line 8: What is this disabling a check for?

halprin (approved):

iam/terraform/environment-specific/bin/deploy-app.sh
codyseibert: shellcheck doesn't like sourcing environment variables from external files.  There is a flag you can use to run in shellcheck like `shellcheck -x` to suppress this error, but the staged plugin doesn't allow custom flag to be passed in.

jimlerza (approved):

PR #2054, opened 2022-04-08
Dependency Update: Un-resolute css-what

halprin:
This fixes the `TypeError: Cannot read property 'Descendant' of undefined` problem after running `npm install`.

The transitive dependency `css-what` was locked to the `5.*.*` version a year ago.  Some other transitive dependencies are now requiring `css-what` version `6.*.*`, but because we force resolve version `5.*.*`, these dependencies are unmet and therefore are `undefined`.

Does anyone remember why `css-what` was locked to version `5.*.*` in the first place?


TomElliottFlexion (approved):

codyseibert (approved):

halprin:
Looks like it may have been pinned due to https://github.com/facebook/create-react-app/issues/11081

PR #2055, opened 2022-04-08
Story 9464: Multiple Clinic Letters for Different Addresses

halprin:
Addressed the requirements in the story, but also...

- Tests
  - Make a new describe block for all the clinic letter tests.
  - Remove the second petitioner from the existing clinic letter tests because they will be covered in our new tests.
  - Added new tests for multiple petitioners based on the tests in the story.
- Actual code
  - Made a dedicated function that computes when a clinic letter is needed since it is complicated.
  - Moved the NOTR generation for the second petitioner out of a helper function and converted it to just check if the addresses are different.
  - Use the combine PDF utility more.


shared/src/business/useCases/serveCaseToIrs/serveCaseToIrsInteractor.js
```diff
@@ -278,46 +357,69 @@ const generatePaperNoticeForContactSecondary = async ({
...
-    secondaryPdfDoc,
-    secondaryPdfDoc.getPageIndices(),
-  );
-  pdfDoc.insertPage(1, coverPageDocumentPages[0]);
-
-  const pdfDataBuffer = await pdfDoc.save();
-  return Buffer.from(pdfDataBuffer);
 };
 
+// const generateNotrForSecondaryContactIfDifferent = async ({
```
codyseibert: was this supposed to be commented out?

codyseibert (approved):

TomElliottFlexion (approved):

shared/src/business/useCases/serveCaseToIrs/serveCaseToIrsInteractor.js
```diff
@@ -151,32 +153,51 @@ const generateNoticeOfReceipt = async ({ applicationContext, caseEntity }) => {
...
   }
 
-  const caseConfirmationPdfName =
-    caseEntity.getCaseConfirmationGeneratedPdfFileName();
-
-  const isProSe = caseEntity.privatePractitioners.length === 0;
-
-  if (preferredTrialCity && isProSe) {
+  let clinicLetter;
+  const isPrimaryContactProSe =
```
rachelschneiderman, line 184: There is also a method off of the Case entity (isUserIdRepresentedByPrivatePractitioner) that just checks if the user is represented, that may be helpful here?

PR #2057, opened 2022-04-08
9474: Prevent terminal user from seeing links to sealed records

jamesobrooks:


halprin (approved):
Looks good!  But I would wait on merging this into `test` since `test` will be reset soon-ish.

PR #2059, opened 2022-04-11
OpEx 1058: Ignore Changes to lambda_config

halprin:
# OpEx 1058

When we deploy, Terraform checks the current state of the Cognito pool and notices that it has some Lambda triggers, but the Terraform IaC states zero lambda triggers.  This leads Terraform to delete the Lambda triggers.  This leads to a) new users being unable to successfully sign-up during the time between a deploy and the color switch, or b) new users being unable to successfully sign-up ever if the prod deploy failed.

We now tell Terraform to ignore the `lambda_config` configuration which leads it to _not_ delete the Lambda triggers on deploy.  This results in the Lambda triggers remaining all the way until the color switch (which is when they switch out to the new color).

_FYI: a lot of Flexion exp environments are missing the Cognito Lambda triggers ostensibly because they've failed their deployment after the deploy job but before the color-switch job.  After this is merged and your deploy fails due to this new verify check, manually add the Lambda triggers and the next deployment will work correctly._

## Before

Notice the `-` (minus, i.e. delete) on the `lambda_config` block.

```
  # module.ef-cms_apis.aws_cognito_user_pool.pool will be updated in-place
  ~ resource "aws_cognito_user_pool" "pool" {
        id                         = "us-east-1_FJ6j52QGj"
        name                       = "efcms-test"
        tags                       = {}
        # (14 unchanged attributes hidden)



      ~ email_configuration {
          ~ from_email_address     = "\"U.S. Tax Court\" <noreply@test.ef-cms.ustaxcourt.gov>" -> "U.S. Tax Court <noreply@test.ef-cms.ustaxcourt.gov>"
            # (3 unchanged attributes hidden)
        }

      - lambda_config {
          - post_authentication = "arn:aws:lambda:us-east-1:350455059537:function:cognito_post_authentication_lambda_test_green" -> null
          - post_confirmation   = "arn:aws:lambda:us-east-1:350455059537:function:cognito_post_confirmation_lambda_test_green" -> null
        }



        # (8 unchanged blocks hidden)
    }
```

## After

Notice the `lambda_config` block is missing which means it will not be deleted.

```
  # module.ef-cms_apis.aws_cognito_user_pool.pool will be updated in-place
  ~ resource "aws_cognito_user_pool" "pool" {
        id                         = "us-east-1_FJ6j52QGj"
        name                       = "efcms-test"
        tags                       = {}
        # (14 unchanged attributes hidden)



      ~ email_configuration {
          ~ from_email_address     = "\"U.S. Tax Court\" <noreply@test.ef-cms.ustaxcourt.gov>" -> "U.S. Tax Court <noreply@test.ef-cms.ustaxcourt.gov>"
            # (3 unchanged attributes hidden)
        }




        # (9 unchanged blocks hidden)
    }
```

### DIFF
--- a/scripts/verify-cognito-lambda-triggers.sh
+++ b/scripts/verify-cognito-lambda-triggers.sh
@@ -13,7 +13,7 @@ ENV="${1}"
 
 cognito_user_pool_name="efcms-${ENV}"
 
-user_pool_has_both_lambda_triggers=$(aws cognito-idp list-user-pools --max-results=60 | jq '.UserPools[] | select(.Name=="'"${cognito_user_pool_name}"'") | .LambdaConfig | .PostConfirmation and .PostAuthentication')
+user_pool_has_both_lambda_triggers=$(aws cognito-idp list-user-pools --max-results=60 --region=us-east-1 | jq '.UserPools[] | select(.Name=="'"${cognito_user_pool_name}"'") | .LambdaConfig | .PostConfirmation and .PostAuthentication')
 
 if [ "${user_pool_has_both_lambda_triggers}" != 'true' ]; then
   echo "ERROR: Cognito user pool ${cognito_user_pool_name} is missing either the PostConfirmation or PostAuthentication Lambda trigger"
--- a/scripts/verify-cognito-lambda-triggers.sh
+++ b/scripts/verify-cognito-lambda-triggers.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+# Usage
+#   smoketest to verify the Cognito pool retains its Lambda triggers
+
+# Arguments
+#   - $1 - the environment [dev, stg, prod, exp1, exp1, etc]
+
+( ! command -v jq > /dev/null ) && echo "jq must be installed on your machine." && exit 1
+[ -z "${1}" ] && echo "ERROR: The env to run smoketest is missing from the \$1 argument.  An example value of this includes [dev, stg, prod... ]" && exit 1
+
+ENV="${1}"
+
+cognito_user_pool_name="efcms-${ENV}"
+
+user_pool_has_both_lambda_triggers=$(aws cognito-idp list-user-pools --max-results=60 | jq '.UserPools[] | select(.Name=="'"${cognito_user_pool_name}"'") | .LambdaConfig | .PostConfirmation and .PostAuthentication')
+
+if [ "${user_pool_has_both_lambda_triggers}" != 'true' ]; then
+  echo "ERROR: Cognito user pool ${cognito_user_pool_name} is missing either the PostConfirmation or PostAuthentication Lambda trigger"
+  exit 1
+fi
+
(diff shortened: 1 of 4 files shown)


jamesobrooks (approved):

TomElliottFlexion (approved):

codyseibert (approved):

rachelschneiderman (commented):
Is there a test that can be added that would have caught this issue before it was deployed to prod?

mrinsin (approved):

halprin:
> Is there a test that can be added that would have caught this issue before it was deployed to prod?

Good idea!  Yeah, there probably could be.

rachelschneiderman (approved):

TomElliottFlexion:
Let's make sure to backmerge this to `test` individually and explicitly

scripts/verify-cognito-lambda-triggers.sh
```diff
@@ -0,0 +1,23 @@
...
+
+ENV="${1}"
+
+cognito_user_pool_name="efcms-${ENV}"
+
+user_pool_has_both_lambda_triggers=$(aws cognito-idp list-user-pools --max-results=60 | jq '.UserPools[] | select(.Name=="'"${cognito_user_pool_name}"'") | .LambdaConfig | .PostConfirmation and .PostAuthentication')
+
+if [ "${user_pool_has_both_lambda_triggers}" != 'true' ]; then
+  echo "ERROR: Cognito user pool ${cognito_user_pool_name} is missing either the PostConfirmation or PostAuthentication Lambda trigger"
+  exit 1
```
jamesobrooks, line 20: Is exiting the right move here? If for some reason we get to this point, wouldn't we want to create triggers for the current color which would then be replaced during `switch-colors`?

halprin: Hmmm.  I think I rather fail than self-heal.  I want the correct behavior to be in Terraform, but we may never know if we don't have the correct behavior if this script self-heals a bug.

jamesobrooks: I just wonder what is different then I guess. If we somehow don't have the triggers, then we have put ourselves in the exact situation that's being corrected right? An environment that was up, but we've broken Cognito? And it's checking for something that wasn't created by terraform in this case, but by a js script that was run in a previous run of the workflow.

halprin: Correct, we'll be back in the same situation.  But this time we'll have an explicit error to help us understand what is wrong.  This is instead of us not knowing something is amiss.

rachelschneiderman (commented):
Do we know why this is a problem now? It looks like the work done to move the triggers out of tf to the js script was done many months ago

halprin:
> Do we know why this is a problem now? It looks like the work done to move the triggers out of tf to the js script was done many months ago

Was this the first failed production deploy that we never subsequently got working before the deploy window closed?  This also affected any users who signed up during a deploy which actually happened a few weeks ago when this issue was first brought up by Tenille.

PR #2060, opened 2022-04-11
9474: Prevent terminal user from seeing links to sealed records

jamesobrooks:
Adds a check as to whether the document is sealed as part of determining whether to show document links to the terminal user.


halprin (approved):

rosiehsmith (approved):

web-client/src/presenter/computeds/Public/publicCaseDetailHelper.test.js
```diff
@@ -366,6 +366,28 @@ describe('publicCaseDetailHelper', () => {
           isFileAttached: false,
           isLegacyServed: true,
         },
+        {
```
rachelschneiderman: This is a pretty large test already. Would it be possible to pull these expectations out into a test that is named more specifically for this new functionality?

jamesobrooks: @rachelschneiderman, sounds good. Just pushed that change.

rachelschneiderman: Lovely! Thanks so much!

rachelschneiderman (approved):

halprin (approved):

PR #2095, opened 2022-04-22
Devex-1031: Fix PDF Seed Data part 2

rosiehsmith:
https://trello.com/c/vPfKwonv/1031-fix-pdf-seed-data

Part 2 of adding actually existent keys from `noop-documents-local-us-east-1` to `efcms-local` documents


rosiehsmith:
Closing because don't need to test vs prod-like data

PR #2099, opened 2022-04-23
Dependency Updates

codyseibert:
I pushed a new docker version 1.4.0 to both flexion and the tax court.

I'm doing a final test on exp4 before this is good to merge

- this change required a 2 second wait after logging in for some reason to get the smoketests to stop randomly failing.


cypress-smoketests/support/pages/local-login.js
```diff
@@ -13,6 +13,8 @@ exports.getUserToken = async username => {
 exports.login = email => {
   cy.visit(`/log-in?code=${email}`);
   cy.get('.progress-indicator').should('not.exist');
+  // eslint-disable-next-line cypress/no-unnecessary-waiting
+  cy.wait(2000);
```
halprin: Is there something that we should `get` that only exists once we fully loaded?  I.e. is there more stuff that loads after the `.progress-indicator` disappears that we should also wait for?

package.json
```diff
@@ -89,32 +89,32 @@
...
     "npm-force-resolutions": "0.0.10",
     "npm-run-all": "^4.1.5",
     "object-sizeof": "^1.6.3",
-    "openapi-types": "^10.0.0",
+    "openapi-types": "^11.0.0",
     "pa11y-ci": "^3.0.1",
     "parse-full-name": "^1.2.6",
     "pdf-lib": "^1.17.1",
-    "pdfjs-dist": "~2.12.313",
+    "pdfjs-dist": "2.12.313",
```
halprin, line 105: We want to pin the version to `2.12.313` exactly?

web-client/src/app.jsx
```diff
@@ -314,7 +314,10 @@ const app = {
     initializeSocketProvider(cerebralApp, applicationContext);
     router.initialize(cerebralApp, route);
 
-    ReactDOM.render(
+    const container = window.document.querySelector('#app');
+    const root = createRoot(container);
+
+    root.render(
```
halprin, line 320: What's this do?

halprin (commented):
I see a few other things other than dependency updates: removing an integration test, changing the root rendering of the private and public frontend, and change some of our CSS.  Was that on purpose?

web-client/src/app.jsx
codyseibert: this is a change related to a react 18 upgrade.  If you don't change the way we render, internally react will fallback on react 17 logic cc: @rosiehsmith knows more about this

package.json
codyseibert: yeah, this was documented in the `docs/team-process.md` section of the docs.  if we upgrade pdfjs-dist, the smoketests start failing because the library seems to have a bug in it.  I'm just locking it down for now since the next person to do a patch upgrade will run into the same issue

cypress-smoketests/support/pages/local-login.js
codyseibert: I could try to maybe wait for something related to the work queue items to show up.  It's a bit more challenging because depending on who you log in with you'll need to wait for different page components.  I can try to find a work around for this wait.

codyseibert:
> I see a few other things other than dependency updates: removing an integration test, changing the root rendering of the private and public frontend, and change some of our CSS. Was that on purpose?

The integration test was removed because it was constantly failing and I couldn't figure out why.  We've never had a test verify CSS styling before, so this test seems out of scope for our normal testing approaches.

web-client/src/app.jsx
rosiehsmith: yup! we were getting this error in the console locally 
Screen Shot 2022-04-20 at 8 15 22 PM
[reference](https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#updates-to-client-rendering-apis)

halprin: Oh!  Thanks for pointing that out to me.

jimlerza (approved):

PR #2109, opened 2022-04-27
9342: Auto-generate Order for Amended Petition and Filing Fee

jamesobrooks:
Implements [story 9342](https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/flexion/ef-cms/9342).

Also refactored the serveCaseToIrsInteractor.js to extract the shared code between all the auto-generated documents into generateDraftDocument.js.
And also refactored serveCaseWithSystemGeneratedDocuments.test.js to simplify set-up and validation of the test through a single object.


shared/src/business/useCases/serveCaseToIrs/generateDraftDocument.js
```diff
@@ -0,0 +1,35 @@
+const {
+  replaceBracketed,
+} = require('../../../business/utilities/replaceBracketed');
+
+/**
+ * @param {object} providers the providers object containing applicationContext, caseEntity, document, replacements
+ * @param providers.applicationContext api applicationContext
```
TomElliottFlexion: I should've caught this on the first PR, but these params all need types

shared/src/business/useCases/serveCaseToIrs/generateDraftDocument.js
```diff
@@ -0,0 +1,35 @@
+const {
+  replaceBracketed,
+} = require('../../../business/utilities/replaceBracketed');
+
+/**
+ * @param {object} providers the providers object containing applicationContext, caseEntity, document, replacements
```
rachelschneiderman, line 6: Are there tests for this new file?

shared/src/business/useCases/serveCaseToIrs/serveCaseToIrsInteractor.js
```diff
@@ -516,69 +512,46 @@ const serveCaseToIrsInteractor = async (
...
-      .getUseCaseHelpers()
-      .addDocketEntryForSystemGeneratedOrder({
-        applicationContext,
-        caseEntity,
-        systemGeneratedDocument: {
-          ...orderForAmendedPetition,
-          content,
-        },
-      });
+    await generateDraftDocument({
```
rachelschneiderman, line 526: Should this method be put onto applicationContext.getUseCaseHelpers? That way this interactor tests can verify it was called and the actual implementation can be unit tested.

shared/src/business/useCases/serveCaseToIrs/serveCaseToIrsInteractor.js
```diff
@@ -516,69 +512,46 @@ const serveCaseToIrsInteractor = async (
...
-      .getUseCaseHelpers()
-      .addDocketEntryForSystemGeneratedOrder({
-        applicationContext,
-        caseEntity,
-        systemGeneratedDocument: {
-          ...orderForFilingFee,
-          content,
-        },
-      });
+    await generateDraftDocument({
```
rachelschneiderman, line 515: Really simple and accurate name for this method 👏

shared/src/business/useCases/serveCaseToIrs/serveCaseToIrsInteractor.test.js
```diff
@@ -977,6 +981,45 @@ describe('serveCaseToIrsInteractor', () => {
...
+    expect(
+      await applicationContext.getUseCaseHelpers()
+        .addDocketEntryForSystemGeneratedOrder.mock.calls[0][0]
+        .systemGeneratedDocument,
+    ).toMatchObject({
+      content: `&nbsp;&nbsp;&nbsp;&nbsp;The Court filed on ${receivedDate}, a document as the petition of the above-named petitioner(s) at the docket number indicated. That docket number MUST appear on all documents and papers subsequently sent to the Court for filing or otherwise. The document did not comply with the Rules of the Court as to the form and content of a proper petition. <br/><br/>&nbsp;&nbsp;&nbsp;&nbsp;Accordingly, it is <br/><br/>&nbsp;&nbsp;&nbsp;&nbsp;ORDERED that on or before ${mockTodayPlus60}, petitioner(s) shall file a proper amended petition. If, by ${mockTodayPlus60}, petitioner(s) do not file an Amended Petition, the case will be dismissed or other action taken as the Court deems appropriate.`,
+    });
+  });
+
+  it('should replace brackets in orderForAmendedPetitionAndFilingFee content with the served date of the petition, and today plus sixty twice', async () => {
```
rachelschneiderman: Seems like some of these tests need to be updated to reflect the change from servedDate to filingDate?

halprin: Hmmmm, yes.  Good catch.  Time for some spelunking to see why these didn't fail.

shared/src/business/useCases/serveCaseToIrs/serveCaseToIrsInteractor.js
jamesobrooks: @rachelschneiderman, my thinking was that since we don't need it anywhere else right now, it's a concern only of this use case. So the unit tests of the interactor are testing it. And they test it well because I can make the tests fail as expected by changing the `generateDraftDocument` code.

shared/src/business/useCases/serveCaseToIrs/serveCaseToIrsInteractor.test.js
jamesobrooks: disregard

shared/src/business/useCases/serveCaseToIrs/serveCaseToIrsInteractor.js
rachelschneiderman: Ok, I'm alright with not adding it to app context. I think there should still be unit tests around that new use case though.

shared/src/business/useCases/serveCaseToIrs/serveCaseToIrsInteractor.test.js
halprin: Done!

shared/src/business/useCases/serveCaseToIrs/generateDraftDocument.js
halprin: Done!

jimlerza (approved):

TomElliottFlexion (approved):

shared/src/business/useCases/serveCaseToIrs/generateDraftDocument.test.js
```diff
@@ -0,0 +1,81 @@
...
+} = require('../../utilities/DateHandler');
+const {
+  INITIAL_DOCUMENT_TYPES,
+  SYSTEM_GENERATED_DOCUMENT_TYPES,
+} = require('../../entities/EntityConstants');
+const { Case } = require('../../entities/cases/Case');
+const { generateDraftDocument } = require('./generateDraftDocument');
+const { MOCK_CASE } = require('../../../test/mockCase');
+
+describe('generateDraftDocument', () => {
```
jamesobrooks: @rachelschneiderman, here's the new test file.

rachelschneiderman: Oh ok! I thought we agreed yesterday to try the sociable testing strategy instead? What changed your mind?

jamesobrooks: I'm honestly confused. I thought we agreed to talk about trying that out, but if you're okay with it here, I can just revert that commit. Honestly the test basically does nothing. There's a single path through the code, so we're validating that the `replaceBracketed` function works because it would make absolutely no sense to mock that, because the only other thing is to call a function that I've now mocked with the result.

jamesobrooks: To be clear, the existing tests are very sociable. We're just covering a lot of code several times—from different test files—it would seem.

jimlerza (approved):

mmarcotte (approved):

PR #2116, opened 2022-04-28
Add Decision Document for Blue/Green Migrations

mmarcotte:
In order to wrap up https://github.com/ustaxcourt/ef-cms/issues/1223 this PR simply adds documentation for the big decision to make use of Blue/Green migrations during deployments where data structures change. Feedback is welcome!


codyseibert (approved):
it seems pretty accurate to me I'd say

jimlerza (approved):

PR #2118, opened 2022-04-28
Bug 9486 consolidated case list

ujahio:
Solution to BUG [9486](https://github.com/flexion/ef-cms/issues/9486)


web-client/integration-tests/docketClerkVerifiesConsolidatedCases.test.js
```diff
@@ -0,0 +1,108 @@
...
+const trialLocation = `Boise, Idaho, ${Date.now()}`;
+
+const overrides = {
+  caseCaption: 'Mona Schultz, Petitioner',
+  docketNumberSuffix: 'L',
+  preferredTrialCity: trialLocation,
+  trialLocation,
+};
+
+describe('Case Consolidation Journey', () => {
```
TomElliottFlexion: Beautiful

TomElliottFlexion (approved):

rosiehsmith (approved):

web-client/integration-tests/docketClerkVerifiesConsolidatedCases.test.js
```diff
@@ -0,0 +1,108 @@
...
+    cerebralTest.docketNumber = cerebralTest.leadDocketNumber =
+      caseDetail.docketNumber;
+  });
+
+  loginAs(cerebralTest, 'docketclerk@example.com');
+  docketClerkUpdatesCaseStatusToReadyForTrial(cerebralTest);
+  docketClerkOpensCaseConsolidateModal(cerebralTest);
+  docketClerkSearchesForCaseToConsolidateWith(cerebralTest);
+
+  it('login as a petitioner and create the case to consolidate with', async () => {
```
mmarcotte, line 52: Are we still logged in as the docket clerk here?

ujahio: @mmarcotte resolved!

web-client/integration-tests/journey/docketClerkVerifiesConsolidatesCases.js
```diff
@@ -0,0 +1,10 @@
+export const docketClerkVerifiesConsolidatesCases = cerebralTest => {
+  return it('Docket clerk verifies consolidates cases', () => {
+    expect(cerebralTest.getState('caseDetail')).toHaveProperty(
```
rachelschneiderman, line 6: Would it be worth it to add a step in this test to route to case detail to ensure we are on the case we expect? Or adding an expectation for the expected docket number? Just a thought

web-client/integration-tests/docketClerkVerifiesConsolidatedCases.test.js
```diff
@@ -0,0 +1,108 @@
...
+const trialLocation = `Boise, Idaho, ${Date.now()}`;
+
+const overrides = {
+  caseCaption: 'Mona Schultz, Petitioner',
+  docketNumberSuffix: 'L',
+  preferredTrialCity: trialLocation,
+  trialLocation,
+};
+
+describe('Docket Clerk verifies Consolidated Cases', () => {
```
rachelschneiderman, line 29: This is an extremely thorough test file 👏 👏 👏 Nice job!

rachelschneiderman (approved):

web-client/integration-tests/docketClerkVerifiesConsolidatedCases.test.js
shumway-tm: Fixed to have the correct actor explicitly logging in at the appropriate steps!

web-client/integration-tests/journey/docketClerkVerifiesConsolidatesCases.js
shumway-tm: Thank you for the suggestion! I added a change here to check for that. However one difference in the implementation is that instead of checking that we are on the expected docketNumber I instead check that the case is part of the same consolidated group by checking the lead docket number associated with that case and comparing it to the lead docket number created at the beginning of the test.

shumway-tm: I think that is appropriate because the docket number can be any case in the group, as long as it is in the same consolidated case group. If I am wrong or misunderstood please let me know!

rachelschneiderman (approved):

PR #2133, opened 2022-05-02
Bug 9486 intermediate to test

shumway-tm:
Addressed feedback for editing a case caption and subsequently verifying consolidated cases are still in state.


codyseibert (approved):

web-client/integration-tests/docketClerkVerifiesConsolidatedCases.test.js
```diff
@@ -105,4 +106,13 @@ describe('Docket Clerk verifies Consolidated Cases', () => {
...
   docketClerkVerifiesConsolidatesCases(cerebralTest);
+
+  const updatedCaseCaption = 'I am the new case caption!';
+  updateACaseCaption(cerebralTest, updatedCaseCaption);
+  it('docketClerk updates Case Caption', async () => {
+    const caseDetail = cerebralTest.getState('caseDetail');
+    expect(caseDetail.caseCaption).toBeDefined();
+    expect(caseDetail.caseCaption).toEqual(updatedCaseCaption);
+  });
+  docketClerkVerifiesConsolidatesCases(cerebralTest);
```
jamesobrooks: :pushpin: I think the name of this function should be `docketClerkVerifiesConsolidatedCases`, but I know that file wasn't created in this PR. Maybe something that can be corrected at some point. Or maybe I'm missing the point of the file.

jamesobrooks (approved):

web-client/integration-tests/journey/docketClerkConsolidatesCaseThatCannotBeConsolidated.js
```diff
@@ -11,7 +11,7 @@ export const docketClerkConsolidatesCaseThatCannotBeConsolidated =
         'Place of trial is not the same',
       ]);
       expect(
-        cerebralTest.getState('caseDetail.consolidatedCases'),
-      ).toBeUndefined();
+        cerebralTest.getState('caseDetail.consolidatedCases').length,
+      ).toEqual(0);
```
jamesobrooks, line 15: ❓ I am confused by this change. Was this test failing before? I don't see any changes that should have changed the behavior such that we were getting an undefined thing before, and now we're getting an empty array.

shumway-tm: So the nature of the bug fix was that consolidated cases were disappearing from state after using any of the case functions in the Case Information Tab. Our fix was to, by default, set them again in state in all of the associated sequences. Because of this fix the property consolidatedCases on caseDetail can no longer be undefined. It is an empty array, indicating that there are no associated consolidated cases.

web-client/integration-tests/docketClerkVerifiesConsolidatedCases.test.js
shumway-tm: Good catch! We have updated the name of the file. Small typo.

PR #2137, opened 2022-05-02
9492: Bugfix related to message attachments

rosiehsmith:


web-client/src/presenter/computeds/messageModalHelper.js
```diff
@@ -6,7 +6,7 @@ export const messageModalHelper = (get, applicationContext) => {
 
   const caseDetail = get(state.caseDetail);
   const screenMetadata = get(state.screenMetadata);
-  const { attachments } = get(state.modal.form);
+  const attachments = get(state.modal.form.attachments) || [];
```
jimlerza: It was my understanding that the reasoning behind a wholesale, app-wide find-and-squash of such fallback instantiation is that we shouldn't be returning undefined (and therefore fallback is not necessary). This makes sense when the method being called is custom (and can be modified to never return undefined), but this was a `get()`. 

I'm concerned that we might have we removed other fallback instantiations from other methods that may return undefined, or even from other `get()`s. How can we find these before they're found in the field?

codyseibert (approved):

web-client/src/presenter/computeds/messageModalHelper.js
rosiehsmith: Good question, we have a path that may be a different solution than to default this to [], which is possibly just hiding the actual problem in how the `messageModalHelper` does not have access to the all the fields that it requires due to how the sequence is calling actions. Looking into that now instead.

jimlerza (approved):

PR #2155, opened 2022-05-05
DevEx - refactor interactor calling another interactor

codyseibert:
This was a refactoring found when trying to work on 9483, but we are keeping this a separate PR to reduce the blast radius. 
 We original tried to follow an approach to never call interactors from other interactors (clean architecture decision).  I noticed we had an interactor in the front end which was calling 2 other endpoints which goes against this approach.

- [x] tested on `test` environment by: @ttlenard 
- [x] merged to `test` on 5/6/2022


jimlerza (approved):

shared/src/business/useCases/caseNote/getUserCaseNoteForCasesInteractor.js
```diff
@@ -23,16 +24,32 @@ exports.getUserCaseNoteForCasesInteractor = async (
...
+  if (userEntity.isChambersUser()) {
+    const judgeUser = await applicationContext
+      .getUseCaseHelpers()
+      .getJudgeInSectionHelper(applicationContext, {
+        section: userEntity.section,
+      });
+    if (judgeUser) {
+      ({ userId } = judgeUser);
+    }
+  }
```
mmarcotte: this code seems identical to what is in 

* `shared/src/business/useCases/caseNote/deleteUserCaseNoteInteractor.js`,
* `shared/src/business/useCases/caseNote/getUserCaseNoteInteractor.js`. 
* `shared/src/business/useCases/caseNote/updateUserCaseNoteInteractor.js`
 
Does it make sense to put that into a `useCaseHelper`?

codyseibert: yeah good catch, I think I can abstract that away

shared/src/business/useCaseHelper/getUserIdForNote.js
```diff
@@ -0,0 +1,39 @@
+const { User } = require('../entities/User');
+
+/**
+ * returns the userId of who this note should be stored on.  This is mainly
+ * used for chambers users since they all share the same user note with the judge
+ * of their section.
+ *
+ * @param {object} applicationContext the application context
+ * @param {object} obj the options argument
+ * @param {string} obj.section the section to fetch the judge from
```
mmarcotte:
```
 * @param {string} obj. userIdMakingRequest the unique identifier of a user
```

PR #2156, opened 2022-05-05
DevEx - Refactor interactor calling interactor

codyseibert:
We original tried to follow an approach to never call interactors from other interactors (clean architecture decision). I noticed we had an interactor in the front end which was calling 2 other endpoints which goes against this approach.


TomElliottFlexion (approved):
Backmerge from `staging` should fix up the extra files

shared/src/business/useCases/caseNote/getUserCaseNoteForCasesInteractor.test.js
```diff
@@ -22,20 +22,23 @@ describe('getUserCaseNoteForCasesInteractor', () => {
...
   };
 
   beforeEach(() => {
     mockCurrentUser = mockJudge;
     mockNote = MOCK_NOTE;
 
-    applicationContext.getCurrentUser.mockImplementation(() => mockCurrentUser);
+    applicationContext.getCurrentUser.mockImplementation(
+      () => new User(mockCurrentUser),
+    );
```
halprin, line 35: I don't think we need to mock this any more since we mock the usecase helper below.

shared/src/business/useCases/caseNote/getUserCaseNoteInteractor.test.js
```diff
@@ -42,7 +44,7 @@ describe('Get case note', () => {
   });
 
   it('throws an error if the entity returned from persistence is invalid', async () => {
-    applicationContext.getCurrentUser.mockReturnValue(mockJudge);
+    applicationContext.getCurrentUser.mockReturnValue(new User(mockJudge));
```
halprin, line 47: ~We mock the current user for a bunch of these tests, but in the other tests for the user case notes (i.e. `getUserCaseNoteForCasesInteractor.js`) we only mock the new usecase helper.  So in these tests we're calling the real usecase helper I would assume.  Should we be consistent?  I think this also gets into the social versus unsocial unit tests.~

halprin (commented):
More review coming after `staging` merges into `test`.

shared/src/business/useCases/caseNote/getUserCaseNoteForCasesInteractor.test.js
TomElliottFlexion: I believe we still need the mock for the authorization check

halprin: Oh, duh.  Nothing to see here folks.  Move along, move along.

shared/src/business/useCases/caseNote/getUserCaseNoteInteractor.test.js
halprin: Oops, this comment doesn't make any sense anymore.

halprin: I can't select the line because it is too far below the change in this PR, but line 51 - 53.  That's mocking the old interactor that isn't being called.  So we can change that to the usecase helper, right?

shared/src/business/useCases/caseNote/getUserCaseNoteForCasesInteractor.test.js
TomElliottFlexion: Alt Text

codyseibert:
closing because github keeps showing tons of file changes even though I've merged staging into this branch.  I made a new branch and merged test into it.

PR #2160, opened 2022-05-06
bugfix - adding fallbacks onto the entity for userNotes and calendaredCases

codyseibert:
the bug:
- login as a trialclerk1@example.com
- open any new trial session that is a working copy (mobile, Alabama)
- uI crashes


halprin (approved):
Seems OK to add the "or" logic.

mrinsin (commented):
Is it worth adding unit tests around these changes so that if someone unknowingly deletes this code, the tests can fail?

PR #2166, opened 2022-05-09
DevEx - Refactoring our Router Logic

codyseibert:
Adding some decoration to potentially fix a race condition in our UI related to changing routes. We use riot-router and it works by using an event listener to the window object when the push state occurs, which can cause two of our routes to run in parallel. This causes our UI to get into bad states where the url in the browser says /case-detail, but we are actually viewing the trial-session page. These race conditions also cause our integration tests and smoke tests to become very flaky.


jimlerza (approved):

mmarcotte (approved):
🙌 thanks for this!

halprin (approved):

PR #2172, opened 2022-05-11
devex-1060: Change approach to using count instead of an empty policy

rosiehsmith:
For reference: CircleCI [workflow](https://app.circleci.com/pipelines/github/flexion/ef-cms/34996/workflows/fcf7bab0-b33d-4096-8cf9-fca12dd8dd08/jobs/221420)


halprin (approved):
Amazing!

PR #2175, opened 2022-05-11
Smoketests Timeout Fix Attempt: try adding videoUploadOnPasses set to false 

rosiehsmith:
Error: ERR_TIMED_OUT on browser_init is still an open [issue](https://github.com/cypress-io/cypress/issues/17627) but one suggestion that's working according to some recent posts is setting this flag. Either way, when smoketests pass it doesn't seem necessary for us to have videos to watch and reduces the test run by about a minute looking at exp5 [before](https://app.circleci.com/pipelines/github/flexion/ef-cms/35007/workflows/1446f173-755f-4e96-98c4-003a12eddd66) and [after](https://app.circleci.com/pipelines/github/flexion/ef-cms/35008/workflows/231395b7-1a1b-4193-91d6-82b75310a1e5) 

Looking at a couple failures [1](https://app.circleci.com/pipelines/github/ustaxcourt/ef-cms/3399/workflows/8f828322-bc46-47b6-a19d-518da5bbb01a/jobs/35041) and [2](https://app.circleci.com/pipelines/github/ustaxcourt/ef-cms/3390/workflows/59f4884d-eb3a-42ad-ba0d-b62db239b5c8/jobs/34921) on `test`, we're still timing facing this timeout.

### DIFF
--- a/cypress-smoketests.json
+++ b/cypress-smoketests.json
@@ -5,6 +5,7 @@
     "toConsole": true
   },
   "video": true,
+  "videoUploadOnPasses": false,
   "fixturesFolder": "cypress-smoketests/fixtures",
   "integrationFolder": "cypress-smoketests/integration",
   "pluginsFile": "cypress-smoketests/plugins/index.js",
--- a/cypress-smoketests.json
+++ b/cypress-smoketests.json
@@ -18,5 +18,5 @@
   "viewportHeight": 900,
   "retries": 3,
   "chromeWebSecurity": false,
-  "videoCompression": 15
+  "videoCompression": 10
 }


cypress-smoketests.json
```diff
@@ -5,6 +5,7 @@
     "toConsole": true
   },
   "video": true,
+  "videoUploadOnPasses": false,
```
halprin, line 8: My understanding is this only disables the uploading of the video to Cypress' SASS offering.  The video still get's created though.  Right???

halprin: Just stumbled on [this issue](https://github.com/cypress-io/cypress/issues/4345) that sadly confirms my understanding.  And here's the [documentation](https://docs.cypress.io/guides/references/configuration#Videos).

halprin: @rgoshen and I late last week added the `videoCompression` key to this config file.  Maybe that value isn't small enough.  Maybe decrease it to `10`?

rosiehsmith: sad. nevermind then.

rosiehsmith: woops i didn't see your last comment. yeah, fair point about that being for the cypress dashboard which we aren't using. but by default it's set to true anyway and setting it to false still seems to reduce the time for the tests? running it one more time to see if the time still remains lower.

updated `videoCompression` to 10

rosiehsmith: it's also interesting to me that that issue is from [2019](https://github.com/cypress-io/cypress/issues/4345) and the comments on the open issue, suggesting this fix, are from [8 days ago](https://github.com/cypress-io/cypress/issues/17627#issuecomment-1116136937)

halprin: That person's really sure.  I guess we can try both...?

rosiehsmith: I'm not sure 🤷  feels weird to do both probably. Wish we could get it to fail on an experimental

rachelschneiderman (approved):

cypress-smoketests.json
rosiehsmith: Even though this is merged, thought this was pertinent! In the merged workflow, it looks like it's not compressing the videos anymore since the specs passed with the `videoUploadOnPasses` flag. [Before](https://app.circleci.com/pipelines/github/ustaxcourt/ef-cms/3399/workflows/8f828322-bc46-47b6-a19d-518da5bbb01a/jobs/35041) and [after](https://app.circleci.com/pipelines/github/ustaxcourt/ef-cms/3402/workflows/4c6b2b81-c69b-454f-8d29-ecbf593b0900/jobs/35132) workflows. Seems like this [paragraph](https://docs.cypress.io/guides/guides/screenshots-and-videos#:~:text=When%20using%20the,option%20to%20false.) in the docs helps support that 🧐

PR #2185, opened 2022-05-16
Devex 974 to Staging

rachelschneiderman:


shared/src/business/useCases/getCasesForUserInteractor.js
```diff
@@ -1,41 +1,43 @@
 const { CASE_STATUS_TYPES } = require('../entities/EntityConstants');
+const { compareISODateStrings } = require('../utilities/sortFunctions');
 const { UserCase } = require('../entities/UserCase');
 
 /**
- * getOpenConsolidatedCasesInteractor
+ * getCasesForUserInteractor
  *
  * @param {object} applicationContext the application context
  * @returns {object} the open cases data
```
halprin: Small nit and sorry, should have captured this on the PR to `test`.  `@returns` should be updated because more than the open cases are now being returned.

web-api/src/cases/getCasesForUserLambda.js
```diff
@@ -6,9 +6,9 @@ const { genericHandler } = require('../genericHandler');
  * @param {object} event the AWS event object
```
halprin, line 6: Another small nit in a similar vein.  Line 4 mentions this is used to fetch all open cases, but it also returns closed cases too.  I can't comment directly on line 4 since it's outside of what GitHub lets me.

halprin (approved):

shared/src/business/useCases/getCasesForUserInteractor.js
rachelschneiderman: No problem, I really don't mind comments on staging PRs! Sometimes it just happens that we miss the PR to test... Thanks for the review, I'll update that now!

halprin (approved):

jimlerza:
Here is the original card:
https://trello.com/c/NVC15vYH

PR #2218, opened 2022-05-25
9491 sort my messages

shumway-tm:


ujahio:
created an intermediate branch to resolve the conflict #2221

PR #2221, opened 2022-05-25
9491 intermediate to test

ujahio:
Resolves some merge conflicts to test environment


web-client/integration-tests/adcClerkViewsMessagesJourney.test.js
```diff
@@ -0,0 +1,296 @@
...
+    });
+
+    loginAs(cerebralTest, 'docketclerk@example.com');
+    createNewMessageOnCase(cerebralTest, {
+      subject: message2Subject,
+      toSection: 'adc',
+      toUserId: testAdcId,
+    });
+
+    loginAs(cerebralTest, 'docketclerk@example.com');
```
rachelschneiderman: Do we need to log in again as 'docketclerk' when we are already logged in on 74?

web-client/integration-tests/adcClerkViewsMessagesJourney.test.js
```diff
@@ -0,0 +1,296 @@
...
+  MESSAGE_QUEUE_TYPES.forEach(messageQueue => {
+    const testAdcId = '6805d1ab-18d0-43ec-bafb-654e83405416';
+    const petitionsClerkId = '4805d1ab-18d0-43ec-bafb-654e83405416';
+    let beforeInboxMessageCount = 0;
+    let beforeOutboxMessageCount = 0;
+    let beforeCompletedMessageCount = 0;
+
+    loginAs(cerebralTest, 'adc@example.com');
+    it('get before counts for all section message boxes', async () => {
+      await getUserMessageCount(cerebralTest, 'inbox', messageQueue);
```
rachelschneiderman: Could we make use of the value returned from getUserMessageCount like on line 109 to get the length of the  messages in the provided messageQueue?

web-client/integration-tests/adcClerkViewsMessagesJourney.test.js
```diff
@@ -0,0 +1,296 @@
...
+      const newSortedMessages = [...messages].sort((a, b) =>
+        b.createdAt.localeCompare(a.createdAt),
+      );
+
+      expect(messages).toEqual(newSortedMessages);
+    });
+  });
+});
+
+const validateMessageOrdering = (actualMessages, expectedMessageSubjects) => {
```
rachelschneiderman: 💭 Could we make use of the `expect.arrayContaining` method in jest to enumerate what we expect the message ordering to be? That way if the test fails because the messages are not ordered as we expect, it may be more obvious why the sorting is not as we expected.

web-client/integration-tests/adcClerkViewsMessagesJourney.test.js
```diff
@@ -0,0 +1,296 @@
...
+  expect(pointer).toEqual(expectedMessageSubjects.length);
+};
+
+const markMessageAsComplete = async (
+  context,
+  messages,
+  messageCompletedSubject,
+  completedMessageSubject,
+) => {
+  const foundMessage = messages.find(
```
rachelschneiderman, line 300: ❓ Is there a way to save the `messageId`s from each of the messages created during test setup? That way we can be sure we are grabbing the message we intend off of state. I see that the date is included in the subject, which is a good way of uniquely identifying the messages! Just wondering if there is a more direct way.

web-client/src/presenter/actions/setDefaultTableSortAction.test.js
```diff
@@ -65,41 +66,40 @@ describe('setDefaultTableSortAction', () => {
       },
     });
 
-    expect(state.tableSort.sortField).toEqual('completedAt');
+    expect(state.tableSort.sortField).toEqual('createdAt');
     expect(state.tableSort.sortOrder).toEqual(DESCENDING);
   });
 
-  it('should not set the tableSort if role is not ADC', async () => {
+  it('the completed should be sorted by createdAt DESCENDING for adc user', async () => {
```
rachelschneiderman: I think there's a discrepency between the test name and the expectations here, line 73 says the completed messages should be sorted by `createdAt` but the expectation says `completedAt`.

web-client/src/presenter/actions/setDefaultTableSortAction.test.js
```diff
@@ -65,41 +66,40 @@ describe('setDefaultTableSortAction', () => {
...
         presenter,
       },
       props: {
         box: 'inbox',
-        queue: 'my',
       },
       state: {
         user: {
-          role: 'adc',
+          role: 'docketclerk',
```
rosiehsmith, line 108: Could you explicitly set the tableSort here in the `arrange` portion of the test so we can be sure this will fail? Thanks!

web-client/src/presenter/actions/setDefaultTableSortAction.js
```diff
@@ -10,14 +10,13 @@ import { state } from 'cerebral';
...
-  // Story 9490 requires the sorting to only apply to the ADC Section boxes.  So these if statements exist to limit the scope for now.  These if statements can be removed in future stories that expand the scope.
   const userRole = get(state.user.role);
-  if (props.queue !== 'section' || userRole !== 'adc') {
+  if (userRole !== 'adc') {
     store.unset(state.tableSort);
     return;
   }
 
-  // different tables require different sorting
+  // different tables require different default sorting
```
rachelschneiderman: Nit: y'all added some really awesome tests for this functionality, is this comment needed?

web-client/src/views/Messages/MessagesIndividualCompleted.jsx
```diff
@@ -1,21 +1,80 @@
+import {
+  ALPHABETICALLY_ASCENDING,
+  ALPHABETICALLY_DESCENDING,
+  CHRONOLOGICALLY_ASCENDING,
+  CHRONOLOGICALLY_DESCENDING,
+} from './sortConstants';
```
rachelschneiderman: ❓ Why were these added to a `sortConstants` over adding them to the existing applicationContext constants?

rosiehsmith (approved):

web-client/integration-tests/adcClerkViewsMessagesJourney.test.js
ujahio: Yes. We've made that change.

web-client/src/presenter/actions/setDefaultTableSortAction.test.js
ujahio: Good catch

web-client/src/presenter/actions/setDefaultTableSortAction.js
ujahio: Made the change.

web-client/src/views/Messages/MessagesIndividualCompleted.jsx
ujahio: No particular reason. We have migrated those items to application context.

web-client/src/presenter/actions/setDefaultTableSortAction.test.js
ujahio: Good catch, added the default sort order

web-client/integration-tests/adcClerkViewsMessagesJourney.test.js
ujahio: Yes. There is a way to use ids. We have made the appropriate refactor to use ids.

web-client/integration-tests/adcClerkViewsMessagesJourney.test.js
ujahio: After some research, we think expect.arrayContaining guarantees order. Part of the tests we performed was altering the ideal order and the tests returned a false positive.

PR #2224, opened 2022-05-26
9490 - Sorting ADC Section Messages

codyseibert:


web-client/src/presenter/state.js
```diff
@@ -330,6 +334,10 @@ export const baseState = {
     todaysOrdersSort: [],
   },
   showValidation: false,
+  tableSort: {
+    sortField: 'createdAt',
+    sortOrder: 'asc',
```
mmarcotte, line 339: feels nitpicky, but should this 'asc' be making use of  `web-client/src/presenter/presenterConstants.js`?

ujahio: You're correct. I'll address shortly

mmarcotte: ... updating in future work.

mmarcotte (approved):

PR #2231, opened 2022-05-31
9488 adc filtering to test

codyseibert:


web-client/src/presenter/computeds/formattedMessages.js
```diff
@@ -21,12 +22,62 @@ export const formattedMessages = (get, applicationContext) => {
...
+      completedByFilter ? message.completedBy === completedByFilter : true,
+    );
+  }
+
+  const caseStatuses = uniq(map(filteredMessages, 'caseStatus'));
+  const toUsers = uniq(map(filteredMessages, 'to'));
+  const fromUsers = uniq(map(filteredMessages, 'from'));
+  const fromSections = uniq(map(filteredMessages, 'fromSection'));
+  const toSections = uniq(map(filteredMessages, 'toSection'));
+
```
jamesobrooks: ⛏️  I would remove this blank line.

web-client/src/presenter/computeds/formattedMessages.test.js
```diff
@@ -146,4 +150,338 @@ describe('formattedMessages', () => {
...
+          toSection: 'petitionsclerk',
+        }),
+      ]),
+    );
+    expect(result).toMatchObject({
+      completedByUsers: expect.arrayContaining(['bob']),
+    });
+  });
+
+  it('the return the messages unfiltered if no filters are set', () => {
```
jamesobrooks, line 330:
```suggestion
  it('should return the messages unfiltered if no filters are set', () => {
```

web-client/src/presenter/computeds/formattedMessages.test.js
```diff
@@ -146,4 +150,338 @@ describe('formattedMessages', () => {
...
+          message: 'This is a test message',
+          to: 'petitionsclerk',
+          toSection: 'petitionsclerk',
+        }),
+      ]),
+    );
+    expect(result.messages.length).toEqual(2);
+  });
+
+  it('the return the completed messaged unfiltered if no filters are set', () => {
```
jamesobrooks, line 368:
```suggestion
  it('should return the completed messaged unfiltered if no filters are set', () => {
```

jamesobrooks (commented):
I'm not sure we should do the filtering in `formattedMessages`. We (@Absolutestunna and @shumway-tm I believe) pulled sorting out of there because it was doing too much. While filtering does seem more related to the work that `formattedMessages` is intended to do than sorting is, I think it would be better to have a separate computed and sequence similar to sorting.

web-client/src/presenter/computeds/formattedMessages.test.js
```diff
@@ -146,4 +150,342 @@ describe('formattedMessages', () => {
...
   });
+
+  it('the filter dropdown values should be set correctly for the messages data', () => {
+    const result = runCompute(formattedMessages, {
+      state: {
+        messageBoxToDisplay: {
+          box: 'outbox',
+        },
+        messages: [
+          {
```
rosiehsmith, line 165: 💭  With all these messages in tests, maybe it'd be worth adding a `mockMessages` file like we have with `mockCase.js`

rosiehsmith (approved):

web-client/src/presenter/computeds/formattedMessages.js
```diff
@@ -1,83 +1,61 @@
...
-
   const hasMessages = messages.length > 0;
 
-  return {
-    caseStatuses,
-    completedByUsers,
-    completedMessages: filteredCompletedMessages,
-    fromSections,
-    fromUsers,
+  let showFilters = role === USER_ROLES.adc;
```
jamesobrooks, line 27: :pick: I might name this one `isFilterableAndSortable` or something else to indicate that it affects both things (notably line 56).

jamesobrooks (approved):

web-client/src/presenter/computeds/formattedMessages.js
```diff
@@ -1,32 +1,59 @@
...
+    });
+
+    sharedComputedResult = {
+      ...sharedComputedResult,
+      ...messageFilterResults.filterValues,
+      ...completedMessageFilterResults.filterValues,
+      completedMessages:
+        completedMessageFilterResults.filteredCompletedMessages,
+      messages: messageFilterResults.filteredMessages,
+      showFilters: true,
```
rachelschneiderman: Show filters would already be true right?

web-client/src/presenter/computeds/formattedMessages.test.js
```diff
@@ -146,4 +154,240 @@ describe('formattedMessages', () => {
       messages: [],
     });
   });
+
+  it('the filter dropdown values should be set correctly for the messages data', () => {
```
rachelschneiderman: Nit: What does correctly mean for this test? I don't see filterValues in the expectation at all. Is there a way the test name can be made more descriptive so if/when it fails it's a little more indicative of what we are expecting the behavior to be?

web-client/src/presenter/utilities/processFormattedMessages.test.js
```diff
@@ -451,4 +454,114 @@ describe('processFormattedMessages', () => {
       ]);
     });
   });
+
+  describe('applyFiltersToCompletedMessages', () => {
+    it('returns the filteredValues and expected filtered completed messages', () => {
```
rachelschneiderman: Nit: Could we add to the test description that we are filtering by completed user Rick?

web-client/src/presenter/utilities/processFormattedMessages.test.js
```diff
@@ -451,4 +454,114 @@ describe('processFormattedMessages', () => {
...
+        },
+      });
+
+      expect(result.filteredCompletedMessages.length).toEqual(1);
+      expect(result.filterValues.completedByUsers).toEqual(['rick']);
+    });
+  });
+
+  describe('applyFiltersToMessages', () => {
+    it('returns the expected filterValues and filtered messages', () => {
```
rachelschneiderman: Nit: Same here, what is the test doing and/or why do we expect the result?

web-client/src/presenter/computeds/formattedMessages.js
codyseibert: yeah good point

PR #2245, opened 2022-06-03
9489 - Filtering ADC my messages

codyseibert:
depends on #2244


web-client/src/views/Messages/MessagesIndividualInbox.jsx
```diff
@@ -18,19 +19,56 @@ const {
 
 export const MessagesIndividualInbox = connect(
   {
+    caseStatuses: state.formattedMessages.caseStatuses,
```
rachelschneiderman, line 22: Nit: Can formattedMessages just be pulled in once and then properties accessed off of it?

rachelschneiderman (approved):

web-client/src/views/Messages/MessagesIndividualInbox.jsx
mmarcotte: @rachelschneiderman -- I agree, but in the interest of time i'm going to merge this. Perhpas add it to devex?

PR #2248, opened 2022-06-03
9502: Display Consolidated Status on Message Views

rachelschneiderman:
My Inbox:
image

My Sent:
image

My Completed:
image

Section Inbox:
image

Section Sent:
image

Sent Completed:
image

Tooltip for consolidated case:
image

Tooltip for lead case:
image


shared/src/business/useCases/processStreamRecords/processCaseEntries.test.js
```diff
@@ -72,7 +72,7 @@ describe('processCaseEntries', () => {
     expect(caseRecord).toEqual('Case');
   });
 
-  it('should index a case docket entry mapping record', async () => {
+  it('should index a case docket entry mapping record in order to create a parent-child relatioship between a case and a docket entry', async () => {
```
rosiehsmith: nit: spelling on relatiooooooship

shared/src/business/useCases/processStreamRecords/processCaseEntries.test.js
```diff
@@ -85,6 +85,19 @@ describe('processCaseEntries', () => {
     expect(caseDocketEntryMappingRecord).toEqual('CaseDocketEntryMapping');
   });
 
+  it('should index a case message mapping record in order to create a parent-child relationship between a case and a message', async () => {
```
rosiehsmith, line 88: wonderfully named tests so far - clear as to what the intent is in the code. 👍 ⚡

shared/src/persistence/elasticsearch/helpers/formatDocketEntryResult.js
```diff
@@ -0,0 +1,33 @@
+const AWS = require('aws-sdk');
+
+exports.formatDocketEntryResult = ({ caseMap, hit, sourceUnmarshalled }) => {
+  const casePk = hit['_id'].split('_')[0];
+  const docketNumber = casePk.replace('case|', ''); // TODO figure out why docket number isn't always on a DocketEntry
```
rosiehsmith: is this to be addressed?

web-client/src/views/Messages/MessagesSectionOutbox.jsx
```diff
@@ -1,23 +1,15 @@
...
-  ASCENDING,
-  CHRONOLOGICALLY_ASCENDING,
-  CHRONOLOGICALLY_DESCENDING,
-  DESCENDING,
-} = applicationContext.getConstants();
-
 export const MessagesSectionOutbox = connect(
   {
     caseStatuses: state.formattedMessages.caseStatuses,
+    constants: state.constants,
```
rosiehsmith, line 11: nice use of constants off state 👍

rosiehsmith (approved):

shared/src/business/entities/Message.test.js
```diff
@@ -60,12 +122,10 @@ describe('Message', () => {
...
 
-    it('creates an invalid Message with an empty subject', () => {
+    it('should be false when subject is an empty string', () => {
       const message = new Message(
         {
-          from: 'gg',
-          fromSection: PETITIONS_SECTION,
-          fromUserId: '6805d1ab-18d0-43ec-bafb-654e83405416',
+          ...mockMessage,
           subject: ' ',
```
shumway-tm: Just a question! But is this an empty string if it has a space?

shumway-tm (approved):

codyseibert:
I'd recommend we load test this as much as possible when it's deployed to test.  We've seen issues in the past with the parent-child joins for docket entries, so let's work on verifying the cluster won't have issues when multiple internal court users are hitting their dashboard around the same time.

shared/src/business/entities/Message.test.js
rachelschneiderman: Great point! Updated the description and added another test. Thanks!

rosiehsmith: Extra info: It's worth noting that the trimming of empty spaces is happening in `JoiValidationDecorator` in the `instanceHandler`

web-api/elasticsearch/efcms-message-mappings.js
```diff
@@ -1,5 +1,11 @@
 module.exports = {
   properties: {
+    case_relations: {
+      relations: {
+        case: 'message',
+      },
+      type: 'join',
+    },
     'caseStatus.S': {
```
codyseibert, line 9: I thought we would need to add leadDocketNumber to this message-mappings.js file.  I think we did something similar for docket-entry-mapping.js where we had to put `caseCaption` into that mapping to allow us to query for it in the future.  Just a thought... if everything seems to be working and tests are good, maybe it's not an issue.

shared/src/business/entities/Message.test.js
shumway-tm: Ahhh okay so in this case whitespaces in a string would functionally be an empty string. Thanks for that information!

shared/src/persistence/elasticsearch/helpers/formatMessageResult.js
```diff
@@ -0,0 +1,32 @@
...
+        return true;
+      }
+    });
+  }
+
+  if (foundCase) {
+    const foundCaseUnmarshalled = AWS.DynamoDB.Converter.unmarshall(foundCase);
+
+    return {
+      leadDocketNumber: foundCaseUnmarshalled.leadDocketNumber,
```
codyseibert: why do we do this?  shouldn't the ...foundCaseUnmarshalled directly below already have that property?

codyseibert (approved):

web-api/elasticsearch/efcms-message-mappings.js
rachelschneiderman: That's a good question. I could be wrong so please correct me if I am! The reason we didn't add it into `efcms-message` mappings is because we only need to know if the parent case has a `leadDocketNumber` and if so, what the value is. We don't need to query the message index for that value. Does that make sense?

shared/src/persistence/elasticsearch/helpers/formatMessageResult.js
rachelschneiderman: Good catch, you're right! Thanks!

PR #2252, opened 2022-06-06
9508 - Update caseless petition

codyseibert:


shumway-tm (approved):

web-client/src/styles/custom.scss
```diff
@@ -1984,3 +1984,17 @@ button.change-scanner-button {
   background-size: 30% 30%;
   outline: 3px solid #005ea2;
 }
+
+.text-right {
```
rachelschneiderman: 🔧 There is already a text-right class in `utils.scss`!

web-client/src/styles/custom.scss
```diff
@@ -1984,3 +1984,17 @@ button.change-scanner-button {
...
+  text-align: right;
+}
+
+.width-0 {
```
rachelschneiderman: 🔧 I believe USWDS also defines this as a class!

web-client/src/styles/custom.scss
codyseibert: the utils.scss is only used for the pdf html generators, so we also needed it here I think.

codyseibert: I removed this class; we are not even using it at this point.  I think we accidentally committed it.

web-client/src/styles/custom.scss
codyseibert: ty! I removed this class; we are not even using it at this point.  I think we accidentally committed it.

mrinsin (approved):

rachelschneiderman (approved):

PR #2271, opened 2022-06-13
Dependency updates for week of 6/14/22

ujahio:
Minor updates to ```puppeteer-core```, ```puppeteer``` and ```broadcast-channel``` introduced some breaking changes. More information in team process docs.

Migration performed on [exp2](https://app.circleci.com/pipelines/github/flexion/ef-cms/35851/workflows/92cee98f-bb5f-4093-b0ea-8529318405d2/jobs/228788)


package.json
```diff
@@ -298,19 +295,22 @@
     "prettier": "^2.6.2",
     "prop-types": "^15.8.1",
     "pug-loader": "^2.4.0",
+    "qs": "^6.10.5",
```
rachelschneiderman: Nit: I know technically webpack bundles all dependencies regardless of which section they are in but I believe "qs" is used for more than just development!

package.json
```diff
@@ -298,19 +295,22 @@
...
+    "react-idle-timer": "^5.4.0",
```
rachelschneiderman: Same as above :)

web-client/src/presenter/computeds/formattedWorkQueue.js
```diff
@@ -20,7 +20,7 @@ export const workQueueItemsAreEqual = (first, second) => {
  * @param {*} applicationContext our UI application context
  * @param {*} now optional now value used for large tables to avoid recalculating
  * @param {*} yesterday optional yesterday value used for large tables to avoid recalculating
- * @returns {string} string formatted date as a string
+ * @returns {string} formatted date as a string
```
rachelschneiderman, line 23: Nice catch!

rachelschneiderman (approved):

docs/team-process.md
```diff
@@ -167,6 +167,7 @@ Below is a list of dependencies that are locked down due to known issues with se
...
 
-#### Validating Updates
--  After changes are made to any dependencies, deploy to an exp environment to verify that all tests pass!
-    - Be sure the deploy [runs a migration](./additional-resources/blue-green-migration.md#manual-migration-steps) to verify the updates do not affect the migration workflow.
+-`puppeteer` and `puppeteer-core`: temporily locked to ^14.1.2. The minor update to 14.3.0 causes jest timeout issues. See [build](https://app.circleci.com/pipelines/github/flexion/ef-cms/35766/workflows/4f92d2a3-b923-4f9b-8e22-e4c54d69f95b/jobs/228023).
+
+-`broadcast-channel`: temporily locked to ^4.12.0. The minor update to ^4.13.0 introduces a breaking change to how esbrowser function processes the browser alias of process module. See [build](https://app.circleci.com/pipelines/github/flexion/ef-cms/35774/workflows/ea7e7274-b979-4ee4-b7ff-edee692117f2/jobs/228087) for more information about memory leakage in cypress as well.
```
rosiehsmith: Was wondering if it should be set to ~4.12.0 or 4.12.0 specifically in the package.json rather than using the ^ which will still allow it to be upgraded to the minor version 4.13.0, I believe

rosiehsmith:
Could you link to a migration workflow? Thanks 😄

docs/team-process.md
```diff
@@ -167,6 +167,7 @@ Below is a list of dependencies that are locked down due to known issues with se
 
 - It'd be good to keep an eye on `s3rver` for when it exceeds 3.7.1. We have a patch in place for called `s3rver+3.7.1.patch` in order to address the high severity issue exposed by `s3rver`'s dependency on `busboy` 0.3.1, which relies on `dicer` that actually has the [security issue](https://github.com/advisories/GHSA-wm7h-9275-46v2). Unfortunately, `busboy` >0.3.1, aka ^1.0.0, is incompatible with s3rver which is why there's a patch in place to make it compatible.
 
-#### Validating Updates
--  After changes are made to any dependencies, deploy to an exp environment to verify that all tests pass!
-    - Be sure the deploy [runs a migration](./additional-resources/blue-green-migration.md#manual-migration-steps) to verify the updates do not affect the migration workflow.
```
halprin, line 172: I think we should still keep this guidance, yes?  This is still a good thing to do.

docs/team-process.md
halprin: This is the same for the bullet point above.  `^14.1.2` will include `14.3.0`.

docs/team-process.md
rosiehsmith: Ah, this is something Josh was cleaning up since I had moved it to line 152 inside the Library Update Steps, so this one is a duplicate now

halprin: Oh, OK!  Nice.

PR #2275, opened 2022-06-13
3569: Enable Creating Docket Entries on Consolidated Cases from the Lead Case

halprin:
# [3569](https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/flexion/ef-cms/3569)

This is a _big_ PR.  👎🏻  Please, _please_, we are happy to provide a walkthrough if you want one.  🙂

## Summary of Changes

- Created a feature flag that defaults to off, but is turned on when running locally.
- Frontend
  - Modified `ConfirmInitiateServiceModal.jsx` modal to allow checking of cases in the consolidated case group.  This is the main modal that ties a bunch of the helpers/computeds together.
  - Modified how we call the backend interactor with different arguments.
  - New CSS to make the checkbox text appear the default color when the checkbox is disabled instead of greyed out.  This is an accessibility requirement.
- Backend
  - Greatly modified `fileAndServeCourtIssuedDocumentInteractor.js` to propagate the docket entry to other, passed in cases.
  - Updated `updateDocketEntryPendingServiceStatus` to conditionally update the docket entry only if it already exist.  If it doesn't exist, we do nothing through a "silent failure".
  - Updated downstream helpers to support multiple cases.
- Tests up and down the line.
- Small improvements/refactors here and there.


shared/src/business/useCases/courtIssuedDocument/fileAndServeCourtIssuedDocumentInteractor.js
```diff
@@ -48,242 +53,331 @@ exports.fileAndServeCourtIssuedDocumentInteractor = async (
...
-    }
-
-    const serviceStampDate = formatDateString(
-      docketEntryEntity.servedAt,
-      'MMDDYY',
-    );
-
-    const newPdfData = await addServedStampToDocument({
+  if (ENTERED_AND_SERVED_EVENT_CODES.includes(docketEntryEntity.eventCode)) {
+    await closeCaseAndUpdatedTrialSessionForEnteredAndServedDocuments({
```
codyseibert, line 324: I'm not sure what this is doing.  So are there certain event codes that close the case basically?

shared/src/business/useCases/courtIssuedDocument/fileAndServeCourtIssuedDocumentInteractor.js
```diff
@@ -48,242 +53,331 @@ exports.fileAndServeCourtIssuedDocumentInteractor = async (
...
-      serviceStampType = 'Entered and Served';
-    }
-
-    const serviceStampDate = formatDateString(
-      docketEntryEntity.servedAt,
-      'MMDDYY',
-    );
-
-    const newPdfData = await addServedStampToDocument({
+  if (ENTERED_AND_SERVED_EVENT_CODES.includes(docketEntryEntity.eventCode)) {
```
codyseibert, line 323: I know this has been in the code base a while, but I wish this was named something else such as EVENT_CODES_THAT_CLOSE_CASE or something 🤷

shared/src/persistence/dynamo/documents/updateDocketEntryPendingServiceStatus.js
```diff
@@ -6,18 +6,29 @@ exports.updateDocketEntryPendingServiceStatus = async ({
...
-    },
-    Key: {
-      pk: `case|${docketNumber}`,
-      sk: `docket-entry|${docketEntryId}`,
-    },
-    UpdateExpression: 'SET #isPendingService = :status',
-    applicationContext,
-  });
+  try {
+    await client.update({
```
codyseibert, line 10: what was the reasoning behind adding all this conditional expression logic?  I wouldn't think our code base should be invoking this method unless this is a real docket entry right?

codyseibert (approved):

shared/src/persistence/dynamo/documents/updateDocketEntryPendingServiceStatus.js
TomElliottFlexion: Since this is called whether or not the docket entry operations are successful, we have no (efficient) way of knowing whether or not these docket entries were created on the consolidated sub-cases. Since we need to get them out of pending status either way, a conditional update seemed like the best way to go. The try catch is just to avoid throwing an error if the conditional update "fails"

shared/src/business/useCases/courtIssuedDocument/fileAndServeCourtIssuedDocumentInteractor.js
halprin: We discussed that too.  I think it was inertia that stopped us from changing it.

web-client/src/presenter/actions/CaseConsolidation/updateCaseCheckboxAction.test.js
```diff
@@ -0,0 +1,113 @@
...
+              ...MOCK_CASE,
+              checked: changedCheckValue,
+              leadDocketNumber: customizedDocketNumberOne,
+            },
+            {
+              ...MOCK_CASE,
+              checked: unchangedCheckValue,
+              docketNumber: customizedDocketNumberTwo,
+              leadDocketNumber: customizedDocketNumberOne,
+            },
```
rosiehsmith, line 37: ⛏️❓(nit and a question) I might be missing something, but is this third case giving us any more information into a different code path in the logic? It looks to me like it would give us the same info as the first case.

TomElliottFlexion: Since the lead case always remains checked, we need two additional cases to test the checkboxes on sub-cases. This third case is the _second_ sub-case.

rosiehsmith: Ah thanks!

PR #2278, opened 2022-06-17
9501 - Consolidated Coversheet to Test

codyseibert:


shared/src/business/useCases/generateCoverSheetData.test.js
```diff
@@ -0,0 +1,487 @@
...
+        documentTitle: undefined,
+        documentType: mockDocumentType,
+      },
+      filingDateUpdated: false,
+    });
+
+    expect(result.documentTitle).toBe(mockDocumentType);
+  });
+
+  it('should use documentType as documentTitle if documentTitle is undefined', async () => {
```
TomElliottFlexion: Should this make reference to the consolidated cases?

TomElliottFlexion, line 441: Should this make reference to the consolidated cases?

shared/src/business/useCases/docketEntry/fileCourtIssuedDocketEntryInteractor.js
```diff
@@ -31,128 +31,160 @@ exports.fileCourtIssuedDocketEntryInteractor = async (
     throw new UnauthorizedError('Unauthorized');
   }
 
-  const { docketEntryId, docketNumber } = documentMeta;
+  const { docketEntryId, docketNumbers, subjectDocketNumber } = documentMeta;
```
TomElliottFlexion: Nit: Consider checking the feature flag here as well, and pruning down this list of consolidated docket numbers to just the lead/subject case if false

TomElliottFlexion, line 34: Nit: Consider checking the feature flag here as well, and pruning down this list of consolidated docket numbers to just the lead/subject case if false

web-client/integration-tests/journey/docketClerkAddsDocketEntryForHearingExhibitsFromDraftOnLeadCase.js
```diff
@@ -0,0 +1,76 @@
...
+        key: 'filingDateYear',
+        value: '2000',
+      },
+    );
+
+    await cerebralTest.runSequence('saveCourtIssuedDocketEntrySequence');
+
+    await cerebralTest.runSequence('consolidatedCaseCheckboxAllChangeSequence');
+
+    for (let docketNumber of getDocketNumbersToUncheck()) {
```
TomElliottFlexion, line 66: slick

TomElliottFlexion, line 66: slick

shared/src/business/useCases/docketEntry/fileCourtIssuedDocketEntryInteractor.js
```diff
@@ -31,128 +31,160 @@ exports.fileCourtIssuedDocketEntryInteractor = async (
...
-    assigneeId: user.userId,
-    assigneeName: user.name,
-    section: user.section,
-    sentBy: user.name,
-    sentBySection: user.section,
-    sentByUserId: user.userId,
-  });
+    const docketEntry = caseEntity.getDocketEntryById({
+      docketEntryId,
+    });
```
halprin: Let's move this code lower and closer to where we use it in the `if(!docketEntry)` statement.

web-client/src/presenter/actions/CourtIssuedDocketEntry/submitCourtIssuedDocketEntryActionHelper.test.js
```diff
@@ -0,0 +1,61 @@
...
+      getDocketNumbers: () => ['101-20'],
+      subjectDocketNumber: '101-20',
+    });
+
+    expect(
+      applicationContext.getUseCases().addCoversheetInteractor,
+    ).toHaveBeenCalled();
+  });
+
+  it('should NOT create a coversheet if the eventCode is a coversheet required document', async () => {
```
TomElliottFlexion:
```suggestion
  it('should NOT create a coversheet if the eventCode is not a coversheet required document', async () => {
```
```suggestion
  it('should NOT create a coversheet if the eventCode is a coversheet required document', async () => {
```

TomElliottFlexion:
```suggestion
  it('should NOT create a coversheet if the eventCode is not a coversheet required document', async () => {
```
```suggestion
  it('should NOT create a coversheet if the eventCode is a coversheet required document', async () => {
```

shared/src/business/useCases/docketEntry/fileCourtIssuedDocketEntryInteractor.js
```diff
@@ -31,128 +31,160 @@ exports.fileCourtIssuedDocketEntryInteractor = async (
...
-      sentByUserId: user.userId,
-    },
-    { applicationContext },
-  );
+  const isUnservable = UNSERVABLE_EVENT_CODES.includes(documentMeta.eventCode);
 
-  if (isUnservable) {
-    workItem.setAsCompleted({ message: 'completed', user });
-  }
+  for (let docketNumber of docketNumbers) {
```
halprin: For each case, we run the multiple saves asynchronously, but we iterate over each case synchronously.  Especially for large consolidated case groups, we should asynchronously iterate over each case.  That's what we did in `fileAndServeCourtIssuedDocumentInteractor`.

web-client/src/presenter/actions/CourtIssuedDocketEntry/submitCourtIssuedDocketEntryToConsolidatedGroupAction.js
```diff
@@ -0,0 +1,29 @@
...
+export const submitCourtIssuedDocketEntryToConsolidatedGroupAction = async ({
+  applicationContext,
+  get,
+}) => {
+  await submitCourtIssuedDocketEntryActionHelper({
+    applicationContext,
+    docketEntryId: get(state.docketEntryId),
+    form: get(state.form),
+    getDocketNumbers: () => {
+      return flow(
```
TomElliottFlexion: this is cool

TomElliottFlexion, line 22: this is cool

shared/src/business/useCases/docketEntry/fileCourtIssuedDocketEntryInteractor.js
halprin: Was any thought put into moving the similar code between this interactor and the `fileAndServeCourtIssuedDocumentInteractor` we touched into a usecase helper?

halprin (commented):
Some initial thoughts.  More to come later probably.

web-client/src/views/ConfirmInitiateSaveModal.jsx
```diff
@@ -0,0 +1,60 @@
...
+    documentTitle: props.documentTitle,
+    formattedCaseDetail: state.formattedCaseDetail,
+    serveCourtIssuedDocumentFromDocketEntrySequence:
+      sequences.serveCourtIssuedDocumentFromDocketEntrySequence,
+    submitCourtIssuedDocketEntrySequence:
+      sequences.submitCourtIssuedDocketEntrySequence,
+    updateCaseCheckbox: sequences.updateCaseCheckboxSequence,
+    waitingForResponse: state.progressIndicator.waitingForResponse,
+  },
+  function ConfirmInitiateServiceModal({
```
TomElliottFlexion: Minor missed rename

TomElliottFlexion, line 22: Minor missed rename

web-client/src/views/ConfirmInitiateSaveModal.jsx
```diff
@@ -0,0 +1,60 @@
...
+    formattedCaseDetail,
+    submitCourtIssuedDocketEntrySequence,
+    updateCaseCheckbox,
+  }) {
+    const [isSubmitting, setIsSubmitting] = useState(false);
+    return (
+      <ModalDialog
+        cancelLabel="No, Take Me Back"
+        cancelSequence={cancelSequence}
+        className="confirm-initiate-service-modal"
```
TomElliottFlexion, line 36: also here

TomElliottFlexion: also here

web-client/src/views/ConfirmInitiateServiceModal.jsx
```diff
@@ -4,6 +4,67 @@ import { connect } from '@cerebral/react';
 import { props, sequences, state } from 'cerebral';
 import React from 'react';
 
+export const ConsolidatedCasesCheckboxes = ({
```
TomElliottFlexion, line 7: This is much better, thanks!

TomElliottFlexion, line 7: This is much better, thanks!

TomElliottFlexion (approved):

shared/src/business/useCases/docketEntry/fileCourtIssuedDocketEntryInteractor.js
codyseibert: I may just keep this interactor as is if that's cool.  I think since the front end will fetch the feature flag right before it submits, it means that the front end should always send the proper format to the backend. at least that's what I think🤞

shared/src/business/useCases/docketEntry/fileCourtIssuedDocketEntryInteractor.js
codyseibert: I haven't checked if there were ways to re-use code.  I can try double checking.

codyseibert: I checked the code and the code paths seem to do pretty different things.  I feel drying these up would result in more complexity

shared/src/business/useCases/docketEntry/fileCourtIssuedDocketEntryInteractor.js
```diff
@@ -31,128 +31,163 @@ exports.fileCourtIssuedDocketEntryInteractor = async (
...
+          isOnDocketRecord: true,
+          judge: documentMeta.judge,
+          numberOfPages,
+          scenario: documentMeta.scenario,
+          serviceStamp: documentMeta.serviceStamp,
+          trialLocation: documentMeta.trialLocation,
+          userId: user.userId,
+        },
+        { applicationContext },
+      );
```
halprin: You'll need to add `docketNumber: caseEntity.docketNumber`.  Same for inside the `editState`.  I.e....


```javascript
editState: JSON.stringify({
    ...documentMeta,
}),
```

You may even need to override `docketEntryId` in the `editState`.  @TomElliottFlexion, do you know why overrode `docketEntryId` in `editState` in our interactor?

shared/src/business/useCases/docketEntry/fileCourtIssuedDocketEntryInteractor.js
```diff
@@ -31,128 +31,163 @@ exports.fileCourtIssuedDocketEntryInteractor = async (
...
+          assigneeId: null,
+          assigneeName: null,
+          associatedJudge: caseToUpdate.associatedJudge,
+          caseStatus: caseToUpdate.status,
+          caseTitle: Case.getCaseTitle(caseEntity.caseCaption),
+          docketEntry: {
+            ...docketEntryEntity.toRawObject(),
+            createdAt: docketEntryEntity.createdAt,
+          },
+          docketNumber: caseToUpdate.docketNumber,
```
halprin: Small nit: Change `caseToUpdate` to `caseEntity`.  That way it uses the validated data.

shared/src/business/useCases/docketEntry/fileCourtIssuedDocketEntryInteractor.test.js
```diff
@@ -232,4 +236,54 @@ describe('fileCourtIssuedDocketEntryInteractor', () => {
...
+    await fileCourtIssuedDocketEntryInteractor(applicationContext, {
+      documentMeta: {
+        docketEntryId: LEAD_CASE.docketEntries[0].docketEntryId,
+        docketNumbers: [
+          MOCK_CONSOLIDATED_1_CASE_WITH_PAPER_SERVICE.docketNumber,
+          LEAD_CASE.docketNumber,
+        ],
+        documentType: 'Trial Exhibits',
+        eventCode: 'TE',
+        subjectDocketNumber: '109-19',
```
halprin: Shouldn't this be the `LEAD_CASE.docketNumber`?

shared/src/business/useCases/docketEntry/fileCourtIssuedDocketEntryInteractor.test.js
```diff
@@ -232,4 +236,54 @@ describe('fileCourtIssuedDocketEntryInteractor', () => {
...
+      },
+    });
+
+    expect(
+      applicationContext
+        .getUseCaseHelpers()
+        .updateCaseAndAssociations.mock.calls[0][0].caseToUpdate.docketEntries.find(
+          docketEntry => docketEntry.eventCode === 'TE',
+        ),
+    ).toBeDefined();
```
halprin: We should expect other things like ensuring that the `docketNumber` was updated for the different case.

shared/src/business/useCases/docketEntry/fileCourtIssuedDocketEntryInteractor.js
```diff
@@ -31,128 +31,163 @@ exports.fileCourtIssuedDocketEntryInteractor = async (
     throw new UnauthorizedError('Unauthorized');
   }
 
-  const { docketEntryId, docketNumber } = documentMeta;
+  const { docketEntryId, docketNumbers, subjectDocketNumber } = documentMeta;
```
halprin: What do you think about refactoring how the frontend passes data to the backend like how we did in our PR?  We changed it to separate the form data from the docket entry ID, docket numbers, and the subject case docket number.  Right now, the frontend and this interactor combines it all together under `documentMeta`.

halprin (commented):
More!

shared/src/business/useCases/docketEntry/fileCourtIssuedDocketEntryInteractor.js
TomElliottFlexion: Good catch

TomElliottFlexion: No, I'm not sure, it seems kind of redundant now?

TomElliottFlexion: It should be the same as the `documentMeta.docketEntryId`. Only the `docketNumber` should need to be overwritten

codyseibert: the addDocketEntry method will set the docketNumber of the docketEntry.  So since the only time we call updateDocketEntry is for the subject case, idk if this is even needed.

jamesobrooks: I want to say we moved away from using things from documentMeta once we'd evaluated it once already.

shared/src/business/useCases/docketEntry/fileCourtIssuedDocketEntryInteractor.js
```diff
@@ -31,128 +31,163 @@ exports.fileCourtIssuedDocketEntryInteractor = async (
...
+          }),
+        );
+      } else {
+        saveItems.push(
+          applicationContext.getPersistenceGateway().saveWorkItem({
+            applicationContext,
+            workItem: rawValidWorkItem,
+          }),
+        );
+      }
```
jamesobrooks, line 177: 🌱 This block of code exhibits one of the few things that does make me think TypeScript would be a good addition. Not that it would necessarily be an overall good decision to implement, but I think this could be much clearer. We have these two gateways that both pertain to `WorkItem`s, but we just have to know that they are both there and what the consequences of the two different functions are. If we could have an `interface` or a `type` which would let us just call the function we need on the the `WorkItem` and not have to know about these separate gateways. Of course we could implement something that allows similar behavior, but probably not as cleanly.

shared/src/business/useCases/docketEntry/fileCourtIssuedDocketEntryInteractor.js
shumway-tm: Thanks for the feedback. docketNumbers and subjectDocketNumber have been separated from documentMeta.

PR #2279, opened 2022-06-17
Devex 950 - Simplify running smoke tests locally

jamesobrooks:
# Make Use of Secrets Manager and Deploy Tables

This PR is the first step in simplifying the process of running smoketests locally pointing to deployed environments.

## Add New NPM Scripts to Test the Current Color if Desired

Two new NPM scripts are defined:
```
cypress:smoketests:current-color
cypress:smoketests:current-color:open
```

These pass a parameter to the new script described below.

## Add Script to Set Up the Local Environment Variable

Currently the `ENV` variable must be exported prior to running the script. Additionally, as is common to many scripts, it is advisable to export the AWS environment variables or have some other method of authenticating with AWS.

This new script retrieves secrets from AWS Secrets Manager and determines the value of the `DEPLOYING_COLOR` variable to be passed to Cypress via the `--env` option. If there is an argument passed into the script, we use `CURRENT_COLOR` so that you can test the currently deployed code in an environment.

## Future Work

I plan to extend the use of this new script to the other `cypress:[specifiers*]` NPM scripts and/or reduce the number of scripts and rely on arguments to determine which Cypress tests to run, and how to run them.


TomElliottFlexion (approved):

halprin (approved):

mmarcotte (approved):
looks excellent for tying things together. thank you!

PR #2282, opened 2022-06-17
3569: Make endpoint async

TomElliottFlexion:
Updates `file-and-serve-court-issued-docket-entry` endpoint to be async. Fixes associated tests. 

Should be less ugly after staging is backmerged to test.


jamesobrooks (approved):

rgoshen (commented):
Reviewed and approved it because already been reviewed by team.  We had to hard reset USTC/test due to unresolvable conflicts.

PR #2293, opened 2022-06-22
Devex 1051: Don't re-create the ES Cluster if it isn't needed

rachelschneiderman:
See [trello](https://trello.com/c/v1wiy84J/1051-dont-re-create-the-es-cluster-if-it-isnt-needed) card for deployment steps


rachelschneiderman:
I realized one of the env variables the migration cron lambda is using was not brought in properly. Closing this PR while I redeploy and verify it's now correctly configured.

PR #2295, opened 2022-06-22
3569: Fix mistakes

TomElliottFlexion:
- Await call to `getFeatureFlagInteractor` (don't know how we missed this)
- Check flag prior to calling `fileAndServeCourtIssuedDocumentInteractor`, just to be double sure
- Fixed `alertSuccess` message not being set correctly.
- TESTS!

### DIFF
--- a/web-client/integration-tests/journey/docketClerkAddsAndServesDocketEntryFromOrder.js
+++ b/web-client/integration-tests/journey/docketClerkAddsAndServesDocketEntryFromOrder.js
@@ -7,6 +7,7 @@ import { withAppContextDecorator } from '../../src/withAppContext';
 export const docketClerkAddsAndServesDocketEntryFromOrder = (
   cerebralTest,
   draftOrderIndex,
+  testFormBehavior = true,
 ) => {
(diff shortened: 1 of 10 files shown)


web-client/integration-tests/journey/docketClerkAddsAndServesDocketEntryFromOrder.js
```diff
@@ -72,110 +74,112 @@ export const docketClerkAddsAndServesDocketEntryFromOrder = (
...
-    // eventCode: OCA
-    await cerebralTest.runSequence(
-      'updateCourtIssuedDocketEntryFormValueSequence',
-      {
-        key: 'eventCode',
-        value: 'OCA',
-      },
-    );
+    if (testFormBehavior) {
+      // eventCode: OCA
```
rosiehsmith, line 78: Nit: are all the eventCode comments cleanup-worthy?

rosiehsmith (approved):

web-client/integration-tests/journey/docketClerkAddsAndServesDocketEntryFromOrder.js
TomElliottFlexion: Those do seem redundant, don't they?

PR #2296, opened 2022-06-22
Admin script for re-serving documents that were served within a specified timeframe

jimlerza:
Added a script we can run to re-serve documents that were served while the IRS Service Agent email address was on SES suppression list.

For this, we are assuming the admin running the script has already determined the timeframe to search within for documents to re-serve.


shared/admin-tools/email/resend-service-email.js
```diff
@@ -0,0 +1,82 @@
...
+      docketEntryId,
+    });
+  } else {
+    await sendServedPartiesEmails({
+      applicationContext,
+      caseEntity,
+      docketEntryId,
+      servedParties: { electronic: [] },
+    });
+  }
```
jamesobrooks: Do we want to be re-sending service to users other than the IRS superuser?

jamesobrooks (approved):
If this exists specifically to handle the documents that were not served to the IRS, I wonder if it should include the ability to exclude parties, or maybe even only be capable of serving the IRS superuser.

TomElliottFlexion (approved):

halprin (commented):
I'd add an npm script to `package.json` to call your new script.  There seem to be other admin tools that are there already.

shared/admin-tools/email/resend-service-email.js
jimlerza: I believe this is being achieved by the empty array (there on line 65) - the IRS superuser is always appended, so since the array is otherwise empty, only the IRS superuser is served.

PR #2304, opened 2022-06-23
Bug 9522: Unable to download all cases on a trial session

halprin:
Forcing the resolution of `async` to version `2.6.4` broke either the `archiver` dependency when ZIPing all the cases or the `aws-sdk` dependency when uploading the ZIP file.  `async` was originally force resolved because of a vulnerability in `artillery` (so it was a transitive dependency), but it no longer depends on a vulnerable version.  So we don't need this forced resolution anymore.

Because this edits the dependencies, I'd like to get #2303 merged first and then back-merged into `test` before this one merges.

Many thanks to @rosiehsmith, @codyseibert, @jamesobrooks, @mmarcotte, and @jimlerza for helping triage this bug!


rosiehsmith (approved):

rosiehsmith:
Just for my own sanity reasons, wanted to note that git bisect appears to have worked in helping narrow down the commits because in commit 9ea93d27b8d328190 (the one I ended up at), it appears that in the package-lock `artillery` had updated its dependency on `async` to 2.6.4. We'd previously already had the transitive dependency, with a ^ in front of 2.6.4, but from what I can see it was using `async` 1.5.2 for `artillery` prior to that. That is to say, wow, what a web of things. 

My bad for the diversion onto the sparticuz/chrome-aws-lambda, that was the commit preceding the one above (598ef). I think that I marked it as bad during git bisect because it failed to compile for separate reasons, not related to the trial sessions download.

A takeaway I have from this is that during our dependency update rotations, we could also go through the transitive dependencies and see if they're still necessary as their parent packages get updated. Particularly the ones that were added originally due to security vulnerabilities.

TomElliottFlexion (approved):

halprin:
Closing in lieu of PR #2309.

PR #2309, opened 2022-06-24
Bug 9522: Unable to download all cases on a trial session

halprin:
Forcing the resolution of `async` to version `2.6.4` broke either the `archiver` dependency when ZIPing all the cases or the `aws-sdk` dependency when uploading the ZIP file.  `async` was originally force resolved because of a vulnerability in `artillery` (so it was a transitive dependency), but it no longer depends on a vulnerable version.  So we don't need this forced resolution anymore.

Because this edits the dependencies, I'd like to get #2303 merged first and then back-merged into `test` before this one merges.

Many thanks to @rosiehsmith, @codyseibert, @jamesobrooks, @mmarcotte, and @jimlerza for helping triage this bug!

This is the same as #2304, but GitHub was stuck with a bad diff.


codyseibert (approved):

maddenf (approved):
Looks good.

PR #2313, opened 2022-06-26
Redact the Authorization header from logs

mmarcotte:
Add `request.headers.Authorization` to the list of keys to redact before logging them to Cloudwatch. We don't want to log bearer tokens to cloudwatch nor index them in Kibana for security reasons.

### DIFF
--- a/web-api/src/createLogger.js
+++ b/web-api/src/createLogger.js
@@ -14,7 +14,11 @@ const { cloneDeep, unset } = require('lodash');
 
 const redact = format(logEntry => {
   const copy = cloneDeep(logEntry);
-  ['user.token', 'request.headers.authorization'].forEach(k => unset(copy, k));
+  [
+    'user.token',
+    'request.headers.authorization',
+    'request.headers.Authorization',
+  ].forEach(k => unset(copy, k));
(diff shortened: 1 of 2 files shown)


web-api/src/createLogger.js
```diff
@@ -14,7 +14,11 @@ const { cloneDeep, unset } = require('lodash');
 
 const redact = format(logEntry => {
   const copy = cloneDeep(logEntry);
-  ['user.token', 'request.headers.authorization'].forEach(k => unset(copy, k));
+  [
```
rachelschneiderman, line 17: Is there a unit test that can be written to cover this? 

It doesn't look like we are setting the little 'a' authorization header, can that be removed?

halprin (commented):
Looks good, but I also like the idea of writing a simple unit test in `createLogger.test.js`.

web-api/src/createLogger.js
mmarcotte: will add tests, thanks!

however, pretty sure the lowercase `a` is used (and redacted) in the regular api logs.

rachelschneiderman (approved):

PR #2314, opened 2022-06-27
9520 extra large trial sessions

rgoshen:
Change endpoint '/case-notes/batch-cases/${docketNumbers}/user-notes' from get to post.  Removed docketNumbers from URL into request body.


web-client/src/presenter/actions/TrialSession/updateTrialSessionAction.js
```diff
@@ -17,26 +18,37 @@ export const updateTrialSessionAction = async ({
...
   props,
 }) => {
-  const startDate = // AAAA-BB-CC
-    (props.computedDate &&
-      applicationContext
-        .getUtilities()
-        .prepareDateFromString(props.computedDate)
-        .toISOString()) ||
-    null;
+  console.log('startDate before: ', props.computedStartDate);
```
mrinsin, line 21: Nit: seems like some logs got leftover (line 26 as well!)

web-api/src/app.js
```diff
@@ -745,8 +745,8 @@ const { validatePdfLambda } = require('./documents/validatePdfLambda');
  * case-notes
  */
 {
-  app.get(
-    '/case-notes/batch-cases/:docketNumbers/user-notes',
+  app.post(
```
mrinsin, line 748: Just verifying that the `get` had to change to a `post` because we needed a different way to pass the `Docket Numbers` through? What was the issue with passing them as query params? (This is mostly to satiate my curiosity :D )

mrinsin (approved):

web-api/src/app.js
rgoshen: Normally, this probably wouldn't have been a big deal, but on trial sessions that can have a lot of trials (say in this instance 125 or more) caused a 414 error to be thrown.  So, to make sure the URI did not exceed the limit of the gateway, it was easier to pass in the body vs. query params.

jamesobrooks: @mrinsin My understanding of the way [API Gateway](https://docs.aws.amazon.com/apigateway/latest/developerguide/limits.html) works, our URL is artificially limited in length. It could be that I read that documentation wrong and query params have a different effect on API Gateway than long paths do.

Alternatively we could leave the request as a GET instead of changing to POST and still include a request body. I don't particularly have any issues with that here since this API isn't exactly high on the Richardson maturity model of REST.

mrinsin: Ahh thank y'all for the explanation! @jamesobrooks no worries, I'm not feeling too strongly about changing the request especially now that y'all helped me get a deeper understanding of the reason behind the change! 

Side question - can we really do a GET request with a body? I'd love to learn how if so - I feel like I tried it before and couldn't actually access the body until I changed it to a POST!

jamesobrooks: Well, I don't know if the libraries we're using would support it out of the box.

halprin: You can.  According to [RFC 7231, section 4.3.1](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.1), a request body for `GET` is acceptable.

> A payload within a GET request message has no defined semantics; sending a payload body on a GET request might cause some existing implementations to reject the request.

Meaning that the HTTP specification doesn't have any predefined semantics, but we _can_.

The earlier [RFC 2616](https://www.rfc-editor.org/rfc/rfc2616) contained language that led many to believe that a request body shouldn't be included in a `GET` request.  That language has been removed in the newer RFC 7231.

Because of that earlier language, many libraries artificially forbid request bodies in `GET` requests.

mrinsin: Awesome, thanks for knowledge sharing, y'all!

rgoshen: Thanks @halprin.  I also thought you could only get access to the body using POST.

jamesobrooks: Thanks @halprin for a much better answer than mine.

rachelschneiderman (commented):
Is there a test that could be added to cover this bug for the future?

jamesobrooks:
> Is there a test that could be added to cover this bug for the future?

@rachelschneiderman Unless I'm mistaken about the consequences of the following [from AWS documentation](https://docs.aws.amazon.com/apigateway/latest/developerguide/limits.html), I don't really see a feasible way to test this.
>When authorization is enabled on a method, the maximum length of the method's ARN (for example, arn:aws:execute-api:{region-id}:{account-id}:{api-id}/{stage-id}/{method}/{resource}/{path}) is 1600 bytes. The path parameter values (whose size is determined at runtime) can cause the ARN length to exceed the limit. When this happens, the API client receives a 414 Request URI too long response.

It seems as though the ARN does not have a determinable length, and we have avoided the entire issue by not including the list of docket numbers in the URI.

TomElliottFlexion (approved):

rgoshen:
Closing this pull request because it shows changed files that were not touched under this bug.

PR #2332, opened 2022-06-29
9519 Docket: Add Standalone Remote as Trial Session Location

rosiehsmith:
- Adds trial location option "Standalone Remote" for scenarios type F (HEAR, FTRL, PTRL, TRL) and type G (NTD) in dropdown when creating court issued nonstandard forms. These types both display the trial location on the form.
- Docket entry modified from "at [City/State]" to "in standalone remote session"

image

image
image

### DIFF
--- a/shared/src/business/entities/courtIssuedDocument/CourtIssuedDocumentTypeG.js
+++ b/shared/src/business/entities/courtIssuedDocument/CourtIssuedDocumentTypeG.js
@@ -9,6 +9,7 @@ const {
 const { formatDateString, FORMATS } = require('../../utilities/DateHandler');
 const { JoiValidationConstants } = require('../JoiValidationConstants');
 const { replaceBracketed } = require('../../utilities/replaceBracketed');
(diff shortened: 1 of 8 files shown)


shared/src/business/entities/courtIssuedDocument/CourtIssuedDocumentTypeG.js
```diff
@@ -24,6 +25,18 @@ CourtIssuedDocumentTypeG.prototype.init = function init(rawProps) {
 };
 
 CourtIssuedDocumentTypeG.prototype.getDocumentTitle = function () {
+  if (this.trialLocation === TRIAL_SESSION_SCOPE_TYPES.standaloneRemote) {
```
codyseibert, line 31: nit: maybe see if some of this logic can be extracted to a shared util function between these two document types.

codyseibert (approved):

rachelschneiderman (approved):

PR #2353, opened 2022-07-06
DevEx 1068 - Reduce Build Memory Usage

codyseibert:
I went with an approach of refactoring away webpack for the public and private UIs of this application.  Webpack is pretty slow, so I'm going with esbuild which seems to reduce the build time by 2 minutes and prevents the memory usage from spiking in circle.


codyseibert:
this was merged to test last week.  so far I haven't seen any issues with this refactoring.

TomElliottFlexion (approved):

PR #2374, opened 2022-07-13
USTC-1810 - Define a timeout for the info rotator lambda

jimlerza:
Going with a large timeout (60s), at least at first, to give our lambda time to churn through a few months of snapshots the first time it runs. We'll reduce this later.


codyseibert (approved):
you could also increase the default memory which will make it run faster since it means more CPU as well, I've read 700mb or so is a good amount

PR #2388, opened 2022-07-27
Dependency updates 07/25/22

TomElliottFlexion:
Patch/Minor Updates: 
+ esbuild@0.14.50
+ esbuild-css-modules-plugin@2.5.1
+ core-js@3.24.0
+ artillery@2.0.0-22
+ eslint-plugin-jsx-a11y@6.6.1
+ esbuild-sass-plugin@2.3.1
+ @vendia/serverless-express@4.10.0
+ sass@1.54.0
+ jsdoc@3.6.11
+ webpack@5.74.0
+ aws-sdk@2.1182.0
+ @aws-sdk/credential-provider-node@3.137.0
+ @pdftron/pdfjs-express-viewer@8.7.0
+ @vendia/serverless-express@4.10.1

Notes:
- Removed `ajv` as a direct dependency since it only seems to be required as a transitive dependency again
- Added `terser` to resolutions block to avoid a vulnerability via `html-webpack-plugin`

Migration: https://app.circleci.com/pipelines/github/flexion/ef-cms/36652/workflows/ea95b3f1-d0e6-4f18-9768-9c8d39bc32dc


package.json
```diff
@@ -330,7 +330,7 @@
     "stylelint-config-standard": "^26.0.0",
     "stylelint-config-standard-scss": "^5.0.0",
     "swagger-cli": "^4.0.4",
-    "webpack": "^5.73.0",
```
rosiehsmith, line 333: maybe a question for the team, but can we remove `webpack` now that we're using `esbuild` or are we still using it for something?

.circleci/config.yml
```diff
@@ -56,7 +56,7 @@ commands:
           name: Reset package-lock
           command: 'git checkout -- package-lock.json'
       - save_cache:
-          key: v22-npm-{{ checksum "package.json" }}-{{ checksum "package-lock.json" }}
```
rosiehsmith: 💭 Just thinking about this, but maybe we should revisit when and why we bump this version. To my memory, I think as long as the `package-lock` is actually different due to a difference in updated packages, we shouldn't need to bump the cache key every round of updates. It should automatically use the new version since the checksums will be different. If we want to force it to use a new cache key despite the package-lock's checksum not changing then we have to bump it up.

docs/team-process.md
```diff
@@ -151,7 +151,7 @@ If dependencies have no patch, replace it with an alternative, or wait for the l
 
 #### Caveats
 
-Below is a list of dependencies that are locked down due to known issues with security, integration problems within DAWSON, etc. Feel free to try and update any of these items in the list, please be aware of the issue that's documented and ensure it's been resolved.
+Below is a list of dependencies that are locked down due to known issues with security, integration problems within DAWSON, etc. Try to update these items but please be aware of the issue that's documented and ensure it's been resolved.
```
rosiehsmith, line 154: Better actionable language 🙌 This reads like me...

rosiehsmith (approved):

package.json
TomElliottFlexion: @codyseibert Any reason you can think of that we shouldn't remove `webpack`?

.circleci/config.yml
TomElliottFlexion: That's a good point, I was just blindly doing this because the docs said to; I'll test and update the docs accordingly

rosiehsmith: Thank you!

package.json
TomElliottFlexion: Looks like we still use for the API

jimlerza (approved):

PR #2419, opened 2022-08-10
Improve documentation for blue-green migrations

mmarcotte:
We are documenting a more straightforward approach to manually triggering a blue-green migration. 

We are also including any manual steps that are still required to perform a migration, manual or automated. Mainly, you need to ensure that the destination table does not exist before kicking off the CircleCI job.


docs/additional-resources/blue-green-migration.md
```diff
@@ -4,24 +4,62 @@
...
+    If you do not, then it exists, and you must delete the table.
+
+3. If it exists, delete the **destination table** in both `east` and `west`. You can do this via the AWS Console or AWS CLI:
+
+    ```bash
+    $ aws dynamodb delete-table --table-name efcms-dev-alpha --region us-west-1
+    $ aws dynamodb delete-table --table-name efcms-dev-alpha --region us-east-1
+    ```
+
+    NOTE: after deleting the `west` table, you may have to wait a few minutes before you can delete the `east` table.
```
rosiehsmith, line 85: We also need to delete the destination elasticsearch domain

docs/additional-resources/blue-green-migration.md
```diff
@@ -30,12 +68,17 @@ Some key steps, excluding automated tests, include:
...
 
-1. Remove a migration script item from the environment's `efcms-<ENV>-deploy` DynamoDB table that is expected to migrate, from the `migrationsToRun.js` file. For example, if `migrationsToRun.js` includes a script called `0001-test-script.js`, delete that item from the DynamoDB table.
-2. Kick off a CircleCI workflow, which will follow the steps above for Automated Migration Steps.
+1. Go into the environment's deploy table `efcms-<ENV>-deploy`
+2. Change the `migrate` flag to `true`
+3. Identify the source table and destination table. Both should be the same value, and that is what the application is currently using as it its main data store. If they are both `alpha`, then the destination table is `beta`. If they are both `beta`, the destination table is `alpha`. If they are different, inspect at the current color Lamdba's environment config to determine the dynamo table that the application is currently using as its data store.
+4. Switch to the `west-1` region.
```
rgoshen: Nit: maybe explain here why to delete from west first

rgoshen (approved):

docs/additional-resources/blue-green-migration.md
```diff
@@ -30,12 +68,17 @@ Some key steps, excluding automated tests, include:
...
 
-
 ## Manual Migration Steps
 
 The application kicks off a migration automatically if it detects migrations that need to be run in the codebase that haven't yet been run upon that environment. In order to force a migration, perform the following manual steps. You might do this if you were doing a package update that might impact the migration and wanted to test it fully.
 
-1. Remove a migration script item from the environment's `efcms-<ENV>-deploy` DynamoDB table that is expected to migrate, from the `migrationsToRun.js` file. For example, if `migrationsToRun.js` includes a script called `0001-test-script.js`, delete that item from the DynamoDB table.
-2. Kick off a CircleCI workflow, which will follow the steps above for Automated Migration Steps.
+1. Go into the environment's deploy table `efcms-<ENV>-deploy`
+2. Change the `migrate` flag to `true`
```
rosiehsmith, line 106: will this successfully work since `is-migration-needed` in [setup-for-blue-green-migration.sh](https://github.com/ustaxcourt/ef-cms/blob/staging/setup-for-blue-green-migration.sh#L15) will return `SKIP_MIGRATION` true if that environment has already run all the expected migration scripts?

rosiehsmith: i'm probably missing something here

rosiehsmith: ~~based on looking at elasticsearch in the AWS console, the `test` domains have no indexing rate despite the CI workflow saying that it's on the reindexing step. that leads me to believe that the old documentation is correct and in order to trigger a full migration and subsequent reindex, one of the expected migration scripts must be deleted from the environment. 🤔~~

hmm, scratch that, i think i wasn't looking far enough back in time, possibly?

rosiehsmith: I might be missing something, but wouldn't it be faster to delete one of the migration scripts rather than running all migrations in the `migrationsToRun` file which seems to be what would happen by toggling the `migrate` and `destination-table-version` flags? Or, is the intention here that we would want to document how to force a *full* migration, aka all the migration scripts in the `migrationsToRun` file?

docs/additional-resources/blue-green-migration.md
jimlerza: I don't think this is true anymore - I think it won't exist because it's not being created if it's not needed.

jimlerza (approved):

docs/additional-resources/blue-green-migration.md
mmarcotte: Cando.

docs/additional-resources/blue-green-migration.md
mmarcotte: @rosiehsmith -- great question. Maybe it would be easier, maybe not. I think it might be a matter of preference.

Another way to think about it is if you want to force a migration, but not actually have the data get altered, you could use the steps that I'm adding into the documentation. It's conceivable that a migration should not be run more than once (I think this has been the case in the past). 

I think I'm going to compromise and keep both steps in the documentation since they both work. Thanks for the questions. 🙌

rosiehsmith: That makes sense. Thanks for the follow up!

docs/additional-resources/blue-green-migration.md
```diff
@@ -4,24 +4,30 @@
...
 
-Additionally, for deployments that require change in the data schema or broad stroke changes to the database in order for the application to function, we build an empty copy of the data stores (DynamoDB and Elasticsearch). Then we perform a migration of information from the old data store into the newly created one. This migration passes all of the data through a lambda to update and verify the data before saving it into a new data store.
+Additionally, for deployments that require change in the data schema or elasticsearch mappings in order for the application to function, we build an empty copy of the data stores (DynamoDB and Elasticsearch). Then we perform a migration of information from currently running DynamoDB database table (*source table*) into the newly created, empty one (*destination table*). This migration passes all of the data through a lambda to update and verify the data before saving it into the destination table.
```
rosiehsmith, line 7: ❤️

rosiehsmith (approved):

PR #2420, opened 2022-08-10
Bug 9569 duplicate trial session notices generation

shumway-tm:
### Changes of Note:
- Details around how this bug occurred can be found in the Zenhub link above.  In order to prevent possible duplicate calendar Trial Session events from being kicked off and processed, a check for previously processed or completed Trial Sessions was added. 

- Additionally, the retry count for the queue feeding the set-trial-session lambda was increased to 3 before passing messages into the dead-letter queue. 

- More robust logging has been implemented at different points to more easily debug failures in the future.

### DIFF
--- a/shared/src/business/useCases/trialSessions/generateNoticesForCaseTrialSessionCalendarInteractor.test.js
+++ b/shared/src/business/useCases/trialSessions/generateNoticesForCaseTrialSessionCalendarInteractor.test.js
@@ -89,9 +89,9 @@ describe('generateNoticesForCaseTrialSessionCalendarInteractor', () => {
       .getJobStatus.mockResolvedValue({});
   });
 
-  it('should return and do nothing if the job is already processing', async () => {
+  it('should return and do nothing if the job is already processed', async () => {
     applicationContext.getPersistenceGateway().getJobStatus.mockResolvedValue({
-      [docketNumber]: 'processing',
+      [docketNumber]: 'processed',
     });
     await generateNoticesForCaseTrialSessionCalendarInteractor(
       applicationContext,
@@ -108,7 +108,7 @@ describe('generateNoticesForCaseTrialSessionCalendarInteractor', () => {
       interactorParamObject,
     );
     expect(
-      applicationContext.getPersistenceGateway().setJobAsProcessing,
+      applicationContext.getPersistenceGateway().setJobStatus,
     ).toHaveBeenCalled();
   });
 
--- a/shared/src/business/useCases/trialSessions/generateNoticesForCaseTrialSessionCalendarInteractor.test.js
+++ b/shared/src/business/useCases/trialSessions/generateNoticesForCaseTrialSessionCalendarInteractor.test.js
@@ -86,13 +86,15 @@ describe('generateNoticesForCaseTrialSessionCalendarInteractor', () => {
     };
     applicationContext
       .getPersistenceGateway()
-      .getJobStatus.mockResolvedValue({});
+      .getTrialSessionJobStatusForCase.mockResolvedValue({});
   });
 
   it('should return and do nothing if the job is already processed', async () => {
-    applicationContext.getPersistenceGateway().getJobStatus.mockResolvedValue({
-      [docketNumber]: 'processed',
-    });
+    applicationContext
+      .getPersistenceGateway()
+      .getTrialSessionJobStatusForCase.mockResolvedValue({
+        [docketNumber]: 'processed',
+      });
     await generateNoticesForCaseTrialSessionCalendarInteractor(
       applicationContext,
       interactorParamObject,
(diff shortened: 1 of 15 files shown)


rosiehsmith (approved):

rosiehsmith (approved):

shared/src/business/useCases/trialSessions/generateNoticesForCaseTrialSessionCalendarInteractor.test.js
```diff
@@ -108,7 +108,7 @@ describe('generateNoticesForCaseTrialSessionCalendarInteractor', () => {
       interactorParamObject,
     );
     expect(
-      applicationContext.getPersistenceGateway().setJobAsProcessing,
+      applicationContext.getPersistenceGateway().setJobStatus,
     ).toHaveBeenCalled();
```
mmarcotte: Seems like we should also add a test here to indicate that `it should re-attempt the job if it is already processing`. In other words, introduce a test to handle what happens when the job is processing.

shared/src/business/useCases/trialSessions/setNoticesForCalendaredTrialSessionInteractor.test.js
```diff
@@ -114,13 +124,81 @@ describe('setNoticesForCalendaredTrialSessionInteractor', () => {
...
+      .getPersistenceGateway()
+      .getTrialSessionProcessingStatus.mockResolvedValueOnce('processing');
+
+    await setNoticesForCalendaredTrialSessionInteractor(applicationContext, {
+      trialSessionId,
+    });
+    expect(
+      applicationContext.getPersistenceGateway()
+        .setTrialSessionProcessingStatus,
+    ).not.toHaveBeenCalled();
```
mmarcotte, line 137: I think the more accurate function to test is

```
applicationContext .getMessageGateway()
      .sendSetTrialSessionCalendarEvent(...)
```

This is the function that actually triggers the process of creating the notices, right?

shared/src/business/useCases/trialSessions/setNoticesForCalendaredTrialSessionInteractor.test.js
```diff
@@ -114,13 +124,81 @@ describe('setNoticesForCalendaredTrialSessionInteractor', () => {
...
+    applicationContext
+      .getPersistenceGateway()
+      .getTrialSessionProcessingStatus.mockResolvedValueOnce('complete');
+
+    await setNoticesForCalendaredTrialSessionInteractor(applicationContext, {
+      trialSessionId,
+    });
+    expect(
+      applicationContext.getPersistenceGateway()
+        .setTrialSessionProcessingStatus,
```
mmarcotte: ... same with this line.

web-api/src/applicationContext.js
```diff
@@ -1057,8 +1060,8 @@ const {
   setForHearingInteractor,
 } = require('../../shared/src/business/useCases/trialSessions/setForHearingInteractor');
 const {
-  setJobAsProcessing,
-} = require('../../shared/src/persistence/dynamo/trialSessions/setJobAsProcessing');
+  setJobStatus,
+} = require('../../shared/src/persistence/dynamo/trialSessions/setJobStatus');
```
mmarcotte: Hmm, should this function be renamed to `setTrialSessionJobStatusForCase` or something? Right now it feels a little confusing for `setJobStatus` and `getJobStatus` as it's not clear what the nature of the jobs are. And that name could potentially be claimed by a more generic purpose. Open to suggestions.

mmarcotte (changes requested):
This is looking really good, and my comments are largely nitpicky. They could be handled with follow-up devex, but I think they deserve attention and response before merging.

shared/src/business/useCases/trialSessions/generateNoticesForCaseTrialSessionCalendarInteractor.test.js
shumway-tm: We created an additional test for reattempted jobs!

shared/src/business/useCases/trialSessions/setNoticesForCalendaredTrialSessionInteractor.test.js
shumway-tm: Addressed.

shared/src/business/useCases/trialSessions/setNoticesForCalendaredTrialSessionInteractor.test.js
shumway-tm: Addressed.

web-api/src/applicationContext.js
shumway-tm: I agree. We renamed both the setter and the getter for the case status!

mmarcotte (approved):

jimlerza (approved):

PR #2421, opened 2022-08-11
DevEx - Reduce Calendared Trial Session Return Size

codyseibert:
When dealing with very large trial sessions (250+ cases), we have seen the api gateway throw 502 errors.  This is probably due to some cases have a LOT of docket entries, so if you are trying to view the trail session that contains a case with a lot of docket entries, this api endpoint return the case and every entry along with it.  The front end logic only cares about a particular docket entry for this page, so this PR introduces a filter to only return the PMT, non stricken entries in the returned payload to reduce the size and speed up the request.

This is a screenshot of the request for a trial session with 103 cases.  Notice the response size is about 189kb:
Screen Shot 2022-08-11 at 3 16 52 PM

after deploying 44kb
Screen Shot 2022-08-24 at 10 49 45 AM

### DIFF
--- a/shared/src/business/useCases/trialSessions/getCalendaredCasesForTrialSessionInteractor.js
+++ b/shared/src/business/useCases/trialSessions/getCalendaredCasesForTrialSessionInteractor.js
@@ -26,6 +26,7 @@ exports.getCalendaredCasesForTrialSessionInteractor = async (
     .getPersistenceGateway()
     .getCalendaredCasesForTrialSession({
       applicationContext,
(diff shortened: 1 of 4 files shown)


mrinsin (commented):
Could these new lines be covered by tests?

mrinsin (approved):

jimlerza (approved):

TomElliottFlexion (approved):

mmarcotte (approved):

PR #2424, opened 2022-08-11
#9575 adjust the permissions for the update petitioner cases lambda

mmarcotte:
- remove the permissions to send email;
- add the ability to publish to the sqs queue that sends emails


jimlerza (approved):

jimlerza:
Should we also merge this branch into staging?

PR #2436, opened 2022-08-15
9544 blocked cases consolidated

rgoshen:
This PR addresses [Story ustaxcourt/ef-cms#8219](https://github.com/flexion/ef-cms/issues/9544). It adds either a Lead Case icon with tool tip to a lead case in a consolidated group or a group icon and tool tip to a member case in a consolidated group to a blocked case on the Blocked Cases Report.


rosiehsmith:
FYI the base branch of this is to `staging`, not `test` in case you meant it to be to `test` 😄

web-client/src/presenter/computeds/blockedCasesReportHelper.js
```diff
@@ -32,12 +33,28 @@ export const blockedCasesReportHelper = (get, applicationContext) => {
     blockedCasesFormatted = blockedCases
       .sort(applicationContext.getUtilities().compareCasesByDocketNumber)
       .map(blockedCase => {
+        const inConsolidatedGroup = !!blockedCase.leadDocketNumber;
```
mrinsin, line 36: At this point, we are using this exact same logic in three helpers. You could extract this logic to a utility method and call that method from all three places to DRY up your code!

mrinsin (commented):
Lookin good! Added a note for refactoring

web-client/src/presenter/computeds/blockedCasesReportHelper.js
```diff
@@ -32,12 +33,28 @@ export const blockedCasesReportHelper = (get, applicationContext) => {
...
+          inConsolidatedGroup &&
+          blockedCase.leadDocketNumber === blockedCase.docketNumber;
+
+        if (inConsolidatedGroup) {
+          if (inLeadCase) {
+            consolidatedIconTooltipText = 'Lead case';
+          } else {
+            consolidatedIconTooltipText = 'Consolidated case';
+          }
+        }
```
jamesobrooks, line 47: This seems to be a little more complicated than it needs to be since `inLeadCase` already includes `inConsolidatedGroup` on line 38.
```suggestion
        consolidatedIconTooltipText = inConsolidatedGroup && 'Consolidated case';
        consolidatedIconTooltipText = inLeadCase && 'Lead case';
```

Honestly I still find this syntax a bit esoteric, but it keeps the nested if away. Alternatively, you could leave out line 38 and leave this code as-is or even make a slight change to the code I've proposed and achieve the same thing.

PR #2438, opened 2022-08-15
9544 blocked cases consolidated

rgoshen:
This PR addresses https://github.com/flexion/ef-cms/issues/9544. It adds either a Lead Case icon with tool tip to a lead case in a consolidated group or a group icon and tool tip to a member case in a consolidated group to a blocked case on the Blocked Cases Report.


web-client/src/presenter/computeds/blockedCasesReportHelper.js
```diff
@@ -32,12 +33,28 @@ export const blockedCasesReportHelper = (get, applicationContext) => {
     blockedCasesFormatted = blockedCases
       .sort(applicationContext.getUtilities().compareCasesByDocketNumber)
       .map(blockedCase => {
+        const inConsolidatedGroup = !!blockedCase.leadDocketNumber;
```
mrinsin: At this point, we are using this exact same logic in three helpers. You could extract this logic (lines 36-48) to a utility method and call that method from all three places to DRY up your code!

mrinsin (changes requested):
Added a comment for refactoring but otherwise looks good to me!

web-client/src/presenter/computeds/blockedCasesReportHelper.js
rgoshen: I agree...Tim and I will see if we address that

rgoshen: Thank you

web-client/src/presenter/computeds/blockedCasesReportHelper.js
```diff
@@ -32,12 +33,28 @@ export const blockedCasesReportHelper = (get, applicationContext) => {
...
+          inConsolidatedGroup &&
+          blockedCase.leadDocketNumber === blockedCase.docketNumber;
+
+        if (inConsolidatedGroup) {
+          if (inLeadCase) {
+            consolidatedIconTooltipText = 'Lead case';
+          } else {
+            consolidatedIconTooltipText = 'Consolidated case';
+          }
+        }
```
jamesobrooks: This seems to be a little more complicated than it needs to be since `inLeadCase` already includes `inConsolidatedGroup` on line 38.
```
        consolidatedIconTooltipText = inConsolidatedGroup && 'Consolidated case';
        consolidatedIconTooltipText = inLeadCase && 'Lead case';
```

Honestly I still find this syntax a bit esoteric, but it keeps the nested if away. Alternatively, you could leave out line 38 and leave this code as-is or even make a slight change to the code I've proposed and achieve the same thing.

codyseibert (approved):

web-client/src/presenter/computeds/blockedCasesReportHelper.js
jamesobrooks: Actually, no, that suggestion won't work, because the second one could set to false. It would need to be a ternary like:
```suggestion
        consolidatedIconTooltipText = inLeadCase ? 'Lead case' : inConsolidatedGroup ? 'Consolidated case' : null;
```

web-client/src/presenter/computeds/blockedCasesReportHelper.js
rgoshen: done...will do a devex task to update other previous stories

web-client/src/presenter/computeds/blockedCasesReportHelper.js
rgoshen: I am not sure how this makes it more readable.  Personal opinion, a nested ternary operator is harder to read and if we are going to leave in the nested if, why then the need to nest a ternary operator.

mrinsin (approved):

web-client/src/presenter/computeds/utilities/addConsolidatedProperties.js
```diff
@@ -0,0 +1,27 @@
+/**
+ * This can add consolidatedToolTip, inConsolidatedGroup flag and/or inLeadCase flag to passed in object item.
+ *
+ * @param object
+ * @returns object
+ */
+
+export const addConsolidatedProperties = object => {
```
mrinsin, line 8: Looks great, thanks for doing this! Would love to see some tests around it too but PR approved :D

web-client/src/presenter/computeds/blockedCasesReportHelper.js
```diff
@@ -1,10 +1,11 @@
+import { addConsolidatedProperties } from './utilities/addConsolidatedProperties';
 import { state } from 'cerebral';
 
 export const blockedCasesReportHelper = (get, applicationContext) => {
   const blockedCases = get(state.blockedCases);
 
   let blockedCasesFormatted = [];
-  let consolidatedIconTooltipText = null;
+  // let consolidatedIconTooltipText = null;
```
mrinsin: maybe leftover by accident?

web-client/src/presenter/computeds/blockedCasesReportHelper.js
```diff
@@ -33,28 +34,14 @@ export const blockedCasesReportHelper = (get, applicationContext) => {
...
-          } else {
-            consolidatedIconTooltipText = 'Consolidated case';
-          }
-        }
-
+        const blockedCaseWithConsolidatedProperties =
+          addConsolidatedProperties(blockedCase);
         return {
-          ...setFormattedBlockDates(blockedCase),
+          ...setFormattedBlockDates(blockedCaseWithConsolidatedProperties),
```
mrinsin, line 39: 🎉

mrinsin (approved):

web-client/src/presenter/computeds/utilities/addConsolidatedProperties.js
codyseibert: we added some tests 🥳

web-client/src/presenter/computeds/utilities/addConsolidatedProperties.test.js
```diff
@@ -0,0 +1,39 @@
+import { addConsolidatedProperties } from './addConsolidatedProperties';
```
mrinsin, line 1: 🎉 🎉 🎉 🎉

web-client/src/presenter/computeds/utilities/addConsolidatedProperties.js
mrinsin: Awesome! Thanks

web-client/src/presenter/computeds/blockedCasesReportHelper.js
rosiehsmith: I definitely see the sentiment to make this clearer, but it's refreshing my memory that originally with the first helper, we went through this same thought process. Nested if => Ternary => nested if for readability 🤷

rosiehsmith (approved):

PR #2452, opened 2022-08-18
Devex 1028 contribution guide

jamesobrooks:
# DevEx Issue 1028

[Add detail to contribution guide](https://trello.com/c/c3hY53N4/1028-add-detail-to-contribution-guide)

## Additional Information

- Bugs/Issue reporting
- PR procedures
- Prerequisites

## Inspiration 

This is generally modeled after the [React contribution guide](https://reactjs.org/docs/how-to-contribute.html).

# Questions

Did I cover all tests that people not involved with Flexion or the Court can run?
Are there more prerequisites and should we get more specific about the Node version?
Can we ensure that outside contributions have had pre-commit hooks run?


CONTRIBUTING.md
```diff
@@ -17,3 +21,48 @@ the [CC0 1.0 Universal public domain dedication](https://creativecommons.org/pub
...
+
+From the [Issues page](https://github.com/ustaxcourt/ef-cms/issues) you can click the `New issue` button or go straight to the [list of templates](https://github.com/ustaxcourt/ef-cms/issues/new/choose).
+
+### Security-related Issues
+
+For security-sensitive issues, please do not file a public issue. Rather, please email [Dawson support](dawson.support@ustaxcourt.gov).
+
+## Proposing Changes
+
+We welcome pull requests, but we do still want those changes to be related to an [Issue](#reporting-issues). Please file one prior to submitting a pull request.
```
maddenf, line 41: This was well stated

CONTRIBUTING.md
```diff
@@ -17,3 +21,48 @@ the [CC0 1.0 Universal public domain dedication](https://creativecommons.org/pub
...
+## Pull Request Procedures
+
+Please follow these steps before submitting a pull request:
+
+1. Fork the repository and create a branch from `staging`.
+1. Run `npm install` in the repository root.
+   1. Do not delete `package-lock.json`.
+1. If reasonable for the changes being made, write tests.
+1. Run `npm run lint:fix` in the repository root.
+1. Ensure that all tests run with the following scripts pass:
```
maddenf: We might want to restate the following line:  "Ensure that all tests run with the following scripts pass"
Might be better to say:  "Ensure that all tests run and that the following scripts pass"

jamesobrooks: @maddenf I didn't want to say all tests because many tests are intended to be run via ci tools against deployed artifacts. So maybe just: `Run the following scripts, ensuring no failures:`?

CONTRIBUTING.md
```diff
@@ -17,3 +21,48 @@ the [CC0 1.0 Universal public domain dedication](https://creativecommons.org/pub
...
+   1. `npm run test:api`
+   1. `npm run test:client`
+   1. `npm run test:shared`
+   1. `npm run test:pa11y`
+
+For non-trivial changes, PR's should be initiated into the `test` branch. Others can be initiated straight into the `staging` branch.
+
+### Active Pull Requests
+
+We will be civil, but we are opinionated about the architecture and coding standards in the project. Please do not be offended if the team requests changes or otherwise provides feedback. We may use [emoji's](https://github.com/erikthedeveloper/code-review-emoji-guide) in our comments.
```
rosiehsmith: I wonder if we can incorporate our existing [code review documentation](https://github.com/ustaxcourt/ef-cms/blob/staging/docs/additional-resources/code-review.md). Some of it is outdated but a lot of it is good content, as it discusses the code review approach and what we look for which I appreciate in its explicitness.

It makes sense to me that you'd put a few sentences here under CONTRIBUTING, but maybe the language could be more like:

> As code reviewers, we provide feedback on pull requests so we can maintain the quality of our code, but do so with respect, kindness, and an open mind. We want code reviews to be a space for discussion and questions, while explaining our reasoning for suggestions.

rosiehsmith:
For this question:
> Can we ensure that outside contributions have had pre-commit hooks run?

I believe the closest we can request is that `npm install` when they start contributing (which I think you covered) because then `husky` will be installed on the `postinstall` step, so their next commit should run the `pre-commit` file from `husky`

CONTRIBUTING.md
rosiehsmith: Probably should have explained my reasoning 😆  - I think it would be helpful to specify expectations for what we provide feedback for and that could even be integrated into how the code might be written in the first place. That's how some of the existing code review docs could be helpful in fleshing that out. The main part that I'm unsure about is the "please do not be offended" because to me it seems like we're putting the responsibility on them without clarifying the expectations enough around coding standards and architecture. I might be missing where the coding standards are further specified, though. Happy to talk about it irl if you want!

jamesobrooks: @rosiehsmith, I'm coming back around to this. I'm definitely open to discussing changing this section. I had thought that at some point we would want to link to other documentation which is why I asked about whether we could get the github pages set up for the court repo to host docsify. Of course we can still link out if that is not going to happen.

jamesobrooks:
> For this question:
> 
> > Can we ensure that outside contributions have had pre-commit hooks run?
> 
> I believe the closest we can request is that `npm install` when they start contributing (which I think you covered) because then `husky` will be installed on the `postinstall` step, so their next commit should run the `pre-commit` file from `husky`

That's kind of my concern. I don't know a way to trigger the hooks if there was a commit with the `--no-verify` flag. Certainly we could do things like copy the code in the changed files, apply it to a clean branch and run a commit without the flag just to see. But I doubt we want to put time into scripting something like that unless for some reason we start getting lots of contributions from outside the team.

jimlerza (approved):

CONTRIBUTING.md
mmarcotte: Sounds good @jamesobrooks . Let's go with that.

CONTRIBUTING.md
mmarcotte: @jamesobrooks We [enabled docsify for the Court repo](https://ustaxcourt.github.io/ef-cms/). ✅

CONTRIBUTING.md
jamesobrooks:
```suggestion
1. Run the following scripts, ensuring no failures:
```

jamesobrooks: @mmarcotte, I updated it.

CONTRIBUTING.md
jamesobrooks: @mmarcotte, thanks! I'll get a change pushed to refer to that documentation.

jimlerza (approved):

PR #2466, opened 2022-08-22
DevEx - refactoring the app context into multiple files

codyseibert:
- the application context was too many lines which made saving the files and editing a pain, splitting up the useCases, useCaseHelpers, and persistenceGateway into separate files.
- include some additional persistence methods inside the validation decorate


web-api/src/getPersistenceGateway.js
```diff
@@ -0,0 +1,651 @@
...
+    createJobStatus,
+    createMessage,
+    createNewPetitionerUser,
+    createNewPractitionerUser,
+    createOrUpdatePractitionerUser,
+    createOrUpdateUser,
+    createTrialSession,
+    createTrialSessionWorkingCopy,
+    deleteKeyCount,
+    fetchPendingItems,
```
rosiehsmith, line 498: this one is also in `getuseCaseHelpers` - is that necessary?

rosiehsmith: by this one i mean `fetchPendingItems`

rosiehsmith (approved):
🎉 nice refactoring

web-api/src/getPersistenceGateway.js
codyseibert: good catch, I think I had some extra persistence files in the wrong file

PR #2468, opened 2022-08-23
Opex 1075 color switch gate

jamesobrooks:
# [OpEx 1075](https://trello.com/c/yDZxWemV/1075-1075-earlier-deployments)

As an engineer, in order to not to stay up late for a deployment and more easily triage a deployment failure during business hours, I need to do a deployment earlier in the week but then do the color-switch automatically at Sunday midnight.

## Solution

We add a new approval job which is only used in the `build-and-deploy-with-context` workflow. This is similar to the one used to wait for a reindex to finish. The notable difference is that we are not waiting for state, rather for time. So the `aws_cloudwatch_event_rule` does not need to start out disabled, nor does it need to run repeatedly to determine state. It will only trigger based on the cron expression described below.

> NOTE:\
> The approval job can be manually triggered at any time.

### `* 5,6 ? * SUN *`

You may note that the `?` seems non-standard. This is an AWS thing and it was required to even successfully create the rule. It is set to run at any time between 5:00 AM and 6:59 AM UTC. It will run every minute as this is the maximum granularity , but should only ever actually trigger once unless there is some problem communicating with CircleCI. The reason for selecting that time is to prevent having to change the cron when time changes twice a year.

> NOTE:\
> The cron currently reflects `MON-SAT` rather than `SUN` for running in the exp environment for development.

## Migration Deploy

https://app.circleci.com/pipelines/github/flexion/ef-cms/37139/workflows/a3b779e1-2b6a-4821-8657-97618df09513

This deploy completed on 24 August at 1:10 AM Eastern after having waited since 2:07 PM Eastern.

## Non-migration Deploy After Rename

I renamed a couple of things so ran this build to ensure that the refactor did not interfere.

https://app.circleci.com/pipelines/github/flexion/ef-cms/37173/workflows/82e4058e-68a5-4b5d-a2d1-193bf116f016


rgoshen (approved):

web-api/switch-colors-cron-terraform/main/main.tf
```diff
@@ -0,0 +1,14 @@
...
+terraform {
+  backend "s3" {
+  }
+
+  required_providers {
+    aws = "3.54.0"
+  }
+}
+
+data "aws_caller_identity" "current" {}
```
jimlerza: Newline please!

jimlerza (approved):

web-api/switch-colors-cron-terraform/main/main.tf
jamesobrooks:
```suggestion
data "aws_caller_identity" "current" {}

```

shumway-tm (approved):

jimlerza (approved):

PR #2472, opened 2022-08-24
9566 Remove character limit in Case Notes 

mrinsin:
image


shared/src/business/useCases/validateCaseDetailInteractor.js
```diff
@@ -6,9 +7,18 @@ const { CaseQC } = require('../entities/cases/CaseQC');
  * @param {object} applicationContext the application context
  * @param {object} providers the providers object
  * @param {object} providers.caseDetail the case data to validate
+ * @param {object} providers.useCaseEntity the flag indicating what kind of entity to use
```
jamesobrooks, line 10: 🤔 I might rename `useCaseEntity`. I thought it was like "YUCE CASE entity", not "YOUS case entity" if that makes any sense. Maybe `shouldUseCaseEntity`?

jamesobrooks (approved):
🌱 I feel like a validation error around a 9000 character limit is probably not very actionable for a user that actually experiences it. If someone pastes in text, it seems a bit awkward to have to figure out where 9000 characters cuts off.

If we have people actually experience this, we might want to find a way to indicate what portion of the note is too much, or maybe offer a way to attach a document for the note.

Anyway, for now this is good, although I do think there's one property that could be renamed to be a bit more clear.

jimlerza (approved):

PR #2473, opened 2022-08-24
DevEx - Reduce trial session data size

codyseibert:
When dealing with very large trial sessions (250+ cases), we have seen the api gateway throw 502 errors.  This is probably due to some cases have a LOT of docket entries, so if you are trying to view the trail session that contains a case with a lot of docket entries, this api endpoint return the case and every entry along with it.  The front end logic only cares about a particular docket entry for this page, so this PR introduces a filter to only return the PMT, non stricken entries in the returned payload to reduce the size and speed up the request.

This is a screenshot of the request for a trial session with 103 cases.  Notice the response size is about 189kb:
Screen Shot 2022-08-11 at 3 16 52 PM

after deploying 44kb
Screen Shot 2022-08-24 at 10 49 45 AM


shared/src/business/useCases/trialSessions/getCalendaredCasesForTrialSessionInteractor.test.js
```diff
@@ -55,4 +55,41 @@ describe('getCalendaredCasesForTrialSessionInteractor', () => {
...
+              ...MOCK_CASE.docketEntries[0],
+              eventCode: 'PMT',
+              isStricken: false,
+            },
+            {
+              ...MOCK_CASE.docketEntries[0],
+              eventCode: 'PMT',
+              isStricken: true,
+            },
+          ],
```
jamesobrooks, line 81: The array here doesn't seem to fit with the test's name. I would have thought to see docket entries with event codes that we expect to be excluded, but instead it seems like we're checking for the stricken one to be excluded even though the event code is right?

jimlerza (approved):

rosiehsmith (approved):

jamesobrooks (approved):

PR #2471, opened 2022-08-23
9497 identify consolidated cases in qc display

rgoshen:
This PR addresses PO Review comments on lead case and consolidated case icons not show up on Individual Docket QC Processed tab when docket clerk uploads a PDF.

PR #2475, opened 2022-08-24

rgoshen:
This PR was created due to [PR#2471](https://github.com/ustaxcourt/ef-cms/pull/2471) contained files not part of the story 9497 or PO feedback.  This PR addresses PO Review comments on lead case and consolidated case icons not showing up on Individual Docket QC Processed tab when docket clerk uploads a PDF.


shared/src/business/entities/OutboxItem.test.js
```diff
@@ -124,6 +124,6 @@ describe('OutboxItem', () => {
     expect(outboxItem.docketEntry.documentType).toEqual(
       'Proposed Stipulated Decision',
     );
-    expect(outboxItem.leadDocketNumber).not.toBeUndefined();
+    expect(outboxItem.leadDocketNumber).toEqual('101-18');
```
jamesobrooks, line 127: 👍

shared/src/business/useCases/courtIssuedDocument/fileAndServeCourtIssuedDocumentInteractor.test.js
```diff
@@ -285,6 +285,40 @@ describe('fileAndServeCourtIssuedDocumentInteractor', () => {
...
+    const caseRecordWithLeadDocketNumber = {
+      ...MOCK_CASE,
+      docketEntries: [mockDocketEntryWithWorkItem],
+      leadDocketNumber,
+    };
+
+    applicationContext
+      .getPersistenceGateway()
+      .getCaseByDocketNumber.mockReturnValueOnce(caseRecordWithLeadDocketNumber)
+      .mockReturnValueOnce(caseRecordWithLeadDocketNumber);
```
jamesobrooks, line 299: ❓ Why do we use mockReturnValueOnce two times with the same return but no default?

shared/src/business/useCases/courtIssuedDocument/fileAndServeCourtIssuedDocumentInteractor.test.js
```diff
@@ -496,16 +530,22 @@ describe('fileAndServeCourtIssuedDocumentInteractor', () => {
...
+      })
+      .mockReturnValueOnce({
         ...caseRecord,
-        petitioners: [
-          {
-            ...caseRecord.petitioners[0],
-            serviceIndicator: 'None',
-          },
-        ],
+        petitioners,
```
jamesobrooks, line 548: ❓ Same question here about mockReturnValueOnce. We are using it twice to return the same thing, and then not providing a default.

shared/src/business/useCases/courtIssuedDocument/fileAndServeCourtIssuedDocumentInteractor.test.js
rgoshen: My understanding is mockReturnValueOnce is needed to be called twice in a row because in the interactor itself getCaseByDocketNumber is invoked once for subjectCase and then invoked a second time in the try catch block.  Since it gets invoked a second time, when the test passes in caseRecordWithLeadDocketNumber, the second invocation causes the leadDocketNumber to become undefined and therefore fails.

shared/src/business/useCases/courtIssuedDocument/fileAndServeCourtIssuedDocumentInteractor.test.js
rgoshen: I am not quite sure about this one.  This did not originally fail when we first started updating this test.  However, when I commented it out, the test failed because it expected the promise to reject and throw an error.  But, the test instead resolved the promise to a value of undefined.

shared/src/business/useCases/courtIssuedDocument/fileAndServeCourtIssuedDocumentInteractor.test.js
jamesobrooks: After talking through this, I believe the reason we are doing this is because the default mock would persist into the later tests and have unintended consequences.

shared/src/business/useCases/courtIssuedDocument/fileAndServeCourtIssuedDocumentInteractor.test.js
jamesobrooks: Same response from above applies:
> After talking through this, I believe the reason we are doing this is because the default mock would persist into the later tests and have unintended consequences.

So this test would potentially have been affected by the previous one, and would in turn potentially affect other tests whether they exist yet or not.

jamesobrooks (approved):

PR #2483, opened 2022-08-26
DevEx 998: Complete transition to Luxon from Moment

maddenf:
Refactored startOfWeek code to use a method defined in DateHandler.js called createDateAtStartOfWeekEST().

Also removed hard coded date formats and replaced with constants defined in DateHandler.js.

Note:  'DDD' which was originally being used hard coded, should be identical to 'MMMM d, yyyy' defined as MONTH_DAY_YEAR in DateHandler.js.  For this reason I just replaced 'DDD' with that constant.


mrinsin (approved):
🎉 🎉 🎉

jimlerza (approved):

mmarcotte (approved):

PR #2500, opened 2022-08-31
DevEx - Refactor Interactors to use TypeScript 

codyseibert:
refactoring our interactors to be in typescript


jamesobrooks (approved):
I did some limited testing in exp3, creating cases, orders, notices, a trial session, even consolidating cases and changing proceeding type. All seemed to work fine.

mmarcotte (approved):

mmarcotte:
@codyseibert -- so it really looks the main effort was to refactor into typescript in order to get the parameters of each of the interactors into a typed structure. Everything else pretty much remained the same?

jimlerza (approved):

codyseibert:
> @codyseibert -- so it really looks the main effort was to refactor into typescript in order to get the parameters of each of the interactors into a typed structure. Everything else pretty much remained the same?

yes, the goal was to start with interactors first, then move into the persistence layer.  Right now, just the interactors by themself don't give much value since most of the functions we call in the business domain are interactors -> persistence.  So we will see a lot more value when we get to the persistence layer.  We decided to do a layer at a time so the PR wasn't too big.

rosiehsmith (approved):

PR #2511, opened 2022-09-02
9564 story

mrinsin:
### With filter "Exhibits":

image

### With filter "All documents" (default):

image


web-client/src/styles/overrides.scss
```diff
@@ -11,6 +11,12 @@ option:first-child {
   font-weight: initial;
 }
 
+#inline-select {
```
TomElliottFlexion, line 14: Nit: Does this need to be in overrides? Is there a more specific stylesheet that could be used?

TomElliottFlexion: Actually I can't find a better place either

TomElliottFlexion (approved):

web-client/src/styles/overrides.scss
mrinsin: We added it to overrides because these styles technically 'override' the USWDS styling in the `select` element! Is there somewhere that would make more sense? Happy to move it!

mrinsin: Oops I didnt see your second comment until now lol.. ok np! TY for reviewing!

PR #2524, opened 2022-09-06
9559 UX Feedback: Remove asterisk appended to 'Other' in trial clerk dropdown

TomElliottFlexion:
"Other" no longer has a an asterisk:

Screen Shot 2022-09-06 at 2 50 27 PM

### DIFF
--- a/shared/src/business/entities/trialSessions/NewTrialSession.js
+++ b/shared/src/business/entities/trialSessions/NewTrialSession.js
@@ -27,7 +27,7 @@ NewTrialSession.prototype.init = function init(
 NewTrialSession.VALIDATION_ERROR_MESSAGES = {
   ...TrialSession.VALIDATION_ERROR_MESSAGES,
   alternateTrialClerkName:
-    'A valid Alternate Trial Clerk name must be provided if "Other" is selected',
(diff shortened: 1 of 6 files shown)


mrinsin (approved):

shared/src/business/entities/trialSessions/NewTrialSession.js
```diff
@@ -27,7 +27,7 @@ NewTrialSession.prototype.init = function init(
 NewTrialSession.VALIDATION_ERROR_MESSAGES = {
   ...TrialSession.VALIDATION_ERROR_MESSAGES,
   alternateTrialClerkName:
-    'A valid Alternate Trial Clerk name must be provided if "Other*" is selected',
+    'A valid Alternate Trial Clerk name must be provided if "Other" is selected',
```
rosiehsmith: This probably doesn't matter, but should the validation message's "Alternate Trial Clerk" be lowercase like in `SessionAssignmentsForm`?

rosiehsmith (approved):

shared/src/business/entities/trialSessions/NewTrialSession.js
TomElliottFlexion: Good catch, thank you! That would be better

PR #2521, opened 2022-09-06
9497 identify consolidated cases in qc display fixed

rgoshen:
This is a resubmit of [2513](https://github.com/ustaxcourt/ef-cms/pull/2513) and [2514](https://github.com/ustaxcourt/ef-cms/pull/2514).  They were closed due too many files not associated with the story [9497](https://github.com/flexion/ef-cms/issues/9497).  The original branch inadvertently back merged testing into the story branch.  This branch was reverted to a commit prior to the back merge.

PR #2525, opened 2022-09-06
9497 identify consolidated cases in qc display test

rgoshen:
This PR replaces [2521](https://github.com/ustaxcourt/ef-cms/pull/2521).  The readme file is the only file that was not altered during this story.


shared/src/business/useCases/processStreamRecords/processWorkItemEntries.js
```diff
@@ -1,8 +1,61 @@
...
+  applicationContext,
+  workItemRecords,
+}) => {
+  if (!workItemRecords.length) return;
+
+  applicationContext.logger.debug(
+    `going to index ${workItemRecords.length} work item records`,
+  );
+
+  const indexWorkItemEntry = async workItemRecord => {
```
mmarcotte: Looks like we can remove the `async` here as the function itself is synchronous.

shared/src/business/useCases/processStreamRecords/processWorkItemEntries.js
```diff
@@ -1,8 +1,61 @@
...
+        NewImage: {
+          ...workItemNewImage,
+          ...caseWorkItemMappingRecord,
+        },
+      },
+      eventName: 'MODIFY',
+    };
+  };
+
+  const indexRecords = await Promise.all(
```
mmarcotte: ... and this probably isn't a Promise.all since `indexWorkItemEntry` is synchronous.

shared/src/business/useCases/courtIssuedDocument/serveCourtIssuedDocumentInteractor.js
```diff
@@ -54,7 +57,7 @@ const completeWorkItem = async ({
 
   await applicationContext.getPersistenceGateway().putWorkItemInUsersOutbox({
     applicationContext,
-    section: user.section,
+    section: user.section ? user.section : DOCKET_SECTION,
```
jimlerza: I'm curious about the real-life use case that required this change. Was this required for tests? I'd think any court user with access to serve court-issued documents would have a section.

shared/src/business/useCases/processStreamRecords/processWorkItemEntries.js
jimlerza: It appears this doesn't need to be asynchronous.

rgoshen: You are absolutely correct, and I have removed it.  Thanks for catching it.

shared/src/business/useCases/courtIssuedDocument/serveCourtIssuedDocumentInteractor.js
rgoshen: This fixes a bug when pdf is uploaded and save entry without serving and then later served by clicking on the link within docket record.  Without the ternary, when you serve it from the docket record, it shows up under the docket clerk's Document QC Processed tab, but does not show up under the Section Document QC Processed tab.  This will fix that issue.

rgoshen: fixed petitionsclerk issue and no longer need this line, reverted back to original line of just using user.section only

shared/src/business/useCases/processStreamRecords/processWorkItemEntries.ts
```diff
@@ -1,14 +1,64 @@
...
+          ...caseWorkItemMappingRecord,
+        },
+      },
+      eventName: 'MODIFY',
+    };
+  };
+
+  const indexRecords = await Promise.all(
+    workItemRecords.map(indexWorkItemEntry),
+  );
```
mmarcotte: indexWorkItemEntry is a synchronous function, and I don't think we need to use `await Promise.all( ... )`

shared/src/business/useCases/processStreamRecords/processWorkItemEntries.js
rgoshen: this file should have been deleted

mmarcotte: yep; i had started this review a week ago 😅

shared/src/business/useCases/processStreamRecords/processWorkItemEntries.ts
rgoshen: good catch...i updated to remove the promise.all()

mmarcotte: thanks rick!

mmarcotte (approved):

PR #2526, opened 2022-09-07
2379 - adding some ux feedback to fix text and icons

codyseibert:
- show different icon depending on if the added docket numbers was set or not
- changing the modal text to be static

### DIFF
--- a/web-client/src/presenter/computeds/createOrderHelper.js
+++ b/web-client/src/presenter/computeds/createOrderHelper.js
@@ -22,9 +22,10 @@ export const createOrderHelper = (get, applicationContext) => {
   const isLeadCase = caseDetail.leadDocketNumber === caseDetail.docketNumber;
 
   return {
+    addDocketNumbersButtonIcon: addedDocketNumbers ? 'edit' : 'plus-circle',
     addDocketNumbersButtonText: addedDocketNumbers
-      ? 'Edit docket numbers to the coversheet'
-      : 'Add docket numbers to the coversheet',
+      ? 'Edit docket numbers in the cover sheet'
+      : 'Add docket numbers to the cover sheet',
     documentToEdit,
     isEditing,
     pageTitle,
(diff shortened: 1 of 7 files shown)


web-client/src/presenter/computeds/createOrderHelper.js
```diff
@@ -22,9 +22,10 @@ export const createOrderHelper = (get, applicationContext) => {
   const isLeadCase = caseDetail.leadDocketNumber === caseDetail.docketNumber;
 
   return {
+    addDocketNumbersButtonIcon: addedDocketNumbers ? 'edit' : 'plus-circle',
     addDocketNumbersButtonText: addedDocketNumbers
-      ? 'Edit docket numbers to the coversheet'
-      : 'Add docket numbers to the coversheet',
+      ? 'Edit docket numbers in the cover sheet'
```
TomElliottFlexion, line 27: Nit: 
```suggestion
      ? 'Edit docket numbers on the cover sheet'
```

web-client/src/presenter/computeds/createOrderHelper.test.js
```diff
@@ -87,7 +88,8 @@ describe('createOrderHelper', () => {
     });
 
     expect(result.addDocketNumbersButtonText).toEqual(
-      'Edit docket numbers to the coversheet',
+      'Edit docket numbers in the cover sheet',
```
TomElliottFlexion, line 91:
```suggestion
      'Edit docket numbers on the cover sheet',
```

TomElliottFlexion (approved):

web-client/src/presenter/computeds/createOrderHelper.js
codyseibert: ux asked for `to` and `in` specifically

web-client/src/presenter/computeds/createOrderHelper.test.js
codyseibert: same as above, ux asked for `to` and `in` specifically

web-client/src/presenter/computeds/createOrderHelper.js
TomElliottFlexion: Obviously I'll defer, but `on` would be correct

PR #2530, opened 2022-09-09
9497 identify consolidated cases in qc display test

rgoshen:


rgoshen:
too many files will close

PR #2532, opened 2022-09-09
9497 identify consolidated cases in qc display to test

rgoshen:
This PR fixes the issue with logging in as a petitions clerk, serving a court issued document (upload a PDF) and having it appear under the correct section Document QC outbox.


rgoshen:
try this one more time it has 9 more files than it should

PR #2540, opened 2022-09-12
9497 identify consolidated cases in qc display to test

rgoshen:
This PR fixes the issue with logging in as a petitions clerk, serving a court issued document (upload a PDF) and having it appear under the correct section Document QC outbox.  In serveCourtIssuedDocumentInteractor.js, changed how user was being called.  The old way got current user from application context, however, this user entity did not have a section attribute.  So, once the interactor got the current user, then used the user id to make a call to DynamoDB to get the whole entity to include the user's section.


jimlerza (approved):

rgoshen:
closing due to merging into wrong branch

PR #2520, opened 2022-09-06
9559: alternate trial clerk

TomElliottFlexion:
[PR to test](https://github.com/ustaxcourt/ef-cms/pull/2512)
image

PR #2512, opened 2022-09-02
9559 alternate trial clerk

rgoshen:
This PR addresses [Story 9559](https://github.com/flexion/ef-cms/issues/9559). An alternate trial clerk field has been added when "Other" is selected in the Trial Clerk drop down for Session Assignments in Trial Sessions.

Screen Shot 2022-09-02 at 1 22 56 PM

PR #2544, opened 2022-09-13
9559 alternate trial clerk

rgoshen:
[PR to staging](https://github.com/ustaxcourt/ef-cms/pull/2512)
image

Resubmission from [PR2520](https://github.com/ustaxcourt/ef-cms/pull/2520)


mrinsin (approved):

shared/src/business/utilities/getFormattedTrialSessionDetails.setPretrialMemorandumFilter.test.js
```diff
@@ -0,0 +1,174 @@
...
+      userId: '7805d1ab-18d0-43ec-bafb-654e83405416',
+    };
+
+    let mockCase;
+
+    beforeEach(() => {
+      applicationContext
+        .getPersistenceGateway()
+        .getCaseByDocketNumber.mockImplementation(() => mockCase);
+    });
```
mmarcotte, line 34: I'm not sure you're getting much out of this `beforeEach`. The first time it fires before the first test, `mockCase` is still undefined. The second time it fires before the second test, `mockCase` is what the first test had defined, etc.

Plus, I don't think that this Persistence Gateway method is invoked in the function you are testing. Seems safe to remove.

mmarcotte (changes requested):

shared/src/business/utilities/getFormattedTrialSessionDetails.setPretrialMemorandumFilter.test.js
rgoshen: removed

mmarcotte (approved):
thanks @rgoshen !

PR #2557, opened 2022-09-15
9493: File paper document on lead case in consolidated group

rachelschneiderman:
This PR covers the following tasks for this story:
- Add paper filing docket entry to non-lead cases in consolidated group, save for later and THEN serve
- Add paper filing docket entry to non-lead cases in consolidated group, save and serve
- Hide checkboxes for filing docket entry on other cases in group if the paper filing is a case decision document type

There is a lot of refactoring, testing, and beautifying left to do but with 39 files changed we figured it was time for a PR :)


web-client/terraform/main/setup_dynamsoft.sh
```diff
@@ -26,4 +26,4 @@ sudo sed -i "s|DYNAMSOFT_RESOURCE_PATH|${dynamsoft_url}/Dynamic%20Web%20TWAIN%20
 sudo sed -i "s|DYNAMSOFT_PRODUCT_KEYS|${dynamsoft_product_keys}|" /tmp/dynamsoft.webtwain.config.js.tpl
 
 # copy back from /tmp to main nginx html folder
-sudo cp /tmp/dynamsoft.webtwain.config.js.tpl Dynamic\ Web\ TWAIN\ SDK\ 17.2.5/Resources/dynamsoft.webtwain.config.js
\ No newline at end of file
+sudo cp /tmp/dynamsoft.webtwain.config.js.tpl Dynamic\ Web\ TWAIN\ SDK\ 17.2.5/Resources/dynamsoft.webtwain.config.js
```
codyseibert, line 29: was a new line character removed from this .sh script?  I think all .sh scripts should end with a blank line

web-client/integration-tests/docketRecordJourney.test.js
```diff
@@ -507,6 +507,8 @@ describe('Docket Clerk Verifies Docket Record Display', () => {
         }),
       ]),
     );
+
+    console.log('the docketNumber', cerebralTest.docketNumber);
```
codyseibert: console log left over

shared/src/business/useCases/docketEntry/addPaperFilingInteractor.ts
```diff
@@ -75,85 +98,126 @@ export const addPaperFilingInteractor = async (
...
+        docketNumber: docketNo,
+      });
+
+    let aCaseEntity = new Case(aCase, { applicationContext });
+    caseEntities.push(aCaseEntity);
+  }
+
+  let filedByFromLeadCase;
+  //why raw case? who knows. couldnt think of a name
+  for (const rawCase of caseEntities) {
```
codyseibert: maybe just call it caseEntity instead of rawCase since it's an actual entity at this point I think

shared/src/business/useCases/docketEntry/addPaperFilingInteractor.ts
```diff
@@ -75,85 +98,126 @@ export const addPaperFilingInteractor = async (
...
+        userId: user.userId,
       },
-      docketNumber: caseToUpdate.docketNumber,
-      docketNumberWithSuffix: caseToUpdate.docketNumberWithSuffix,
-      inProgress: isSavingForLater,
-      isRead: user.role !== ROLES.privatePractitioner,
+      { applicationContext, petitioners: rawCase.petitioners },
+    );
+
+    if (rawCase.docketNumber === rawCase.leadDocketNumber) {
```
codyseibert: do you think we should just have a case entity method called `isLeadCase()` which does this logic?

shared/src/business/useCases/document/serveExternallyFiledDocumentInteractor.ts
```diff
@@ -1,149 +1,325 @@
...
-  }: { docketEntryId: string; docketNumber: string },
+    docketNumbers,
+    subjectCaseDocketNumber,
+  }: {
+    docketEntryId: string;
+    docketNumbers: string[];
+    subjectCaseDocketNumber: string;
+  },
 ) => {
+  console.log('serveExternallyFiledDocumentInteractor');
```
codyseibert: extra log

shared/src/business/useCases/docketEntry/addPaperFilingInteractor.ts
mrinsin: Good idea! You okay with us adding that as a DOD refactoring item, since we still have some stuff to do for this story?

rosiehsmith (approved):

PR #2559, opened 2022-09-19
DevEx - Fix issue with TS husky staged file checker

codyseibert:
- the husky pre-commit was failing on IApplicationContext not found, this pr fixes that


.eslintrc.js
```diff
@@ -47,6 +47,12 @@ module.exports = {
         'jest/expect-expect': 'off',
       },
     },
+    {
+      files: ['**/*.ts', '**/*.tsx'],
+      rules: {
+        'no-undef': 'off',
```
mrinsin, line 53: Is there an alternative fix for this than turning off the lint rule (or a devex task to address this later)? It almost seems counterintuitive to use TS but turn off this rule!!

mrinsin (approved):

.eslintrc.js
codyseibert: Screen Shot 2022-09-19 at 5 23 34 PM

so typescript will actually give a different warning if you try to invoke a function or use something that was not defined, so we don't even need eslint to catch these types of errors since typescript will find it for us.

mrinsin: Ooh nice!! TY for showing me

jimlerza (approved):

PR #2561, opened 2022-09-20
Local ARM64 support

jasonfu9966:
Updating start-dynamo and start-elasticsearch scripts to support M1 Macs. Also add a clause in readme about Airplay / API port conflict.


codyseibert (approved):

docs/running-locally.md
```diff
@@ -84,6 +84,11 @@ All of the applications dependencies are managed via our `package.json` and `pac
 
 `npm install`
 
+### 💻 MacOS Monterey+
```
rachelschneiderman, line 87: Great update! Thanks!

rachelschneiderman (commented):
I am a little concerned about modifying startup scripts for specific hardware. Has it been proved this is an issue with the machine and not an issue with a dependency like Java?

jasonfu9966:
> I am a little concerned about modifying startup scripts for specific hardware. Has it been proved this is an issue with the machine and not an issue with a dependency like Java?

Good question! After double-checking the staging branch, I have confirmed that both dynamo and elasticsearch both fail on api load. Strangely, dynamo does not fail to start up when the start-dynamo script is started on its own, but does fail when run in context. This change fixes that.

For the elasticsearch changes, Rick and Cody both ran into issues on their M1 Macs and this change fixes that as well.

Would it help if I included a clause on the documentation about how I fixed the Java problems?

jimlerza (approved):

rachelschneiderman:
> > I am a little concerned about modifying startup scripts for specific hardware. Has it been proved this is an issue with the machine and not an issue with a dependency like Java?
> 
> Good question! After double-checking the staging branch, I have confirmed that both dynamo and elasticsearch both fail on api load. Strangely, dynamo does not fail to start up when the start-dynamo script is started on its own, but does fail when run in context. This change fixes that.
> 
> For the elasticsearch changes, Rick and Cody both ran into issues on their M1 Macs and this change fixes that as well.
> 
> Would it help if I included a clause on the documentation about how I fixed the Java problems?

Got it! Thanks! I think it would be good to document whatever fixes you had to do, thank you!

PR #2562, opened 2022-09-20
9592 hybrid case filter

rgoshen:


web-client/src/presenter/computeds/formattedTrialSessionDetails.js
```diff
@@ -24,6 +24,13 @@ export const formattedTrialSessionDetails = (get, applicationContext) => {
         .formatPhoneNumber(formattedTrialSession.chambersPhoneNumber);
     }
 
+    formattedTrialSession.isHybridSession =
+      formattedTrialSession.sessionType === 'Hybrid';
```
rachelschneiderman: 🔧 Could the entity constant `SESSION_TYPES` be used here instead of hardcoding?

web-client/src/presenter/computeds/formattedTrialSessionDetails.js
```diff
@@ -45,6 +52,10 @@ export const formattedTrialSessionDetails = (get, applicationContext) => {
       const hasNoActiveCases =
         isEmpty(allCases) || isEqual(allCases, inactiveCases);
 
+      if (formattedTrialSession.formattedEligibleCases.length === 0) {
```
rachelschneiderman: 🤔 I'm wondering if it would be easier to tell when/if the filter was disabled if this check and the check on line 30 were combined?

web-client/src/presenter/computeds/formattedTrialSessionDetailsForFilteredEligibleCases.js
```diff
@@ -0,0 +1,31 @@
+import { formattedTrialSessionDetails } from './formattedTrialSessionDetails';
+import { state } from 'cerebral';
+
+export const formattedTrialSessionDetailsForFilteredEligibleCases = (
+  get,
+  applicationContext,
+) => {
+  const trialSessionDetails = formattedTrialSessionDetails(
```
rachelschneiderman, line 9: 🔧 Instead of re-computing this value, I believe it can be retrieved off of state!

web-client/src/presenter/computeds/formattedTrialSessionDetails.js
rgoshen: Good catch...I thought I had used EntityConstants every where I could.  Thanks.

web-client/src/presenter/computeds/formattedTrialSessionDetailsForFilteredEligibleCases.test.js
```diff
@@ -0,0 +1,145 @@
...
+          screenMetadata: {
+            eligibleCasesFilter: {
+              hybridSessionFilter: undefined,
+            },
+          },
+        },
+      },
+    );
+
+    expect(result.formattedEligibleCases).toHaveLength(4);
```
rachelschneiderman: ⛏️ Could the expected value be the length of the mock data being returned, in case it's ever changed?

web-client/src/views/TrialSessionDetail/TrialSessionDetail.jsx
```diff
@@ -59,6 +60,30 @@ export const TrialSessionDetail = connect(
                 tabName="EligibleCases"
                 title="Eligible Cases"
               >
+                {formattedTrialSessionDetails.isHybridSession && (
+                  <div className="grid-row grid-col-9">
+                    <div className="grid-col-1 padding-top-05 margin-right-3">
+                      <h3 id="filterHybridSessionHeading">Filter by</h3>
```
rachelschneiderman: 🔧 To make the filters more accessible, I believe we are now making the text `<label>` elements so they can be read by screenreaders. We did something similar for another story in `web-client/src/views/DocketRecord/DocketRecordHeader.jsx`, line 71.

web-client/src/presenter/computeds/formattedTrialSessionDetailsForFilteredEligibleCases.js
```diff
@@ -0,0 +1,31 @@
...
+  const filter = get(
+    state.screenMetadata.eligibleCasesFilter.hybridSessionFilter,
+  );
+
+  trialSessionDetails.formattedEligibleCases =
+    trialSessionDetails.formattedEligibleCases.filter(eligibleCase => {
+      const allSmallCases = /^(S+)/;
+      const regularCasesOnly = /^(?!S+)/;
+
+      if (filter === 'Small') {
```
rachelschneiderman, line 19: 🔧 Is there an entity constant that could be used here?

web-client/integration-tests/filterAHybridTrialSession.test.js
```diff
@@ -0,0 +1,90 @@
...
+import { docketClerkViewsNewTrialSession } from './journey/docketClerkViewsNewTrialSession';
+import { docketClerkViewsTrialSessionList } from './journey/docketClerkViewsTrialSessionList';
+import { loginAs, setupTest, uploadPetition } from './helpers';
+import { petitionsClerkSubmitsCaseToIrs } from './journey/petitionsClerkSubmitsCaseToIrs';
+import { petitionsClerkViewsAHybridTrialSessionFilteredEligibleCases } from './journey/petitionsClerkViewsAHybridTrialSessionsFilteredEligibleCases';
+
+const cerebralTest = setupTest();
+const { CASE_TYPES_MAP } = applicationContext.getConstants();
+
+// create a trial session
```
rachelschneiderman: 🤔 Are these comments still needed?

web-client/src/presenter/computeds/formattedTrialSessionDetails.js
rgoshen: Thanks...done

rachelschneiderman (commented):
Nice job! Everything is incredibly well tested and the PR is a great size, it was easy to review :)

web-client/src/views/TrialSessionDetail/TrialSessionDetail.jsx
rgoshen: Thanks for pointing this out.  I had this as a label at one point for accessibility.  I must have accidentally reverted it back at some point.

web-client/src/presenter/computeds/formattedTrialSessionDetailsForFilteredEligibleCases.js
rgoshen: Thanks... I did not know there were.  This would have made my life simpler and more productive, since I have not worked much with REGEX.  But it was a good learning experience.

web-client/integration-tests/filterAHybridTrialSession.test.js
rgoshen: nope thanks

web-client/src/presenter/computeds/formattedTrialSessionDetailsForFilteredEligibleCases.test.js
rgoshen: done...thanks

web-client/src/presenter/computeds/formattedTrialSessionDetailsForFilteredEligibleCases.js
rgoshen: The thought process was to not alter the original computed (formattedTrialSessionDetails) since it is used elsewhere and therefore may break other portions of the app.  So, the filter was split out into its own computed and calls the formattedTrialSessionDetails because the filter is only needed on this view.  If you just pull it off of state, the filter no longer works.  I am not sure how else to do this.  I know this is a little weird, calling one computed from another.

web-client/src/presenter/computeds/formattedTrialSessionDetails.js
```diff
@@ -24,6 +25,12 @@ export const formattedTrialSessionDetails = (get, applicationContext) => {
         .formatPhoneNumber(formattedTrialSession.chambersPhoneNumber);
     }
 
+    formattedTrialSession.isHybridSession =
+      formattedTrialSession.sessionType === SESSION_TYPES.hybrid;
+
+    formattedTrialSession.disableHybridFilter =
```
rachelschneiderman, line 31: I think this is missing the check for if the sessionType is hybrid?

rgoshen: In the unit test lines 59-71 test if the session type is set to Regular and lines 73-85 test if session type is set to Hybrid.

rachelschneiderman: You're totally right, I was confusing the variables! Apologies

rachelschneiderman (approved):

PR #2572, opened 2022-09-22
9493 UX Feedback and DOD part 1

rosiehsmith:
Includes:
- DOD work to make an `isLeadCase` method off the `Case` entity https://github.com/flexion/ef-cms/pull/9624
- Address UX Feedback that event codes Motion and Notice are multi-docketable and display properly on the coversheet. Using these event codes matches how the paper filing document type dropdown is populated via internal rather than external event codes, while still excluding 5 event codes for case decisions.

image


shared/src/business/entities/EntityConstants.js
```diff
@@ -432,15 +432,15 @@ const TRACKED_DOCUMENT_TYPES = {
...
-  ...Object.values(DOCUMENT_EXTERNAL_CATEGORIES_MAP),
+  ...Object.values(DOCUMENT_INTERNAL_CATEGORIES_MAP),
 ])
-  .filter(externalEvent => externalEvent.caseDecision)
+  .filter(internalEvent => internalEvent.caseDecision)
   .map(x => x.eventCode);
 
-const MULTI_DOCKET_EXTERNAL_FILING_EVENT_CODES = flatten([
-  ...Object.values(DOCUMENT_EXTERNAL_CATEGORIES_MAP),
+const MULTI_DOCKET_FILING_EVENT_CODES = flatten([
```
rachelschneiderman, line 440: Nice rename 👏

rachelschneiderman (approved):

PR #2582, opened 2022-09-26
Dependency updates

jamesobrooks:
# Dependency Updates

### react-number-format

NumberFormat was deprecated and NumericFormat was the proper replacement for our single use. 
[Migration guide](https://s-yadav.github.io/react-number-format/docs/migration/)

### stylelint-config-idiomatic-order

Some styles have changed their idiomatic order.

## Minor
@aws-sdk/credential-provider-node
@aws-sdk/node-http-handler
@aws-sdk/protocol-http
@aws-sdk/signature-v4
@typescript-eslint/eslint-plugin
@typescript-eslint/parser
aws-sdk
core-js
esbuild-sass-plugin
eslint
luxon
ts-jest

# Migration Build

Here is a build with a migration:
[build-and-deploy](https://app.circleci.com/pipelines/github/flexion/ef-cms/37920/workflows/83ec9fd1-7c07-4a34-b004-0432e695a403)

# GitHub Actions

Adds the workflow file name to the name so that it is displayed on the `Actions` page.

Also sets up the workflows to run on push to a branch named `dependency-updates`. If this is undesirable, I can revert that change.


.github/workflows/api.yml
```diff
@@ -1,6 +1,10 @@
-name: Node.js CI
+name: Node.js CI - api
 
-on: [pull_request]
+on:
+  pull_request:
+  push:
+    branches:
+      - dependency-updates
```
rachelschneiderman, line 7: So if we push to a branch called `dependency-updates` this action will run?

jamesobrooks: > So if we push to a branch called `dependency-updates` this action will run?

Yes. I can certainly take it out, but I found it useful to run all the pre-deploy tests on each major change separately.

ujahio (approved):

.github/workflows/api.yml
jimlerza: I like it!

jimlerza (approved):

PR #2595, opened 2022-09-29
9511 printable trial session copy

ujahio:


shared/src/proxies/trialSessions/generatePrintableTrialSessionCopyReportProxy.js
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * generatePrintableTrialSessionCopyReportInteractorProxy
+ *
+ * @param {object} applicationContext the application context
+ * @param {object} providers the providers object
+ * @param {string} providers.docketNumber the optional trialSessionId filter
+ * @returns {Promise<*>} the promise of the api call
+ */
+exports.generatePrintableTrialSessionCopyReportInteractor = (
```
jimlerza, line 11: Nit: missing "Interactor" in the filename. Should be `generatePrintableTrialSessionCopyReportInteractorProxy.js`

jimlerza (approved):

shared/src/proxies/trialSessions/generatePrintableTrialSessionCopyReportProxy.js
ujahio: @jimlerza For the proxies's file name convention, we don't, and haven't been adding, `interactor` to the name of the file. The function name has the word `interactor` in it, just not the file name.

mmarcotte (approved):

PR #2600, opened 2022-09-29
9511: Printable Trial Session Copy - Intermediate to Test

shumway-tm:
# [9511: Printable Trial Session Copy](https://github.com/flexion/ef-cms/issues/9511)

## New PDF Preview Page

> NOTE:
> The content of the new pdf is defined to be a printable, black-and-white, copy of the same data displayed on the non-printable working copy page. Any future changes to the non-printable working copy page should be made with this in mind.

## Image Compare Changes

Since the new pdf has the option to include or not include case notes, we needed to verify two versions of the pdf. Additionally, the header for pages 2+ can grow to be quite large, and if for some reason the alignment or content changes, it can push the page number down into the display area for the content, so we needed to verify both the first and second pages. Changes to the image compare files allow for a developer to run the commands listed in `docs/testing.md` one time to generate multiple files and specify a filename and page number for verification. Previously only the first page would ever be checked, and we would error out after creating a single image, even if we were trying to create multiple new pdf's.

> NOTE:
> There seems to be some difference in how different machines react to the change to `Dockerfile-pdf-testing`. On my machine (James) running the build from that file unchanged stopped working and I had to update `deb10u5` to `deb10u6`, but on another machine `deb10u6` does not work.


image-compare-pdfs.js
```diff
@@ -7,57 +7,86 @@ const { PNG } = require('pngjs');
...
+  {
+    fileName: 'Printable_Trial_Session_Working_Copy_Without_Case_Notes.pdf',
+    pageNumber: 2,
+  },
+  { fileName: 'Receipt_of_Filing.pdf', pageNumber: 1 },
+  { fileName: 'Standing_Pretrial_Order_For_Small_Case.pdf', pageNumber: 1 },
+  { fileName: 'Standing_Pretrial_Order.pdf', pageNumber: 1 },
+  { fileName: 'Trial_Calendar.pdf', pageNumber: 1 },
+  { fileName: 'Trial_Session_Planning_Report.pdf', pageNumber: 1 },
 ];
```
rosiehsmith, line 55: 💭 instead of altering `pdfs` from an array of strings to an object where only 2 exceptions have a `pageNumber` of 2, what if you had a default `pageNumber` of 1, unless the pdf is either 'Printable_Trial_Session_Working_Copy_With_Case_Notes' or 'Printable_Trial_Session_Working_Copy_Without_Case_Notes'? that could be within the loop itself. just a thought, let me know if you wanna chat about it.

shared/src/business/utilities/DateHandler.js
```diff
@@ -13,6 +13,7 @@ const FORMATS = {
   MMDDYYYY_DASHED: 'MM-dd-yyyy',
   MONTH_DAY_YEAR: 'MMMM d, yyyy',
   MONTH_DAY_YEAR_WITH_DAY_OF_WEEK: 'DDDD',
+  SHORT_MONTH_DAY_YEAR: 'MMM d, yyyy',
```
rosiehsmith, line 16: what's the reason for adding this? is this in the AC or design? maybe i missed it somewhere

ujahio: @rosiehsmith 
This was added to save spaces in the pdf header. We found out there wasn't enough space for the combination of the longest edge cases like a spelled out city (e.g Washington, District of Columbia); spelled out date (December 25th, 2022); Judges name (Gustafson) and page number. 

After a conversation with the UI/UX team, the best decision then was to have a short version of the date to accommodate for the long edge cases.

rosiehsmith: ah i see, thanks josh. seems like there might be a benefit to having that documented somewhere, whatcha think @liadmccabe?

web-client/src/presenter/sequences/openPrintableTrialSessionWorkingCopyModalSequence.js
```diff
@@ -0,0 +1,13 @@
+import { clearModalStateAction } from '../actions/clearModalStateAction';
+import { setShowModalFactoryAction } from '../actions/setShowModalFactoryAction';
+import { state } from 'cerebral';
+
+const setCaseNotesFlagDefaultAction = ({ store }) => {
+  store.set(state.modal['caseNotesFlag'], true);
+};
```
rosiehsmith: can you split this action out into its own file and add a test please?

web-client/src/views/TrialSessionWorkingCopy/PrintableTrialSessionWorkingCopyModal.jsx
```diff
@@ -0,0 +1,92 @@
...
+                  htmlFor="caseNotesIncluded-yes"
+                  id="case-notes-included-yes"
+                >
+                  Yes
+                </label>
+              </div>
+              <div className="usa-radio usa-radio__inline">
+                <input
+                  aria-describedby="irs-verified-notice-radios"
+                  checked={caseNotesFlag === false}
```
rosiehsmith: could this just be `!caseNotesFlag`?

web-client/src/presenter/actions/TrialSessionWorkingCopy/preparePrintableFormattedCasesAction.js
```diff
@@ -0,0 +1,23 @@
...
+    temporaryFormattedCases.push(formattedCase);
+    if (formattedCase.leadCase) {
+      temporaryFormattedCases = temporaryFormattedCases.concat(
+        formattedCase.consolidatedCases,
+      );
+    }
+  });
+
+  formattedCases = temporaryFormattedCases;
+  return { formattedCases };
```
rosiehsmith, line 22: would something like this make sense? maybe it's only more readable to me. trying to figure out a way to not use a temp variable but i guess that could be ok. the following assumes that we'd only expect `formattedCases` to contain one lead case. and you'd have to change the test's expectation to include `expect.arrayContaining`...food for thought i guess!

```suggestion
    let { formattedCases } = props;

  const leadCaseConsolidatedCases = formattedCases.filter(
    formattedCase => formattedCase.leadCase,
  )[0].consolidatedCases;

  leadCaseConsolidatedCases.forEach(consolidatedCase => {
    formattedCases.push(consolidatedCase);
  });

  return { formattedCases };
```

web-client/src/presenter/actions/TrialSessionWorkingCopy/getTrialSessionWorkingCopyCaseNotesFlagAction.js
```diff
@@ -0,0 +1,13 @@
...
+/**
+ * getTrialSessionWorkingCopyCaseNotesFlagAction
+ *
+ * @param {object} providers.get the cerebral get function
+ * @returns {Object} caseNotesFlag
+ */
+
+export const getTrialSessionWorkingCopyCaseNotesFlagAction = ({ get }) => {
+  const caseNotesFlag = get(state.modal.caseNotesFlag);
+  return { caseNotesFlag };
```
rosiehsmith: Nit: is there a more accurate, descriptive name for this? looks like it's being used for showing case notes on the printable session copy so like, `includeCaseNotes` or `showCaseNotes` maybe?

web-client/src/presenter/actions/TrialSessionWorkingCopy/getFormattedTrialSessionCasesAction.js
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * getFormattedTrialSessionCasesAction
+ *
+ * @param {object} providers.get the cerebral get function
+ * @returns {Object} formattedCases
+ */
+
+export const getFormattedTrialSessionCasesAction = ({ get }) => {
+  const formattedCases =
+    get(state.trialSessionWorkingCopyHelper).formattedCases || [];
```
rosiehsmith, line 12: Does this need to be defaulted to [] since the helper also seems to be defaulting it to []? i might be wrong, not sure

web-client/src/presenter/actions/PendingItems/setTitleForGlobalReportAction.js
```diff
@@ -3,10 +3,17 @@ import { state } from 'cerebral';
...
+ * @param {string} headerTitle the name of the header title
+ * @returns {Promise} async action
  */
-export const setTitleForGlobalReportAction = ({ store }) => {
-  store.set(state.screenMetadata.headerTitle, 'Pending Report');
-};
+export const setTitleForGlobalReportAction =
+  headerTitle =>
+  /**
+   * sets the state.screenMetaDate.headerTitle based on the scoped page
```
rosiehsmith: Nit: screenMetaData sp

rosiehsmith: It also looks like the jsdocs could be updated :)

shared/src/business/utilities/documentGenerators/constants/printableTrialSessionWorkingCopyConstants.js
```diff
@@ -0,0 +1,462 @@
+const FORMATTED_CASES = [
+  {
+    calendarNotes:
+      'Spicy jalapeno bacon ipsum dolor amet laborum nostrud eiusmod, ex drumstick quis eu dolore excepteur fatback. Qui sirloin andouille chuck. Jowl velit capicola quis elit meatloaf.',
```
rosiehsmith, line 4: MEAT! 🍖 

we do have a folder `shared/src/test` that contains constants that end up being used for testing purposes, would this be a good place to put this file? though that folder could probably be renamed to something more descriptive than `test` haha

shared/src/business/utilities/documentGenerators/printableWorkingCopySessionList.js
```diff
@@ -0,0 +1,53 @@
+const {
+  reactTemplateGenerator,
+} = require('../generateHTMLTemplateForPDF/reactTemplateGenerator');
+const { generateHTMLTemplateForPDF } = require('../generateHTMLTemplateForPDF');
+
+const printableWorkingCopySessionList = async ({
```
rosiehsmith, line 6: yay! ❤️

rosiehsmith:
Still making my way through the files, but nice job y'all!

shared/src/business/useCases/trialSessions/generatePrintableTrialSessionCopyReportInteractor.ts
```diff
@@ -0,0 +1,83 @@
...
+import { UnauthorizedError } from '../../../errors/errors';
+/**
+ * generatePrintableTrialSessionCopyReportInteractor
+ *
+ * @param {object} applicationContext the application context
+ * @param {object} providers the providers object
+ * @param {string} providers.judge the optional judge filter
+ * @param {string} providers.docketNumber the optional docketNumber filter
+ * @returns {Array} the url of the document
+ */
```
rosiehsmith, line 18: nit: jsdocs

shared/src/business/useCases/trialSessions/generatePrintableTrialSessionCopyReportInteractor.test.ts
```diff
@@ -0,0 +1,144 @@
...
+const {
+  generatePrintableTrialSessionCopyReportInteractor,
+} = require('./generatePrintableTrialSessionCopyReportInteractor');
+const { ROLES } = require('../../entities/EntityConstants');
+
+describe('generatePrintableTrialSessionCopyReportInteractor', () => {
+  let mockUser;
+  let mockTrialSession;
+
+  const TRIAL_SESSION = {
```
rosiehsmith: Nit and half an idea: could try using test data in `mockTrial` here instead - might need to add some fields though since it looks like the structure isn't quite the same

rosiehsmith: I tried using `MOCK_TRIAL_REGULAR` from `mockTrial.js` and that seems to work

shared/src/business/utilities/pdfGenerator/documentTemplates/PrintableWorkingCopySessionList.jsx
```diff
@@ -0,0 +1,206 @@
...
+const generateSelectedFilterList = filters => {
+  const selectedFilters = [];
+  const filterKeys = Object.keys(filters).filter(
+    filterKey => filterKey !== 'showAll',
+  );
+  const userFilterSelection = filterKeys.filter(
+    filterKey => filters[filterKey],
+  );
+
+  orderedFilterMap.map(orderedFilterKey => {
```
rosiehsmith: Nit: my `deepscan` plugin actually pointed this out - since nothing is being used from the return of this `map`, it could be a `forEach`

web-client/src/presenter/actions/TrialSessionWorkingCopy/preparePrintableFormattedCasesAction.js
jamesobrooks: @rosiehsmith The current implementation groups the member cases with the lead case, but if I am reading yours correctly, it would push member cases to the end of the formatted cases array which would be ungrouping the consolidated cases.

rosiehsmith: ah i see, i didn't realize that was part of what was required for this action. rather than pushing to the end of `formattedCases` it could be a `splice` or something to be inserted after the leadCase. meh

web-client/src/presenter/sequences/openPrintableTrialSessionWorkingCopyModalSequence.js
shumway-tm: Done!

web-client/src/views/TrialSessionWorkingCopy/PrintableTrialSessionWorkingCopyModal.jsx
shumway-tm: Done!

web-client/src/presenter/actions/TrialSessionWorkingCopy/getTrialSessionWorkingCopyCaseNotesFlagAction.js
shumway-tm: Renamed to showCaseNotes throughout all files.

shared/src/business/utilities/DateHandler.test.js
```diff
@@ -342,6 +342,15 @@ describe('DateHandler', () => {
       );
       expect(result).toBe('8:40 pm ET');
     });
+
+    it('creates a formatted EST time using DateHandler internal format "TIME_TZ"', () => {
```
jamesobrooks, line 346: Change this name.

shared/src/proxies/trialSessions/generatePrintableTrialSessionCopyReportProxy.js
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * generatePrintableTrialSessionCopyReportInteractorProxy
+ *
+ * @param {object} applicationContext the application context
+ * @param {object} providers the providers object
+ * @param {string} providers.docketNumber the optional trialSessionId filter
+ * @returns {Promise<*>} the promise of the api call
+ */
+exports.generatePrintableTrialSessionCopyReportInteractor = (
```
rosiehsmith, line 11: nit: jsdocs

rosiehsmith (approved):

web-client/src/presenter/actions/TrialSessionWorkingCopy/preparePrintableFormattedCasesAction.js
jamesobrooks: @rachelschneiderman, Rosie mentioned you might have an idea about this that would maybe let us not use the swap variable, but be simple to read. I hadn't used a splice because I felt like keeping track of an index for the splice is error-prone for the future and harder to keep track of in cognitive complexity terms.

shared/src/business/utilities/pdfGenerator/documentTemplates/PrintableWorkingCopySessionList.jsx
jamesobrooks: Done

shared/src/business/useCases/trialSessions/generatePrintableTrialSessionCopyReportInteractor.test.ts
jamesobrooks: Done

shared/src/business/useCases/trialSessions/generatePrintableTrialSessionCopyReportInteractor.ts
jamesobrooks: Done

web-client/src/presenter/actions/PendingItems/setTitleForGlobalReportAction.js
jamesobrooks: Done

web-client/src/views/TrialSessionWorkingCopy/PrintableTrialSessionWorkingCopyPreviewPage.jsx
```diff
@@ -0,0 +1,45 @@
...
+import { connect } from '@cerebral/react';
+import { sequences, state } from 'cerebral';
+import React from 'react';
+
+export const PrintableTrialSessionWorkingCopyPreviewPage = connect(
+  {
+    gotoTrialSessionWorkingCopySequence:
+      sequences.gotoTrialSessionWorkingCopySequence,
+    headerTitle: state.screenMetadata.headerTitle,
+    navigateBackSequence: sequences.navigateBackSequence,
```
rachelschneiderman: 🔧 Looks like this isn't used, can it be removed?

web-client/src/presenter/actions/TrialSessionWorkingCopy/getTrialSessionWorkingCopyDataAction.js
```diff
@@ -0,0 +1,13 @@
+import { state } from 'cerebral';
+
+/**
+ * getTrialSessionWorkingCopyDataAction
+ *
+ * @param {object} providers.get the cerebral get function
+ * @returns {Object} filters, sessionNotes
+ */
+
+export const getTrialSessionWorkingCopyDataAction = ({ get }) => {
```
rachelschneiderman: Nit: Could this be condensed into a get call in the action where these objects are used? Typically we pass `props` around from action to action in a sequence when they are being mutated. Since the action that consumes these values grabs them from props I think they could just be pulled off of state in that action to make things a bit cleaner!

web-client/src/presenter/actions/TrialSessionWorkingCopy/preparePrintableFormattedCasesAction.js
rachelschneiderman: I was thinking this looked familiar to how `CaseListPetitioner` renders consolidated cases and when I looked into how that was done, it seems a lot of the sorting is done server-side and the client mostly just renders whatever it's handed back. I was just thinking it may be possible to DRY up consolidated case list logic a bit if we could reuse some of that code!

web-client/src/presenter/actions/TrialSessionWorkingCopy/getFormattedTrialSessionDetailsAction.js
```diff
@@ -0,0 +1,13 @@
+import { state } from 'cerebral';
+
+/**
+ * getFormattedTrialSessionDetailsAction
+ *
+ * @param {object} providers.get the cerebral get function
+ * @returns {Object} formattedTrialSessionDetails
+ */
+
+export const getFormattedTrialSessionDetailsAction = ({ get }) => {
```
rachelschneiderman: 🔧 Same thing here, can the action that uses `formattedTrialSessionDetails` just pull that value off of state?

web-client/src/presenter/actions/TrialSessionWorkingCopy/getTrialSessionWorkingCopyCaseNotesFlagAction.js
```diff
@@ -0,0 +1,13 @@
+import { state } from 'cerebral';
+
+/**
+ * getTrialSessionWorkingCopyCaseNotesFlagAction
+ *
+ * @param {object} providers.get the cerebral get function
+ * @returns {Object} caseNotesFlag
+ */
+
+export const getTrialSessionWorkingCopyCaseNotesFlagAction = ({ get }) => {
```
rachelschneiderman: 🔧 Do we need an action for this? Can the action that needs this value pull it directly off of state?

web-client/src/presenter/actions/TrialSessionWorkingCopy/generatePrintableTrialSessionCopyReportAction.js
```diff
@@ -0,0 +1,95 @@
...
+      inConsolidatedGroup: formattedCase.inConsolidatedGroup,
+      irsPractitioners: formattedCase.irsPractitioners,
+      leadCase: formattedCase.leadCase,
+      privatePractitioners: formattedCase.privatePractitioners,
+      trialStatus,
+      userNotes: formattedCase.userNotes,
+    };
+  });
+
+  const formattedEstimatedEndDateFull = applicationContext
```
rachelschneiderman, line 47: Nit: Can this be standardized to either define all calls to this method as a variable or as a function?

web-client/src/presenter/actions/PendingItems/setTitleForGlobalReportAction.js
```diff
@@ -3,10 +3,17 @@ import { state } from 'cerebral';
...
- * @param {object} providers the providers object
- * @param {object} providers.store the cerebral store object used for setting pendingItems
- * @param {object} providers.props the pendingItems to set
+ * @param {string} headerTitle the name of the header title
+ * @returns {Promise} async action
  */
-export const setTitleForGlobalReportAction = ({ store }) => {
-  store.set(state.screenMetadata.headerTitle, 'Pending Report');
-};
+export const setTitleForGlobalReportAction =
```
rachelschneiderman: 💭 Typically when we have actions that accept arguments I believe we name them `<name>FactoryAction` to help indicate that. What do you think about updating this function name to indicate that?

web-api/swagger.json
```diff
@@ -4814,6 +4814,85 @@
         }
       }
     },
+    "/trial-sessions/{trialSessionId}/printable-working-copy": {
```
rachelschneiderman, line 4817: 👏 👏 👏 Thanks for doing this!

shared/src/business/utilities/pdfGenerator/documentTemplates/PrintableWorkingCopySessionList.jsx
```diff
@@ -0,0 +1,206 @@
...
+    code: 'takenUnderAdvisement',
+    name: 'Taken Under Advisement',
+  },
+  {
+    code: 'statusUnassigned',
+    name: 'Status Unassigned',
+  },
+];
+
+const generateSelectedFilterList = filters => {
```
rachelschneiderman: 🔧 There is quite a bit of logic in here, my understanding is that in general we want to try and keep as much logic out of view files as possible so they can have a single responsibility and because they aren't tested. Would it be possible to move some of these functions and constants into other files where they can be tested?

web-client/src/presenter/actions/TrialSessionWorkingCopy/preparePrintableFormattedCasesAction.js
jamesobrooks: @rachelschneiderman, we could look into it, but this scenario is a bit odd. The cases have come to the non-printable working copy and had formatting applied. The story is to make a printable version of the same view, so we start with those formatted cases. We decided to have the client copy the member cases into the array at the top level to avoid complex logic in the html for the pdf (which does not benefit from the ability to utilize helpers).

I feel like doing what you're suggest could be good, but would require some refactoring of the non-printable working copy.

web-client/src/views/TrialSessionWorkingCopy/PrintableTrialSessionWorkingCopyPreviewPage.jsx
shumway-tm: Resolved!

web-client/src/presenter/actions/TrialSessionWorkingCopy/getTrialSessionWorkingCopyDataAction.js
shumway-tm: Resolved! We moved them to `generatePrintableTrialSessionCopyReportAction`, which is where they are actually used.

PR #2603, opened 2022-09-29
9610 hybrid case type counts

rgoshen:


rgoshen:
files not part of story are in the PR

PR #2604, opened 2022-09-29
9493 Court User: Filing a paper document in a consolidated case

rosiehsmith:
Relevant, smaller PRs into the story branch itself:
https://github.com/flexion/ef-cms/pulls?q=is%3Apr+9493+

This covers the following 2 workflows for paper filing:
- Save and serve on adding a paper filing
- Save for later, then serve on adding a paper filing


shared/src/business/entities/EntityConstants.js
```diff
@@ -435,6 +436,18 @@ const TRACKED_DOCUMENT_TYPES = {
...
+  ...Object.values(DOCUMENT_INTERNAL_CATEGORIES_MAP),
+])
+  .filter(internalEvent => internalEvent.caseDecision)
+  .map(x => x.eventCode);
+
+const MULTI_DOCKET_FILING_EVENT_CODES = flatten([
+  ...Object.values(DOCUMENT_INTERNAL_CATEGORIES_MAP),
+])
+  .filter(internalEvent => !internalEvent.caseDecision)
+  .map(x => x.eventCode);
```
jamesobrooks, line 450: :camping: I wonder if we should consider a TypeScript class or interface for event codes at this point. It might be easier to manage if there are functions on a type to get back a specific subset that we need.

shared/src/business/useCaseHelper/consolidatedCases/formatConsolidatedCaseCoversheetData.js
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Formats consolidated cases coversheet data
+ *
+ * @param {object} providers.applicationContext the applicationContext
+ * @param {object} providers.caseEntity the case entity
+ * @param {object} providers.coverSheetData the coversheet data
+ * @param {object} providers.docketEntryEntity the docketEntryEntity
+ * @returns {object} coversheet data with consolidated cases
+ */
+exports.formatConsolidatedCaseCoversheetData = async ({
```
jamesobrooks, line 12: 💭  I think elsewhere we've begun removing sorting and filtering from the `format` function, and this doesn't seem to be doing any actual formatting, so I think we should consider renaming this function.

shared/src/business/useCases/docketEntry/addPaperFilingInteractor.test.ts
```diff
@@ -103,12 +96,11 @@ describe('addPaperFilingInteractor', () => {
     expect(
       applicationContext.getUseCaseHelpers().serveDocumentAndGetPaperServicePdf
         .mock.calls[0][0].docketEntryId,
-    ).toEqual('c54ba5a9-b37b-479d-9201-067ec6e335bb');
+    ).toEqual(mockPrimaryDocumentFileId);
```
jamesobrooks, line 93: :+1:

shared/src/business/useCases/docketEntry/addPaperFilingInteractor.test.ts
```diff
@@ -251,7 +248,7 @@ describe('addPaperFilingInteractor', () => {
       applicationContext.getPersistenceGateway().updateCase,
     ).toHaveBeenCalled();
     expect(
-      applicationContext.getPersistenceGateway().updateCase.mock.calls[1][0]
+      applicationContext.getPersistenceGateway().updateCase.mock.calls[0][0]
```
jamesobrooks, line 294: These sorts of changes to tests always make me a bit concerned. Are we testing the same thing we used to be testing since we're now looking at a different call? Or was the test looking at the wrong thing before? No answer necessary, I just wanted to throw it out there.

shared/src/business/useCases/docketEntry/addPaperFilingInteractor.test.ts
```diff
@@ -352,4 +352,60 @@ describe('addPaperFilingInteractor', () => {
...
+        isSavingForLater: true,
+        primaryDocumentFileId: 'c54ba5a9-b37b-479d-9201-067ec6e335bb',
+      });
+
+      expect(
+        applicationContext.getPersistenceGateway().saveWorkItem,
+      ).toHaveBeenCalledTimes(mockConsolidatedGroup.length);
+    });
+
+    it('should still save only one copy of the document to s3', async () => {
```
jamesobrooks, line 509: 👍

jamesobrooks (commented):
I'm not all the way through yet, just had a couple of thoughts I wanted to get out in case I get distracted.

shared/src/business/entities/cases/Case.isLeadCase.test.js
```diff
@@ -0,0 +1,32 @@
+const { isLeadCase } = require('./Case');
+
```
mrinsin, line 2: 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉

mrinsin (approved):

shared/src/business/useCases/docketEntry/addPaperFilingInteractor.test.ts
rosiehsmith: yeah, i felt weird about this too at first and my commit message helped clue me in (yay past me!) - "update expectation to check the first call to `updateCase` which is the entity that is passed into the updateCaseAutomaticBlock call". i'm pretty certain that the test was wrong, because previously in `addPaperFilingInteractor` we were calling `updateCaseAndAssociations` twice, which was seemingly for no reason so now it's only calling it once.

good call out!

shared/src/business/useCaseHelper/consolidatedCases/formatConsolidatedCaseCoversheetData.js
rosiehsmith: it is still sorting and filtering - are you asking that we remove that part? i think i need some clarity. if we were to remove sorting and filtering from this into a separate file, any ideas on what the new name of this existing file would be without sorting/filtering?

rosiehsmith: or maybe an example of a file where it's doing what you're referencing could be helpful

shared/src/business/entities/EntityConstants.js
rosiehsmith: this is an interesting idea! should this be a larger discussion for the team since only interactors have been converted so far and this could turn into a larger snowball of a decision?

shared/src/business/useCaseHelper/consolidatedCases/formatConsolidatedCaseCoversheetData.js
jamesobrooks: I believe `web-client/src/presenter/computeds/formattedMessages.js` used to do sorting, filtering, and formatting, and we pulled out the sorting/filtering.

TomElliottFlexion (approved):

shumway-tm (approved):

jimlerza (approved):

PR #2612, opened 2022-09-30
9610 hybrid case type counts

rgoshen:
This PR adds a small case count and regular case count in addition to the total completed count to Eligible Cases tab for a trial session.

Screen Shot 2022-10-03 at 17 09 04


mmarcotte:
@rgoshen  -- can you please remove `package-lock.json` from the list of files changed? 🙏  thanks!

web-client/src/presenter/computeds/trialSessionDetailsHelper.js
```diff
@@ -1,19 +1,46 @@
+import {
```
rachelschneiderman: 🔧 I think entity constants in the web-client should be pulled off of `applicationContext.getConstants()` instead of directly imported

rgoshen: Done

rachelschneiderman (approved):

mmarcotte (approved):

mmarcotte (approved):

PR #2617, opened 2022-10-04
Handle broken pdf-testing builds

jamesobrooks:
# Problem

Some developers cannot use specific versions of `ghostscript`. I suspect an issue with the mirror.

# Solution

Conditionally install whichever version of `ghostscript` will work for the developer's machine.

# Docs

Adds ensuring that `ghostscript` and `graphicsmagick` versions are still supported to the dependency updates instructions.


TomElliottFlexion (approved):

docs/team-process.md
```diff
@@ -121,28 +121,29 @@ Every week we rotate responsibility for updating dependencies. As an open-source
 #### Library Update Steps
 
 1. `npm update`: Update to current minor versions of all libraries. These shouldn't include any breaking changes, but still might, so it's best to verify with smoke tests in AWS.
-
-2. `npm outdated`: Informs us of major version updates that we need to update manually. Often there are breaking API changes that require refactoring.
-
-3. `npm audit`: Informs us of known security vulnerabilities. If transitive dependencies are vulnerable, use the resolutions block in `package.json` to specify version overrides.
+1. `npm outdated`: Informs us of major version updates that we need to update manually. Often there are breaking API changes that require refactoring.
```
ujahio, line 123: Just noticed the listing are all "1"'s.

jamesobrooks: Yes, this is so that additions don't require* reordering.
https://www.markdownguide.org/basic-syntax/#ordered-lists

*Technically, md only requires that the first item be a `1`. But it's much more readable if they're not out of order.

1. `1.`
5. `5.`
3. `3.`
9. `9.`

jimlerza (approved):

Dockerfile-pdf-testing
```diff
@@ -8,8 +8,13 @@ RUN sed -i '/deb http:\/\/deb.debian.org\/debian stretch-backports main/d' /etc/
 RUN apt-get -o Acquire::Check-Valid-Until=false update
 
 RUN apt-get install -y \
-  graphicsmagick=1.4+really1.3.35-1~deb10u1 \
-  ghostscript=9.27~dfsg-2+deb10u5
+  graphicsmagick=1.4+really1.3.35-1~deb10u1
+
+# Install secondary if primary does not work
+RUN apt-get install -y \
+  ghostscript=9.27~dfsg-2+deb10u5 \
```
rgoshen, line 15: I think Cody found out that installing graphicsmagick also installs ghostscript automatically.

rgoshen (approved):

docs/team-process.md
ujahio: @jamesobrooks Learned something new. Thanks

Dockerfile-pdf-testing
jamesobrooks: @codyseibert is that happening? How can I tell, since completely commenting out the `ghostscript` install seems to do nothing? From what I can see, it _depends_ on libgraphicsmagick-q16-3 >= 1.3.5, which in turn _recommends_ ghostscript.

jamesobrooks:
@jimlerza I had to deal with some merge conflicts in the docs. I figured you might want another look.

jimlerza (approved):

PR #2618, opened 2022-10-04
9633 BUG: Petitions Clerks and Docket Clerks receive validation error if they use the Calculate links when entering Statistics on a Petition

mrinsin:
The source of the bug is due to the major update to `react-number-format`. The prop `isNumericString` needed to be migrated to `valueIsNumericString`. See this reference: https://s-yadav.github.io/react-number-format/docs/migration/#isnumericstring---valueisnumericstring

We added to a cypress end to end test to cover this bug. We used this kind of test since the bug is only found via UI interactions, and jest integration tests would bypass what causes the actual bug.


rachelschneiderman (approved):
Fixed!

jimlerza (approved):

ujahio (approved):

shumway-tm (approved):

TomElliottFlexion (approved):

PR #2621, opened 2022-10-04
Devex 1071 enzyme replacement

rgoshen:
This PR is for [DEVEX1071](https://trello.com/c/vdZODDGU/1071-substitute-enzyme-for-an-alternative-library) replacing all Enzyme tests in DAWSON.  Enzyme is no longer supported.

- removed all pdfGenerator/documentTemplates enzyme tests which are already convered by tests in image-compare-pdfs (verifies content looks as we want) as well as documentGenerators tests (see closed [PR2581](https://github.com/ustaxcourt/ef-cms/pull/2581))
- removed all component enzyme tests as these are subcomponents of image tested documentTemplates (in image-compare-pdfs) (see closed [PR2581](https://github.com/ustaxcourt/ef-cms/pull/2581))
- removed all email enzyme tests and replaced them with tests that generate pdfs so the same image comparison tests can be ran as pdf generator image comparison tests
- updated the docker container for Apple Silicon so any dev with a M1 Mac can run the image comparison tests locally


update-pdf-images.sh
```diff
@@ -4,6 +4,6 @@
 
 echo "running npm ci... this may take a while"
 
-npm ci
+npm i
```
jimlerza: I think this could result in updates being applied to dependencies, which would be unexpected as it is not the purpose of this script.

jimlerza (approved):

update-pdf-images.sh
rgoshen: The current version of base cypress is based on Node 14 which does not support apple silicon.  So, with updating to Node 16 for this particular docker and the package-lock based on Node 14, npm would need to install everything.  We are in the process of updating everything to Node 16 and then maybe we can change this back to 'ci' then.

jimlerza: I think you just need to run `npm i` once to generate a `package-lock` containing the updated packages, after which running `npm ci` is sufficient to install those updated packages.

I'll rephrase - my understanding of the dependency updates process is that the dev in charge of dependency updates for the week will be the one that runs `npm i` so they can carefully monitor incoming changes. I don't think we want other scripts, whose purpose is not explicitly "update dependencies," updating dependencies, unless there is some kind of warning that "running this script puts you on dependency updates for the week."

rgoshen: I changed it back to `npm ci`

rgoshen:
Closing PR due to too many files changes b/c of merging latest staging into branch

PR #2627, opened 2022-10-04
Devex 1071 enzyme replacement to test

rgoshen:
This PR is for [DEVEX1071](https://trello.com/c/vdZODDGU/1071-substitute-enzyme-for-an-alternative-library) replacing all Enzyme tests in DAWSON. Enzyme is no longer supported.

    removed all pdfGenerator/documentTemplates enzyme tests which are already convered by tests in image-compare-pdfs (verifies content looks as we want) as well as documentGenerators tests (see closed https://github.com/ustaxcourt/ef-cms/pull/2581)
    removed all component enzyme tests as these are subcomponents of image tested documentTemplates (in image-compare-pdfs) (see closed https://github.com/ustaxcourt/ef-cms/pull/2581)
    removed all email enzyme tests and replaced them with tests that generate pdfs so the same image comparison tests can be ran as pdf generator image comparison tests
    updated the docker container for Apple Silicon so any dev with a M1 Mac can run the image comparison tests locally


update-pdf-images.sh
```diff
@@ -4,6 +4,6 @@
 
 echo "running npm ci... this may take a while"
 
-npm ci
+npm i
```
jamesobrooks, line 7: Is this intentional?

Dockerfile-pdf-testing
```diff
@@ -1,15 +1,23 @@
-FROM cypress/base:14.19.0
+FROM cypress/base:16.17.1
```
jamesobrooks, line 1: Are we upgrading this everywhere? The base image version is tied to the version of Node that we are on, which is 14.19.0.

jamesobrooks (commented):
I'm not done yet, but had a couple of questions.

update-pdf-images.sh
rgoshen: Yes. This forces the reinstallation of the NPM packages otherwise on the M1 it would come back missing packages.

Dockerfile-pdf-testing
rgoshen: Yes, we are also in the process of upgrading to Node 16 everywhere.  We are taking on the Devex task next.  Also, this docker file was tested on two Apple silicon macs and also on an Intel mac and both ran the pdf image compare tests without any issues.

TomElliottFlexion (approved):

PR #2628, opened 2022-10-05
9610 hybrid case type counts to test

rgoshen:
This addresses an issue of when a user does not have TRIAL_SESSION_QC_COMPLETE permission, do not show the counts of regular and small cases.  The `showSmallAndRegularQcComplete` in `trialSessionDetailsHelper.js` was created to fix this.


rgoshen:
two files untouched by stroy are included in pr

PR #2633, opened 2022-10-06
9610-fix for bottom border of hybrid trial session filter when not a …

rgoshen:
Addressed wrong bottom margin for non petitions clerk users.


codyseibert (approved):

jasonfu9966 (approved):
yas

PR #2651, opened 2022-10-11
9627: Adding a STIN from the Review Petition page creates a Docket Index number for the STIN

jamesobrooks:
# [9627](https://github.com/flexion/ef-cms/issues/9627)

When adding the STIN on the wizard initially, we set the index to 0. When adding initial documents we were generating the next index number when we should not assign a real index number.

### DIFF
--- a/shared/src/business/entities/cases/Case.js
+++ b/shared/src/business/entities/cases/Case.js
@@ -1018,6 +1018,10 @@ Case.prototype.addDocketEntry = function (docketEntryEntity) {
 
     if (updateIndex) {
       docketEntryEntity.index = this.generateNextDocketRecordIndex();
+    } else if (
+      docketEntryEntity.eventCode === INITIAL_DOCUMENT_TYPES.stin.eventCode
+    ) {
+      docketEntryEntity.index = 0;
     }
   }
 
--- a/shared/src/business/entities/cases/Case.js
+++ b/shared/src/business/entities/cases/Case.js
@@ -1021,8 +1021,12 @@ Case.prototype.addDocketEntry = function (docketEntryEntity) {
     } else if (
       docketEntryEntity.eventCode === INITIAL_DOCUMENT_TYPES.stin.eventCode
     ) {
-      docketEntryEntity.index = 0;
+      throw new Error('STIN documents should not be on the docket record.');
     }
+  } else if (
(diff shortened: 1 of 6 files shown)


TomElliottFlexion (approved):

shumway-tm (approved):

ujahio (approved):

rgoshen (approved):

shared/src/business/useCaseHelper/initialFilingDocuments/updateInitialFilingDocuments.js
```diff
@@ -92,9 +92,10 @@ exports.updateInitialFilingDocuments = async ({
   caseEntity,
   caseToUpdate,
 }) => {
+  const PETITION_KEY = 'petitionFile';
   const initialDocumentTypesWithoutPetition = omit(
     INITIAL_DOCUMENT_TYPES_MAP,
-    'petitionFile',
+    PETITION_KEY,
```
jamesobrooks, line 98: I'm willing to remove this change. I thought the name would add a little more meaning to why the string is what it is.

jamesobrooks:
@rachelschneiderman, thanks for the feedback. Let me know if I've missed anything.

shared/src/business/entities/cases/Case.js
```diff
@@ -1018,7 +1018,15 @@ Case.prototype.addDocketEntry = function (docketEntryEntity) {
 
     if (updateIndex) {
       docketEntryEntity.index = this.generateNextDocketRecordIndex();
+    } else if (
```
rachelschneiderman, line 1021: 💭 I wonder if it would be more testable and reliable to put code like this into the Joi validation for DocketEntry? Using a Joi conditional we could specify that when `eventCode` is `STIN`, `isOnDocketRecord` has to be false

Just a thought, let me know what you think!

rachelschneiderman: Oooh same with the index for `STIN` has to be 0? That way if we ever try to persist that value the validation will throw an error. Again, just a thought.

rachelschneiderman (approved):

shared/src/business/entities/cases/Case.js
jamesobrooks: I thought the intent of using TypeScript was to move away from Joi validations. Happy to add it if you think it's important though.

PR #2652, opened 2022-10-11
9596 - Petitioner: Add Consolidated Card to Case Information Screen

shumway-tm:
### Non-mobile view

image

### Mobile view on iPhone SE

1233F460-6B55-47CD-A47E-D4D9983D05DE

### Mobile view on iPhone XR

image


rosiehsmith (approved):
cool

rosiehsmith:
One nit to take or leave: it would be nice if the vertical alignment of the mobile view matched between the docket number and the case title

mrinsin:
> One nit to take or leave: it would be nice if the vertical alignment of the mobile view matched between the docket number and the case title

Done (thank you @shumway-tm !) Added screenshots for mobile view (and all also iPhone SE view which is xs)

PR #2657, opened 2022-10-12
9511: Printable Trial Session Copy - Address PO Feedback

shumway-tm:


shared/src/business/utilities/documentGenerators/printableWorkingCopySessionList.js
```diff
@@ -7,6 +7,7 @@ const printableWorkingCopySessionList = async ({
   applicationContext,
   data,
 }) => {
+  console.log('User heading:::::', data.userHeading);
```
mrinsin: Log!

mrinsin (approved):
Looks good to me! Just one left over log

rgoshen (approved):

shared/src/business/utilities/documentGenerators/printableWorkingCopySessionList.js
shumway-tm: Removed it. Good eye!

PR #2661, opened 2022-10-13
9493 story to test

rosiehsmith:
includes PO feedback in this PR https://github.com/flexion/ef-cms/pull/9647
along with other staging to test differences


web-client/src/presenter/actions/CaseConsolidation/shouldSetupConsolidatedCasesAction.js
```diff
@@ -1,24 +1,42 @@
 import { state } from 'cerebral';
 
 /**
- * determines the path to be taken based on form data
+ * determines the path to be taken based on helper property
  *
  * @param {object} providers the providers object
  * @param {object} providers.get the cerebral get object
- * @param {object} providers.applicationContext the applicationContext object
```
jamesobrooks, line 8: Don't we usually have application context listed in the JSDOCS when it is used?

web-client/src/presenter/actions/CaseConsolidation/shouldSetupConsolidatedCasesAction.js
```diff
@@ -1,24 +1,42 @@
...
+    ...SINGLE_DOCKET_RECORD_ONLY_EVENT_CODES,
+  ];
+
+  let { eventCode } = get(state.form);
+
+  if (!eventCode) {
+    ({ eventCode } = caseDetail.docketEntries.find(
+      doc => doc.docketEntryId === docketEntryId,
+    ));
+  }
```
jamesobrooks, line 35: ❓ I'm curious as to why sometimes `eventCode` is on state and others it's not, so we just get it from the document. Could we not always get it from the document? Or should it be in state in both cases?

web-client/src/presenter/actions/CaseConsolidation/shouldSetupConsolidatedCasesAction.test.js
```diff
@@ -28,26 +27,58 @@ describe('shouldSetupConsolidatedCasesAction', () => {
       },
       state: {
         form: {
-          eventCode: SINGLE_DOCKET_RECORD_ONLY_EVENT_CODES[0],
+          eventCode: 'M083',
```
jamesobrooks: :question: Would it be too much to parameterize this test like:
```
it.each(SINGLE_DOCKET_RECORD_ONLY_EVENT_CODES)('should return the no path when the eventCode is one of SINGLE_DOCKET_RECORD_ONLY_EVENT_CODES', async () => {
```

web-client/src/presenter/actions/CaseConsolidation/shouldSetupConsolidatedCasesAction.test.js
```diff
@@ -28,26 +27,58 @@ describe('shouldSetupConsolidatedCasesAction', () => {
...
         },
       },
     });
 
     expect(pathNoStub).toHaveBeenCalled();
   });
 
-  it('should return the yes path when the eventCode is NOT one of SINGLE_DOCKET_RECORD_ONLY_EVENT_CODES', async () => {
+  it('should return the no path when the eventCode is one of ENTERED_AND_SERVED_EVENT_CODES', async () => {
```
jamesobrooks: ❓ Similar to above comment, could we do this:
```suggestion
  it.each(ENTERED_AND_SERVED_EVENT_CODES)('should return the no path when the eventCode is one of ENTERED_AND_SERVED_EVENT_CODES', async () => {
```

rosiehsmith: hmmm...maybe i'm not awake yet, but how then do we reference each ENTERED_AND_SERVED_EVENT_CODES to pass into the actual test?

rosiehsmith: Ah, instead of `it.each` I think we'd instead have to use a `forEach` 🚀

web-client/src/presenter/actions/CaseConsolidation/shouldSetupConsolidatedCasesAction.js
mrinsin: The `eventCode` at this point on `state.form` could have changed from what the original docket entry's `eventCode` was, which is why we try to grab that value from state - however this action is also called from the flow where a document is served from the `DocumentViewer` screen. In that flow, there is no `state.form` populated and this was causing eventCode to be undefined. So as of now it seems like both checks are needed! LMK if you'd like to chat about this in a zoom!

rachelschneiderman (approved):

TomElliottFlexion (approved):

rgoshen (approved):

PR #2662, opened 2022-10-13
9627 stin docket record index

jamesobrooks:
# [9627](https://github.com/flexion/ef-cms/issues/9627)

When adding the STIN on the wizard initially, we set the index to 0. When adding initial documents we were generating the next index number when we should not assign a real index number.


shared/src/business/entities/cases/Case.addDocketEntry.test.js
```diff
@@ -4,7 +4,25 @@ const {
 const { Case } = require('./Case');
 
 describe('addDocketEntry', () => {
-  it('attaches the docket entry to the case', () => {
+  it('should throw when docket entry is a STIN', () => {
```
jamesobrooks: I'm now wondering if these titles should be different.
```suggestion
  it('should throw when docket entry to be added to the docket record is a STIN', () => {
```

shared/src/business/entities/cases/Case.addDocketEntry.test.js
```diff
@@ -24,4 +42,29 @@ describe('addDocketEntry', () => {
       userId: 'irsPractitioner',
     });
   });
+
+  it("should assign docket entry index of '0' to the STIN", () => {
```
jamesobrooks: Following with the clarification of the above test:
```suggestion
  it("should assign docket entry index of '0' to the STIN when not being added to the docket record", () => {
```

shared/src/business/entities/cases/Case.addDocketEntry.test.js
mrinsin: I feel like that may not be necessary as the method under test is conveying the adding of the docket entry! (but thats just my opinion)

shared/src/business/entities/cases/Case.addDocketEntry.test.js
mrinsin: This rename makes sense to me!!

shared/src/business/useCaseHelper/initialFilingDocuments/updateInitialFilingDocuments.js
```diff
@@ -92,9 +92,10 @@ exports.updateInitialFilingDocuments = async ({
   caseEntity,
   caseToUpdate,
 }) => {
+  const PETITION_KEY = 'petitionFile';
```
mrinsin, line 95: Nice, I wonder if this string is used elsewhere in the code and if this would be worth moving to `EntityConstants`

mrinsin (approved):

rgoshen (approved):

shared/src/business/useCaseHelper/initialFilingDocuments/updateInitialFilingDocuments.js
jamesobrooks: @mrinsin there is a use in `CaseInternal.js` for validation, but there are several similar strings in use there too. Maybe a separate PR for that change if we want it? I don't feel strongly either way.

mrinsin: I don't feel too strongly about the changes being in this PR! 👍 to different PR

mmarcotte (approved):

PR #2697, opened 2022-10-21
9587 practitioner detail tabs

rgoshen:
This PR addresses UX comments:

1. No green success banner when successfully uploading a file
2. Spell out the "Other" option instead of the using abbreviations when a "Certificate of Good Standing" is selected as a type.

This PR changes the `US_STATES_OTHER` from an array of abbreviations in `EntityConstants.j` file to an object with the abbreviations being the key. Then, it updates all applicable files that uses that constant.

upload_practitioner_document_good_standing


jasonfu9966 (approved):
it makes me uncomfortable to see all the merging files in here, even though me didn't touch them during this story.

otherwise, nice work.

PR #2699, opened 2022-10-24
BUG 9660 Sealed Case information Display

shumway-tm:
# [Bug 9660](https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/flexion/ef-cms/9660)

### Solution

For external users, Petitioners, Private Practitioners, and IRS Practitioners, in consolidated groups that have no association with other cases in the group (irrespective of seal status), these users will no longer receive a full Case entity in the JSON returned when hitting the consolidated-cases endpoint in the case-detail screen. 

They now receive PublicCase entities for the cases they are unassociated to that are in the same consolidated 
group as a case they are associated with.

Behavior for internal users at this point, is unchanged. Behavior for the IRS SuperUser, is unchanged as well.

----

### Petitioner 

46A961F8-6B28-441A-986A-B036E637C5CC

#### Private Practitioner

210A935E-FDA8-4FF7-A779-289B6A2DDE9B

### IRS Practitioner

8F2B247B-9E31-48BF-82B5-E5289538669C


shared/src/business/useCaseHelper/consolidatedCases/formatPublicCase.test.ts
```diff
@@ -0,0 +1,103 @@
...
+  //     expect(
+  //       applicationContext.getPersistenceGateway().getCaseByDocketNumber.mock
+  //         .calls[0][0],
+  //     ).toEqual({
+  //       applicationContext,
+  //       docketNumber: '123-19',
+  //     });
+  //   });
+
+  it.skip('Should return a Not Found error if the case does not exist', async () => {
```
mrinsin: Looks like this test is skipped! And the previous one is commented out!

shared/src/business/useCaseHelper/consolidatedCases/formatPublicCase.ts
```diff
@@ -0,0 +1,48 @@
+import { NotFoundError } from '../../../errors/errors';
+import { PublicCase } from '../../entities/cases/PublicCase';
+import {
+  caseContactAddressSealedFormatter,
+  caseSealedFormatter,
+} from '../../utilities/caseFilter';
+import { decorateForCaseStatus } from '../../useCases/getCaseInteractor';
+import { isSealedCase } from '../../entities/cases/Case';
+
+// TODO: WRITE UNNNNNNIIIIIIITTTTTTT TEST!!!!!!!!!!!
```
mrinsin: nit: comment!

shared/src/business/useCaseHelper/consolidatedCases/formatPublicCase.ts
```diff
@@ -0,0 +1,48 @@
...
+  rawCaseRecord?: TCase;
+  docketNumber: string;
+}) => {
+  if (!rawCaseRecord.docketNumber && !rawCaseRecord.entityName) {
+    const error = new NotFoundError(`Case ${docketNumber} was not found.`);
+    error.skipLogging = true;
+    throw error;
+  }
+  rawCaseRecord.isSealed = isSealedCase(rawCaseRecord);
+  if (isSealedCase(rawCaseRecord)) {
```
mrinsin, line 26: Could this if condition check for `rawCaseRecord.isSealed`, and remove the duplicative call to `isSealedCase(rawCaseRecord)`?

shared/src/business/useCases/getConsolidatedCasesByCaseInteractor.ts
```diff
@@ -12,12 +14,51 @@ export const getConsolidatedCasesByCaseInteractor = async (
...
+
   const consolidatedCases = await applicationContext
     .getPersistenceGateway()
     .getCasesByLeadDocketNumber({
       applicationContext,
       leadDocketNumber: docketNumber,
     });
 
-  return Case.validateRawCollection(consolidatedCases, { applicationContext });
+  if (
```
mrinsin, line 26: Nit: could this new logic be unit tested?

shared/src/business/useCaseHelper/consolidatedCases/formatPublicCase.ts
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * formatPublicCase takes a rawCase and formats to a public case
+ *
+ * @param {object} providers the providers object
+ * @param {string} providers.applicationContext the applicationContext
+ * @param {string} providers.docketNumber the docket number of the case to get
+ * @param {object} providers.rawCaseRecord the rawCaseRecord
+ * @returns {object} the validated public case data
+ */
+export const formatPublicCase = ({
```
mrinsin, line 19: Nice extraction of logic!

shared/src/business/useCases/public/getPublicCaseInteractor.ts
```diff
@@ -26,22 +20,5 @@ export const getPublicCaseInteractor = async (
...
-
-  rawCaseRecord = caseContactAddressSealedFormatter(rawCaseRecord, {});
-  rawCaseRecord = decorateForCaseStatus(rawCaseRecord);
-
-  const publicCaseDetail = new PublicCase(rawCaseRecord, {
-    applicationContext,
-  });
-
-  return publicCaseDetail.validate().toRawObject();
+  return formatPublicCase({ applicationContext, docketNumber, rawCaseRecord });
```
mrinsin: 🎉

shumway-tm:
@mrinsin I meant to mark this as a draft. Unit tests incoming and then we will address your comments! Thank you for looking at it so quickly!

mrinsin (commented):
Didn't realize this was a draft oops....please ignore my comments

shared/src/business/useCaseHelper/consolidatedCases/formatPublicCase.ts
jamesobrooks: @mrinsin this one was a bit odd. `caseSealedFormatter` grabs several properties from the `rawCaseRecord`, including `isSealed` and returns it. It is possible given the logic in `Case.js` for `isSealed` to be falsy and `sealedDate` to be truthy and we would still treat the case as sealed. I do not know if this is a state that we would ever naturally end up in, but I think the previous logic probably accounted for this. We also did not want to make changes to other areas for this bug, although I certainly would not fault someone for wanting to track down whether that is a valid scenario, and if not, correct it.

We went with this new logic to hopefully make it more clear, although a bit verbose, that we are setting `isSealed` regardless.

rachelschneiderman: This seems like a good opportunity to tighten some of the validation logic for a Case. It also seems funky to me that it can get in that state (`isSealed` === false and `sealedDate` is defined) since we have methods off the Case to seal and unseal where both values are updated set. I feel like more investigation into how this is happening is warranted, maybe we aren't using that method somewhere where we should be

shared/src/business/useCases/getConsolidatedCasesByCaseInteractor.test.ts
```diff
@@ -3,53 +3,92 @@ import { MOCK_USERS } from '../../test/mockUsers';
...
+        },
+      ]);
+
+      expect(
+        applicationContext.getPersistenceGateway().getCasesByLeadDocketNumber,
+      ).toHaveBeenCalled();
+    });
   });
+
+  [petitionerUser, privatePractitionerUser, irsPractitionerUser].forEach(
```
mrinsin, line 62: 🎉

mrinsin (approved):

shared/src/business/useCaseHelper/consolidatedCases/formatPublicCase.test.ts
ujahio: Tests updated!

shared/src/business/useCaseHelper/consolidatedCases/formatPublicCase.ts
ujahio: Comments removed and tests adjusted

shared/src/business/useCaseHelper/consolidatedCases/formatPublicCase.ts
ujahio: Ok we are going to time box the investigation to see how cases are being set (with or without the sealedDate reset) and perhaps create a devex task or DOD item for the bug.

PR #2703, opened 2022-10-24
BUG 9648: When a Lead case is removed from a trial, member cases no longer display on the Session Copy Page

rosiehsmith:
Screen Shot 2022-10-24 at 2 54 16 PM

addresses flexion#9648

### DIFF
--- a/web-client/src/presenter/computeds/trialSessionWorkingCopyHelper.caseNotes.test.js
+++ b/web-client/src/presenter/computeds/trialSessionWorkingCopyHelper.caseNotes.test.js
@@ -1,0 +1,131 @@
+import { MOCK_CASE } from '../../../../shared/src/test/mockCase';
+import {
+  STATUS_TYPES,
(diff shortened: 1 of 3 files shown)


TomElliottFlexion (approved):

ujahio (approved):

shumway-tm:
Although the bug doesn't mention it (unless I missed it), this looks like it also fixes the same bug in the Printable Session Copy!🔥🎉💪🏼

shumway-tm (approved):

shumway-tm:
> Although the bug doesn't mention it (unless I missed it), this looks like it also fixes the same bug in the Printable Session Copy!🔥🎉💪🏼

Just wanted to clarify! I don't believe there is a published bug for the Printable Session Copy. The behavior of the Printable Session Copy directly uses the formatting of cases from the Trial Session Session Copy, so this bug was manifesting in the Printable Session Copy as well before this fix!

PR #2704, opened 2022-10-25
Bug 9660

ujahio:
# [Bug 9660](https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/flexion/ef-cms/9660)

### Note

 For external users, Petitioners, Private Practitioners, and IRS Practitioners, in consolidated groups, that have no association with other member cases, these users will no longer receive a full Case entity in the JSON returned when hitting the consolidated-cases endpoint in the case-detail screen. 

 Instead, these users will receive PublicCase entities for the cases they are unassociated with but are in the same consolidated group as a cases they are associated with.

 Behavior for both internal users and the IRS SuperUser remain unchanged.

----

### Petitioner 

46A961F8-6B28-441A-986A-B036E637C5CC

#### Private Practitioner

210A935E-FDA8-4FF7-A779-289B6A2DDE9B

### IRS Practitioner

8F2B247B-9E31-48BF-82B5-E5289538669C


jimlerza (approved):

shared/src/business/useCaseHelper/consolidatedCases/formatPublicCase.test.ts
```diff
@@ -0,0 +1,125 @@
...
+      applicationContext,
+      rawCaseRecord: { ...mockCase, petitioners: [sealedContactPrimary] },
+    });
+
+    result.petitioners.forEach(petitioner => {
+      expect(petitioner.entityName).toEqual('PublicContact');
+      expect(petitioner).not.toMatchObject({
+        address1: 'address 1',
+        email: 'email 1',
+      });
```
mmarcotte, line 98: is this block even going to run? should it be testing for `result.petitioners[0]` instead of looping through?

shared/src/business/useCaseHelper/consolidatedCases/formatPublicCase.test.ts
```diff
@@ -0,0 +1,125 @@
...
+      applicationContext,
+      rawCaseRecord: mockCase,
+    });
+
+    result.petitioners.forEach(petitioner => {
+      expect(petitioner.entityName).toEqual('PublicContact');
+      expect(petitioner).not.toMatchObject({
+        address1: 'address 1',
+        email: 'email 1',
+      });
```
mmarcotte, line 120: ... same here

shared/src/business/useCases/getConsolidatedCasesByCaseInteractor.ts
```diff
@@ -12,12 +14,51 @@ export const getConsolidatedCasesByCaseInteractor = async (
...
       leadDocketNumber: docketNumber,
     });
 
-  return Case.validateRawCollection(consolidatedCases, { applicationContext });
+  if (
+    ![
+      ROLES.petitioner,
+      ROLES.privatePractitioner,
+      ROLES.irsPractitioner,
+    ].includes(user.role)
```
mmarcotte: This might be easier to maintain and read:

```User.isInternalUser(user.role) || ROLES.irsSuperuser === user.role```

File: shared/src/business/useCases/public/getPublicCaseInteractor.test.ts [hunk 4]
```diff
@@ -68,12 +68,12 @@ describe('getPublicCaseInteractor', () => {
...
   });
 
-  it('should format the given docket number, removing leading zeroes and suffix', async () => {
+  it('should format the given docket number, removing leading zeroes and suffix', () => {
     applicationContext
       .getPersistenceGateway()
       .getCaseByDocketNumber.mockReturnValue(MOCK_CASE);
 
-    await getPublicCaseInteractor(applicationContext, {
+    getPublicCaseInteractor(applicationContext, {
```
mmarcotte: why was the `await` dropped here?

shared/src/business/useCaseHelper/consolidatedCases/formatPublicCase.test.ts
jamesobrooks: @mmarcotte, that code does execute. We want to make sure that for all petitioners in the array, their PII is not included. Currently there is only one petitioner, but if someone were to edit the test in such a way that there were multiple petitioners, then the added petitioners would also be validated.

shared/src/business/useCases/getConsolidatedCasesByCaseInteractor.ts
shumway-tm: Good catch. We have implemented the change.

shared/src/business/useCases/public/getPublicCaseInteractor.test.ts
shumway-tm: Added the await back in. Maybe an oversight.

jimlerza (approved):

PR #2708, opened 2022-10-25
9587 - practitioner detail tabs

rgoshen:
Addresses UX feedback to have the downloaded file to save as its file name instead of its document id.

This does not affect how the document is stored.  This is still stored and retrieved by its assigned id.

Screen Shot 2022-10-25 at 10 29 01


rgoshen:
closed to merge conflicts

PR #2725, opened 2022-10-27
9680 BUG: Address label is not displaying Country Name in International petitioner contact info

rosiehsmith:
I'm open to suggestions for testing this further as the visual PDF tests already included the country, since they're just using the props passed into the test to generate the PDF.

image


jamesobrooks (approved):
Looks good to me. It seems like there was a difference in naming of input and the attempts to use that input. Is that correct?

PR #2732, opened 2022-10-28
9649 BUG: Removing pending item from a Consolidated Case causes Consolidated Case Card to disappear

shumway-tm:
Fix for [Bug 9649](https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/flexion/ef-cms/9649)

The Consolidated Case Card now remains populated after removing a pending item in the tracked items tab.

+ get and set actions were added into the sequence for removing pending items from the case detail
+ an integration test written for a previous closely related [bug](https://github.com/flexion/ef-cms/issues/9486) was expanded to cover this bug case


web-client/integration-tests/journey/removePendingItemFromCase.js
```diff
@@ -0,0 +1,22 @@
+export const removePendingItemFromCase = (cerebralTest, clerkName) => {
```
rosiehsmith, line 1: Nice, reusing this!

rosiehsmith (approved):

PR #2740, opened 2022-10-31
9687 - Critical Sealed Transcript Clickable Link Causes 403 Error

codyseibert:


rachelschneiderman:
Is there a specific named test that can be added to catch this bug? That way if it fails it will be really clear why 😄

TomElliottFlexion (approved):

codyseibert:
> Is there a specific named test that can be added to catch this bug? That way if it fails it will be really clear why 😄

we added an additional way to put "descriptions" on this test file.  This one test file was a bit different from how we normally write tests, but it'll print the real test description on failure now.

PR #2752, opened 2022-11-02
Bugfix 9654 completing qc

ujahio:
This is a workaround solution for flexion#9654. We are going to create a devex task for properly updating the documentTitle on a docket entry.


mrinsin (commented):
Unit tests for the changes in this PR? 😄

PR #2753, opened 2022-11-02
9588: practitioner detail edit/delete/sort

TomElliottFlexion:


shared/src/business/useCases/practitioners/deletePractitionerDocumentInteractor.test.ts
```diff
@@ -0,0 +1,56 @@
...
+      barNumber,
+      practitionerDocumentFileId,
+    });
+    expect(
+      applicationContext.getPersistenceGateway().deleteDocumentFromS3.mock
+        .calls[0][0],
+    ).toMatchObject({
+      key: practitionerDocumentFileId,
+    });
+  });
```
mmarcotte: It seems like a deviation from the Clean Architecture approach that has gone into this project to mention S3 explicitly. In theory we could use another file store and not have to change this function's name, right? Is `deleteDocumentFile` more fitting?

TomElliottFlexion: Sure thing, renamed to `deleteDocumentFile`

mmarcotte (approved):

PR #2794, opened 2022-11-11
9628 BUG: Blocked label is not removed on the case header when there aren't any Pending Items or Deadlines

rosiehsmith:


rosiehsmith:
Closing this because GitHub is out of sync with reality

PR #2801, opened 2022-11-14
DevEx - More TS Refactoring

codyseibert:


shared/src/persistence/s3/getDocument.test.ts
```diff
@@ -78,11 +84,14 @@ describe('getDocument', () => {
 
     await getDocument({
       applicationContext,
+      docketNumber: '101-29',
```
rachelschneiderman: Can these values be pulled out into named constants?

shared/src/persistence/s3/getDownloadPolicyUrl.ts
```diff
@@ -5,12 +5,17 @@
...
   filename,
   key,
-  useTempBucket,
-}) => {
+  useTempBucket = false,
+}: {
+  applicationContext: IApplicationContext;
+  filename?: string;
+  key: string;
+  useTempBucket?: boolean;
```
rachelschneiderman, line 17: For my understanding, since `useTempBucket` is defaulted to false, does it also need to be marked as optional?

shared/src/persistence/s3/isFileExists.test.ts
```diff
@@ -48,6 +47,7 @@ describe('isFileExists', () => {
       await isFileExists({
         applicationContext,
         key: 'jenny-not-found',
+        useTempBucket: false,
```
rachelschneiderman: Similar question to above, just for my learning 😄 If `useTempBucket` is marked as optional and defaulted to false, why does it have to be added here?

shared/src/persistence/s3/getDownloadPolicyUrl.ts
codyseibert: yeah, if you don't set it as optional, typescript will complain expecting you to set where ever you invoke this function

shared/src/persistence/s3/isFileExists.test.ts
codyseibert: correct, I can remove those from this file since they defaulted to false in the function.

rgoshen (approved):

jasonfu9966 (approved):

PR #2819, opened 2022-11-17
9718 - flipping order of the router to fix bug

codyseibert:


mmarcotte (approved):

jimlerza (approved):

jasonfu9966 (approved):

cypress-integration/integration/docket-clerk-edits-court-issued-document.cy.js
```diff
@@ -0,0 +1,25 @@
+describe('Docket clerk edits a court issued document', function () {
+  before(() => {
+    cy.login('docketclerk');
+  });
+
+  it('upload a court issued document, click on the in progress link, and navigate to the correct page', () => {
+    cy.visit('case-detail/104-20/upload-court-issued');
```
rachelschneiderman, line 7: 🌱 I was reading the "recommended practices" from the cypress docs and noticed they recommend adding `data-cy` attributes to elements we want to query for in tests in order to prevent accidentally breaking tests if we change class names or delete an id. Just a thought for the future, I've been trying to remember to add those 😄

rachelschneiderman (approved):
Thanks for adding a test that will catch this in the future!!! 👏 👏 👏

PR #2827, opened 2022-11-21
9628 BUG: Blocked label is not removed on the case header when there aren't any Pending Items or Deadlines

rosiehsmith:
flexion#9628


rosiehsmith:
Wrong branch.

PR #2855, opened 2022-11-24
Intermittent cypress test fix

rosiehsmith:
This fixes intermittency with the `docket-clerk-edits-court-issued-document` test.

Essentially the retry attempts were not working because the`before` block logged in the user only on the first try and not again, so the retries would just be stuck on the login page. Also, added a check that the page is no longer on the `add-docket-entry` screen after saving the added docket entry since watching the cypress failure videos show that the newly added docket entry doesn't appear in the list of "In Progress" on Document QC, possibly because the entry wasn't actually saved yet.

See a run where it successfully retries then passes: https://app.circleci.com/pipelines/github/flexion/ef-cms/39507/workflows/8a06d6ef-38d8-4feb-8e4d-0d26fd0dccec/jobs/261315

A video where it retries multiple times and fails [here](https://output.circle-artifacts.com/output/job/9e83ae77-f9f3-4235-97df-553f0a7b7095/artifacts/0/root/project/cypress-integration/videos/docket-clerk-views-consolidated-case.cy.js.mp4)


rachelschneiderman (approved):
WOW! Nice debugging

Clap

ujahio (approved):

PR #2859, opened 2022-11-28
9715 Bugfix: Add extracted utility method to applicationContextPublic

rachelschneiderman:
Test for this bug will be coming in another PR. I wanted to make a PR ASAP so Tenille could be unblocked from testing other stories on `ustc/test`.


mrinsin (approved):
🎉

PR #2871, opened 2022-12-01
Bugfix 9654 intermediate to test

ujahio:


shared/src/business/utilities/getModifiedDocumentTitleWithAdditionalInfo.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * returns a modified document title when adding new information to a previous document title
+ * especially in the case of qc'ing a docket entry
+ *
+ * @param {string} documentMetaData name of the feature
+ * @returns {string} document title after modification
+ */
+
+const getModifiedDocumentTitleWithAdditionalInfo = documentMetaData => {
```
rosiehsmith, line 9: hey i know this is a draft, but just a heads up, we have this other existing utility that seems to do the same thing! `getDocumentTitleWithAdditionalInfo.js`

PR #2877, opened 2022-12-05
Bugfix 9654 docket entry

ujahio:
Notes for this solution:

1. Created a new helper function `getDescriptionDisplay` that calculates a non-persistence docket entry presentation title analogous to documentTitle. This helper function contains rules for presentation of docket entry's title in scenarios like generating coversheet titles, message subject line titles, descriptionDisplay e.t.c.
2. Made correction that guarantees that we are setting the documentTitle of the originally referenced docket entry (previousDocument) without duplicated instances of a docket entry's document's title.
3. This solution also closes devex [1104](https://trello.com/c/oW0I6rkB/1104-standardize-creation-and-retrieval-of-document-title-for-docket-entries).


shared/src/business/utilities/getDescriptionDisplay.test.js
```diff
@@ -0,0 +1,84 @@
...
+  it('returns descriptionDisplay as documentType with no added information', () => {
+    const expected = 'doc type';
+    documentMetaData = {
+      ...documentMetaData,
+      documentTitle: undefined,
+      documentType,
+    };
+    const result = getDescriptionDisplay(documentMetaData);
+
+    expect(result).toEqual(expected);
```
mmarcotte: This like a different approach than other tests in the codebase by using `expected` and not simply using that value in the expect at the bottom.

```

    const result = getDescriptionDisplay({
      documentType,
    });

    expect(result).toEqual('doc type');
```
this could be easier to read and understand. What do you think?

mmarcotte (approved):

shumway-tm (approved):

shared/src/business/utilities/getDescriptionDisplay.test.js
jasonfu9966: We do actually see this pattern all over our tests, but I agree it's easier to read/understand if it's not unnecessarily extracted out.

jasonfu9966 (approved):

shared/src/business/utilities/getDescriptionDisplay.test.js
mmarcotte: You're right. There is a mix of patterns in the tests, and we do see expected. It just seems like that's used for more involved checks of objects rather than strings.

ujahio: @mmarcotte Made the changes as suggested. Thanks for bringing it up

PR #2890, opened 2022-12-06
9745 bugfix

codyseibert:


TomElliottFlexion (approved):

shared/src/business/useCases/courtIssuedDocument/fileAndServeCourtIssuedDocumentInteractor.test.ts
```diff
@@ -344,6 +356,33 @@ describe('fileAndServeCourtIssuedDocumentInteractor', () => {
     expect(expectedDocketEntry.isOnDocketRecord).toBe(true);
   });
 
+  it('should set the filingDate to be today even if it was already set on the past', async () => {
```
rosiehsmith, line 359: what's the point of this unit test when there's an integration test covering the bug?

rosiehsmith: i could totally be missing something.

jasonfu9966: we had a fruitful discussion about this and decided that we're going to employ a unit test instead of an integration test. it's difficult to run the test repeatedly without reseeding the data. the solutions we found made this smell like it should be a unit test.

thanks for your question!

PR #2894, opened 2022-12-07
9684 Message Display: Display Place and Date of Trial Session When Case is Calendared 

mrinsin:
Reindex required! 
Since the story requires a reindex without a migration, or any other special deploy steps, apply these changes to the following flexion branches:
   - [x] experimental1
   - [ ] experimental2
   - [ ] experimental3
   - [ ] experimental4
   - [x] experimental5
   - [x] test

image


web-client/src/presenter/computeds/formattedMessages.test.js
```diff
@@ -488,4 +491,117 @@ describe('formattedMessages', () => {
...
+        },
+        messages: [mockCalendaredMessage],
+        screenMetadata: {},
+        user: {
+          role: 'adc',
+        },
+      },
+    });
+
+    expect(result.messages[0].showTrialInformation).toBeFalsy();
```
ujahio: ❓ Since the test is specifically expecting a boolean, should the expect specifically reflect that? 
`false` vs `falsy`

ujahio (approved):

web-client/src/presenter/computeds/formattedMessages.test.js
mrinsin: Good idea, on it!

PR #2897, opened 2022-12-08
3287:  Update trial status types

TomElliottFlexion:
Requires `setup-updated-trial-session-types-flag.sh` script to be run to initialize feature flag.
Runs `0003-update-trial-session-working-copy-status` migration


rgoshen (approved):

shared/src/business/useCases/trialSessions/generatePrintableTrialSessionCopyReportInteractor.ts
```diff
@@ -45,6 +46,12 @@ export const generatePrintableTrialSessionCopyReportInteractor = async (
     throw new UnauthorizedError('Unauthorized');
   }
 
+  const updatedTrialSessionTypes = await applicationContext
```
rachelschneiderman: 🔧 This doesn't necessarily indicate to me that it's a boolean value. Could it be renamed to express that a little more clearly? For example, `isTrialSessionFilterFeatureEnabled` (not a great name, just an example!)

web-client/src/views/TrialSessionWorkingCopy/WorkingCopySessionList.jsx
```diff
@@ -107,14 +109,17 @@ export const WorkingCopySessionList = connect(
...
-              />
-            );
-          })}
+          <tbody>
+            {trialSessionWorkingCopyHelper.formattedCases.map(item => {
+              return (
+                <CaseListRowTrialSession
+                  formattedCase={item}
+                  key={item.docketNumber}
+                  trialSessionWorkingCopy={trialSessionWorkingCopy}
```
rachelschneiderman, line 118: 🔧 Looks like this component already pulls in this helper off of state, is there a reason it needs to be explicitly passed in?

shared/src/business/entities/EntityConstants.js
```diff
@@ -100,6 +100,10 @@ const ALLOWLIST_FEATURE_FLAGS = {
   STAMP_DISPOSITION: {
     key: 'stamp-disposition-enabled',
   },
+  UPDATED_TRIAL_STATUS_TYPES: {
+    disabledMessage: 'Currently using legacy trial status types.',
```
rachelschneiderman, line 101: Is this being used anywhere?

shared/src/business/entities/EntityConstants.js
```diff
@@ -1228,16 +1232,74 @@ const SECTIONS = sortBy([
...
-];
+const TRIAL_STATUS_TYPES = {
+  setForTrial: {
+    deprecated: true,
+    displayOrder: 999,
+    label: 'Set for Trial',
+  },
+  dismissed: {
+    deprecated: true,
+    displayOrder: 999,
```
rachelschneiderman, line 1240: Why are some of these all '999'? Is that to force them to the bottom of the list?

web-api/migration-terraform/main/lambdas/migrations/0003-update-trial-session-working-copy-status.js
```diff
@@ -0,0 +1,43 @@
...
+        switch (item.caseMetadata[aCase].trialStatus) {
+          case 'aBasisReached':
+            item.caseMetadata[aCase].trialStatus = 'basisReached';
+            break;
+          case 'takenUnderAdvisement':
+            item.caseMetadata[aCase].trialStatus = 'submittedCAV';
+            break;
+        }
+      }
+    }
```
rachelschneiderman, line 45: 🔧 I think this migration should validate items it updates, we have a `validateWithLogging` method that I believe is typically used for migrations!

shared/src/business/useCases/trialSessions/generatePrintableTrialSessionCopyReportInteractor.ts
TomElliottFlexion: Good point, I went with `areUpdatedTrialSessionTypesEnabled`?

web-client/src/views/TrialSessionWorkingCopy/WorkingCopySessionList.jsx
TomElliottFlexion: The `trialSessionWorkingCopy` itself is being passed in, while the component pulls the `trialSessionWorkingCopyHelper` from state. The `CaseListRowTrialSession` could probably also just pull the `trialSessionWorkingCopy` computed from state itself, but it's noticeably faster this way

shared/src/business/entities/EntityConstants.js
TomElliottFlexion: A few places; `setDefaultWorkingCopyValuesAction`, `trialSessionWorkingCopyHelper`, `gotoTrialSessionWorkingCopySequence`

shared/src/business/entities/EntityConstants.js
TomElliottFlexion: It's a fallback, those should never be used since they're deprecated after the flag is switched. We could easily take them out

web-api/migration-terraform/main/lambdas/migrations/0003-update-trial-session-working-copy-status.js
TomElliottFlexion: Oh shoot, do we still have to validate in individual migrations in addition to running `0000-validate-all-items`?

shared/src/business/entities/EntityConstants.js
TomElliottFlexion: Oh, you meant the `disabledMessage`...

TomElliottFlexion: Not explicitly, but `getFeatureFlagValueFactoryAction` would log it anytime the flag is false

shared/src/business/useCases/trialSessions/generatePrintableTrialSessionCopyReportInteractor.ts
rachelschneiderman: Love it! Thanks!

web-client/src/views/TrialSessionWorkingCopy/WorkingCopySessionList.jsx
rachelschneiderman: Ah ok thanks!

shared/src/business/entities/EntityConstants.js
rachelschneiderman: I could be wrong, looking at that action it seems like it won't set the alert warning because `ignorePath` is set to true.

web-client/src/views/TrialSessionWorkingCopy/CaseListRowTrialSession.jsx
```diff
@@ -58,12 +61,27 @@ const getCaseRow = ({
...
-              <option key={key} value={key}>
-                {value}
-              </option>
-            ))}
+            <option value="">-{unassignedLabel}-</option>
+            {Object.keys(trialStatusOptions).map(key => {
+              if (updatedTrialSessionTypesEnabled) {
+                if (
+                  !trialStatusOptions[key].deprecated ||
+                  trialSessionWorkingCopyStatus === key
```
rachelschneiderman, line 69: Can this `trialSessionWorkingCopyStatus` be reused on line 55? And to clarify, this is the case trial status on the working copy, right?

web-api/migration-terraform/main/lambdas/migrations/0003-update-trial-session-working-copy-status.js
rachelschneiderman: We don't necessarily have to! I am in favor because it shortens the feedback loop and also can help us debug when there is a validation error, especially if there are multiple migration scripts. My understanding of the `0000-validate-all-items` script up to this point was it is a safety measure to ensure we aren't migrating any invalid data but I could have misunderstood.

TomElliottFlexion: Fair enough, thanks for clarifying! Adding it now

web-client/src/views/TrialSessionWorkingCopy/CaseListRowTrialSession.jsx
TomElliottFlexion: Correct on both counts!

shared/src/business/entities/EntityConstants.js
TomElliottFlexion: You are not wrong! All things being equal, I'd prefer to simply keep the `disabledMessage` for consistency's sake, but I could be persuaded otherwise

web-client/src/views/TrialSessionWorkingCopy/CaseListRowTrialSession.jsx
TomElliottFlexion: Correction: this `trialSessionWorkingCopyStatus` cannot be reused as the `key` on 55, as this refers to the location on the trialSessionWorkingCopy in which the `trialSessionWorkingCopyStatus` should be saved

mmarcotte (approved):

PR #2902, opened 2022-12-09
DevEx - Removing API Smoketests

codyseibert:
removing the api smoke tests which have been failing on node14 for years without anyone knowing


jasonfu9966 (approved):
nice catch! strange catch. what are the circumstances around this that made it fly under the radar for so long? was the result unreported?

jimlerza (approved):

jimlerza:
> nice catch! strange catch. what are the circumstances around this that made it fly under the radar for so long? was the result unreported?

The axios call was always returning an error, but for some reason the node process itself was returning a 0, which made it "pass" in circleci. As of node 16, the process returns a 1, which results in a failure in circleci.

.circleci/config.yml
```diff
@@ -484,6 +484,7 @@ jobs:
           name: Admin User Setup
           command: |
             if [ "${CIRCLE_BRANCH}" != "prod" ]; then
+              . ./scripts/load-environment-from-secrets.sh
```
mrinsin, line 487: Do we still need to source the vars from secrets? I know this was part of our debugging effort but I may have lost context at some point!

codyseibert: yeah, I think at one point I mentioned we might be able to remove those, but I'm pretty sure that fixed an issue where it kept trying to create the admin user with admin@example.com when the secrets expected it to create ustcadmin@example.com.  Without this change the build creates the wrong admin user name.

mrinsin: Oh that's right - thank you for explaining it to me again!

mrinsin (approved):

scripts/data-import/judge/bulkImportJudgeUsers.helpers.js
```diff
@@ -39,7 +39,7 @@ const init = async (csvFile, outputMap) => {
           } else {
             const services = await getServices();
             endpoint = `${
-              services[`gateway_api_${process.env.DEPLOYING_COLOR}`]
+              services[`gateway_api_${process.env.CURRENT_COLOR}`]
```
rachelschneiderman: This is breaking deploys to experimental environments, I believe we want to import judge users into the `deploying_color` so we can run the smoketests with judges populated. Was this an accidental commit during debugging?

rachelschneiderman: Since Cody is out this week, I made this change 😄

TomElliottFlexion (approved):

mmarcotte (approved):

mmarcotte:
Merging before tests pass (as they will be re-run) and this change only affects CircleCI's `config.yml`

PR #2905, opened 2022-12-13
9684 Message Display: Display Place and Date of Trial Session When Case Status = Calendared

mrinsin:
Reindex required! 
   - [x] staging
   - [x] irs
   - [x] dev
   - [x] prod
   - [x] test

image


shared/src/business/entities/Message.js
```diff
@@ -17,37 +17,49 @@ function Message() {
 }
 
 Message.prototype.init = function init(rawMessage, { applicationContext }) {
+  messageDecorator(this, rawMessage, { applicationContext });
+};
+
+const messageDecorator = (
```
TomElliottFlexion, line 23: This is nice, I like this

shared/src/business/utilities/abbreviateState.test.js
```diff
@@ -0,0 +1,12 @@
+const { abbreviateState } = require('./abbreviateState');
+
+describe('abbreviateState', () => {
```
TomElliottFlexion, line 3: Might it be worth adding a test around multi-word cities? e.g. 'Las Vegas, Nevada'

mrinsin: Ooh thats a great idea, on it!

web-client/src/views/Messages/MessagesIndividualInbox.jsx
```diff
@@ -157,7 +157,15 @@ export const MessagesIndividualInbox = connect(
                   <td className="message-queue-row max-width-25">
                     {message.caseTitle}
                   </td>
-                  <td className="message-queue-row">{message.caseStatus}</td>
+                  {!message.showTrialInformation && (
+                    <td className="message-queue-row">{message.caseStatus}</td>
```
TomElliottFlexion, line 161: These could be if...elses, but that's just a matter of preference

TomElliottFlexion (approved):

web-client/src/views/Messages/MessagesIndividualInbox.jsx
mrinsin: We tried it as an if/else but didn't find it readable (to our eyes at least) so we decided to go down this route.. good suggestion tho!

mmarcotte (approved):

PR #2910, opened 2022-12-14
Devex ts more ts refactor

ujahio:


jimlerza (approved):

types/TEntity.d.ts
```diff
@@ -426,3 +426,16 @@ type TContact = {
...
+  continued: boolean;
+  dismissed: boolean;
+  recall: boolean;
+  rule122: boolean;
+  setForTrial: boolean;
+  settled: boolean;
+  showAll: boolean;
+  statusUnassigned: boolean;
+  takenUnderAdvisement: boolean;
+};
```
rosiehsmith, line 442: Just curious but is having one large file with the type definitions something that's a norm? This files looks like it's getting pretty long, I just don't know if that's something that's going to have an impact at all.

rosiehsmith (approved):

PR #2916, opened 2022-12-19
9667 Auto-Generate Deadline - Order for Amended Petition (OAP)

rosiehsmith:
flexion#9667


web-client/integration-tests/petitionsClerkCaseDeadlineFromOrderForAmendedPetition.test.js
```diff
@@ -0,0 +1,290 @@
...
+        'updateCourtIssuedDocketEntryFormValueSequence',
+        {
+          key: 'eventCode',
+          value:
+            SYSTEM_GENERATED_DOCUMENT_TYPES.orderForAmendedPetition.eventCode,
+        },
+      );
+
+      await cerebralTest.runSequence(
+        'updateCourtIssuedDocketEntryFormValueSequence',
```
rachelschneiderman: Nit: These could be combined into form data below

shared/src/business/useCases/courtIssuedDocument/fileAndServeCourtIssuedDocumentInteractor.test.ts
```diff
@@ -184,35 +186,135 @@ describe('fileAndServeCourtIssuedDocumentInteractor', () => {
...
+      },
+      subjectCaseDocketNumber: caseRecord.docketNumber,
+    });
+
+    expect(
+      applicationContext.getUseCaseHelpers().autoGenerateDeadline,
+    ).not.toHaveBeenCalled();
+  });
+
+  it('should create a deadline on the subject case when docket entry is an Order For Amended Petition', async () => {
```
rachelschneiderman: NIt: Could these tests that check for OAP/OF generating a deadline be combined into one? My thinking is since there is an entity method checking the event code against a list of pre-defined codes that should generate a deadline, one test to make sure that method is called is sufficient

rachelschneiderman (approved):

TomElliottFlexion (approved):

mmarcotte (approved):

PR #2917, opened 2022-12-19
9742 Recent Messages: Update w/ Trial Session Info

mrinsin:
image

### DIFF
--- a/web-client/src/presenter/computeds/formattedMessages.js
+++ b/web-client/src/presenter/computeds/formattedMessages.js
@@ -15,6 +15,20 @@ export const formattedMessages = (get, applicationContext) => {
     tableSort,
   });
 
+  const { STATUS_TYPES } = applicationContext.getConstants();
+
+  messages.forEach(message => {
+    message.showTrialInformation =
+      message.caseStatus === STATUS_TYPES.calendared;
+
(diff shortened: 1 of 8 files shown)


web-client/src/presenter/computeds/formattedMessages.js
```diff
@@ -17,6 +15,8 @@ export const formattedMessages = (get, applicationContext) => {
     tableSort,
   });
 
+  const { STATUS_TYPES } = applicationContext.getConstants();
```
rosiehsmith, line 18: i like that you moved this declaration closer to its usage.

web-client/src/views/WorkQueue/RecentMessagesInbox.jsx
```diff
@@ -76,7 +76,15 @@ export const RecentMessagesInbox = connect(
                   <td className="message-queue-row">
                     <span>{item.caseTitle}</span>
                   </td>
-                  <td>{item.caseStatus}</td>
+                  {!item.showTrialInformation && (
```
rosiehsmith: Nit: I think this could be a ternary

rosiehsmith (approved):

shumway-tm (approved):

PR #2919, opened 2022-12-19
9758 - New endpoint to only fetch judge sessions

codyseibert:
Adding a new endpoint for fetching just the sessions associated with a judge instead of returning every single trial session.

### DIFF
--- a/shared/src/business/useCases/trialSessions/getTrialSessionsForJudgeInteractor.ts
+++ b/shared/src/business/useCases/trialSessions/getTrialSessionsForJudgeInteractor.ts
@@ -1,0 +1,37 @@
+import {
+  ROLE_PERMISSIONS,
+  isAuthorized,
+} from '../../../authorization/authorizationClientService';
+import { TrialSession } from '../../entities/trialSessions/TrialSession';
+import { UnauthorizedError } from '../../../errors/errors';
+
+/**
+ * getTrialSessionsForJudgeInteractor
+ *
+ * @param {object} applicationContext the application context
+ * @returns {Array<TrialSession>} the trial sessions returned from persistence
+ */
+export const getTrialSessionsForJudgeInteractor = async (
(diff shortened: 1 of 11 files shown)


mmarcotte (approved):

shared/src/business/useCases/trialSessions/getTrialSessionsForJudgeInteractor.ts
```diff
@@ -0,0 +1,37 @@
...
+import { TrialSession } from '../../entities/trialSessions/TrialSession';
+import { UnauthorizedError } from '../../../errors/errors';
+
+/**
+ * getTrialSessionsForJudgeInteractor
+ *
+ * @param {object} applicationContext the application context
+ * @returns {Array<TrialSession>} the trial sessions returned from persistence
+ */
+export const getTrialSessionsForJudgeInteractor = async (
```
rachelschneiderman, line 14: 🔧  What do you think about `getTrialSessionsByPresidingJudgeInteractor`? Just so it's clear that not only authenticated judges might use this endpoint

codyseibert: since the interactor requires a judgeId to work, I'm not sure adding Presiding helps me understand the intent any better.  I'd be willing to change it to getTrialSessionsByJudgeUserIdInteractor if that helps?

rachelschneiderman (approved):

shared/src/business/useCases/trialSessions/getTrialSessionsForJudgeInteractor.ts
rachelschneiderman: Gotcha, I was thinking it was more indicative that the trial sessions being returned were those that the provided judge ID was assigned to. I don't feel super strongly though.

rgoshen (approved):

PR #2923, opened 2022-12-20
9759 trial session logout bug

codyseibert:


web-api/src/applicationContext.js
```diff
@@ -711,6 +712,7 @@ module.exports = (appContextUser, logger = createLogger()) => {
         getDocumentTypeForAddressChange,
         getFormattedCaseDetail,
         getStampBoxCoordinates,
+        getTrialSessionStatus,
```
mmarcotte: feels a little persnickety, but I think this might better belong inside `shared/src/business/dto/TrialSessionInfoDto.ts` or a member function of TrialSession... Happy for this to be a future refactoring task that really targets trimming down functions in `getUtilities()` into places where they make better sense. It just feels like a catch all toolbox at the moment. What do you think?

codyseibert: sounds ok since this helper function isn't used in the frontend and only used from other business layer functions / entities.

web-client/src/presenter/state.js
```diff
@@ -303,6 +303,7 @@ export const baseState = {
     showMobileMenu: false,
     showUsaBannerDetails: false,
   },
+
```
TomElliottFlexion: Nit

TomElliottFlexion (approved):

rgoshen (approved):

PR #2926, opened 2022-12-21
devex: remove reindex setup from circle workflow

mmarcotte:
As per parking lot discussion on Tuesday, December 20th, 2022, we decided to remove the setup for reindex workflow from CircleCI. Seemed quicker to complete the work than create a card and vote on it. 

This avoids the confusion of having this run on a production-sized dataset.

* If you need to perform a reindex on a production sized dataset, refer to the [Blue/Green Migration Steps](https://github.com/ustaxcourt/ef-cms/blob/staging/docs/additional-resources/blue-green-migration.md)
* IF you need to perform a reindex on a development dataset, refer to the [web-api/reindex-dynamodb-records.js](https://github.com/ustaxcourt/ef-cms/blob/staging/web-api/reindex-dynamodb-records.js) script.


jimlerza (approved):

.circleci/config.yml
```diff
@@ -572,24 +565,23 @@ jobs:
           command: |
             cat "./env-for-circle/${CIRCLE_BRANCH}.sh" >> $BASH_ENV
             source $BASH_ENV
-            echo "export REINDEX_FLAG=$(./scripts/dynamo/get-reindex-flag.sh $ENV)" >> $BASH_ENV
             echo "export MIGRATE_FLAG=$(./scripts/dynamo/get-migrate-flag.sh $ENV)" >> $BASH_ENV
```
rachelschneiderman: I think we would want this entire step to go away, the `reindex` step defined started on line 553.

.circleci/config.yml
```diff
@@ -909,12 +899,6 @@ jobs:
             if [ $MIGRATE_FLAG == true ]; then
               npm run migration:cleanup -- $ENV
             fi
-      - run:
-          name: Cleanup Deploy Table After Reindex
-          command: |
-            if [ $REINDEX_FLAG == true ]; then
-              npm run reindex:cleanup -- $ENV
```
rachelschneiderman, line 916: Can this script be removed from package.json?

.circleci/config.yml
mmarcotte: Hmm, I think _that_ `reindex` step needs to stay as that's the long-running wait for reindex to finish step that occurs after a migration. 

However, per your comment I did remove this step b75f006b55e15cb6ada8674e97e63c0d895f36b0.

I'm going to run this over on another environment to ensure everything is 🅰️ 👌

rachelschneiderman: I believe the long running step you mentioned is the `wait-for-reindex` which is the approval job defined in the workflows :) 

The `reindex` job will now run enable the `wait-for-reindex` cron and run the `reindex-dynamo-records` script as long as the `migrate` flag is false, I don't think we want that but I could be wrong!

mmarcotte: You are 100% correct, thank you!!! Doing some more test runs before requesting your re-review.

rachelschneiderman (approved):

PR #2931, opened 2022-12-21
Devex: Remove unnecessary setTimeout in integration tests (Part 1 of many)

rachelschneiderman:
Remove setting timeout in tests if the timeout value being set was 30 seconds, that is already defaulted in the jest config file for integration tests, also cleanup test files.


codyseibert:
nice, I never understood why we had those to begin with

mmarcotte (approved):

PR #2951, opened 2022-12-26
Devex: Remove unnecessary timeouts

rachelschneiderman:
Remove setting timeout in tests if the old timeout value being set was 30 seconds, that is already defaulted in the jest config file for integration tests, also cleanup test files


mrinsin (approved):
👑

rgoshen (approved):

TomElliottFlexion (approved):

mmarcotte (approved):

PR #2955, opened 2022-12-27
9759 Refactor Trial Session Status

codyseibert:
this change requires a blue green migration


TomElliottFlexion (approved):

shared/src/business/entities/trialSessions/TrialSession.js
```diff
@@ -481,11 +511,12 @@ const isStandaloneRemoteSession = function (sessionScope) {
  * @returns {TrialSession} the trial session entity
  */
 TrialSession.prototype.setAsClosed = function () {
-  this.isClosed = true;
```
rachelschneiderman, line 484: It looks like there are other places in the code where setting `isClosed` has been removed. I'm good with that and think it's a great idea but in that case, I think that value should be removed off the entity too, since we can no longer rely on it being a valid/up-to-date value.

mmarcotte (approved):

shared/src/business/entities/trialSessions/TrialSession.js
TomElliottFlexion: Forgive me if I'm missing something, but it has been removed, up on what was line 56. It's also removed in the migration

codyseibert: I just removed it from the DTO, but there should only be one place is exists at this point which is in the helper function used in the data migration

shared/src/business/entities/trialSessions/TrialSession.js
```diff
@@ -14,9 +16,29 @@ const {
...
 const { JoiValidationConstants } = require('../JoiValidationConstants');
 
+const getTrialSessionStatus = session => {
+  const allCases = session.caseOrder || [];
+  const inactiveCases = allCases.filter(
+    sessionCase => sessionCase.removedFromTrial === true,
+  );
+
+  if (
+    session.isClosed ||
```
rachelschneiderman: This should be removed now, right?

web-api/migration-terraform/main/lambdas/migrations/0004-set-session-status.test.js
```diff
@@ -0,0 +1,164 @@
...
+        isCalendared: false,
+        sessionStatus: 'New',
+      },
+    ]);
+  });
+
+  it('should set the trial-session sessionStatus to open when calendared', async () => {
+    const items = [mockSession];
+    const results = await migrateItems(items);
+    console.log('results', results);
```
rachelschneiderman: Nit: Log is left here

web-api/migration-terraform/main/lambdas/migrations/0004-set-session-status.test.js
```diff
@@ -0,0 +1,164 @@
...
+      },
+      {
+        pk: 'case|102-20',
+        sk: 'case|102-20',
+      },
+    ]);
+  });
+
+  it('should set the trial-session sessionStatus to new when NOT calendared', async () => {
+    const items = [{ ...mockSession, isCalendared: false }];
```
rachelschneiderman: Nit: Can status be explicitly set to `undefined` (or anything that is NOT new) here so we ensure the test doesn't pass for the wrong reasons if the mockSession is accidentally changed?

web-api/migration-terraform/main/lambdas/migrations/0004-set-session-status.test.js
```diff
@@ -0,0 +1,164 @@
...
+      {
+        ...mockSession,
+        isCalendared: false,
+        sessionStatus: 'New',
+      },
+    ]);
+  });
+
+  it('should set the trial-session sessionStatus to open when calendared', async () => {
+    const items = [mockSession];
```
rachelschneiderman: Same thing here, can sessionStatus be explicitly set?

web-api/migration-terraform/main/lambdas/migrations/0004-set-session-status.test.js
```diff
@@ -0,0 +1,164 @@
...
+
+    expect(results).toEqual([
+      {
+        ...mockSession,
+        sessionStatus: 'Closed',
+      },
+    ]);
+  });
+
+  it('should set the trial-session sessionStatus to open when all cases removed from the trial session and not standaloneRemote', async () => {
```
rachelschneiderman: Seems like the test name and the expectation are in conflict here

shared/src/proxies/trialSessions/getTrialSessionsForJudgeProxy.js
```diff
@@ -9,6 +9,6 @@ const { get } = require('../requests');
 exports.getTrialSessionsForJudgeInteractor = (applicationContext, judgeId) => {
   return get({
     applicationContext,
-    endpoint: `/judges/${judgeId}/trial-sessions`,
+    endpoint: `/judges/${judgeId}/trial-sessions?fields=trialLocation,trialSessionId,sessionStatus,startDate`,
```
rachelschneiderman, line 12: I love the idea of using select fields in our endpoints, are these being used right now though?

web-api/dynamo-export.js
```diff
@@ -18,5 +18,8 @@ documentClient
   })
   .promise()
   .then(documents => {
+    documents.Items.sort((a, b) => {
```
rachelschneiderman, line 21: Did this end up helping make exporting local data easier to diff? 🥳 🎖️

shared/src/business/entities/trialSessions/TrialSession.js
codyseibert: we use that helper function for the migration script, so we can't remove that function yet

web-api/dynamo-export.js
codyseibert: yeah it was easier to see where things were added / removed I think

shared/src/proxies/trialSessions/getTrialSessionsForJudgeProxy.js
codyseibert: not sure what you mean?  are these properties being used?  the ability to pass `fields` in the query string has existed in our code base for 3 years, but I don't think it's been used at all

shared/src/business/entities/trialSessions/TrialSession.js
rachelschneiderman: Ah, ok. If it's only needed for the migration, I would prefer that logic move into the migration script itself so it gets cleaned up when the migration script is deleted. Just my opinion, I won't hold up the PR for that though.

shared/src/proxies/trialSessions/getTrialSessionsForJudgeProxy.js
rachelschneiderman: My question was, how are the fields being consumed/used by the backend? Right now it appears to me like they aren't being used by the interactor

codyseibert: this logic which looks for fields is in the apiGatewayHelper, so you can pass this query string to any endpoint to filter down the data set.

rachelschneiderman: Got it, thank you!

shared/src/business/entities/trialSessions/TrialSession.js
codyseibert: actually I was mistaken, we still use that logic in the removeCaseFromCalendar method.  When recompute and store the status when you remove a case from a trial session.  I'll need to re-read through this, I think we are missing something

web-api/src/lambdaWrapper.js
```diff
@@ -19,6 +19,8 @@ exports.lambdaWrapper = (lambda, options = {}) => {
     const shouldMimicApiGatewayAyncEndpoint =
       options.isAsync && process.env.NODE_ENV != 'production';
 
+    console.log('here', shouldMimicApiGatewayAyncEndpoint);
```
mmarcotte: mayhaps this is leftover from debugging?

PR #2956, opened 2022-12-27
DevEx-1116 - Log when response payload over 5mb

codyseibert:


TomElliottFlexion (approved):

web-api/src/genericHandler.js
```diff
@@ -97,6 +97,16 @@ exports.genericHandler = (awsEvent, cb, options = {}) => {
         });
       }
 
+      const responsePayload = JSON.stringify(returnResults);
+      const responsePayloadSizeBytes = responsePayload.length * 2;
```
mmarcotte: Just curious why the `* 2;`? 🤔

codyseibert: removed, using the Buffer.byteLength approach instead, thank you!

mmarcotte (approved):

PR #2997, opened 2023-01-04
9749: rearrange trial status options

TomElliottFlexion:
Screen Shot 2022-12-30 at 9 18 42 AM


mmarcotte (approved):

rachelschneiderman (commented):
Has DOD been completed for this story?

rgoshen (approved):

TomElliottFlexion:
Checkboxes checked!

shared/src/business/entities/EntityConstants.js
```diff
@@ -1320,6 +1278,48 @@ const TRIAL_STATUS_TYPES = {
...
+    label: 'Motion',
+  },
+  recall: {
+    deprecated: false,
+    displayOrder: 2,
+    label: 'Recall',
+  },
+  continued: {
+    deprecated: false,
+    displayOrder: 4,
```
rachelschneiderman, line 1294: Does `displayOrder` get used? Since this story was concerning reordering the options, I was expecting the display order to have changed for these items but maybe I'm misunderstanding

TomElliottFlexion: The `displayOrder` defines the order in which these values are displayed in the list of filters, while the order of array defines the order in which they are displayed in the status dropdown

TomElliottFlexion: Could probably be named better

PR #3001, opened 2023-01-05
9594 consolidated visibility for petitioner

codyseibert:
This feature is behind a feature flag.


shumway-tm (approved):

web-api/src/v2/getCaseLambda.test.js
```diff
@@ -172,6 +188,100 @@ describe('getCaseLambda (which fails if version increase is needed, DO NOT CHANG
...
+            address3: 'Under the stairs',
+            city: 'Chicago',
+            phone: '+1 (555) 555-5555',
+            postalCode: '61234',
+            state: 'IL',
+          },
+          email: 'ab@example.com',
+          firmName: 'GW Law Offices',
+          name: 'Test Attorney',
+          serviceIndicator: 'Paper',
```
mmarcotte: could this data be defined once and then re-used rather than copy pasting from below? and same for the other tests. it's a little difficult to to tell why we have two tests, but I think it's because you're trying to prove that the format remains the same whether or not the feature flag is `true` or `false`.

mmarcotte (approved):

PR #3029, opened 2023-01-16
DevEx - Convert DocketEntry to typescript

codyseibert:


shared/src/business/entities/DocketEntry.ts
```diff
@@ -0,0 +1,592 @@
...
+  public workItem: any;
+
+  constructor(
+    rawDocketEntry: any,
+    {
+      applicationContext,
+      petitioners,
+    }: { applicationContext: IApplicationContext; petitioners?: any[] },
+  ) {
+    this.entityName = 'DocketEntry';
```
zachrog, line 123: NitPick: I might set entityName on init rather than the constructor so that the constructor is doing less. 
So -> public EntityName: string = 'DocketEntry'

shared/src/business/entities/DocketEntry.ts
```diff
@@ -0,0 +1,592 @@
...
+    this.init(rawDocketEntry, { applicationContext, petitioners });
+  }
+
+  init(
```
zachrog, line 126: Could init be private? Does anything else besides the cosntructor use init()?

shared/src/business/entities/JoiValidationDecorator.ts
```diff
@@ -120,12 +120,12 @@ function getFormattedValidationErrors(entity) {
   return Object.keys(obj).length === 0 ? null : obj;
 }
 
-export interface TValidationEntity {
+export interface TValidationEntity<T> {
```
zachrog: Love the generic.

Not sure what you want to do in the project but generally I have seen interface names prepended with an "I". So if I had a food interface it would be IFood. Wanted to see if you wanted to establish a similar pattern here. so have IValidationEntity<T>.

shared/src/business/entities/trialSessions/TrialSession.ts
```diff
@@ -573,4 +573,5 @@ export const TrialSession: typeof TrialSessionClass =
   validEntityDecorator(TrialSessionClass);
 
 // eslint-disable-next-line no-redeclare
-export interface TrialSessionClass extends TValidationEntity {}
+export interface TrialSessionClass
```
zachrog, line 570: I like the idea that our entities all implement some validationEntity so we know the common validation functions like .validate().

This interface is a little confusing to me though as we are exporting an interface called "TrialSessionClass" and it is not a class. Also higher in this file we are exporting a class called "TrialSessionClass". 

I think what you may be looking for is a single class that implements TValidationEntity. So something like:

export class TrialSessionClass implements TValidationEntity<whateverType> {
constructor()
validate()
blah()
}

zachrog (approved):
Overall I think it looks pretty good and is moving us in the right direction.

A couple of the things I see which I would want to talk to you about are the need to separate the rawData, Entity, and common Validation functions. Those are the common pieces that I am seeing with all the entities.

A way we could address this for a pizza class would be:
interface IRawPizza{
toppings: string[]
}

interface IValidationEntity {
validate():boolean {}
}

class Pizza extends IValidationEntity, IRawPizza {
 constructor(rawPizza: IRawPizza) {
 this.toppings = rawPizza.toppings
 }
}

shared/src/business/entities/DocketEntry.ts
codyseibert: I'll have to look more into init to figure out what's using it.  I don't know why we don't just run the setup logic in the constructor, but I think the joi validation might have something to do with it.

codyseibert: yeah so init is called from inside the joi validation decorator.

```js
  const factoryHandler = {
    construct(target, args) {
      const entityInstance = new target(...args);
      const proxied = new Proxy(entityInstance, instanceHandler);
      proxied.init && proxied.init(...args);
      return proxied;
    },
  };
  ```
  
  I don't think we can make it private, but I do think at some point when we get more entities refactored to typescript we can find an alternative solution to some of this stuff

shared/src/business/entities/trialSessions/TrialSession.ts
codyseibert: I did originally try this approach, but then our entities light up red unless we implement stub methods for all those required interface's method which we end up overwriting later on which I felt was more dirty than my current approach.  The reason for the class TrialSessionClass and the interface TrialSessionClass is that typescript actually merges both of those interfaces together which allows consumers of this file to know that a TrialSessionClass is actually a combination of the class and the joi validation methods together.

TomElliottFlexion (approved):

jimlerza (approved):

PR #3038, opened 2023-01-23
Admin script: monthly paper vs electronic stats

jimlerza:
Since we run this once a month, it's about time we merge it into staging.

Reworked [this script](https://github.com/ustaxcourt/ef-cms/blob/277-more-scripts/shared/admin-tools/elasticsearch/get-petition-counts.js) to use `searchAll` and only retrieve petitions once.


shared/admin-tools/elasticsearch/get-petition-counts.js
```diff
@@ -0,0 +1,88 @@
...
+    },
+  });
+  return results;
+};
+
+const getCounts = ({ gte, lt, petitions }) => {
+  const petitionsReceivedInTimeframe = petitions.filter(
+    p =>
+      dateStringsCompared(p.createdAt, gte) >= 0 &&
+      dateStringsCompared(p.createdAt, lt) < 0,
```
jimlerza: Is it possible for paper petitions' `createdAt` and `receivedAt` timestamps to be different?

To retrieve the year's petitions, we're doing a range query on `receivedAt`, but then to filter the petitions by month, we're filtering on `createdAt`. Doing it this way, the total count will still match, but the monthly counts could differ slightly if it's possible that a petition could be received on the last day of the month and entered into DAWSON some days later.

rgoshen (approved):

jasonfu9966 (approved):

mmarcotte (approved):

PR #3049, opened 2023-01-26
DEP UPDATES Week of 1-23-23 

codyseibert:
# Before This Change

- Running pdf generation on local computers was failing due to improper docker file setup
- To test pdf generation we would first run the npm script test:pdf-output to create the pdfs and then run node image-compare-pdfs.js to generate images from the pdf and compare
 - When generating pdfs we used the local system date time which ended up on the pdf and always displayed a difference between runs
 - node and npm versions were not enforced

# After This Change
- Remove unnecessary docker file steps for local pdf generation
- Reduce pdf generation down to a single npm command instead of two. New command is npm run test:document-generation
 - Mock out the dateTime for all pdf generation tests so that we can tell if the pdfs are legitimately changing or not.
 - node and npm versions will be enforced with the engines object and npmrc file.

# Why This change

- Pdf generation and testing was split into two unnecessary steps that NEEDED one another to run. It made sense to combine into one single test
- Using the system clock caused pdf differences to always be detected.
- Reduce all of the duplication in the pdf generate steps to a single common function which generates a pdf, converts to image, compares to expected image, and fails a test if they are off by a particular threshold.
- Enforcing Node versions can help reduce package-lock issues and ensure all developers are using the correct version of node and npm

# NOTES
Because we are not hardcoding the dates across all test runs we regenerated all expected.pngs. These files can be ignored.
We refactored all pdf generation tests to use the new generateAndVerifyPdfDiff() function. All files inside business/utilities/documentGeneration are just using this new function. That is the only change that exist in these files.
90% of the updated files in this PR are updating pngs and converting old tests to use generateAndVerifyPdfDiff(), not functional changes to the app.

successful migration run: https://app.circleci.com/pipelines/github/flexion/ef-cms/40867/workflows/f51c7a49-6c1a-4e49-aaa6-37975fc32a6c


rgoshen (approved):
Nice refactoring of the PDF tests. 🎉

Dockerfile
```diff
@@ -1,42 +1,41 @@
...
-RUN apt-get -o Acquire::Check-Valid-Until=false update
-RUN apt-get install -y -t \
-  stretch-backports \
-  openjdk-11-jre-headless=11.0.16+8-1~deb11u1 \
-  openjdk-11-jdk-headless=11.0.16+8-1~deb11u1 \
-  openjdk-11-jre=11.0.16+8-1~deb11u1 \
-  openjdk-11-jdk=11.0.16+8-1~deb11u1 \
+RUN apt-get update
+
+RUN apt-get install -y \
```
jimlerza, line 12: 🥳

jimlerza (approved):
Great refactor!

shared/src/business/utilities/documentGenerators/generateAndVerifyPdfDiff.js
```diff
@@ -0,0 +1,110 @@
...
+  applicationContext,
+} = require('../../test/createTestApplicationContext');
+const {
+  generatePdfFromHtmlInteractor,
+} = require('../../useCases/generatePdfFromHtmlInteractor');
+const { fromPath } = require('pdf2pic');
+const { getChromiumBrowser } = require('../getChromiumBrowser');
+const { PNG } = require('pngjs');
+
+const convertPdfPageToImageFile = async ({ fileName, pageNumber }) => {
```
ujahio, line 14: 👏🏿 👏🏿

ujahio (approved):

PR #3056, opened 2023-01-30
devex-1082: add more metrics for segments

mmarcotte:
to inform further analysis, it would be helpful to know:

* number of pks per segment
* record size per segment?
* what are the pks for the largest segments 
    * by count
    * by recordsize
* how long did the larger segments take to run


rgoshen (approved):

jimlerza (approved):

setup-for-blue-green-migration.sh
```diff
@@ -14,6 +14,33 @@ set -e
   "AWS_SECRET_ACCESS_KEY"
 set +e
 
+function check_opensearch_domain_exists() {
+  ENV=$1
+  NEXT_VERSION=$1
```
jimlerza: I think you mean `$2` here

mmarcotte: 🙏

setup-for-blue-green-migration.sh
```diff
@@ -14,6 +14,33 @@ set -e
...
+
+  aws es describe-elasticsearch-domain --domain-name "efcms-search-${ENV}-${NEXT_VERSION}" --region us-east-1 > /dev/null
```
jimlerza: You may also want to pipe `stderr` to `/dev/null` as well, like so:

```
command > /dev/null 2>&1
```

PR #3063, opened 2023-02-01
devex-1057: Remove CircleCI Environment Variables

mmarcotte:
To wrap up devex-1057, we are consolidating the environment variable gathering from AWS Secrets into a single script that runs to setup the Circle Environment based off of the `$CIRCLE_BRANCH`. 

This script will add all of those secrets to `$BASH_ENV`, which on CircleCI points to `/etc/profile`. This way each subsequent step will make use of `$BASH_ENV` and be aware of all of the pre-calculated values. 

This has the added benefit of injecting commonly regenerated environment variables like `DEPLOYING_COLOR` or `SOURCE_TABLE` into one location. Easy to add more as they are needed! 

Remaining required variables in CircleCI are:

* `AWS_ACCESS_KEY_ID`
* `AWS_SECRET_ACCESS_KEY`
* `CIRCLE_MACHINE_USER_TOKEN`
* `AWS_ACCOUNT_ID`
* `PDF_EXPRESS_LICENSE_KEY` (mostly deprecated)

We made the executive decision that it's easier to maintain/rotate `CIRCLE_MACHINE_USER_TOKEN` if it's stored in one location rather than every lower environment's Secrets Manager resource. It is shared for all environments (lower and prod) and is based off of a User's token (@mmarcotte 's). We could consider adding support for an environment specific token that is stored in Secrets Manager and creating more tokens for the user(s) performing the updates, but that need is not presently apparent.

And it might be time to fully deprecate PDF Express.


jimlerza (approved):

codyseibert (approved):

.circleci/config.yml
```diff
@@ -440,10 +440,7 @@ jobs:
       - run:
           name: Setup Env
           command: |
-            cat "./env-for-circle/${CIRCLE_BRANCH}.sh" >> $BASH_ENV
-            source $BASH_ENV
-            echo "export DEPLOYING_COLOR=$(./scripts/dynamo/get-deploying-color.sh $ENV)" >> $BASH_ENV
```
codyseibert, line 444: 👋 bye bye

PR #3070, opened 2023-02-03
DevEx - Canceled Jobs in Github Actions

codyseibert:
This PR addressed the issue with our github action jobs being canceled for our cypress integration tests.  

### The Root Cause

We are pretty sure the jobs getting canceled are related to CPU or memory usage on the runner.  We found a couple of other posts online that pointed to resource limits inside the container, and we know running cypress along with all our services and UI does require a lot of memory and cpu.  

### The Fix
There were three things we had to do to get these tests consistent:

- use the `macos-latest` larger runner with more cpu and memory
- increase the cypress retries to 4 (it was zero before)
- refactor all `before` calls; instead, we now call all cy commands inside an `it` block.  

The reason for the before change isn't clear to me, but I noticed that even with the retries set to 4, the before commands were still failing on the first try.  I figured there is something in cypress which doesn't retry when running before logic, so I moved the logic into the it statements which has resulted in more consistent tests.  There is an open issue on cypress related to this documenting how before statements do not retry on failure: https://github.com/cypress-io/cypress/issues/19458


rosiehsmith (approved):
This is awesome - thanks Cody and Zach! I noticed there's also a `retries` in `cypress-smoketests.config.js` - is there any need to up that too or is this scoped only to integration tests?

codyseibert:
> This is awesome - thanks Cody and Zach! I noticed there's also a `retries` in `cypress-smoketests.config.js` - is there any need to up that too or is this scoped only to integration tests?

We may need to revisit the smoketests if they also seem to be randomly failing.  I didn't want to address them yet if they don't seem like they are failing.

cypress.config.js
```diff
@@ -25,7 +25,7 @@ module.exports = defineConfig({
     toConsole: true,
   },
   requestTimeout: 60000,
-  retries: 0,
+  retries: 4,
```
zachrog, line 28: Nice. wrapping the Beforeall in an it block should hopefully help the tests run more smoothly.

I am okay with this retry level logic as it will re-run an individual describe block, however we should have some caution as we proceed as certain retry logic can be destructive for the team. Previously I was on a team where e2e tests were re-run from the pipeline level up to 3 times if they failed. This lead to e2e tests that could take an hour to run and the tests to be written with a little less care, as it did not matter as much if an individual test was a little flakey.

zachrog (approved):

cypress.config.js
codyseibert: yeah this used to be set to 4 for the longest time, I reduced it to 0 maybe a month or two ago because I thought these were more consistent, but I guess I was wrong by moving it down to 0.  I personally would like all the retries to be 0 if possible since it would mean our tests are resilient, but I guess the state of e2e tests makes that goal hard.

jimlerza (approved):

mmarcotte (approved):

PR #3076, opened 2023-02-07
Devex 1126: Dangerously Overloaded endpoint

rachelschneiderman:
Response size pre devex-1126:
image

Response size post devex-1126:
image

that's a 84.19% response size reduction!!!


mmarcotte (approved):

TomElliottFlexion (approved):

rgoshen (approved):

jimlerza (approved):

rosiehsmith (approved):

web-client/src/presenter/actions/CaseConsolidation/getConsolidatedCasesByCaseAction.test.js
```diff
@@ -4,68 +4,71 @@ import { presenter } from '../../presenter-mock';
...
     });
 
     expect(
       applicationContext.getUseCases().getConsolidatedCasesByCaseInteractor.mock
-        .calls.length,
-    ).toEqual(0);
+        .calls[0][1],
+    ).toEqual({
+      docketNumber: mockLeadDocketNumber,
+      pickFields: [
```
rosiehsmith, line 61: Thank you for making this explicit!

zachrog (approved):

shared/src/proxies/getConsolidatedCasesByCaseProxy.js
```diff
@@ -10,10 +10,10 @@ const { get } = require('./requests');
  */
 exports.getConsolidatedCasesByCaseInteractor = (
   applicationContext,
-  { docketNumber },
+  { docketNumber, pickFields },
 ) => {
   return get({
     applicationContext,
-    endpoint: `/cases/${docketNumber}/consolidated-cases`,
+    endpoint: `/cases/${docketNumber}/consolidated-cases?fields=${pickFields}`,
```
codyseibert, line 17: I think the approach we normally take is the pass a params object instead of setting the query string manually, like so:

```
return get({
    applicationContext,
    endpoint: '/case-documents/opinion-search',
    params: {
      ...omit(searchParams, 'opinionTypes'),
      opinionTypes: opinionTypesQuery,
    },
  });
  ```

codyseibert, line 17: nit: is pickFields an array, or a string with commas a this point?  I would think pickFields should be an array and the proxy knows how to .join(',').  I was trying to look through the code to find where this comes from but I couldn't find it.

PR #3085, opened 2023-02-09
9755 story with feedback addressed

rosiehsmith:
all the smaller associated PRs here: https://github.com/flexion/ef-cms/pulls?q=is%3Apr+is%3Aclosed+9755

https://github.com/flexion/ef-cms/issues/9755

This PR also contains a few feedback items addressed:
- Issue: When the CSS assigns a work item in the Petitions or Docket QC section display, that item disappears completely from the display regardless of view or filtering. This behavior does not occur when the same operation is done by Petitions or Docket clerk roles. 
- Issue: items processed by CSS are not showing on the Processed view for Section Document QC.


shared/src/business/useCases/editDocketEntry/completeDocketEntryQCInteractor.ts
```diff
@@ -205,10 +207,17 @@ export const completeDocketEntryQCInteractor = async (
     user,
   });
 
+  // if CSS user completes a docket section QC item
```
codyseibert: nit: was this comment intentionally to leave it?  the code below seems straight forward enough to not need this comment imo.  If this is logic we will need up doing everywhere in the app for this type of role, maybe we can move it somehow to be inside the entity; just a thought.

web-client/src/views/WorkQueue/RecentMessagesInbox.jsx
```diff
@@ -85,7 +85,7 @@ export const RecentMessagesInbox = connect(
                     <td className="message-queue-row">{item.caseStatus}</td>
                   )}
                   <td>{item.from}</td>
-                  <td>{item.fromSection}</td>
+                  <td>{item.fromSectionFormatted || item.fromSection}</td>
```
codyseibert, line 88: nit: maybe move to computed

web-client/src/presenter/computeds/formattedWorkQueue.filterWorkItems.test.js
```diff
@@ -331,13 +331,13 @@ describe('filterWorkItems', () => {
     applicationContext.getCurrentUser.mockReturnValueOnce(docketClerk1);
     filterWorkItems({
       applicationContext,
-      section: 'docket',
+      section: DOCKET_SECTION,
```
codyseibert, line 334: 😄

codyseibert (approved):

shared/src/business/useCases/editDocketEntry/completeDocketEntryQCInteractor.ts
rosiehsmith: we're in the process of removing this! i left that in to tell mrinalini yesterday

rosiehsmith: i like the thought of moving it to the entity

web-client/src/views/WorkQueue/RecentMessagesInbox.jsx
mrinsin: Good idea! We are trying to get this PR merged before demo, is it ok to address this piece of feedback in our next PR?

PR #3144, opened 2023-02-17
DevEx 1115: add a new `wait-for-migration` step to the deployment workflow

jimlerza:
[Link to the Trello card](https://trello.com/c/VKK12G8L/1115-move-long-running-run-migration-step-into-wait-for-reindex-operation)

### Summary of this story: 
CircleCI has a one-hour limit on workflow jobs to remain in the free tier, so we made these changes to introduce a `wait-for-migration` job that will support migrations taking longer than an hour.

### Things we did:
- Refactored individual workflow job terraform directories to be more organized and DRY
- Created a new `wait-for-migration` job as a lambda (`migration-status`), created a helper for it, and deprecated the old `wait-for-migration.sh`
- Added a script to abort an unrecoverable migration and set the environment up so that another migration can be performed.

### How things are different:
- Previously, `wait-for-migration.sh` would return an error code if it encountered an error with an ongoing migration, resulting in the Circle workflow appearing as `failed`. Now, if the migration status lambda encounters an error with an ongoing migration, it will cancel the workflow because approval jobs cannot be marked as `failed`. 
- The `migration-terraform` directory has moved to `web-api/workflow-terraform/migration`. Documentation for writing new migrations has been updated.

### Manual steps:

- [x] Immediately after merging this PR, you will need to run the environment-specific terraform to deploy the new roles to IAM:
```
source scripts/env/set-env.zsh ustc-test
cd iam/terraform/environment-specific/main
../bin/deploy-app.sh "$ENV"
```


web-api/workflow-terraform/migration-cron/main/lambdas/migration-status.js
```diff
@@ -0,0 +1,102 @@
...
+
+  // it's possible the queue has caught up but there are still more segments to process
+  // don't approve the job if this is the first consecutive time the queue is empty
+  results.totalActiveJobs = await getSqsQueueCount(workQueueUrl);
+  if (results.totalActiveJobs === -1) {
+    return context.succeed(results);
+  }
+  if (results.totalActiveJobs > 0) {
+    await putMigrationQueueIsEmptyFlag(false);
+  } else {
```
mmarcotte: this may need to `return context.succeed(results);`, and then you can remove the `else { ... }` block

zachrog (approved):

web-api/workflow-terraform/migration-cron/main/lambdas/migration-status.js
jasonfu9966: updated!

PR #3150, opened 2023-02-20
9599 to test

TomElliottFlexion:
Hey we tried a new strategy with this one.

We had a feature flag for 9599. Instead of working in the feature flag throughout the getCasesForUserInteractor.ts we decided to keep the old interactor in  getCasesForUserInteractorOld.ts  and have the new interactor branch off of it. We decided to do this because this is theoretically temporary code while the feature flag is turned off, and when it comes time to remove the feature flag it will be much easier to remove the old implementation. Let us know if you like/hate what we tried to accomplish.


zachrog (approved):

web-client/src/views/WorkQueue/SectionWorkQueueInbox.tsx
```diff
@@ -131,21 +132,11 @@ SectionWorkQueueTable.Row = React.memo(
...
-                  aria-label={item.consolidatedIconTooltipText}
-                  className="fa-icon-blue"
-                  icon="copy"
-                />
-                {item.inLeadCase && (
-                  <span className="fa-inverse lead-case-icon-text">L</span>
-                )}
-              </span>
-            )}
+            <ConsolidatedCaseIcon
```
rachelschneiderman, line 135: 🔧 To me, this reads as though this icon will be rendered for each item in the table, would it be possible to extract the logic that determines whether or not to render the component outside of the component itself?

web-client/src/views/WorkQueue/SectionWorkQueueInbox.tsx
```diff
@@ -131,21 +132,11 @@ SectionWorkQueueTable.Row = React.memo(
...
+              consolidatedIconTooltipText={item.consolidatedIconTooltipText}
+              inConsolidatedGroup={item.inConsolidatedGroup}
+              showLeadCaseIcon={item.inLeadCase}
```
rachelschneiderman, line 138: Typo here? Should this read `item.isLeadCase`?

web-client/src/presenter/computeds/trialSessionWorkingCopyHelper.ts
```diff
@@ -77,7 +77,7 @@ export const trialSessionWorkingCopyHelper = (get, applicationContext) => {
   );
 
   leadAndUnconsolidatedCases.forEach(caseToUpdate => {
-    if (caseToUpdate.leadCase) {
+    if (caseToUpdate.isLeadCase) {
```
rachelschneiderman, line 80: 👏

web-client/src/views/WorkQueue/SectionWorkQueueInbox.tsx
zachrog: Good question. Had the same thought. In this case we are displaying a workItem which is IN a lead case and not a lead case itself. However, we still want to show the lead case icon for this work item. Let us know if you have another idea of maybe how to deal with that.

shared/src/business/useCases/getCasesForUserInteractorOld.ts
```diff
@@ -0,0 +1,153 @@
...
+import { UserCase } from '../entities/UserCase';
+import { compareISODateStrings } from '../utilities/sortFunctions';
+import { partition, uniqBy } from 'lodash';
+
+type TAssociatedCase = {
+  isRequestingUserAssociated: boolean;
+  consolidatedCases?: TCase[];
+} & TCase;
+
+/**
```
ujahio, line 11: 👏🏿 👏🏿

web-client/src/views/WorkQueue/SectionWorkQueueInbox.tsx
zachrog: We went back and forth on this for a little bit. We ended up putting the show/hide logic inside of the ConsolidatedCaseIcon.tsx so that it was easier for other components to use. Not sure which way we prefer.

zachrog: This makes sense. We went back and forth on this one. We thought it might be easier for other components to use if we put the show/hide logic inside of the ConsolidatedCaseIcon.tsx, which comes at the expense of obfuscating the show hide.

rgoshen (approved):

ujahio (approved):

PR #3155, opened 2023-02-22
9595: Visibility Across Consolidated Case Group

TomElliottFlexion:


zachrog (approved):

shared/src/business/entities/cases/Case.isUserPartOfGroup.test.js
```diff
@@ -0,0 +1,83 @@
...
+          irsPractitioners: [],
+          petitioners: [{ contactId: 'xyz' }],
+          privatePractitioners: [{ userId: 'abc' }],
+        },
+      ],
+      userId: 'abc',
+    });
+
+    expect(isPartOfCase).toBe(true);
+  });
```
mmarcotte: I think this is meant to have `{ userId: 'abc' }` be in the `irsPractitioners` array

mmarcotte (changes requested):
Looks great! One pretty straightforward test fix from me.

mmarcotte (approved):
thanks!

jimlerza (approved):

PR #3176, opened 2023-02-24
DevEx 1115: add a new `wait-for-migration` step to the deployment workflow

jimlerza:
[Link to the Trello card](https://trello.com/c/VKK12G8L/1115-move-long-running-run-migration-step-into-wait-for-reindex-operation)

### Summary of this story: 
CircleCI has a one-hour limit on workflow jobs to remain in the free tier, so we made these changes to introduce a `wait-for-migration` job that will support migrations taking longer than an hour.

### Things we did:
- Refactored individual workflow job terraform directories to be more organized and DRY
- Created a new `wait-for-migration` job as a lambda (`migration-status`), created a helper for it, and deprecated the old `wait-for-migration.sh`
- Added a script to abort an unrecoverable migration and set the environment up so that another migration can be performed.

### How things are different:
- Previously, `wait-for-migration.sh` would return an error code if it encountered an error with an ongoing migration, resulting in the Circle workflow appearing as `failed`. Now, if the migration status lambda encounters an error with an ongoing migration, it will cancel the workflow because approval jobs cannot be marked as `failed`. 
- The `migration-terraform` directory has moved to `web-api/workflow-terraform/migration`. Documentation for writing new migrations has been updated.

### Manual steps:

- [x] Immediately after merging this PR, you will need to run the environment-specific terraform to deploy the new roles to IAM:
```
source scripts/env/set-env.zsh ustc-stg
npm run deploy:environment-specific -- "$ENV"
```
- [x] Also immediately after merging this PR, you will need to remove the `reindex_status_lambda_stg` lambda, if it exists:
```
aws lambda delete-function --function-name reindex_status_lambda_stg
```
- [ ] After merging this PR, the two steps above will also need to be performed for every other environment once this code is backmerged into a branch that is deployed there:
    - [ ] Flexion
        - [ ] exp1
        - [x] exp2
        - [ ] exp3
        - [ ] exp4
        - [x] exp5
    - [x] USTC
        - [x] test
        - [x] stg
        - [x] irs
        - [x] dev
        - [x] exp1


rachelschneiderman:
Looking at the manual step instructions, can we still use the script `npm run deploy:environment-specific` or should that be removed with these changes?

jimlerza:
> Looking at the manual step instructions, can we still use the script `npm run deploy:environment-specific` or should that be removed with these changes?

I believe you have to pass in the environment name, so you'd have to invoke it thusly:

```
npm run deploy:environment-specific -- $ENV
```

But yeah that's totally the better way to call it. I've updated the steps.

mmarcotte (approved):

PR #3183, opened 2023-02-28
9599/9597: Consolidated Group Display on My Cases Page

TomElliottFlexion:


web-client/src/ustc-ui/Icon/ConsolidatedCaseIcon.tsx
```diff
@@ -7,19 +7,24 @@ import React from 'react';
  * @param {object} caseItem the case to be passed to the FontAwesomeIcon
  * @returns {object} a react component
  */
-export const ConsolidatedCaseIcon = ({ caseItem }) => {
+export const ConsolidatedCaseIcon = ({
+  consolidatedIconTooltipText,
+  inConsolidatedGroup,
+  showLeadCaseIcon,
+}: {
+  inConsolidatedGroup: string;
```
mmarcotte, line 15: should `inConsolidatedGroup` be a string if it's used as a truthy value on line 20? This would get outputted. Perhaps a `boolean` is better?

web-client/src/ustc-ui/Icon/ConsolidatedCaseIcon.tsx
```diff
@@ -7,19 +7,24 @@ import React from 'react';
...
+  showLeadCaseIcon: boolean;
+}) => {
   return (
-    caseItem.inConsolidatedGroup && (
-      <span
-        className="fa-layers fa-fw"
-        title={caseItem.consolidatedIconTooltipText}
-      >
+    inConsolidatedGroup && (
+      <span className="fa-layers fa-fw" title={consolidatedIconTooltipText}>
```
mmarcotte: should `inConsolidatedGroup` be a `string` if it's used as a truthy value on line 20? This would get outputted. Perhaps a `boolean` is better?

TomElliottFlexion: Nice catch, thanks! It _is_ a boolean

mmarcotte (approved):

PR #3199, opened 2023-03-02
9767 Case Status History

codyseibert:


shared/src/business/entities/cases/Case.test.ts
```diff
@@ -1264,4 +1348,6 @@ describe('Case entity', () => {
       expect(myCase.getFormattedValidationErrors()).toBe(null);
     });
   });
+
+  describe;
```
mmarcotte: Mayhaps this is left in here inadvertently?

mmarcotte (approved):

PR #3204, opened 2023-03-03
DevEx - Adding Abstract JoiValidationEntity

codyseibert:
an attempt to move away from the joiValidationDecorator and instead use an OOP approach with abstract classes.  Making an abstract JoiValidationEntity will allow all entities to extend this type of validation entity and get the expected methods such as validate without having to add a bunch of extra typescript hacks.

So instead of having to call 
```js
joiValidationDecorator(
  TrialSessionClass,
  
  ...
  
 and ...

export const TrialSession: typeof TrialSessionClass &
  TStaticValidationMethods<TRawTrialSession> =
  validEntityDecorator(TrialSessionClass);

// eslint-disable-next-line no-redeclare
export interface TrialSessionClass
  extends IValidationEntity<TrialSessionClass> {}
  
  ```
  
inside every single entity, we'd just `extend JoiValidationEntity`

### CODE REVIEW

File: shared/src/business/entities/JoiValidationEntity.ts [hunk 1]
```diff
@@ -0,0 +1,158 @@
...
+
+const setIsValidated = obj => {
+  Object.defineProperty(obj, 'isValidated', {
+    enumerable: false,
+    value: true,
+    writable: false,
+  });
+};
+
+/**
```
TomElliottFlexion, line 14: Nit: jsDocs

TomElliottFlexion (approved):

rgoshen (approved):

shared/src/business/entities/JoiValidationEntity.ts
```diff
@@ -0,0 +1,238 @@
...
+    const rules = this.getValidationRules();
+
+    Object.defineProperty(this, 'schema', {
+      enumerable: false,
+      value: joi.object().keys(rules),
+      writable: true,
+    });
+  }
+
+  abstract getValidationRules(): any;
```
zachrog, line 139: Is there any reason we need to have a function which fetches the validation rules? Instead could we just set the schema property directly? 

So schema would be an abstract property that all classes implementing JoiValidationEntity would set on their own.

codyseibert: when you put the schema directly on the object, it tries to get stored to dynamodb when we persist.  We could potentially have the schema be returned from a getter function, but I wanted to keep it off of the object as a public field.  You can't use abstract private in typescript.

zachrog (approved):

mmarcotte (approved):

PR #3205, opened 2023-03-04
Devex 1082 refactor overloaded pks

mmarcotte:
Start archiving old user and section outboxes onto primary keys that are more exclusive. By having fewer records sharing the same pk, migrations and glue jobs encounter fewer throttles and take less time. While these archived records aren't used by the UI at the moment, we are keeping them around for now. And we are adding a time-to-live (ttl) on the records that make use of `user-outbox|...` and `section-outbox|...` so they don't stick around longer than 8 days.


codyseibert (approved):
👍

rgoshen (approved):

PR #3240, opened 2023-03-10
9882: Fix overloaded cases endpoint to test

mmarcotte:


rachelschneiderman:
Is this for the IRS practitioner bug Tenille was mentioning earlier this week?

mmarcotte:
> Is this for the IRS practitioner bug Tenille was mentioning earlier this week?

it sure is. 

I was trying to troubleshoot it with this PR. I've since created [an issue](https://github.com/flexion/ef-cms/issues/9882).

TomElliottFlexion (approved):

shared/src/business/useCases/getCasesForUserInteractor.ts
```diff
@@ -90,7 +91,18 @@ const sortAndFilterCases = (nestedCases, caseType: 'open' | 'closed') => {
       } else {
         return compareISODateStrings(b.createdAt, a.createdAt);
       }
-    });
+    })
+    .map(nestedCase => ({
+      ...new UserCase(nestedCase).toRawObject(),
+      consolidatedCases: nestedCase.consolidatedCases
```
rachelschneiderman, line 97: Previously to solve this problem, we have made DTOs that specify the fields needed for that view and provided them in the request so that `apiGatewayHelper` can use them to filter the data being returned. Could we make use of that strategy here at all?

mmarcotte: Possibly, but more work would have to be made to either `apiGatewayHelper` or the interactor (or even the whole endpoint) because this interactor currently returns two objects:

```
return { closedCaseList: sortedClosedCases, openCaseList: sortedOpenCases };
```

Previously this endpoint only returned UserCase records since that is what consumes this endpoint. Once we added consolidated cases, the data became polluted with case meta data of the consolidated cases. Personally, it doesn't feel desirable to ever include that superfluous information.

rachelschneiderman: Thanks, that makes sense. The more I thought about it, the more I thought adding this to the UserCase entity makes more sense.

Do any tests need to be added or updated after this change?

mmarcotte: Good idear, I've included one to cover ensuring `docketEntries` will be undefined.

rachelschneiderman (approved):

PR #3241, opened 2023-03-10
9861 Add Motion Filter to Docket Record

rosiehsmith:
image
image

flexion#9861


shared/src/business/useCases/createCaseFromPaperInteractor.ts
```diff
@@ -102,7 +102,6 @@ export const createCaseFromPaperInteractor = async (
     { applicationContext },
   ).validate();
 
-  // invoke the createCase interactor
```
rachelschneiderman: Nice catch

rosiehsmith: that's owed to those who made this change on staging :)

shared/src/business/entities/EntityConstants.ts
```diff
@@ -413,6 +413,14 @@ export const INTERNAL_DOCUMENTS_ARRAY = flatten(
   Object.values(DOCUMENT_INTERNAL_CATEGORIES_MAP),
 );
 
+export const MOTION_EVENT_CODES = [
+  ...DOCUMENT_INTERNAL_CATEGORIES_MAP['Motion'].map(entry => {
```
rachelschneiderman, line 417: During the amicus brief story, we discovered that internal and external filing events motion categories are different (one may be a subset of the other, I'm honestly not sure). Is there any concern that using just the internal list here will miss any e-filed motions?

web-client/integration-tests/journey/docketClerkAddsDocketEntryForMotion.js
```diff
@@ -0,0 +1,39 @@
...
+    await cerebralTest.runSequence('gotoAddPaperFilingSequence', {
+      docketNumber: cerebralTest.docketNumber,
+    });
+
+    const formFields = {
+      dateReceivedDay: 12,
+      dateReceivedMonth: 2,
+      dateReceivedYear: 2002,
+      eventCode: 'M000',
+      freeText: 'making my way downtown, walking fast, all tests pass',
```
rachelschneiderman, line 15: 😆

web-client/integration-tests/journey/docketClerkAddsDocketEntryForMotion.js
```diff
@@ -0,0 +1,39 @@
...
+      hasOtherFilingParty: true,
+      otherFilingParty: 'Vanessa Carlton',
+    };
+
+    for (const [key, value] of Object.entries(formFields)) {
```
rachelschneiderman, line 20: 👏

rachelschneiderman (approved):

shared/src/business/entities/EntityConstants.ts
rosiehsmith: Seems like a good thing to check. I did notice that internal contained the generic "Motion" document type, M000, while external didn't.

rosiehsmith: Okay, I got the difference between internal and external motions, and it looks like internal has an another 9 motion types: ['M000', 'M031', 'M032', 'M048', 'M057', 'M079', 'M082', 'M086', 'M123']. Do you think it seems valid to use internal then since it has more types? All 128 motion types for external exist within internal.

rachelschneiderman: Thanks for doing that investigation, sounds like it's good to merge then

PR #3257, opened 2023-03-13
9796: QC Display: Add Date/Place of Trial to Calendared Work Items

shumway-tm:
- Display the date and place of trial session on the QC display when the case the work item is attached to is Calendared.

#### My Inbox

image

---

#### My Processed (Outbox)

717BD00E-BA50-4D0B-9BA7-7CD2C40CFB93

---

#### Section Inbox

A1BBC275-4931-4519-A092-689EA6DA7AF5

---

#### Section Processed (Outbox)

42BE036C-3507-4B67-AED8-0FB3F63A3631

---

#### Section In Progress

ED223278-95BE-462F-9181-600823F3EDA1


shared/src/persistence/dynamo/workitems/updateAttributeOnDynamoRecord.ts
rosiehsmith, line 1: ❤️

shared/src/business/entities/OutboxItem.ts
```diff
@@ -22,7 +22,8 @@ export class OutboxItemClass {
   public leadDocketNumber: string;
   public section: string;
   public assigneeId: string;
-  public trialDate: string;
+  public trialDate?: string;
```
rosiehsmith, line 25: trialDate is optional on OutboxItem now?

web-api/elasticsearch/efcms-work-item-mappings.js
```diff
@@ -13,7 +13,7 @@ module.exports = {
       type: 'join',
     },
     'completedAt.S': {
-      type: 'text',
+      type: 'date',
```
rosiehsmith, line 16: Nice.

rosiehsmith (approved):

shared/src/business/entities/OutboxItem.ts
shumway-tm: This was more to bring it in line with this field being optional on a WorkItem based on the assumption that an OutboxItem is simply a completed WorkItem. If that is incorrect I am happy to change it back!

shumway-tm: Also on further thought, I believe it is optional and allow(null) in the Joi Validation Rules if that makes a difference.

rosiehsmith: Makes sense! Thanks for clarifying that.

PR #3271, opened 2023-03-16
9803 Bug: Page count not updated after editing docket entry meta

rachelschneiderman:


rosiehsmith (approved):

shared/src/business/useCases/docketEntry/updateDocketEntryMetaInteractor.ts
```diff
@@ -191,20 +171,16 @@ export const updateDocketEntryMetaInteractor = async (
...
+    const { numberOfPages } = await applicationContext
       .getUseCaseHelpers()
       .removeCoversheet(applicationContext, {
         docketEntryId: originalDocketEntry.docketEntryId,
         docketNumber: caseEntity.docketNumber,
-        filingDateUpdated,
       });
-    caseEntity.updateDocketEntry(updatedDocketEntry);
+
+    docketEntryEntity.setNumberOfPages(numberOfPages);
```
rosiehsmith, line 181: Is this the main bug fix?

rachelschneiderman: No, good question, there are a lot of changes here.

The main bug fix was switching from `pdfDoc.getPages().length` to `pdfDoc.getPageCount()`. I observed that when removing or adding pages to the pdf document, the getPages method was often off by one, I did some digging into the source code and issues for documentation on the issue and found a few old "resolved" reports of the same behavior but nothing concrete.

rosiehsmith: Ah okay, good to know - thanks. I thought that was clean up.

PR #3306, opened 2023-03-21
9858: display all documents

TomElliottFlexion:


web-client/src/presenter/computeds/formattedDocketEntries.ts
```diff
@@ -246,6 +246,9 @@ export const getFormattedDocketEntry = ({
       ? 'Unseal to the public and parties of this case'
       : 'Unseal to the public'
     : 'Seal to the public';
+  formattedResult.toolTipText = !formattedResult.isFileAttached
```
rosiehsmith, line 249: Missed this on the last PR - worth adding a test here?

rosiehsmith (approved):

web-client/src/presenter/computeds/formattedDocketEntries.ts
TomElliottFlexion: Good call, it is

rgoshen (approved):

mmarcotte (approved):

PR #3310, opened 2023-03-22
 9806 Bugfix - When associating a swing session, the associated trial session does not indicate that it is a part of a swing session

mrinsin:
- Made a helper `associateSwingTrialSessions` out of an old interactor to make it reuseable
- `associateSwingTrialSessions` updates swing session info on BOTH trial sessions in question
- create and update interactors for trial sessions now call this helper
- removing current trial session from swing session dropdown list if `state.trialSessionId` is defined


ujahio (approved):

shared/src/business/useCases/trialSessions/createTrialSessionInteractor.test.ts
```diff
@@ -107,6 +107,24 @@ describe('createTrialSessionInteractor', () => {
     expect(result.isCalendared).toEqual(true);
   });
 
+  it('shoud associate swing trial sessions when needed', async () => {
```
rosiehsmith: Nit: typo on should & could this be a little more descriptive - like `should associate swing trial sessions when the trial session has a swing session` or something?

rosiehsmith (approved):

web-client/src/presenter/computeds/formattedTrialSessions.test.ts
```diff
@@ -337,6 +344,27 @@ describe('formattedTrialSessions', () => {
     expect(formatTrialSessionDisplayOptions).toHaveBeenCalled();
   });
 
+  it('removes the current trial session from the sessionsByTerm when state.trialSessionId is defined', () => {
```
rosiehsmith, line 347: ❤️

shared/src/business/useCases/trialSessions/createTrialSessionInteractor.test.ts
mrinsin: Good idea! TY

PR #3320, opened 2023-03-23
9700 duplicate work items to test

codyseibert:


shared/src/business/useCases/editDocketEntry/completeDocketEntryQCInteractor.test.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable jest/valid-expect */
```
TomElliottFlexion: Just curious about this

TomElliottFlexion (approved):

shared/src/business/useCases/editDocketEntry/completeDocketEntryQCInteractor.test.ts
codyseibert: good catch, fixing

TomElliottFlexion: Haha sorry I thought it was intentional!

PR #3332, opened 2023-03-23
Devex 1148 Move policy.json to shared/terraform/bin

rosiehsmith:
https://trello.com/c/TFbEKZET/1148-move-policyjson-to-shared-terraform-bin


rachelschneiderman (approved):

jimlerza (changes requested):
I think you will also need to change the path to `policy.json` on [line 42 of `shared/terraform/bin/create-bucket.sh`](https://github.com/flexion/ef-cms/blob/devex-1148-policy-json/shared/terraform/bin/create-bucket.sh#L42)

See [line 7 of `shared/terraform/bin/init.sh`](https://github.com/flexion/ef-cms/blob/devex-1148-policy-json/shared/terraform/bin/init.sh#L7) for an example of how to determine the path to `shared/terraform/bin` when you don't know the current working directory (and it might be different, as is the case when running terraform).

jimlerza (approved):

mmarcotte (approved):

PR #3339, opened 2023-03-24
Fix esbuild crashing when attempting to hot reload with an error

zachrog:
Previously, esbuild would crash when trying to hot reload when: 
1) the front-end is running
2) there is a javascript error (syntax, compile, yadda yadda)

now it doesn't

edit:
also included in this PR is a fix for `e2e-cypress-public` tests failing by switching the runner from ubuntu to macos, and also fixed a failing test due to erroneously-set timeouts in the same test suite.

### DIFF
--- a/.github/workflows/e2e-cypress-public.yml
+++ b/.github/workflows/e2e-cypress-public.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   e2e_cypress_public:
-    runs-on: macos-latest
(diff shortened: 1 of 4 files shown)


codyseibert (approved):

rachelschneiderman (approved):
wow

rosiehsmith:
Thank you for fixing this!

mmarcotte (approved):

PR #3351, opened 2023-03-27
Like Console.log()... but Pink!

zachrog:
Small utility function to use in debugging that prints to any console in pink and puts json Object in a pretty format. Has the exact same signature as console.log();

Pasted Graphic

### DIFF
--- a/shared/src/tools/pinkLog.ts
+++ b/shared/src/tools/pinkLog.ts
@@ -1,0 +1,22 @@
+/**
+ *
+ */
+export function pinkLog(...args): void {
+  const pink: string = '\x1b[35m';
+  const resetToDefault: string = '\x1b[0m';
+  const interpolateCharacter: string = '%s';
+
+  for (let i = 0; i < args.length; i++) {
+    if (i === args.length - 1) {
+      console.log(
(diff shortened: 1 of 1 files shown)


shared/src/tools/pinkLog.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ *
+ */
+export function pinkLog(...args): void {
+  const pink: string = '\x1b[35m';
```
jasonfu9966, line 5: but this is magenta.
please review.

jasonfu9966 (changes requested):

codyseibert:
seems useful when trying to find stuff in the long output logs of our system.  I may even say we could have this pinkLog imported in our applicationContext and have it decorate console with console.pink if we are running locally.  This would prevent us from needing to manually import pinkLog in every file we want to debug with it.

shumway-tm (approved):

shumway-tm:
The days of camouflaged console logs in our api log are finally numbered!

jasonfu9966 (approved):

mmarcotte (approved):

PR #3360, opened 2023-03-28
9809: Add Activity Report item to dropdown, make dropdown item only visible for judges and their chambers

rachelschneiderman:


shared/src/authorization/authorizationClientService.ts
jasonfu9966, line 1: I like it.

web-client/src/presenter/computeds/reportMenuHelper.test.ts
```diff
@@ -0,0 +1,82 @@
...
+          permissions: undefined,
+        },
+      });
+
+      expect(result.showActivityReport).toBeFalsy();
+    });
+  });
+
+  describe('pageIsReports', () => {
+    it('should show a border under the Reports tab to indicate it is the active tab when the current page is Case Deadlines Report', () => {
```
jasonfu9966, line 60: nit: Maybe "should recognize when the current page is Case Deadlines Report" ? I don't seem to see anything here about showing a border.

jasonfu9966 (approved):

web-client/src/presenter/computeds/reportMenuHelper.test.ts
rachelschneiderman: I agree with you!

I think the reason the test name was written this way was to provide a "because"/business need statement to help explain the need for that property - it's only used in the menu to determine if a class should be applied or not. That's the reason I left it, I liked that the test was explaining why the property was needed, what it's used for, in case the test ever fails we can understand why that property is needed. 

What are your thoughts on that?

jasonfu9966: makes sense! approved.

PR #3367, opened 2023-03-29
9828: Penalty/Deficiencies: Remove Positive Value Validation

rgoshen:
- Negative values can be entered in any of the deficiency or penalty fields for a case
- Negative values are calculated correctly along with the value of other penalty fields to arrive at the total
- When entering a negative value, upon typing a "-" (minus sign), the user is presented with a blue confirmation message confirming their intention to enter a negative value

---

image

---

image

---

image

---

image


web-client/src/views/StartCaseInternal/CalculatePenaltiesModal.tsx
```diff
@@ -48,7 +53,14 @@ export const CalculatePenaltiesModal = connect(
         {penalties &&
           penalties.map((penalty, index) => (
             // eslint-disable-next-line react/no-array-index-key
-            <div className="margin-top-3" key={index}>
+            // <div className="margin-top-3" key={index}>
```
rosiehsmith: Nit: little clean up here for comments and warning on naming the key as the array index

rosiehsmith (approved):
yee

web-client/src/views/StartCaseInternal/CalculatePenaltiesModal.tsx
shumway-tm: Thanks! We switched the key to the only unique value available to us on the penalty object.

PR #3381, opened 2023-03-30
9983: BUG: Unable to file non-standard H documents that point to an iteration type document

rosiehsmith:
flexion#9983 

image
image


rosiehsmith:
going to make a different pr with a new branch once po feedback addressed and re-reviewed

PR #3396, opened 2023-04-03
Flakey GA Tests

rachelschneiderman:
Disclaimer: This PR alone will not fix our flakey test issues.

What this PR accomplishes:
- Standardize how node dependencies are installed in each github action workflow - install as part of its own step
- Unify `CIRCLECI` and `CI` environment variables, remove `CIRCLECI` from project
- Remove unneccessary wait-for scripts that are run as part of the integration test GA suite, the run-local does those already
- Add --transpile-only flag to ts-node commands to [reduce memory usage](https://dev.to/aspecto/how-to-reduce-ram-consumption-by-6x-when-using-ts-node-4d8p) when running ts-node
- Clean up run-local script


rachelschneiderman:
Closing while I fix tests

PR #3400, opened 2023-04-04
Monitor Github Action Runners

rachelschneiderman:
This action will output a few different graphs - CPU, RAM, I/O usage and step timing - so we can monitor resource use for a particular workflow.

Good find, @codyseibert!!!

https://github.com/runforesight/workflow-telemetry-action[

[Example output](https://github.com/flexion/ef-cms/actions/runs/4609970137)


mrinsin (approved):
Very cool!

jimlerza (approved):

PR #3439, opened 2023-04-11
Revert 9865 and 9967

TomElliottFlexion:
Since 9865 and 9867 are no longer going forward, this will pull those changes out of test


mrinsin (approved):
😢

rosiehsmith (approved):

PR #3443, opened 2023-04-11
Fix public integration tests

TomElliottFlexion:
~~With the recent changes to typescript, the `--runInBand` flag was dropped from integration tests running in CI, which now causes the tests to run concurrently instead of sequentially. This sometimes causes collisions in testing when some test data share the same petitioner name.~~

We fixed this by changing one of the test petitioner's names.

### DIFF
--- a/web-client/integration-tests-public/journey/unauthedUserSearchesForSealedCaseByName.ts
+++ b/web-client/integration-tests-public/journey/unauthedUserSearchesForSealedCaseByName.ts
@@ -6,7 +6,7 @@ export const unauthedUserSearchesForSealedCaseByName = cerebralTest => {
     await refreshElasticsearchIndex(3000);
 
     const queryParams = {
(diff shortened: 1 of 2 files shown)


rosiehsmith (approved):
yay!

jasonfu9966 (approved):

rachelschneiderman (commented):
Are we going to add that flag back?

jasonfu9966:
> Are we going to add that flag back?

Pending. Going to talk to Zach about why that flag was dropped.

jasonfu9966:
> Are we going to add that flag back?

This flag was moved to [`jest-integration.config.ts` on line 22 as `maxWorkers: 1`](https://github.com/flexion/ef-cms/blob/0102cf046a8c7499f201363e495f1965ee7eaf7b/web-client/jest-integration.config.ts#L11). It was a red herring.

mmarcotte (approved):

PR #3451, opened 2023-04-12
Remove unused inputs in the Date Range Picker

zachrog:
Creating micro-PR for shared date range picker to be used by two inflight stories: 9723, 9809


ujahio (approved):

TomElliottFlexion (approved):

web-client/src/ustc-ui/DateInput/DateRangePickerComponent.tsx
```diff
@@ -39,16 +51,23 @@ export const DateRangePickerComponent = ({
       datePicker.on(startDatePickerRef.current);
       datePicker.on(endDatePickerRef.current);
       dateRangePicker.on(dateRangePickerRef.current);
+    } else if (
```
rachelschneiderman: What is the reason behind adding this else-if?

jasonfu9966: ....not sure. we don't have any recollection of adding it, although git history says we added it about three weeks ago. 

through brief testing with and without this block present, I have not observed much difference in functionality. I will remove it.

rgoshen (approved):

rachelschneiderman (approved):

mmarcotte (approved):

shumway-tm (approved):

PR #3460, opened 2023-04-18
9994 - Work Items not having lead-docket number

zachrog:
BEFORE this change
- Consolidated Case Icon was missing on the My Document QC Section
- Users were not getting all of their WorkItems in My Document QC Section

AFTER this change
- Whenever we create a new Work Item there is an optional parameter to pass in the Case the work item is based on so that information in the Work Item Entity that it is based off of the case has direct access to it. Now lead docket number should always be saved on Work Items.
- Completed work items do not get a gsi2pk as we do not need to query for them once the work item has been completed.

WHY this change
- Because docket clerks have thousands of Work Items assigned to them we only want to fetch for the uncompleted work items.
- WorkItem Entities have information inside of them which is based on the case they are being created for, such as trialLocation or leadDocketNumber. This is somewhat problematic as we should try and separate our entities and know what is based on what. To make this a little easier to work with we are now passing in a Case into the constructor of a work item as an optional field to help try and always set fields according to what the case is.


shared/src/business/entities/WorkItem.test.ts
```diff
@@ -105,6 +109,38 @@ describe('WorkItem', () => {
       );
       expect(workItem.isValid()).toBe(true);
     });
+
+    it('should set properties on WorkItem that pertain to the Case when a Case entity is passed into the constructor', async () => {
+      aValidWorkItem.associatedJudge = 'This should be overwritten';
```
rachelschneiderman, line 114: Nit: ^ async can be removed

rachelschneiderman (approved):
good job

TomElliottFlexion (approved):

PR #3469, opened 2023-04-20
9989 & 10000: New Trial Session Type - Hybrid S

TomElliottFlexion:


shared/src/business/useCases/trialSessions/runTrialSessionPlanningReportInteractor.ts
```diff
@@ -97,14 +105,17 @@ export const getTrialSessionPlanningReportData = async ({
           .compareISODateStrings(a.startDate, b.startDate);
       });
 
-      const previousTermSessionList = [];
+      const previousTermSessionList: Object[] = [];
```
zachrog: Nit: looking at its usage on line 123 this might be a string[] instead of a Object[].

zachrog (approved):

shared/src/business/useCases/trialSessions/runTrialSessionPlanningReportInteractor.ts
TomElliottFlexion: Dang it, you're right! Wonder why it didn't give me a type error

ujahio (approved):

mmarcotte (approved):

PR #3473, opened 2023-04-20
DevEx: Fix IRS superuser script and add documentation for how to use it

jimlerza:
The `irs-super-user` script stopped working, so I've converted it to use AWS SDK v3 and updated it to utilize our standard environment variables.

I've also added documentation for using the script to generate API tokens that we can use to test API v2 endpoints.


mmarcotte (approved):
love it!

jasonfu9966 (approved):
I have tested this and it works

scripts/irs-super-user.ts
```diff
@@ -0,0 +1,128 @@
+// see ../docs/additional-resources/irs-super-user.md for detailed instructions
+
+const { requireEnvVars } = require('../shared/admin-tools/util');
```
zachrog: Nice switch over to Typescript! Can you make all of the imports/exports es6?

zachrog (changes requested):
Looks good to me just need the imports and exports to be es6

zachrog (approved):

PR #3509, opened 2023-04-28
Refactoring Case Entity to ES6 Class

codyseibert:


zachrog (approved):

shared/src/business/entities/cases/Case.ts
```diff
@@ -24,2501 +23,2441 @@ const {
...
+    this.orderForCds = rawCase.orderForCds || false;
+    this.orderForRatification = rawCase.orderForRatification || false;
+    this.orderToShowCause = rawCase.orderToShowCause || false;
+
+    this.assignArchivedDocketEntries({ applicationContext, rawCase });
+    this.assignStatistics({ applicationContext, rawCase });
+    this.assignCorrespondences({ rawCase });
+  }
+
+  static VALIDATION_RULES = {
```
Absolutestunna, line 383: Hey @codyseibert is this better housed in the entityvalidationConstants file?

ujahio (approved):

shared/src/business/entities/cases/Case.ts
ujahio, line 383: Hey @codyseibert are the case validation rules better housed in the entityvalidationConstants file?

codyseibert: I'm not sure why the validation rules for entities are in that file to begin with.  I would think validation logic related to the entity should live in the file that creates the entity.

zachrog: I think there are some Validation rules which are shared between entities. Such as startDate, endDate validation rules could be reused by many entities and is common so gets extracted to entityValidationConstants. However if the validation rules are not shared anywhere then it should probably just live in the entity itself

shumway-tm (approved):

rgoshen (approved):

mmarcotte (approved):

PR #3540, opened 2023-05-05
9969 Trial Session View: NOTT Reminder

mrinsin:
image

Smaller PRs on story branch:
https://github.com/flexion/ef-cms/pull/10023
https://github.com/flexion/ef-cms/pull/10032

Signifier on trial sessions page:

image

Warning on trial session detail:

image

Persisting the dismissal of alerts per trial session so the alert does not show up again once dismissed.

Modal for dismissal of alert:
image

Users without petitions clerk permissions cannot dismiss modal:
image


web-client/src/presenter/computeds/utilities/set30DayNoticeOfTrialReminder.ts
```diff
@@ -0,0 +1,41 @@
...
+  trialStartDate: any;
+}) => {
+  const { DATE_FORMATS } = applicationContext.getConstants();
+
+  const trialStartDateString: any = applicationContext
+    .getUtilities()
+    .prepareDateFromString(trialStartDate, DATE_FORMATS.MMDDYY);
+
+  const thirtyFiveDaysBeforeTrial: any = trialStartDateString.minus({
+    ['days']: 35,
```
codyseibert: Just a thought / nit, you don't need to address it:

From doing the PR review, I don't think I see that many places, but the hard coded logic for 30-35 living in a utility seems more like business logic and might better live somewhere like in the use case or business entity.  I'm trying to think if this changed from 30-40 days range, how many places would need to be updated?

codyseibert (approved):

web-client/src/presenter/computeds/utilities/set30DayNoticeOfTrialReminder.ts
mrinsin: This is a really good suggestion Cody! I'm going to spend some time trying to move this logic to a more appropriate location. Thanks!

mrinsin: @codyseibert just pushed these changes (in case you are interested in re-reviewing!)

shared/src/authorization/authorizationClientService.ts
```diff
@@ -21,6 +21,7 @@ export const ROLE_PERMISSIONS = {
   COURT_ISSUED_DOCUMENT: 'COURT_ISSUED_DOCUMENT',
   CREATE_ORDER_DOCKET_ENTRY: 'CREATE_ORDER_DOCKET_ENTRY',
   CREATE_USER: 'CREATE_USER',
+  DISMISS_30_DAY_ALERT: 'DISMISS_30_DAY_ALERT',
```
rachelschneiderman: Nit: Could this be made a bit more specific? What is the alert for?

mrinsin: Good idea, I renamed it to `DIMISS_NOTT_REMINDER`, does that sound better?

PR #3551, opened 2023-05-10
Node 18: sparticuz/chromium memory leak fix

rachelschneiderman:


TomElliottFlexion (approved):

package.json
```diff
@@ -152,11 +152,7 @@
     "lint:js/ts": "TIMING=1 eslint --ext .js,.jsx,.ts,.tsx shared/src/ web-api/storage/ web-api/src/ web-client/src/ web-client/integration-tests/ web-client/integration-tests-public/",
     "lint:swagger": "swagger-cli validate web-api/swagger.json && swagger-cli validate docs/api/v1.yaml && swagger-cli validate docs/api/v2.yaml",
     "lint": "run-p lint:swagger lint:css lint:js/ts",
-    "loadtest:high": "./web-api/load-test-trial-session.sh dev us-east-1 150",
-    "loadtest:medium": "./web-api/load-test-trial-session.sh dev us-east-1 125",
-    "loadtest:practitioners": "./web-api/load-test-practitioners.sh dev us-east-1 200",
-    "loadtest:work-queue": "./web-api/load-test-work-queue.sh dev us-east-1 200",
-    "loadtest:small": "./web-api/load-test-trial-session.sh dev us-east-1 100",
```
jimlerza, line 159: Were these removed because they're not useful anymore?

jimlerza (approved):

package.json
rachelschneiderman: We actually can't run them currently because they fail entity validation and don't create data the way DAWSON now expects data to be created. Because they aren't being run in CI and it seems like we haven't run them in awhile, they have become outdated. We figured if we expand our loadtesting framework in CI we could go back and find these as inspiration but right now they are just taking up space as they can't even be run.

mmarcotte (approved):

PR #3560, opened 2023-05-15
DevEx - 1170 separate pdf lambda

codyseibert:
# Overview

Updating our lambdas to use node18 caused a lot of issues due to how coupled all our terraform code is with the layers and pdf generation.  This change pulls all the pdf generation code into a separate lambda that uses the layer and adds support for removing the layers from all the other lambdas via a dynamodb config.  The plan would be to deploy this change, turn the feature flag on, and test.  If everything seems fine, we can then set the layer flag to false so that terraform will not attach the layers to all of the lambdas and redeploy.  After turning the layer flag to off, the feature flag can no longer be used.  After all the layers are removed, another opex card will be created to reduce the memory on a lot of these lambdas since they will not be generating PDFs anymore.

Here is a diagram to explain the overview:

Screenshot 2023-05-15 at 1 34 59 PM

1. The deployment will fail on the new lambda zip unless a .zip is manually uploaded to the buckets.  Use the following before deploying to get a successful deployment:

```
aws s3 cp s3://${EFCMS_DOMAIN}.efcms.${ENV}.us-east-1.lambdas/api_${CURRENT_COLOR}.js.zip s3://${EFCMS_DOMAIN}.efcms.${ENV}.us-east-1.lambdas/pdf_generation_${CURRENT_COLOR}.js.zip &&
aws s3 cp s3://${EFCMS_DOMAIN}.efcms.${ENV}.us-west-1.lambdas/api_${CURRENT_COLOR}.js.zip s3://${EFCMS_DOMAIN}.efcms.${ENV}.us-west-1.lambdas/pdf_generation_${CURRENT_COLOR}.js.zip
```

2. You'll need to update environment specific permissions via terraform

`npm run deploy:environment-specific exp1`

### 3. Feature Flag

This feature flag allows us to toggle on / off if we should be generating PDFs using a separate pdf generation lambda, or continue with our current approach.  Do not turn this flag on if the use-layers-state is set to false for the deploying environment

`scripts/dynamo/setup-use-external-pdf-generation-flag.sh`

### 4. Layer Toggle

This toggle is used to inform terraform if we should be attaching the lambda layer to every lambda, or just to the pdf generation lambda.  The idea is we'd remove this from the code base after we've deployed, turned the feature flag on, and verified everything works ok.

`scripts/dynamo/setup-use-layers-state.sh`


codyseibert:
I deployed this to exp5 and test and verified multiple PDF generated related features with the feature flag on / off, and the feature flag on and layer flag off.

rosiehsmith (approved):

shared/src/persistence/dynamo/documents/a
jimlerza, line 1: Is this supposed to be here?

jimlerza (approved):

shared/src/persistence/dynamo/documents/a
codyseibert: I'll delete it, that must have been a mistake

mmarcotte:
I think this is the `aws s3 cp` command for copying the missing lambda files:

```
aws s3 cp s3://${EFCMS_DOMAIN}.efcms.${ENV}.us-east-1.lambdas/api_${CURRENT_COLOR}.js.zip s3://${EFCMS_DOMAIN}.efcms.${ENV}.us-east-1.lambdas/pdf_generation_${CURRENT_COLOR}.js.zip &&
aws s3 cp s3://${EFCMS_DOMAIN}.efcms.${ENV}.us-west-1.lambdas/api_${CURRENT_COLOR}.js.zip s3://${EFCMS_DOMAIN}.efcms.${ENV}.us-west-1.lambdas/pdf_generation_${CURRENT_COLOR}.js.zip
```

NOTE: the lambda name is `pdf_generation`

PR #3564, opened 2023-05-17
9980 jar progress description

rgoshen:


rosiehsmith (approved):

web-client/integration-tests/judgeActivityReportJourney.test.ts
```diff
@@ -14,12 +29,8 @@ describe('Judge activity report journey', () => {
   it('should disable the submit button on initial page load when form has not yet been completed', async () => {
     await cerebralTest.runSequence('gotoJudgeActivityReportSequence');
 
-    const judgeActivityReportHelper = withAppContextDecorator(
-      judgeActivityReportHelperComputed,
-    );
-
     const { isFormPristine, reportHeader } = runCompute(
-      judgeActivityReportHelper,
+      judgeActivityReportHelper as any,
```
mmarcotte, line 33: what _should_ this be?

mmarcotte (approved):

web-client/integration-tests/judgeActivityReportJourney.test.ts
rgoshen: The judgeActivityReportHelper is decorated withAppContextDecorator, which is not currently typed.  So, this was the best to do until withAppContextDecorator is typed.  Although there is an interface defined for judgeActivityReportHelper, what actually needs to be typed here is withAppContextDecorator.

web-client/integration-tests/journey/docketClerkAddsDocketEntryFromOrder.ts
jimlerza, line 1: Is `jar` necessary in the `jarDecisionTypeDocumentEventCodes` variable? I think it might be more readable without, since `jar` is an acronym.

jimlerza (approved):

web-client/integration-tests/journey/docketClerkAddsDocketEntryFromOrder.ts
rgoshen: The only reason I think I added this was because there are more decision type document event codes than these.  But I don't have a strong opinion so I can remove it.

PR #3593, opened 2023-05-24
10034 to test

rgoshen:


web-client/src/presenter/actions/FileDocument/getFileExternalDocumentAlertSuccessAction.test.ts
```diff
@@ -120,4 +120,20 @@ describe('getFileExternalDocumentAlertSuccessAction', () => {
...
+      modules: {
+        presenter,
+      },
+      props: { fileAcrossConsolidatedGroup: true },
+      state: {},
+    });
+
+    expect(result.output).toMatchObject({
+      alertSuccess: {
+        message: 'Document served to selected cases in group.',
```
TomElliottFlexion: Let's use the new constant here instead of hardcoding the message

web-client/src/presenter/actions/FileDocument/uploadExternalDocumentsAction.ts
```diff
@@ -10,6 +10,18 @@ const addCoversheet = ({ applicationContext, docketEntryId, docketNumber }) => {
     });
 };
 
+const removeSealedCases = consolidatedCases => {
```
TomElliottFlexion: We still need to unit test this

web-client/src/presenter/actions/FileDocument/getFileExternalDocumentAlertSuccessAction.test.ts
rgoshen: done

web-client/src/presenter/actions/FileDocument/uploadExternalDocumentsAction.ts
TomElliottFlexion: Nvm...

TomElliottFlexion (approved):

shared/src/business/entities/cases/PublicCase.ts
```diff
@@ -50,11 +49,7 @@ PublicCase.prototype.init = function init(rawCase, { applicationContext }) {
 
   const currentUser = applicationContext.getCurrentUser();
 
-  if (
-    (currentUser.role === ROLES.irsPractitioner ||
-      currentUser.role === ROLES.privatePractitioner) &&
-    !this.isSealed
-  ) {
+  if (currentUser.role === ROLES.irsPractitioner && !this.isSealed) {
```
mrinsin: Hey, what is the reason behind no longer allowing private practitioners to see parties on a case?

mrinsin (commented):
Great test coverage! Just had one question about something that didn't quite make sense to me!

shared/src/business/entities/cases/PublicCase.ts
rgoshen: Prior to Story 9874, a private practitioner if they were not party to that case, they got the public contact for the entity name and therefore only could see the petitioner name, contactType and contactId.  We are simply reverting it back to that if they are not the practitioner of record on that case as per story 10034.

mrinsin: ok, thanks for explaining that to me!

mrinsin (approved):

mrinsin (approved):

shared/src/business/entities/cases/PublicCase.ts
mrinsin: sorry you will have some conflicts on this file because Rosie and I changed it to be a class to match the other TS entities and our code got merged to test recently!

rgoshen: Ur welcome

PR #3601, opened 2023-05-30
9598 sealed cases feedback to test

TomElliottFlexion:
- Allows irsPractitioners to file FIRST document across all cases in consolidated group
- Displays 'Sealed Case' instead of petitioner names in the (very rare) scenario that the IRS is filing a first document or requesting access to a consolidated group with a sealed case
- Fixes bugs around page docket entry metadata edits during QC


rgoshen (approved):

shared/src/business/useCases/generateCoverSheetData.ts
```diff
@@ -117,7 +117,7 @@ export const generateCoverSheetData = async ({
     docketEntryEntity.eventCode,
   );
   if (
-    isLeadCase(caseEntity) &&
+    caseEntity.leadDocketNumber &&
```
mmarcotte, line 133: just curious what the rationale was to go away from isLeadCase

TomElliottFlexion: Good question! Since QCing externally filed documents has the potential to produce new coversheets, the internal flow now needs to support listing out all consolidated cases on the coversheet. This means that formatConsolidatedCaseCoversheetData function can be called on _any_ consolidated case, not just lead cases.

mrinsin (approved):

PR #3627, opened 2023-06-05
9785:add consolidation icons to deadlines report

rgoshen:
Will need a migration.


web-api/workflow-terraform/migration/main/lambdas/migrations/0007-update-corporate-disclosure-document.ts
```diff
@@ -9,7 +9,7 @@ const isCase = item => {
 };
 
 export const migrateItems = items => {
-  const itemsAfter = [];
+  const itemsAfter: any[] = [];
```
jimlerza, line 12: Do we really not know what type to use here? Wouldn't something like a generic entity or object be better than `any`?

web-api/workflow-terraform/migration/main/lambdas/migrations/0010-add-lead-docket-number-to-case-deadline.ts
```diff
@@ -0,0 +1,39 @@
...
+const applicationContext = createApplicationContext({});
+
+const isCaseDeadlineItem = item => {
+  return (
+    item.pk.startsWith('case-deadline|') && item.sk.startsWith('case-deadline|')
+  );
+};
+
+export const migrateItems = async (items, documentClient) => {
+  const itemsAfter: any[] = [];
```
jimlerza: Same here

web-client/src/presenter/actions/CaseDeadline/getCaseDeadlinesAction.ts
```diff
@@ -1,12 +1,5 @@
 import { state } from 'cerebral';
 
-/**
- * get case deadlines between start and end date
```
jimlerza, line 4: I'm on board with removing jsdoc but I hate to see human-readable descriptions go away, especially when the function name is far less clear than the human-readable description was.

jimlerza (approved):

web-api/workflow-terraform/migration/main/lambdas/migrations/0009-add-trial-location-field-to-work-item.ts
```diff
@@ -17,7 +17,7 @@ const isOutboxItem = item => {
 };
 
 export const migrateItems = async (items, documentClient) => {
-  const itemsAfter = [];
+  const itemsAfter: any[] = [];
```
ujahio: Hey @rgoshen, to Jim's point about types, here's another location that can use a more descriptive type.

web-api/workflow-terraform/migration/main/lambdas/migrations/0010-add-lead-docket-number-to-case-deadline.ts
rgoshen: Good call.  I did not realize the CaseDeadline class could actually be used as the type.

web-api/workflow-terraform/migration/main/lambdas/migrations/0009-add-trial-location-field-to-work-item.ts
rgoshen: Not sure how to do this since the item that can be pushed can either be a workItem or an outboxItem.  So I thought any here would be appropriate.

web-api/workflow-terraform/migration/main/lambdas/migrations/0007-update-corporate-disclosure-document.ts
rgoshen: This one is hard to type.  It can be one of two different record types but neither is instantiated like the other migration scripts, so I am not sure what else they can be other then any.

web-client/src/presenter/actions/CaseDeadline/getCaseDeadlinesAction.ts
rgoshen: added back the jsdoc and added to it

web-api/workflow-terraform/migration/main/lambdas/migrations/0009-add-trial-location-field-to-work-item.ts
rgoshen: corrected

ujahio (approved):

jimlerza (approved):

web-api/workflow-terraform/migration/main/lambdas/migrations/0010-add-lead-docket-number-to-case-deadline.test.ts
```diff
@@ -0,0 +1,101 @@
...
+        sortableDocketNumber: 2018000101,
+      },
+    ];
+
+    const results = await migrateItems(
+      mockConsolidatedCaseDeadlineItem,
+      documentClientMock,
+    );
+
+    expect(results).toEqual(mockConsolidatedCaseDeadlineItem);
```
mmarcotte: Shouldn't `mockConsolidatedCaseDeadlineItem` NOT have `leadDocketNumber`? 

And then we would expect the results to have `leadDocketNumber` because that's what `migrateItems(...)` does?

rgoshen: I modify it by removing the leadDocketNumber from the original, then created a clone of the original and added the leadDocketNUmber to compare to

mmarcotte (approved):

PR #3630, opened 2023-06-06
10018-Story to test

zachrog:
Detects when a draft document is about to be deleted but has already been served.

image


shared/src/business/entities/cases/Case.ts
```diff
@@ -2078,7 +2076,9 @@ export const isLeadCase = rawCase =>
   rawCase.docketNumber === rawCase.leadDocketNumber;
 
 export const caseHasServedDocketEntries = rawCase => {
-  return !!rawCase.docketEntries.some(docketEntry => isServed(docketEntry));
+  return !!rawCase.docketEntries.some(docketEntry =>
```
rgoshen: Quick question: why double bang when the some method for arrays already returns true if there is at least element that meets the condition and returns false otherwise?

zachrog: That is a good question... I don't know! It was here before Rachel and I touched the file, we were just attempting to refactor into using DocketEntry.isServed instead of isServed(). I will go ahead and remove it. The double bang does not seem to have an effect.

rgoshen (approved):

PR #3664, opened 2023-06-14
Devex 1142 Fetch all feature flags at once

mrinsin:


web-client/src/presenter/actions/getAllFeatureFlagsAction.ts
```diff
@@ -0,0 +1,23 @@
+import { state } from 'cerebral';
+
+/**
+ * Gets the value of a feature flag passed into the factory function
```
rosiehsmith: could clean these up

mrinsin: TY!

docs/additional-resources/adding-feature-flags.md
mrinsin, line 1: @jimlerza hope this is ok!

jimlerza (approved):

docs/additional-resources/adding-feature-flags.md
jimlerza: Looks good!

shared/src/business/useCases/docketEntry/addPaperFilingInteractor.test.ts
```diff
@@ -41,7 +41,7 @@ describe('addPaperFilingInteractor', () => {
 
     applicationContext
       .getUseCases()
-      .getFeatureFlagValueInteractor.mockReturnValue(Promise.resolve(true));
+      .getAllFeatureFlagsInteractor.mockReturnValue(Promise.resolve(true));
```
mmarcotte: 🤔 hm, wouldn't these change to an `object` instead of a `boolean`? Maybe I'm missing something.

shared/src/business/useCases/featureFlag/getAllFeatureFlagsInteractor.ts
```diff
@@ -0,0 +1,34 @@
+import { ALLOWLIST_FEATURE_FLAGS } from '../../entities/EntityConstants';
+
+/**
+ * getAllFeatureFlagsInteractor
+ *
+ * @param {object} applicationContext the application context
+ * @param {string} providers.featureFlag the feature flag to get the value for
+ * @returns {boolean} true if result of the persistence method is 'true'; false otherwise
```
mmarcotte, line 8: 🤔 hm, wouldn't this return an `object` instead of a `boolean`? Maybe I'm missing something.

shared/src/business/useCases/docketEntry/addPaperFilingInteractor.test.ts
mrinsin: Good catch, you're totally right! Updating all the tests that do this

shared/src/business/useCases/docketEntry/addPaperFilingInteractor.ts
```diff
@@ -60,9 +59,7 @@ export const addPaperFilingInteractor = async (
 
   const isCaseConsolidationFeatureOn = await applicationContext
     .getUseCases()
-    .getFeatureFlagValueInteractor(applicationContext, {
-      featureFlag: ALLOWLIST_FEATURE_FLAGS.MULTI_DOCKETABLE_PAPER_FILINGS.key,
-    });
+    .getAllFeatureFlagsInteractor(applicationContext);
```
mmarcotte: It looks like `getAllFeatureFlagsInteractor` returns a `key: value` `object` instead of a boolean. Perhaps we could pass in an optional `featureFlag` to the interactor or make use of a useCaseHelper... not sure. It doesn't look compatible at first glance.

mrinsin: I think this one was an oversight that we forgot to update! I noticed it right as I saw your comment (while updating the test files) - fixing it now

mmarcotte (approved):

PR #3667, opened 2023-06-15
DevEx 1122 - 24 hour stream issue

codyseibert:
This adds a check in the streams processing lambda to not process items it has already seen before.  We had issues where doing a deployment 24 hours after a blue / green migration would kick off an entire restream process which would basically crash the elasticsearch cluster and also take a long time.  This allows the stream to just skip over events it has already processed to prevent these issues.


rgoshen (approved):

web-api/src/streams/processStreamRecordsLambda.ts
```diff
@@ -1,18 +1,100 @@
...
+  const twentyFourHours = 24 * 60 * 60 * 1000;
+
+  await applicationContext
+    .getDocumentClient({
+      useMasterRegion: true,
+    })
+    .put({
+      Item: {
+        pk: `stream-event-id|${eventID}`,
+        sk: `stream-event-id|${eventID}`,
```
mmarcotte, line 25: Since we would not want to migrate these records to another DDB table, what do you think about filtering these out in `migration-segments.js`?

codyseibert: @mmarcotte I deployed to exp5 with these FilterExpression changes, I verified beta has a record called stream-event-id|hello, and after the migration, alpha did NOT contain that record

codyseibert: https://app.circleci.com/pipelines/github/flexion/ef-cms/45038/workflows/f03192fd-54fd-44f4-acc8-a2b9c70cfb3d

jimlerza (approved):

PR #3685, opened 2023-06-23
10052 public search date range

rgoshen:
This PR removes "Year start" and "Year end" from the public case search by name and advanced case search by name.  It adds a date range picker.  Allows the user to search by date range, start date only or by end date only and date range is not required.

----------
Public search

Screenshot 2023-06-23 at 08 47 16

----------
Advanced search

Screenshot 2023-06-23 at 08 45 41


rosiehsmith (changes requested):
Looks good - only thing is that on the mobile view the date fields aren't the width of the card like the other fields

rosiehsmith (approved):

rgoshen:
After merging in staging, PR became polluted.

PR #3686, opened 2023-06-23
10052 public search date range

rgoshen:


jimlerza (approved):

shared/src/persistence/elasticsearch/casePublicSearch.ts
```diff
@@ -5,24 +5,13 @@ import { search } from './searchClient';
...
+export const casePublicSearch = async ({ applicationContext, searchTerms }) => {
   const { commonQuery, exactMatchesQuery } = aggregateCommonQueryParams({
     applicationContext,
-    countryType,
-    petitionerName,
-    petitionerState,
-    yearFiledMax,
-    yearFiledMin,
+    ...searchTerms,
   });
```
mmarcotte: It looks like [this function](https://github.com/ustaxcourt/ef-cms/pull/3686/files#diff-406d891cc37977094574710e7929c5e19e2601778605496f01e8f046b2e3a278L19) doesn't take `applicationContext`. Can we remove that and just pass along `searchTerms`?

PR #3707, opened 2023-06-29
Resolve race condition undrafting

mmarcotte:


zachrog (changes requested):
If this is a brand new file could we make it in typescript and use aws-sdk v3 rather than v2?

e.g. import { DynamoDB } from '@aws-sdk/client-dynamodb';

If this is a one time script that will be thrown away I am okay with it being Javascript, but could we make a folder to express that?

jimlerza:
> If this is a one time script that will be thrown away, could we make a folder to express that?

I created [devex-1197](https://trello.com/c/qcNDhaXf) for organizing our admin scripts, perhaps we could utilize the directory structure proposed in that card.

mmarcotte:
@zRogersFlexion -- thanks for the feedback! I've added a test, switched to TS, and used v3 api.

zachrog (approved):

PR #3728, opened 2023-07-07
9998 - Async endpoint conversion - Intermediate

ujahio:
This PR addresses the need to have longer processing time for orders and opinion searches for selected judges per PO feedback.

The work includes:

1. Converting endpoints responsible for  fetching orders and opinions to an async endpoints.
2. Created new sequence for setting data from async calls.
3. Fixed query of closed cases for multiple judges.
4. Beginning of DOD work.


shared/src/business/useCases/judgeActivityReport/getOpinionsFiledByJudgeInteractor.ts
```diff
@@ -87,5 +87,15 @@ export const getOpinionsFiledByJudgeInteractor = async (
 
   const sortedResult = orderBy(result, 'eventCode', 'asc');
 
-  return sortedResult;
+  await applicationContext.getNotificationGateway().sendNotificationToUser({
```
rachelschneiderman, line 90: Why did these endpoints need to be transitioned to async?

shared/src/business/useCases/judgeActivityReport/getOpinionsFiledByJudgeInteractor.ts
```diff
@@ -87,5 +87,15 @@ export const getOpinionsFiledByJudgeInteractor = async (
...
+    applicationContext,
+    clientConnectionId: params.clientConnectionId,
+    message: {
+      action: 'fetch_opinions_complete',
+      opinions: sortedResult,
+    },
+    userId: authorizedUser.userId,
+  });
+
+  return sortedResult || [];
```
rachelschneiderman: Since this is an async endpoint, the result will be returned via websocket message above. I may be incorrect but I believe that means the return value is not consumed and can be removed.

zachrog (approved):

shared/src/business/useCases/judgeActivityReport/getOpinionsFiledByJudgeInteractor.ts
ujahio: @rachelschneiderman 

You are correct. The reason I added them was because integration tests failures. The items return from the call weren't been set in state correctly "in time". I have made the changes to remove the return statements and added `waitForLoadingComponentToHide` to wait for the calls to be resolved.

shared/src/business/useCases/judgeActivityReport/getOpinionsFiledByJudgeInteractor.ts
ujahio: They are being transitioned to async the time restrictions on the lambdas and api gateways is causing the call to crash. Part of the requirements as well is to query for all the orders and opinions per judge within a specific time frame. For longer time frames, the query needs more time to process and give back the results.

PR #3751, opened 2023-07-17
9970 Story: Generate 30 Day Notice of Trial

zachrog:
This PR includes only the functionality needed to implement this story, it does not include any of the refactoring we discussed with Chris and Tenille because it requires a migration and currently migration deploys are breaking smoke tests (nothing is actually broken other than the test itself). The refactoring can be merged separately once that issue has been addressed.


rachelschneiderman (approved):

shared/src/business/entities/cases/Case.removeFromHearing.test.ts
jimlerza, line 1: Why this change?

(Several other files have similar changes, too)

codyseibert: I converted them back to esm imports

TomElliottFlexion (approved):

setup-local-env.sh
```diff
@@ -17,3 +17,4 @@ export EFCMS_DOMAIN=localhost
 export ELASTICSEARCH_HOST=localhost
 export ELASTICSEARCH_ENDPOINT=http://localhost:9200
 export DYNAMODB_ENDPOINT=http://localhost:8000
+export IRS_SUPERUSER_EMAIL=irs-superuser@example.com
```
mmarcotte: can you add a newline please?

mmarcotte (approved):

PR #3762, opened 2023-07-20
adding a smoketests which runs after a complete deployment

codyseibert:


codyseibert:
after discussion various approaches, we decided having smoketest run AFTER the color switch feels wrong.  we decided to just skip this test during a migration.

PR #3784, opened 2023-07-27
10076 - Increasing Coverage

codyseibert:


web-client/src/presenter/actions/CaseInventoryReport/setCustomCaseInventoryFiltersAction.test.ts
```diff
@@ -334,5 +334,43 @@ describe('setCustomCaseInventoryFiltersAction', () => {
         'Roger Waters',
       ]);
     });
+
+    it('should negate the high priority when pass in as props', async () => {
```
TomElliottFlexion:
```suggestion
    it('should negate the highPriority filter when passed in as props', async () => {
```
Feel free to ignore!

TomElliottFlexion (approved):

PR #3795, opened 2023-08-02
10116 health check stability

TomElliottFlexion:
Adds a chron function which is constantly pinging our AWS services in both east/west and storing the applications health in dynamo.

Separately we have a health check which is fetching our applications health from dynamo every 30s to see if we should failover to a different region.

### DIFF
--- a/shared/src/business/useCases/health/getCachedHealthCheckInteractor.ts
+++ b/shared/src/business/useCases/health/getCachedHealthCheckInteractor.ts
@@ -1,0 +1,12 @@
+import { StoredApplicationHealth } from '../../../persistence/dynamo/deployTable/setStoredApplicationHealth';
+
+export const getCachedHealthCheckInteractor = async (
+  applicationContext: IApplicationContext,
+): Promise<StoredApplicationHealth> => {
+  console.log('getCachedHealthCheckInteractor 1: ', process.env.REGION);
+  const appHealth = await applicationContext
+    .getPersistenceGateway()
+    .getStoredApplicationHealth(applicationContext, process.env.REGION!);
+  console.log('getCachedHealthCheckInteractor 2', appHealth);
+  return appHealth;
+};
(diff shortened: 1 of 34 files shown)


zachrog (approved):

shared/src/business/useCases/health/getCachedHealthCheckInteractor.ts
```diff
@@ -0,0 +1,12 @@
+import { StoredApplicationHealth } from '../../../persistence/dynamo/deployTable/setStoredApplicationHealth';
+
+export const getCachedHealthCheckInteractor = async (
+  applicationContext: IApplicationContext,
+): Promise<StoredApplicationHealth> => {
+  console.log('getCachedHealthCheckInteractor 1: ', process.env.REGION);
```
rachelschneiderman: Nit: Console.logs

shared/src/business/useCases/health/setHealthCheckCacheInteractor.test.ts
```diff
@@ -0,0 +1,69 @@
...
+    applicationContext
+      .getUseCases()
+      .getHealthCheckInteractor.mockResolvedValue(applicationHealth);
+  });
+
+  afterAll(() => {
+    process.env.REGION = oldRegion;
+  });
+
+  it('should get service statuses and write an healthy status to persistence when all services are health', async () => {
```
rachelschneiderman: `when all services are healthy`?

TomElliottFlexion: Good catch! Thank you

shared/src/business/useCases/health/getCachedHealthCheckInteractor.ts
TomElliottFlexion: Agreed, don't need these in test

web-api/terraform/api/cron.tf
```diff
@@ -1,5 +1,6 @@
 resource "aws_lambda_function" "check_case_cron_lambda" {
-  count            = var.create_cron
+  # Replace with create_check_case_cron?
```
rachelschneiderman: Looks like this has been done! 🎉 Can the comment be removed?

TomElliottFlexion: TY

PR #3797, opened 2023-08-03
De flakeify docket clerk case inventory report

TomElliottFlexion:


ujahio (approved):

web-client/integration-tests/docketClerkCaseInventoryReport.test.ts
```diff
@@ -19,8 +19,21 @@ describe('case inventory report journey', () => {
     cerebralTest.closeSocket();
   });
 
-  const initialCaseInventoryCounts = {};
-  const createdDocketNumbers = [];
+  const initialCaseInventoryCounts: {
```
rachelschneiderman, line 22: I'm not sure we are buying anything from extracting this to a constant, I'm totally willing to have my mind changed. Since they are only referenced once and don't change I would probably prefer they stay in line.

web-client/integration-tests/docketClerkCaseInventoryReport.test.ts
```diff
@@ -127,60 +140,67 @@ describe('case inventory report journey', () => {
...
+  });
+
+  it('should get the updated new case inventory count for Judge Colvin', async () => {
     await cerebralTest.runSequence('updateScreenMetadataSequence', {
       key: 'associatedJudge',
       value: 'Colvin',
     });
     await cerebralTest.runSequence('submitCaseInventoryReportModalSequence');
-    updatedCaseInventoryCount = cerebralTest.getState(
+    const updatedCaseInventoryCount = cerebralTest.getState(
```
rachelschneiderman, line 167: What was the cause of flakiness with reassigning this variable? Is that something to be aware of in our other tests?

TomElliottFlexion: I don't think that reassigning the variable was causing flakiness directly, it just wasn't necessary after splitting the one giant test into smaller ones

jasonfu9966 (approved):

mmarcotte (approved):

PR #3799, opened 2023-08-03
9998 - Judge Activity Report - Judge Selection

ujahio:
1. Add dropdown for Judges' selection (including `All Judges`). See screenshot below:
Screenshot 2023-07-24 at 4 45 32 PM

2. Generate report for aggregated `Orders`, `Opinions`, `Closed Cases`, `Trial Sessions`, `Submitted/CAV Cases` per judges selection (and for all judges).
3. Add pagination to `Submitted/CAV Cases` table.
4. Create new persistence method for `Opinions` and `Orders` count aggregation.

Feedback Addressed: 
1. Convert Orders and Opinions endpoints to regular endpoints. https://github.com/ustaxcourt/ef-cms/pull/3841
2. Refactor CAV/Submitted Cases interactor. https://github.com/ustaxcourt/ef-cms/pull/3877
3. Refactor `getDocketNumbersByStatusAndByJudge` . https://github.com/ustaxcourt/ef-cms/pull/3877


jasonfu9966 (approved):

shared/src/business/useCases/judgeActivityReport/getCasesClosedByJudgeInteractor.ts
```diff
@@ -1,63 +1,37 @@
...
   if (!isAuthorized(authorizedUser, ROLE_PERMISSIONS.JUDGE_ACTIVITY_REPORT)) {
     throw new UnauthorizedError('Unauthorized');
   }
 
-  const searchEntity = new JudgeActivityReportSearch({
-    endDate,
-    judgeName,
-    startDate,
-  });
+  params.endDate = params.endDate || '';
```
rachelschneiderman: If we default these then the validation done on line 25 would never fail, right? Why do we want to default these values here?

shared/src/business/entities/EntityConstants.ts
```diff
@@ -1541,6 +1559,9 @@ export const MAX_ELASTICSEARCH_PAGINATION = 10000;
 export const MAX_SEARCH_CLIENT_RESULTS = 200;
 export const MAX_SEARCH_RESULTS = 100;
 
+export const TEMP_JUDGE_ID_TO_REPRESENT_ALL_JUDGES_SELECTION =
```
rachelschneiderman: What is this variable used for?

web-api/src/app.ts
```diff
@@ -861,12 +861,12 @@ app.get('/sections/:section/judge', lambdaWrapper(getJudgeInSectionLambda));
     lambdaWrapper(getTrialSessionsForJudgeActivityReportLambda),
   );
   app.post(
-    '/judge-activity-report/opinions',
-    lambdaWrapper(getOpinionsFiledByJudgeLambda),
+    '/async/judge-activity-report/opinions',
```
rachelschneiderman: Why did these endpoints need to become async?

shared/src/persistence/elasticsearch/getCasesClosedByJudge.ts
```diff
@@ -13,16 +17,33 @@ import { search } from './searchClient';
...
   startDate,
-}) => {
+}): Promise<CasesClosedType> => {
   const source = ['status'];
 
-  const { results } = await search({
+  const shouldFilters: QueryDslQueryContainer[] = [];
+
+  if (judges.length) {
+    judges.forEach(judge => {
```
rachelschneiderman: For this report, I believe you can only choose one judge to filter by or display all judges, why not just pass a specific judge ID if the user has selected to run the report for just one judge? I think that would simplify a lot of the querying code.

web-api/src/app.ts
ujahio: Processing of `orders` and `opinions` for `All Judges` for 2+ year range exhaust the 29 seconds timeout for regular lambdas. If  we do end up limiting the search to just one year in a future story, it may be possible to convert this endpoint to a regular lambda.

shared/src/business/entities/EntityConstants.ts
ujahio: This variable was created to keep the type of `judge user id` consistently a string. The `All Judges` selection, from the drop down, does not technically have an id associated to it, but  judge id is needed for filtering the correct trial sessions based on the selected judge.

shared/src/business/useCases/judgeActivityReport/getCasesClosedByJudgeInteractor.ts
ujahio: Good question. It's not necessary at all. I have made the change to remove the defaults.

shared/src/persistence/elasticsearch/getCasesClosedByJudge.ts
ujahio: From my understanding, if we don't specify the judges to make the queries for, the results will return the cases closed for all judges in the db (including legacy judges).

shared/src/business/entities/judgeActivityReport/JudgeActivityReportSearch.ts
```diff
@@ -67,11 +81,33 @@ export class JudgeActivityReportSearch extends JoiValidationEntity {
...
       .description(
-        'The start date to search by, which cannot be greater than the current date, and is required when there is an end date provided',
+        'The page size for calculating the number of CAV and Submitted cases to display.',
       ),
+    searchAfter: joi
+      .number()
+      .description(
+        'The last docket number to be used to make the next set of calls per page number',
+      ),
+    startDate: JoiValidationConstants.ISO_DATE.max('now').description(
```
rachelschneiderman, line 105: Why is start date not required anymore?

ujahio: [JudgeActivityReportSearch.ts](https://github.com/ustaxcourt/ef-cms/pull/3799/files/96a9b0554976f8cb2c91b92a04349e86bfdda5bd#diff-5495bccafdb9396bb5d4653ecd9f9af032d074bc54f330583627f215fb754b2b) is used for the validation for all the calls on this page but not all of those calls require the date parameters.

shared/src/business/useCases/judgeActivityReport/getCasesClosedByJudgeInteractor.ts
rachelschneiderman: Thanks!

web-api/src/app.ts
rachelschneiderman: Ok that does make sense, thanks for explaining! I am still a bit curious, is there anything that can be done to make that endpoint more efficient? It appears as if it's doing something very similar to what advanced opinion search does when you submit a search for all judges without specifying a date range, although I could be wrong.

I see that the `getOpinionsFiledByJudgeInteractor` is using a Promise.all to query elasticsearch for each judge in the case that 'all judges' is selected. Would it be more performant to query for all judges and then filter out legacy judges from the results? Again, I could be totally wrong, just trying to think of a way to make the request more performant for a user.

shared/src/business/entities/EntityConstants.ts
```diff
@@ -210,6 +218,15 @@ export const DOCUMENT_SERVED_MESSAGES = {
   SELECTED_CASES: 'Document served to selected cases in group.',
 };
 
+export const VALIDATION_ERROR_MESSAGES = {
```
rachelschneiderman, line 219: Are these validation error messages only used for JudgeActivityReportSearch? If so, I think I would prefer they live in that entity since EntityConstants is already getting pretty crowded and no one else needs to know about them.

shared/src/business/entities/judgeActivityReport/JudgeActivityReportSearch.ts
```diff
@@ -18,6 +22,11 @@ export class JudgeActivityReportSearch extends JoiValidationEntity {
   public startDate: string;
   public judgeName: string;
   public judgeId: string;
+  public statuses: CAV_AND_SUBMITTED_CASE_STATUS_TYPES;
+  public judges: string[];
```
rachelschneiderman, line 26: Are both `judgeId` and `judges` needed to run this report now? Is there anyway to unify them into one property?

shared/src/business/entities/judgeActivityReport/JudgeActivityReportSearch.ts
rachelschneiderman: But a start date and end date are required to run the report, right? So they must be present in the request in order to consider the report search valid?

shared/src/business/useCases/judgeActivityReport/getCasesByStatusAndByJudgeInteractor.ts
```diff
@@ -48,47 +65,43 @@ const filterCasesWithUnwantedDocketEntryEventCodes = caseRecords => {
...
+        statuses: searchEntity.statuses,
+      },
     });
 
   const rawCaseRecords: RawCase[] = await Promise.all(
-    submittedAndCavCasesResults.map(async result => {
-      return await applicationContext
-        .getPersistenceGateway()
-        .getCaseByDocketNumber({
+    submittedAndCavCasesResults.map(
```
rachelschneiderman: I think it would be possible to remove this call to fetch the full case if the `getDocketNumbersByStatusAndByJudge` query included `leadDocketNumber` as a part of the fields to return for each result!

shared/src/business/useCases/judgeActivityReport/getOpinionsFiledByJudgeInteractor.ts
```diff
@@ -22,43 +23,55 @@ import { orderBy } from 'lodash';
...
  */
 export const getOpinionsFiledByJudgeInteractor = async (
-  applicationContext,
-  {
-    endDate,
-    judgeName,
-    startDate,
-  }: { judgeName: string; endDate: string; startDate: string },
-) => {
+  applicationContext: object,
```
rachelschneiderman: Nit: Can this use the `IApplicationContext` type?

shared/src/business/useCases/judgeActivityReport/getCasesClosedByJudgeInteractor.ts
```diff
@@ -1,63 +1,33 @@
-import { CASE_STATUS_TYPES } from '../../entities/EntityConstants';
 import { InvalidRequest, UnauthorizedError } from '../../../errors/errors';
+import { JudgeActivityReportFilters } from './getTrialSessionsForJudgeActivityReportInteractor';
 import { JudgeActivityReportSearch } from '../../entities/judgeActivityReport/JudgeActivityReportSearch';
 import {
   ROLE_PERMISSIONS,
   isAuthorized,
 } from '../../../authorization/authorizationClientService';
 
-/**
```
rachelschneiderman, line 9: 👏 👏

shared/src/business/entities/EntityConstants.ts
ujahio: Yeah its only used only for [JudgeActivityReportSearch.ts](https://github.com/ustaxcourt/ef-cms/pull/3799/files/96a9b0554976f8cb2c91b92a04349e86bfdda5bd#diff-5495bccafdb9396bb5d4653ecd9f9af032d074bc54f330583627f215fb754b2b). I can move it. No problem

shared/src/business/useCases/judgeActivityReport/getOpinionsFiledByJudgeInteractor.ts
```diff
@@ -22,43 +23,55 @@ import { orderBy } from 'lodash';
...
+  params: JudgeActivityReportFilters,
+): Array<any> => {
```
rachelschneiderman: Looks like this function doesn't return anything now, can the return type be changed to reflect that?

shared/src/business/entities/EntityConstants.ts
rachelschneiderman: Thanks!

shared/src/business/useCases/judgeActivityReport/getCasesByStatusAndByJudgeInteractor.ts
```diff
@@ -48,47 +65,43 @@ const filterCasesWithUnwantedDocketEntryEventCodes = caseRecords => {
...
+      async result =>
+        await applicationContext.getPersistenceGateway().getCaseByDocketNumber({
```
mmarcotte: here do you don't need the `async` or the `await`. The `getCaseByDocketNumber` returns a Promise because it is an aync function And so we can create an array of promises with what they return.

shared/src/persistence/elasticsearch/getDocketNumbersByStatusAndByJudge.ts
```diff
@@ -1,50 +1,64 @@
...
+      };
+      shouldArray.push(associatedJudgeFilters);
+    });
+    const shouldObject: QueryDslQueryContainer = {
+      bool: {
+        should: shouldArray,
+      },
+    };
+    mustFilters.push(shouldObject);
+  }
```
mmarcotte: i'm confused why you're mixing and matching `must` and `should` here. plus, you're not including `minimum_should_match`, and so the `should` isn't required.

why not do 

```
bool: {
  should: [
    {
      match_phrase: {
        'associatedJudge.S': judge,
      },
    },
    ...
  ],
  minimum_should_match: 1
}
```

instead of 

```
bool: {
  must: [
    {
      bool: {
        should: [
          {
            'associatedJudge.S': judge,
          },
        },    
        ...
      ],
    }
  ],
}
```

without the `minimum_should_match: 1`, i believe ES will return more results

web-api/src/app.ts
mmarcotte: Agree with you @rachelschneiderman . We are going to pursue improvements to performance, perhaps in another user story. I believe if we made use of an `aggs` query, we can move this back to a synchronous request. It's a more involved `aggs` query because we need to check if the case does not have certain documents filed, but I think that's doable with the parent/child relationship they have in ES.

shared/src/business/useCases/judgeActivityReport/getCasesByStatusAndByJudgeInteractor.ts
ujahio: Makes sense. I have made the adjustment in this [PR](https://github.com/ustaxcourt/ef-cms/pull/3841).

shared/src/business/useCases/judgeActivityReport/getOpinionsFiledByJudgeInteractor.ts
ujahio: Good catch. This has been resolved in this [PR](https://github.com/ustaxcourt/ef-cms/pull/3841)

shared/src/business/useCases/judgeActivityReport/getOpinionsFiledByJudgeInteractor.ts
ujahio: This has been resolved in this [PR](https://github.com/ustaxcourt/ef-cms/pull/3841)

web-api/src/persistence/elasticsearch/getCasesClosedCountByJudge.ts
```diff
@@ -0,0 +1,86 @@
...
+        [CASE_STATUS_TYPES.closed]: 0,
+        [CASE_STATUS_TYPES.closedDismissed]: 0,
+      };
+
+  const judgeNameToLog =
+    judges.length > 1 ? 'all judges' : `judge ${judges[0]}`;
+
+  applicationContext.logger.info(
+    `Found ${total} closed cases associated with ${judgeNameToLog}`,
+  );
```
mmarcotte: hmm this doesn't quite seem right. isn't `judges` just an array of judges? It could have one, two, three, or all of the judges. I think this can safely be removed.

web-api/src/persistence/elasticsearch/helpers/searchQueryForAggregation.ts
```diff
@@ -0,0 +1,67 @@
...
+        },
+      },
+    });
+  }
+
+  if (params.searchType) {
+    documentFilters.push({
+      terms: { 'eventCode.S': params.documentEventCodes },
+    });
+  }
```
mmarcotte: Is this the only way that `params.searchType` is used? Can it be removed? I see either `opinion` or `order` getting passed in to the function that uses this. Can't we simply rely on `params.documentEventCodes`?

web-api/src/persistence/elasticsearch/getCasesClosedCountByJudge.ts
mmarcotte: Let's consider this for the [next story](https://github.com/flexion/ef-cms/issues/10049) and refactoring opportunities.

web-api/src/persistence/elasticsearch/helpers/searchQueryForAggregation.ts
ujahio: Resolved in this commit: [246dcc0](https://github.com/ustaxcourt/ef-cms/pull/3799/commits/246dcc03369049c305da6dcf8777a5e2b1ecc492)

PR #3829, opened 2023-08-11
10122 Bug: Optimize /cases endpoint response time

rachelschneiderman:
After filtering out user associated cases that are not consolidated, also get a unique list of `leadDocketNumbers`, that way we only fetch the consolidated group once. Also, when getting the cases in the consolidated group don't fetch the full case (correspondences, docketEntries, etc.) just need case record and associated counsel. 

The original response time for the `/cases` endpoint was ~25secs, after the two above optimizations, the response time is ~3secs.


shared/src/business/useCases/getCasesForUserInteractor.ts
```diff
@@ -45,16 +45,17 @@ async function fetchConsolidatedGroupsAndNest({
...
-        .filter(aCase => aCase.leadDocketNumber)
-        .map(aCase =>
-          applicationContext
-            .getPersistenceGateway()
-            .getCasesByLeadDocketNumber({
-              applicationContext,
-              leadDocketNumber: aCase.leadDocketNumber,
-            }),
-        ),
+      uniqBy(
```
TomElliottFlexion, line 48: Nice!

TomElliottFlexion (approved):

ujahio (approved):

mmarcotte (approved):

jimlerza (approved):

PR #3831, opened 2023-08-11
Devex-1139: Continue with B/G Migration Deploy if an empty Destination Table or Cluster exists

mmarcotte:
This PR enables the blue/green migration deployment process to proceed in the event that the destination Opensearch cluster still exists and is empty and/or the destination DynamoDB table exists and is empty. This is simply a time-saving maneuver since deleting the cluster or table only to recreate them eats up around 30-45 minutes every attempt. When performing development and testing around the blue/green migration process, this can easily eat up gobs of time and be really frustrating. 

It isn't something we would likely see in Production or even the Test environment as we almost always only have one cluster in operation at any given time. However, we will often have empty DynamoDB tables present before kicking off a deployment. This should help ease deployments in that scenario.


scripts/elasticsearch/ready-cluster-for-migration.helpers.test.ts
```diff
@@ -0,0 +1,188 @@
...
+      count: mockedCount,
+    },
+  })),
+  indices: {
+    delete: jest.fn(),
+    exists: jest
+      .fn()
+      .mockImplementation(() => ({ statusCode: mockedStatusCode })),
+  },
+};
```
TomElliottFlexion, line 31: Nice, this will be a useful pattern

TomElliottFlexion (approved):

PR #3845, opened 2023-08-16
10116: DNS Failover

TomElliottFlexion:
# In the PR
- Create an active-active regional failover system capable of redirecting API traffic if API Gateway || Lambda || Dynamo goes down.
- Our regional APIs now have health checks attached to them, which will re-route traffic to different regions depending on if our API is returning as "healthy".
- Lays a foundation for an iterative approach for creating custom health checks for the system and then re-routing traffic.

# How it works
- Every 30s we are sending a request to our regional APIs at /public-api/cached-health which triggers a lambda to look up a record in our dynamo database containing our applications health. If the lambda is unable to retrieve the dynamo record or the record shows the application is unhealthy, then the entire API will be marked as unhealthy.
- Separately every 30s we are writing the applications health to the dynamo table so that the health check system can retrieve the application's health rapidly. Currently we are not monitoring any of the application's health but always marking the system as healthy as this indirectly monitors the health of DynamoDB, Lambda, and Route53.

## Things to be aware of
- We have started using SSM Parameter store as a means of easily retrieving values from deployed terraform infrastructure. This is a new paradigm for our app.


web-api/terraform/template/failover.tf
```diff
@@ -0,0 +1,91 @@
+resource "aws_cloudwatch_metric_alarm" "failover_has_occurred_alarm" {
+  alarm_name          = "Failover Alarm"
+  alarm_description   = "If this alarm is going off a regional failover is occurring"
+  comparison_operator = "LessThanThreshold"
+  count               = var.enable_health_checks
+  threshold           = "1"
+  evaluation_periods  = "2"
```
TomElliottFlexion, line 7: Re-evaluate the evaluation period after this alarm has been in test for a week.

shared/admin-tools/user/admin.ts
```diff
@@ -202,9 +211,8 @@ export const createDawsonUser = async ({
     },
   };
 
-  const url = deployingColorUrl ?? `https://api.${EFCMS_DOMAIN}/users`;
   try {
-    await axios.post(url, user, headers);
+    await axios.post(deployingColorUrl, user, headers);
```
mmarcotte, line 215: Huh, why not build the deploying color url here with the `process.env.DEPLOYING_COLOR`? I suppose that refactoring can be considered down the road. It just seems strange to rely on an entire URL that's passed into the function to create a user in the system. I realize you didn't make that change to begin with.

web-api/src/persistence/dynamo/deployTable/getStoredApplicationHealth.ts
```diff
@@ -0,0 +1,16 @@
...
+  applicationContext: IApplicationContext,
+  region: string,
+): Promise<StoredApplicationHealth> => {
+  const result = await getFromDeployTable({
+    Key: {
+      pk: 'healthCheckValue',
+      sk: `healthCheckValue|${region}`,
+    },
+    applicationContext,
+  });
```
mmarcotte, line 14: While I understand the use of the deploy table because it's not application data, I do wonder if the deploy table is the appropriate place to be storing this information. It makes me wonder a broader question about its utility. Couldn't we be storing all information in the main table?

Critically, for this failover story, the Deploy table is not replicated across regions. If Dynamo east were to fail, both regions would be unhealthy.

shared/admin-tools/user/admin.ts
zachrog: We agree with you but it seems out of scope for this story.

We originally touched this as we were trying to remove the uncolored-api but reverted that change. We just updated the signature of this function to make it a little more strict for how you call it.

web-api/src/persistence/dynamo/deployTable/getStoredApplicationHealth.ts
zachrog: This makes sense and currently we are vulnerable to the deploy table going down. Right now our application as a whole and the failover mechanism is vulnerable to dynamo deploy table going down so we have not made the system as a whole more vulnerable but it is something we should be conscience of. We are adding this note to the follow on SPIKE.

mmarcotte (approved):

PR #3874, opened 2023-08-22
Devex 1006 updated

TomElliottFlexion:
- Adds framework for running a local version of Cognito, allowing Cognito triggers to be tested
 - Intended to be used only in the new automated tests, but can be used manually and could be expanded to replace local authentication we currently use
 - Is only enabled when running application via `run start:api:cognito-local` and `run start:client:cognito-local`


package.json
```diff
@@ -343,6 +351,7 @@
     "s3rver": "github:20minutes/s3rver",
     "sass": "^1.66.1",
     "sass-loader": "^13.3.2",
+    "serverless-offline": "^12.0.4",
```
zachrog: Do we need serverless-offline here and in the new cognito-triggers-sls package?

shared/src/business/useCases/auth/authenticateUserInteractor.ts
```diff
@@ -8,12 +8,47 @@
  */
 export const authenticateUserInteractor = async (
   applicationContext: IApplicationContext,
-  { code }: { code: string },
+  { code, cognitoLocal }: { code: string; cognitoLocal?: string },
```
zachrog: I think that authenticateUserInteractor probably should not know about cognitoLocal. We should inject in either a local cognito or a real cognito into applicationContext when we start the app so that when we go to use cognito no interactor needs to know whether or not it is using the real one or the local one.

package.json
TomElliottFlexion: Negatory, removing!

shared/src/business/useCases/auth/authenticateUserInteractor.ts
TomElliottFlexion: Agreed. Moving it to persistence

cruzjone-flexion (approved):

jasonfu9966 (approved):
nice work!

PR #3916, opened 2023-09-01
Deploy 10114 to test

codyseibert:
When a practitioner updates their own contact information (or an admissionsclerk updates a practitioner's info), the interactor now uses an SQS queue to send out an event for each case that needs a change of address added to it.  This is behind a feature flag with a key of `use-change-of-address-lambda`.  Before it sends out the sqs events, it creates a job entry in dynamo which tracks all of the processed / unprocessed cases.  As it processes each case, the job will decrement a counter until it reaches 0, then it will send back the completed notification to the user.

diagram of change 
Screenshot 2023-09-01 at 3 23 29 PM


scripts/dynamo/setup-use-change-of-address-lambda-flag.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash -e
+
+# Sets the redaction acknowledgement enabled feature flag to "false" in the dynamo deploy table
```
rachelschneiderman, line 3: Nit: copy/pasta? Can these just be removed?

shared/src/business/useCaseHelper/generateChangeOfAddressHelper.ts
```diff
@@ -0,0 +1,192 @@
...
+import {
+  ROLES,
+  SERVICE_INDICATOR_TYPES,
+} from '@shared/business/entities/EntityConstants';
+import { TUserContact } from '@shared/business/useCases/users/generateChangeOfAddress';
+import { aggregatePartiesForService } from '@shared/business/utilities/aggregatePartiesForService';
+import { clone } from 'lodash';
+import { generateAndServeDocketEntry } from '@shared/business/useCaseHelper/service/createChangeItems';
+
+/**
```
rachelschneiderman, line 12: Nit: These can be removed too 😄

rachelschneiderman (approved):

PR #3983, opened 2023-09-20
10049 display wip cav submitted cases

ujahio:


shared/src/business/entities/EntityConstants.ts
```diff
@@ -415,6 +415,12 @@ export const EVENT_CODES_REQUIRING_JUDGE_SIGNATURE = without(
   'NOT',
 );
 
+const excludedOrderEventCodes = ['OAJ', 'SPOS', 'SPTO', 'OST'];
```
mmarcotte: I think the name excludedOrderEventCodes becomes confusing in this context of EntityConstants. A rename might be in order, or just removed and the codes moved into the `JUDGE_ACTIVITY_REPORT_ORDER_EVENT_CODES` filter function

TomElliottFlexion: Good point!

ujahio: Ok.. Resolved. [c6ac968](https://github.com/ustaxcourt/ef-cms/pull/3983/commits/c6ac9681ca91d6986ca224cd7ca0148f53c2fc24)

shared/src/business/useCases/judgeActivityReport/getCountOfCaseDocumentsFiledByJudgesInteractor.ts
```diff
@@ -0,0 +1,84 @@
...
+          documentEventCodes: JUDGE_ACTIVITY_REPORT_ORDER_EVENT_CODES,
+          endDate: searchEntity.endDate,
+          judges: searchEntity.judges,
+          startDate: searchEntity.startDate,
+        },
+      });
+
+  const computedAggregatedOrderEventCodes =
+    addDocumentTypeToEventCodeAggregation(orderAggregationCount);
+
```
mmarcotte, line 73: This is an improvement by consolidating this into one interactor! 👏 

I think we can do better by having the client call the endpoint with the event codes upon which it wants to report. So, it might make two calls for the information it needs; or more calls if sometime in the future we need to support different subsets of event codes. Perhaps other UIs will make use of this.

mmarcotte: No need to perform this refactor now; I will document it and open up a PR after this is merged.

web-api/src/persistence/elasticsearch/getCountOfConsolidatedCases.test.ts
```diff
@@ -24,7 +24,7 @@ describe('getCountOfConsolidedCases', () => {
   it('returns a number', async () => {
     (count as jest.Mock).mockReturnValue(5);
 
-    const result = await getCountOfConsolidedCases({
+    const result = await getCountOfConsolidatedCases({
```
mmarcotte, line 27: ty!

web-api/src/persistence/elasticsearch/getCountOfConsolidatedCases.ts
```diff
@@ -1,6 +1,6 @@
 import { count } from './searchClient';
 
-export const getCountOfConsolidedCases = async ({
+export const getCountOfConsolidatedCases = async ({
```
mmarcotte, line 3: ty!

mmarcotte (approved):

shared/src/business/useCases/judgeActivityReport/getCountOfCaseDocumentsFiledByJudgesInteractor.ts
TomElliottFlexion: Agreed!

PR #3992, opened 2023-09-22
hotfix: fix paginator on Judge Activity Report

TomElliottFlexion:
- Fixes a bug seen only on staging 
- Reverts a change made to the calculation of totalCount in getCasesByStatusAndByJudgeInteractor

Bug: 
- totalCount was being calculated incorrectly in getCasesByStatusAndByJudgeInteractor which was causing the judgeActivityReportHelper to incorrectly calculate pageCount. This in turn caused the paginator to cease being shown because the incorrectly calculated pageCount for page 2 of results was < 1 : 
``` 
 const pageCount = Math.ceil(
    totalCountForSubmittedAndCavCases / CAV_AND_SUBMITTED_CASES_PAGE_SIZE,
  );
```


ujahio (approved):

rachelschneiderman (approved):
Nice work! Is there something that could be added to the integration test for this feature to help catch this in the future? 

Here's an example of where we inject in a custom page size for another feature to test pagination:
image

TomElliottFlexion:
Donezo Bubonzo

jimlerza (approved):

PR #4018, opened 2023-09-28
10054: generate ea documents

TomElliottFlexion:


web-client/src/presenter/actions/FileDocument/setDocumentFileForAutoGeneratedEntryOfAppearanceAction.test.ts
```diff
@@ -0,0 +1,47 @@
...
+      'this value should not be modified',
+    );
+    expect(result.state.fileUploadProgress.isUploading).toEqual(true);
+  });
+
+  it('should set isUploading to true and modify the primaryDocumentFile when the pdfPreviewUrl is set', async () => {
+    (global as any).fetch = async () => {
+      return {
+        blob: async () => {
+          return new Blob([], { type: 'pdf' });
```
mmarcotte: check these linting warnings for the test. is there a reason these functions are async?

mmarcotte (approved):

web-client/src/presenter/actions/FileDocument/setDocumentFileForAutoGeneratedEntryOfAppearanceAction.test.ts
TomElliottFlexion: I think we were trying to mimic the behavior of fetch closely- updated the test to use jest mock functions

PR #4032, opened 2023-10-02
Devex fix test serve case to irs

mmarcotte:
While combatting flaky tests, I noticed that some tests do not serve a petition before they are added to Trial Sessions. This results in creating inoperable cases in the deployments where these smoketests run.

So, this attempts to serve the petitions in the the cases for the test before adding them to trial sessions. Additionally, some work was made done to help close the dialog for the scanner install


rachelschneiderman (approved):
Thanks for doing this, nice work. I am curious, does this happen in prod? Should we prevent folks from calendaring an unserved petition in the production code?

mmarcotte:
> Thanks for doing this, nice work. I am curious, does this happen in prod? Should we prevent folks from calendaring an unserved petition in the production code?

My thoughts exactly! @ttlenard said that this would never happen; a clerk would never add a "New" case to a Trial Session.  It does make sense to prohibit that behavior, and I think I'll be doing some follow up work to make this happen.

PR #4045, opened 2023-10-05
9788 - Add Consolidation Icons to Printable Copy of Trial Session Information

ujahio:


rachelschneiderman (approved):

shared/src/business/utilities/getFormattedTrialSessionDetails.ts
```diff
@@ -37,12 +43,24 @@ export const setPretrialMemorandumFiler = ({ caseItem }) => {
   return filingPartiesCode;
 };
 
+// type the return
```
mmarcotte: nit: looks like you've done this. 

I could see creating a type and applying it to the entire function (rather than the params and the return within the function definition itself). Perhaps that's a paradigm we should consider in order to help type ApplicationContext. 🤔

shared/src/business/utilities/htmlGenerator/cards.scss
```diff
@@ -0,0 +1,19 @@
...
+  background: #f0f0f0;
+  font-size: 13px;
+  font-weight: bold;
+}
+
+.card-content {
+  display: flex;
+  align-items: flex-start;
+  padding: 10px;
+}
```
mmarcotte: nit: can you please add a newline at the end?

mmarcotte (changes requested):
just a few nitpicky things. looks great otherwise.

shared/src/business/utilities/getFormattedTrialSessionDetails.ts
rachelschneiderman: You make a good point, we chatted about this a bit. Right now it feels like we made "type soup" but we think this is better (if only slightly) than it was before. We noticed that this function (`formatCase`) is modifying the provided `caseItem` by reference and also returning a new object that is a cloned copy of `caseItem` with a few additional properties. I think once we have more types on the web-client we can start identifying patterns of types we can extract out and use throughout the folder. We can also make an effort to stop modifying items in state by reference so we can confidently type and refactor this function without fearing we created a bug.

mmarcotte (approved):

PR #4062, opened 2023-10-09
10161: Activity Report/Judge Dashboard: Remove Decision Doc Criteria

rachelschneiderman:
Addresses [sorting bug](https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/gh/flexion/ef-cms/10168) in the `CAV` and `Submitted` table.


web-api/workflow-terraform/migration/main/lambdas/migrations/0013-default-case-status-history.ts
```diff
@@ -0,0 +1,18 @@
...
+export const migrateItems = items => {
+  const itemsAfter: TDynamoRecord[] = [];
+
+  for (const item of items) {
+    if (isCaseRecord(item) && !item.caseStatusHistory) {
+      item.caseStatusHistory = [];
+    }
+    itemsAfter.push(item);
+  }
+
```
mmarcotte, line 16: Why do you think we needed to write a migration for this? 

Case.ts already has

```
    this.caseStatusHistory = rawCase.caseStatusHistory || [];
```

which treats it as an empty array if it doesn't have there.

rachelschneiderman: Good question! The Case constructor does default that value to an empty array, but there are places where we return a case _record_, straight from DynamoDB, that has not been passed through the case constructor. We felt this was the best decision so that the joi validation and the definition of case match, regardless of whether or not an item has been passed through the Case constructor.

ujahio (approved):

scripts/run-once-scripts/open-search-playground.ts
```diff
@@ -0,0 +1,108 @@
+import { MAX_ELASTICSEARCH_PAGINATION } from '../../shared/src/business/entities/EntityConstants';
```
ujahio, line 1: ❓ Are/Should we keeping this playground?

ujahio (approved):

package.json
```diff
@@ -223,7 +223,7 @@
     "test:client:unit": "jest --config web-client/jest-unit.config.ts",
     "test:client:unit:ci": "jest --config web-client/jest-unit.config.ts --coverageDirectory=./coverage/${NODE_INDEX} ${TESTFILES}",
     "test:scripts": "jest --config scripts/jest-scripts.config.ts",
-    "test:file": "IS_LOCAL=true CI=false NO_SCANNER=true SKIP_VIRUS_SCAN=true AWS_ACCESS_KEY_ID=S3RVER AWS_SECRET_ACCESS_KEY=S3RVER FILE_UPLOAD_MODAL_TIMEOUT=1 jest --detectOpenHandles --config web-client/jest-integration.config.ts -- ",
+    "test:file": "IS_LOCAL=true CI=false NO_SCANNER=true SKIP_VIRUS_SCAN=true AWS_ACCESS_KEY_ID=S3RVER AWS_SECRET_ACCESS_KEY=S3RVER FILE_UPLOAD_MODAL_TIMEOUT=1 jest --config web-client/jest-integration.config.ts -- ",
```
TomElliottFlexion, line 226: Nice catch

TomElliottFlexion (approved):

mmarcotte (approved):

PR #4093, opened 2023-10-16
Devex 1006 to staging

TomElliottFlexion:
- Adds framework for running a local version of Cognito, allowing Cognito triggers to be tested
 - Intended to be used only in the new automated tests, but can be used manually and could be expanded to replace local authentication we currently use
 - Is only enabled when running application via `run start:api:cognito-local` and `run start:client:cognito-local`


jasonfu9966 (approved):

.cognito/config.json
```diff
@@ -0,0 +1,35 @@
...
+    ]
+  },
+  "KMSConfig": {
+    "credentials": {
+      "accessKeyId": "local",
+      "secretAccessKey": "local"
+    },
+    "region": "local"
+  }
+}
```
mmarcotte: nit: newline

mmarcotte (approved):

PR #4098, opened 2023-10-17
10177: BUG: If a practitioner previews their uploaded EA doc, the Redaction acknowledgement checkbox disappears

TomElliottFlexion:
- Replaces code that checks state for a pdfPreviewUrl with code that checks the generation type set on the form


rachelschneiderman (commented):
Can the feature flag be removed for this feature now?

mmarcotte (approved):

PR #4103, opened 2023-10-18
10177 edge case + remove REDACTION_ACKNOWLEDGEMENT_ENABLED feature flag

TomElliottFlexion:


rachelschneiderman (commented):
Thanks for removing the feature flag! There should be a setup script used to create the feature flag `scripts/dynamo/setup-redaction-acknowledgement-enabled-flag.sh`, can that be repurposed to remove the feature flag from each env's deploy table?

TomElliottFlexion:
Good call, done!

codyseibert (approved):

cruzjone-flexion (approved):

PR #4130, opened 2023-10-26
10169: BUG: Numbered/Bulleted list formatting on Order Generator not displaying appropriately on Preview or after Save

TomElliottFlexion:
Screen Shot 2023-10-25 at 8 52 42 PM
Screen Shot 2023-10-25 at 8 53 03 PM

### DIFF
--- a/shared/src/business/utilities/htmlGenerator/documents/order.scss
+++ b/shared/src/business/utilities/htmlGenerator/documents/order.scss
@@ -20,9 +20,9 @@
       list-style-type: lower-alpha;
     }
 
-    // ul, ul ul {
-    //   list-style: inside disc;
-    // }
+    ul, ul ul {
+      list-style: inside disc;
+    }
 
     #order-signature {
       font-size: 16px;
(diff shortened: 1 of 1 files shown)


shared/src/business/utilities/htmlGenerator/documents/order.scss
```diff
@@ -16,6 +16,14 @@
       line-height: 20px;
     }
 
+    ol ol {
+      list-style-type: lower-alpha;
+    }
+
+    // ul, ul ul {
```
nechama-krigsman: Can this be deleted?

Absolutestunna, line 23: ❓ Hey can this be deleted?

nechama-krigsman (approved):

ujahio (approved):

shared/src/business/utilities/htmlGenerator/documents/order.scss
TomElliottFlexion: That....is not supposed to be commented, good catch

PR #4133, opened 2023-10-26
Dependency Updates - October 23, 2023

jasonfu9966:
updated minor versions
updated ECR image due to awscli version update
[Successful migration](https://app.circleci.com/pipelines/github/flexion/ef-cms/47683/workflows/eda615e7-c655-47c4-aed8-0feadb429e68)


TomElliottFlexion (approved):

TomElliottFlexion:
Nice! Is the new ECR image on the Court lower account, too?

rachelschneiderman (commented):
Looks like the reindex was manually approved. What was the underlying reason it wasn't going to auto-approve?

rachelschneiderman (commented):
Is there an update for the AWS provider used in our Terraform files?

jasonfu9966:
> Nice! Is the new ECR image on the Court lower account, too?

I... believe so? I published the ECR image with the script we have and that should provide the image everywhere, right?

> Looks like the reindex was manually approved. What was the underlying reason it wasn't going to auto-approve?

Yeah, I talked to Jim about this, and the way the wait-for script works, you have to manually approve the step if the destination table has more records than the source table. There was a discrepancy in the number of judges, which I believe is something you saw when you were updating dependencies last week. 

> Is there an update for the AWS provider used in our Terraform files?

There wasn't when I created the first PR, but I see there is one now. I could include that here or leave it. I'm open.

TomElliottFlexion:
> > Nice! Is the new ECR image on the Court lower account, too?
> 
> I... believe so? I published the ECR image with the script we have and that should provide the image everywhere, right?
> 
It depends on which credentials are set while the script is run. It needs to be run once with Flexion creds (for exp envs) and once with USTC lower env creds (for test/staging). Once the image is built, you should just need to re-tag and re-push it using the other account. Happy to help

jasonfu9966:
> > > Nice! Is the new ECR image on the Court lower account, too?
> > 
> > 
> > I... believe so? I published the ECR image with the script we have and that should provide the image everywhere, right?
> 
> It depends on which credentials are set while the script is run. It needs to be run once with Flexion creds (for exp envs) and once with USTC lower env creds (for test/staging). Once the image is built, you should just need to re-tag and re-push it using the other account. Happy to help

Thanks! We got this resolved, and I updated the documentation to be clearer about this.

PR #4136, opened 2023-10-27
10186 get calendared cases overloaded to test

mmarcotte:


shared/src/business/useCases/trialSessions/getEligibleCasesForTrialSessionInteractor.ts
```diff
@@ -34,13 +37,22 @@ export const getEligibleCasesForTrialSessionInteractor = async (
...
+        .getCalendaredCasesForTrialSession({
+          applicationContext,
+          trialSessionId,
+        })
+    ).map(aCase => ({
+      ...aCase,
+      docketEntries: aCase.docketEntries.filter(
+        docketEntry =>
+          docketEntry.eventCode === 'PMT' && !docketEntry.isStricken,
+      ),
```
TomElliottFlexion: Will we still reduce the amount of data passed enough if we don't use a CalendaredCase entity here?

TomElliottFlexion (approved):

shared/src/business/useCases/trialSessions/getEligibleCasesForTrialSessionInteractor.ts
mmarcotte: 🤔 Hmm, you know what? I don't need to do the work of removing these docket entries here because they will be filtered out later in the `EligibleCase` entity. 

I had simply borrowed this code from the interactor that it had been calling, that now removed critical information that this interactor needed.

I'll remove this map and just let the EligibleCase entity do with the case what it wants to do.

shared/src/business/entities/cases/CalendaredCase.test.ts
```diff
@@ -0,0 +1,83 @@
+import { CalendaredCase } from './CalendaredCase';
+import { MOCK_CASE_WITH_SECONDARY_OTHERS } from '../../../test/mockCase';
+import { MOCK_COMPLEX_CASE } from '../../../test/mockComplexCase';
+import { applicationContext } from '../../test/createTestApplicationContext';
+
+describe('CalendaredCase', () => {
+  it('white lists the fields set within the entity, removing those not defined', () => {
```
rachelschneiderman: Nit: allow list?

web-api/src/persistence/dynamo/cases/getCaseByDocketNumber.ts
```diff
@@ -15,9 +15,11 @@ import { queryFull } from '../../dynamodbClientService';
 export const getCaseByDocketNumber = async ({
   applicationContext,
   docketNumber,
+  includeConsolidatedCases = true,
```
rachelschneiderman, line 18: Instead of adding this flag, can we allow the entity or interactor to strip out consolidatedCases?

shared/src/business/entities/cases/CalendaredCase.ts
```diff
@@ -0,0 +1,93 @@
+import { DOCKET_NUMBER_SUFFIXES } from '../EntityConstants';
+import { DocketEntry } from '@shared/business/entities/DocketEntry';
+import { IrsPractitioner } from '../IrsPractitioner';
+import { JoiValidationConstants } from '../JoiValidationConstants';
+import { JoiValidationEntity } from '../JoiValidationEntity';
+import { PrivatePractitioner } from '../PrivatePractitioner';
+import joi from 'joi';
+
+export class CalendaredCase extends JoiValidationEntity {
+  public docketEntries?: DocketEntry[];
```
rachelschneiderman: 👏 👏  👏 for only returning what is needed here. 

Does this need to be an entity? Can the endpoint that returns this define a stricter return type in this shape instead?

shared/src/business/useCases/trialSessions/getCalendaredCasesForTrialSessionInteractor.ts
```diff
@@ -31,14 +31,19 @@ export const getCalendaredCasesForTrialSessionInteractor = async (
...
-  const casesWithMinimalRequiredDocketEntries = cases.map(aCase => ({
-    ...aCase,
-    docketEntries: aCase.docketEntries.filter(
-      docketEntry => docketEntry.eventCode === 'PMT' && !docketEntry.isStricken,
-    ),
-  }));
+  const casesWithMinimalRequiredInformation = cases
+    .map(aCase => ({
+      ...aCase,
+      docketEntries: aCase.docketEntries.filter(
```
rachelschneiderman: I did some research into why we need to know if each case has a Pretrial Memorandum filed on it (that's also not stricken). There was a [story](https://github.com/flexion/ef-cms/issues/9512) implemented that added PMT as a column to the trial session working copy page which indicates if a PMT has been filed on each case and by whom. I think we could better document the business use case here by just returning `servedPartiesCode` for that docket entry if it exists.

web-api/src/persistence/dynamo/cases/getCaseByDocketNumber.ts
mmarcotte: we surely could, it just felt like unnecessary dynamo queries when the first thing it does is remove it from memory.

shared/src/business/entities/cases/CalendaredCase.ts
TomElliottFlexion: Forgive me if I'm misunderstanding: is the alternative to define a stricter return type and manually strip out the superfluous data?

shared/src/business/entities/cases/CalendaredCase.test.ts
TomElliottFlexion: Agreed!

shared/src/business/useCases/trialSessions/getCalendaredCasesForTrialSessionInteractor.ts
TomElliottFlexion: Good point!

TomElliottFlexion: This refactoring has been rolled into the CalendaredCase entity, for now

shared/src/business/entities/cases/CalendaredCase.ts
TomElliottFlexion: After much discussion, we're leaving the CalendaredCase entity in place. I will respond to this comment with another brach including possible refactoring

jasonfu9966 (approved):

PR #4159, opened 2023-11-01
Devex 1187: Statistic

nechama-krigsman:


shared/src/business/dto/cases/ConsolidatedCaseSummary.ts
```diff
@@ -28,10 +27,10 @@ export class ConsolidatedCaseSummary extends JoiValidationEntity {
 
   constructor(rawCase: any) {
     super('ConsolidatedCaseSummary');
+
     this.caseCaption = rawCase.caseCaption;
     this.docketNumber = rawCase.docketNumber;
     this.docketNumberWithSuffix = rawCase.docketNumberWithSuffix;
-    this.entityName = rawCase.entityName;
```
jimlerza, line 34: Nice catch!

Are these ConsolidatedCaseSummary records ever stored in dynamo?

jimlerza (approved):

shared/src/business/dto/cases/ConsolidatedCaseSummary.ts
rachelschneiderman: No, they are not. We double checked by looking at the updateCase persistence method and verified the property consolidatedCases is in the list of fields to omit before persisting.

mmarcotte (approved):

shared/src/business/dto/cases/ConsolidatedCaseSummary.ts
jimlerza: 👍

PR #4160, opened 2023-11-01
Devex 1187 Stamp and Scan entities

nechama-krigsman:


shared/src/business/entities/Scan.ts
```diff
@@ -72,19 +72,17 @@ export class Scan extends JoiValidationEntity {
...
-
-  getErrorToMessageMap() {
-    return Scan.VALIDATION_ERROR_MESSAGES;
-  }
+  static VALIDATION_RULES = joi.object().keys({
+    batches: joi
+      .array()
+      .min(1)
+      .required()
+      .messages({ '*': '#At least one batch is required' }),
```
rachelschneiderman: thinking

Is it just me or does the '#' look like a typo? I see it was already in the error message before but that just looks funky to me.

rachelschneiderman (approved):

shared/src/business/entities/Scan.ts
nechama-krigsman: I was thinking the same. Wanted to ask you if I should just remove it

PR #4163, opened 2023-11-01
OpEx: find cases involving practitioners who work for a specific law firm

jimlerza:
Today we had a request from docket to deliver a list of all cases involving private practitioners employed by a specific law firm. We expect to field more requests like this in the future.


shared/admin-tools/elasticsearch/find-firms-cases.ts
```diff
@@ -0,0 +1,98 @@
...
+                  value: '*Jones*',
+                },
+              },
+            },
+            {
+              wildcard: {
+                'firmName.S': {
+                  value: '*Day*',
+                },
+              },
```
jtdevos: Could these be parameterized w/ command-line arguments?

jtdevos (changes requested):

TomElliottFlexion (approved):

jtdevos (approved):

PR #4176, opened 2023-11-03
Devex 1187: Only format validation errors when getErrorToMessageMap exists on the specific entity

rachelschneiderman:


nechama-krigsman (commented):
Do the tests need to be updated to call getFormattedValidationErrors instead of getValidationErrors?

rachelschneiderman:
> Do the tests need to be updated to call getFormattedValidationErrors instead of getValidationErrors?

Yes! Thank you!

jimlerza (approved):

PR #4179, opened 2023-11-06
OpEx: run setup-test-users and setup-glued-judges after data finishes gluing

jimlerza:
Some fixes for automated glue jobs:

* Toggle streams in both regions when toggling streams on/off
* Permit the s3 sync to run without output
* Run `setup-test-users.sh` after data is finished gluing (but before indexing)
* Run `setup-glued-judges.ts` after data is finished gluing (but before indexing)


jimlerza:
Ideally we really want to wait for indexing to finish and then do a full deployment, but since we currently have no way to wait for indexing to finish, setting up the users before indexing will have to do for now.

ujahio (approved):

jtdevos (approved):

PR #4194, opened 2023-11-07
DX-1204 - My Cases Column Header

ujahio:


web-client/src/views/CaseListTable.tsx
```diff
@@ -76,9 +82,9 @@ export const CaseListTable = connect(
                     <th>
                       <span className="usa-sr-only">Lead Case Indicator</span>
                     </th>
-                    <th>Docket number</th>
-                    <th>Case title</th>
-                    <th>Date filed</th>
+                    <th>Docket No.</th>
+                    <th>Case Title</th>
+                    <th>Filed Date</th>
```
rachelschneiderman, line 87: I know this is part of the design debt card, I apologize for not catching this during refinement. To me, `Date Filed` sounds more grammatically correct. I can post this in the UX channel too if that makes the discussion easier!

rachelschneiderman (approved):

web-client/src/views/CaseListTable.tsx
ujahio: Hey Rachel. UX has decided to go with 'Filed Date'. Here's [the decision](https://flexion.slack.com/archives/CD6DEU37Z/p1699383805391469?thread_ts=1699382687.722429&cid=CD6DEU37Z).

jimlerza (approved):

PR #4198, opened 2023-11-07
Rewriting Flaky Test in Cypress

codyseibert:
The previous test was flaky.  I'm rewriting the same test in cypress since we find cypress a more reliable and easier to debug test runner.


rachelschneiderman (approved):
splendid

TomElliottFlexion (approved):

zachrog (approved):

cypress/cypress-integration/support/commands.ts
```diff
@@ -24,6 +24,10 @@ Cypress.Commands.add('showsSuccessMessage', (shows = true) => {
   }
 });
 
+Cypress.Commands.add('getByTestId', (testId: string) => {
```
rachelschneiderman, line 27: What's the benefit of adding this command?

codyseibert: so that we don't have to keep writing [data-testid=""] all throughout our tests.  It's also hinted at in their best practices section of their docs https://docs.cypress.io/guides/references/best-practices#Real-World-Example

rachelschneiderman: Thanks for the link! I don't love obfuscating such a simple command to be honest, but I'm willing to try it out.

zachrog: I am with Rachel as well and do not love the obfuscating of a simple command. I generally don't like adding commands onto the global cypress object and prefer generic javascript functions. 

Open to changing my mind if you feel strongly about it though Cody.

TomElliottFlexion: I'm with Cody on this one, this is what Cypress Commands are for

codyseibert: I think we are going to try using the helper for now, and if we find it annoying we can just do a quick replace all.  I find the code using the helper much easier to read compared to cy.get('[data-testid="my-test-id"]') vs cy.getByTestId("my-test-id").

zachrog (approved):

split-tests-cypress.ts
```diff
@@ -0,0 +1,18 @@
+import fs from 'fs';
+import shuffleSeed from 'shuffle-seed';
+
+// const testType = process.argv[2] || '';
```
mmarcotte: looks like this can go away.

split-tests-cypress.ts
```diff
@@ -0,0 +1,18 @@
...
+const shuffled = shuffleSeed.shuffle(files, process.env.GITHUB_SHA);
+
+const total = parseInt(process.env.CI_NODE_TOTAL!, 10);
+const index = parseInt(process.env.CI_NODE_INDEX!, 10);
+
+const tests = shuffled
+  .filter((num, i) => i % total === index)
+  .map(file => `./cypress/cypress-integration/integration/${file}`);
+
+console.log(tests.join(','));
```
mmarcotte, line 16: is there value in this console log?

codyseibert: yes, we need to output to STDOUT so that the cypress cli can read it in and use that output

split-tests-cypress.ts
codyseibert: removed

PR #4205, opened 2023-11-08
fix rm in Dockerfile - to test

jasonfu9966:
Sometimes Cypress mistakenly publishes two lists in their package which causes us to get errors. This line removes one of them.
`rm` exits 1 if the file wasn't found, so this change accounts for that.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,7 @@ WORKDIR /home/app
 RUN mkdir -p /usr/share/man/man1
 
 # delete duplicate list if it exists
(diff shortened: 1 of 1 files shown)


TomElliottFlexion (approved):
Nice! How did you figure out Cypress was the problem?

jasonfu9966:
> Nice! How did you figure out Cypress was the problem?

Just google :)

mmarcotte:
this will be backmerged into `test`

PR #4213, opened 2023-11-09
remove logs: more potentially sensitive logs

TomElliottFlexion:
Not sure if all of these need to be fixed


jimlerza (approved):

jasonfu9966 (approved):

nechama-krigsman (approved):

shared/admin-tools/user/rotate-environment-secrets.ts
```diff
@@ -74,6 +74,7 @@ const rotateSecrets = async (environmentName: string): Promise<void> => {
 
   const USTC_ADMIN_PASS = makeNewPassword();
 
+  // Should only be run locally; is it necessary?
```
mmarcotte: Yes, this only run locally, and its purpose is to rotate the passwords and make them available for consumption. This is how we distribute it to the non-technical users on the team. There's always a better way...

mmarcotte (approved):

shared/admin-tools/user/rotate-environment-secrets.ts
TomElliottFlexion: Thank you!

PR #4256, opened 2023-11-14
10188 Bug: The status report portion of the Stamp Order does not display in the Stamp Preview

rachelschneiderman:
image

image


web-client/src/ustc-ui/DateInput/DateSelector.tsx
```diff
@@ -40,19 +43,28 @@ export const DateSelector = ({
...
     if (formGroupInputRef.current) {
       const input = formGroupInputRef.current.querySelector('.usa-date-picker');
       if (disabled) {
+        const myDatePicker =
+          formGroupInputRef.current.querySelector(datePickerId);
+
+        if (!myDatePicker)
+          throw new Error('could not find expected date picker');
+
+        (myDatePicker as HTMLInputElement).value = '';
```
TomElliottFlexion, line 66: Out of curiosity only, is there a reason we have to do this instead of giving the datepicker a default value of ''?

TomElliottFlexion (approved):

web-client/src/ustc-ui/DateInput/DateSelector.tsx
rachelschneiderman: When you apply a stamp to an order, there is an option to 'Clear Optional Fields'. One of the optional fields happens to have a date selector inside of it. This was the cleanest way I could figure out to reset the date selector value because of how USWDS exposes/handles the date picker value.

TomElliottFlexion: Makes sense, thank you!

mmarcotte (approved):

jimlerza (approved):

PR #4303, opened 2023-11-22
10183 story

ujahio:


shared/src/business/useCases/trialSessions/runTrialSessionPlanningReportInteractor.ts
```diff
@@ -37,147 +92,106 @@ export const getTrialSessionPlanningReportData = async ({
...
   term: string;
   year: string;
-}) => {
-  const previousTerms: any[] = [];
-  let currentTerm = term;
-  let currentYear = year;
+}): Promise<{
+  previousTerms: PreviousTerm[];
+  trialLocationData: TrialLocationData[];
+}> => {
```
mmarcotte: This function is trying to a lot and is pretty complicated. Do you think it could/should be broken up into smaller pieces?

zachrog: This is an interesting one. We gave a shot at refactoring it into multiple smaller functions, but to us it just made it more difficult to follow. 

Do you have any suggestions on how to break this up to make it more readable? The only method that I have is a forbidden technique that is frowned upon by the jedi.

zachrog: We extracted out fetching individual trial location data, which helps, but please let us know if you have any additional ideas for refactoring this.

mmarcotte: Thanks Zach. Was mainly just asking the Q to help bring attention to it. 

I could see further refactoring down the road. Would love to chat more about it over Zoom etc.

mmarcotte (approved):

PR #4315, opened 2023-11-27
10077: Bug: deficiency statistics validation

TomElliottFlexion:


shared/src/business/entities/cases/PetitionIncomplete.ts
```diff
@@ -20,7 +20,7 @@ import joi from 'joi';
  *  attempting to add to the system.
  *
  */
-export class CaseInternal extends JoiValidationEntity {
+export class PetitionIncomplete extends JoiValidationEntity {
```
rachelschneiderman: How about `CaseInternalIncomplete` to match `CaseExternalIncomplete`?

TomElliottFlexion: We thought about this, but the entity represents neither a Case nor anything specifically internal. It's a validation entity for an unfiled petition, which is why we felt `Petition` should be in the name

rachelschneiderman: I agree, it's a validation entity for a paper filed petition. Which is why I believe 'internal' was included to make it clear it's related to to `CaseExternalIncomplete`. What about `CaseInternalIncomplete` (to make it clear it's similar to `CaseExternalIncomplete`) or rename both `CaseExternalIncomplete` to `IncompleteElectronicCase` (or something similar) and `CaseInternal` to `IncompletePaperCase`?

TomElliottFlexion: Good points, will make these consistent

jasonfu9966: Thanks for the suggestion! We like the idea of renaming both entities and making everything consistent. For now we're going to revert CaseInternal to its original name and create a devex card for the renaming, since it's starting to ooze past the scope of this card.

rachelschneiderman: Is the rename complex or risky? One of the DOD items for a bug is `Code refactored for clarity and to remove any known technical debt` - I think the idea to rename this entity was a really good one that you both had, I would push for it to be completed as part of DOD.

TomElliottFlexion: Since this is a bug, we'd feel more comfortable making the change in a separate PR. Up next on the agenda

jasonfu9966 (approved):

mmarcotte (approved):

PR #4328, opened 2023-11-29
10197: Bug Fix to Test

cruzjone-flexion:


shared/src/business/entities/User.ts
```diff
@@ -202,6 +202,10 @@ export class User extends JoiValidationEntity {
   getValidationRules() {
     return User.VALIDATION_RULES;
   }
+
+  public setisUpdatingInformation(value: boolean) {
```
nechama-krigsman: Is this method used anywhere?

cruzjone-flexion: shared/src/business/useCases/users/updateUserContactInformationInteractor.ts
line 108

cypress/cypress-integration/integration/practitioners-with-no-cases.cy.ts
```diff
@@ -0,0 +1,60 @@
...
+    cy.get('[data-testid="account-menu-button"]').click();
+    cy.get('[data-testid="my-account-link"]').click();
+    cy.get('[data-testid="contact-info-phone-number"]').should(
+      'contain.text',
+      anotherRandomPhoneNumber,
+    );
+  });
+});
+
+function generateRandomPhoneNumber() {
```
nechama-krigsman, line 47: Would this function be helpful in other places? Wondering if it can be moved to a shared file. Same for the function below.

cruzjone-flexion: Sounds good, will look for a better place to have this method, but will not do it for this Test PR only for Staging branch

nechama-krigsman (approved):

PR #4329, opened 2023-11-29
Design Debt 1161 to test again

TomElliottFlexion:
This time with a new ROLE_PERMISSION for creating new trial sessions, specifically excluding Judges/CAs


web-client/src/presenter/computeds/trialSessionsHelper.ts
```diff
@@ -27,6 +29,7 @@ export const trialSessionsHelper = (get: Get): any => {
 
   return {
     additionalColumnsShown,
+    newTrialSession: permissions.CREATE_TRIAL_SESSION,
```
nechama-krigsman: Wondering if we can give a clearer name here. `newTrialSession` seems to represent if a user has permissions to start a trial session or not.

nechama-krigsman (approved):

web-client/src/presenter/computeds/trialSessionsHelper.ts
TomElliottFlexion: Good call, thanks!

PR #4332, opened 2023-11-29
10197: BUG - User cannot update their contact information

cruzjone-flexion:
[ZenHub Card](https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/gh/flexion/ef-cms/10197)


nechama-krigsman (approved):

shared/src/business/useCases/users/updateUserContactInformationInteractor.test.ts
```diff
@@ -227,6 +232,55 @@ describe('updateUserContactInformationInteractor', () => {
...
+    });
+  });
+
+  it('should not clean up DB and send websocket message if "generateChangeOfAddress" returns undefined', async () => {
+    (generateChangeOfAddress as jest.Mock).mockReturnValue(undefined);
+
+    await updateUserContactInformationInteractor(applicationContext, {
+      contactInfo,
+      userId: mockUser.userId,
+    } as any);
```
mmarcotte, line 269: Just curious why this test / use case is permitted? I recognize that `generateChangeOfAddress` should never return undefined, but I don't understand the rationale that we leave the user's account in a broken state if something like this should happen. Why not send the websocket message and clean up the DB regardless?

mmarcotte (approved):

PR #4333, opened 2023-11-29
Devex 1238 - refactor case creation entities

jasonfu9966:
[Devex 1238](https://trello.com/c/nEWRfaWq/1238-rename-caseinternal-caseexternal-and-caseexternalincomplete-entities) began as just renaming entities to make sense, but upon investigation we noticed that CaseExternalIncomplete was being used in one place where it could be replaced easily with CaseExternal. 

Then we renamed the entities. No other functionality changes.

CaseInternal → PaperPetition
CaseExternal → ElectronicPetition


TomElliottFlexion (approved):

rachelschneiderman (commented):
I missed reviewing this PR, thanks for doing the rename! 

I would say, I prefer using the term 'Case' instead of 'Petition' here because 'Petition' is a type of docket entry while the user in either case is attempting to submit an entire Case to the USTC.

TomElliottFlexion:
> I missed reviewing this PR, thanks for doing the rename!
> 
> I would say, I prefer using the term 'Case' instead of 'Petition' here because 'Petition' is a type of docket entry while the user in either case is attempting to submit an entire Case to the USTC.

Moving away from the term 'Case' is the sole reason for this change. We feel pretty strongly that these are NOT cases, resulting in a fair bit of confusion while trying to work with the code around these. Let's talk about it!

PR #4362, opened 2023-12-04
devex: fix smoketests if practitioner does not have any cases

mmarcotte:
After wiping the data in `test` due to a Glue Job, we are seeing smoketests fail due to the expectation that a case list table exists. However, Private Practitioner 1 does not have any cases, and so the table is not displayed. This PR changes that expectation to other items that would show up on the page regardless of whether or not they have previously created cases.

- [CircleCI Failure](https://app.circleci.com/pipelines/github/ustaxcourt/ef-cms/6658/workflows/5b8bf886-6719-4579-93ad-7d89ee69e572)

### DIFF
--- a/web-client/src/views/CaseListTable.tsx
+++ b/web-client/src/views/CaseListTable.tsx
@@ -130,6 +130,7 @@ export const CaseListTable = connect(
                   defaultActiveTab={openTab}
                 >
                   <Tab
+                    data-testid="open-cases-count"
                     id="tab-open"
                     tabName={openTab}
                     title={`Open Cases (${externalUserCasesHelper.openCasesCount})`}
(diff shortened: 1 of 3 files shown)


TomElliottFlexion (approved):

web-client/src/views/CaseListTable.tsx
```diff
@@ -130,6 +130,7 @@ export const CaseListTable = connect(
                   defaultActiveTab={openTab}
                 >
                   <Tab
+                    data-testid="open-cases-count"
```
rachelschneiderman, line 133: For the future, what do you (and the rest of the team) think of trying to name `data-testid` attributes semantically? I have trying to follow a pattern of `<element_type>-<feature>`. For example, here I might use `tab-open-cases`

PR #4390, opened 2023-12-06
10194: Export pending report

TomElliottFlexion:
Uses [csv-stringify](https://www.npmjs.com/package/csv-stringify)

- Same publisher/repo as [csv-parse](https://www.npmjs.com/package/csv-parse)
- Supports both node and browser (using different APIs)


cypress/cypress-integration/integration/pending-report.cy.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  describe('View pending report', () => {
+    it('should log in as a docket clerk and navigate to the pending report then select chief judge from the dropdown', () => {
+      loginAsDocketClerk();
+      cy.get('[data-testid="dropdown-select-report"]').click();
+      cy.get('[data-testid="select-pending-report"]').click();
+      cy.get('[data-testid="dropdown-select-judge"]').select('Chief Judge');
+    });
+
+    it('should have more rows in the table after clicking the "load more" button than before', () => {
```
mmarcotte: Hmm, this appears to assume that data will be prepopulated in the system. Will this work for a newly created environment that has no data in the pending report?

TomElliottFlexion: We have 113 pending items in our seed data, which should mean that this will always work for an integration test. Is that correct?

mmarcotte: hm, I guess I mistakenly thought this would be run upon a deployed environment... and if we wiped an environment clean, there wouldn't be any items in a pending report. Upon closer look, and thanks to your comment, this is only testing locally upon our seed data.

web-api/src/business/useCases/pendingItems/exportPendingReportInteractor.ts
```diff
@@ -0,0 +1,47 @@
...
+  const formattedPendingItems = pendingDocuments.map(pendingItem =>
+    applicationContext
+      .getUtilities()
+      .formatPendingItem(pendingItem, { applicationContext }),
+  );
+
+  return getCsv(formattedPendingItems);
+};
+
+const getCsv = data => {
```
TomElliottFlexion, line 34: Intentionally left here for now, can be injected if/when this is used elsewhere

TomElliottFlexion: Since config is usage-specific

cypress/cypress-integration/integration/pending-report.cy.ts
```diff
@@ -0,0 +1,46 @@
...
+import { loginAsDocketClerk } from '../../helpers/auth/login-as-helpers';
+import path from 'path';
+
+describe('Pending report', () => {
+  const today = formatNow(FORMATS.MMDDYYYY_UNDERSCORED);
+  const fileName = `Pending Report - Chief Judge ${today}.csv`;
+  const downloadsFolder = Cypress.config('downloadsFolder');
+
+  describe('View pending report', () => {
+    it('should log in as a docket clerk and navigate to the pending report then select chief judge from the dropdown', () => {
```
zachrog: To try and follow some of the best practices we have for cypress I think we would want all of these steps inside of a single it block, or each it block needs to have setup that allows it to run independently from one another.

Over from our 'integration-tests.md':
every `it` statement should be a repeatable test that should be able to run in isolation.  This means your `it` statement shouldn't depend on a previous it statement to pass.  If you write tests that depend on previous it statements, your tests can become flaky.  This also goes against the best practices outlined in the cypress docs.

shared/src/business/utilities/downloadCsv.ts
```diff
@@ -0,0 +1,15 @@
+export const downloadCsv = ({
```
zachrog: Nice Util you got there. 🍾🍾🍾

If this can only be used on the frontend could we move it over to web-client/src?

web-client/src/presenter/computeds/pendingReportListHelper.ts
```diff
@@ -16,9 +16,9 @@ export const pendingReportListHelper = (
 
   const showLoadMore =
     get(state.pendingReports.pendingItems).length < searchResultsCount;
-
-  const showNoPendingItems = searchResultsCount === 0 && hasPendingItemsResults;
   const showSelectJudgeText = !judge;
+  const showNoPendingItems =
+    searchResultsCount === 0 && !hasPendingItemsResults && !!judge;
```
zachrog, line 22: I experienced this the last time I worked over in pendingItems but to me it seems like state.pendingReports.hasPendingItemsResults is not clear to me what it is meant to be doing. Like hasPendingItemsResults seems like it is just whether or not we received results back but really it is doing more than that.

I think what it is meant to be used as is hasReportBeenRun. If you know how many results are in state, if the reportHasBeen run, and the selected judge I think you have all the information about what text to show on the screen. A suggestion I have is to refactor state.pendingReports.hasPendingItemsResults into state.pendingReports.hasReportBeenRun.

zachrog (changes requested):
Only blocking for the cypress tests update.

shared/src/business/utilities/downloadCsv.ts
TomElliottFlexion: Sure thing

cypress/cypress-integration/integration/pending-report.cy.ts
TomElliottFlexion: Good call, fixed

web-client/src/presenter/computeds/pendingReportListHelper.ts
TomElliottFlexion: I think after the updates, this variable is named correctly. We want to hide/show elements based on whether there actually _are_ pending item results, rather than whether the report has been run or not (this is where the bug came from)

zachrog (approved):

PR #4398, opened 2023-12-08
10185: IRS Calendar Admin Lookup

cruzjone-flexion:
[ZenHub Card](https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/gh/flexion/ef-cms/10185)

https://github.com/ustaxcourt/ef-cms/assets/94860350/291a7d3b-eab8-4aa1-995e-7ce42b380237


shared/src/business/useCases/getAllUsersByRoleInteractor.ts
```diff
@@ -0,0 +1,21 @@
...
+  isAuthorized,
+} from '@shared/authorization/authorizationClientService';
+import { UnauthorizedError } from '@web-api/errors/errors';
+
+export const getAllUsersByRoleInteractor = async (
+  applicationContext: IApplicationContext,
+  { roles }: { roles: string[] },
+) => {
+  const authorizedUser = applicationContext.getCurrentUser();
+  if (!isAuthorized(authorizedUser, ROLE_PERMISSIONS.CREATE_TRIAL_SESSION)) {
```
ujahio, line 12: Is it possible to move the authorization check to either its own action (function) or part of the validation action earlier in the sequence its been used?  that way this interactor is cleaner and agnostic to the user fetching roles. 

This interactor seems like an interactor that'll be used in context of other users performing different tasks when fetching roles of users, but please let me know if my assumptions are incorrect.

shared/src/business/useCases/trialSessions/generateTrialCalendarPdfInteractor.test.ts
```diff
@@ -197,6 +197,139 @@ describe('generateTrialCalendarPdfInteractor', () => {
...
+      .getPersistenceGateway()
+      .getTrialSessionById.mockReturnValue({
+        ...MOCK_TRIAL_INPERSON,
+        irsCalendarAdministratorInfo,
+      });
+
+    const result = await generateTrialCalendarPdfInteractor(
+      applicationContext,
+      {
+        trialSessionId: MOCK_TRIAL_INPERSON.trialSessionId!,
```
ujahio: small nit (and question): but why would trialSessionId be possibly undefined? Is it not already defined in  `MOCK_TRIAL_INPERSON` object?

web-client/src/presenter/computeds/TrialSession/getAllIrsPractitionersForSelectHelper.test.ts
```diff
@@ -0,0 +1,58 @@
...
+describe('getAllIrsPractitionersForSelectHelper', () => {
+  const TEST_IRS_PRACTITIONERS: any[] = [];
+  const getAllIrsPractitionersForSelectHelper = withAppContextDecorator(
+    getAllIrsPractitionersForSelectHelperComputed,
+    {
+      ...applicationContext,
+    },
+  );
+
+  it('should return an object from the helper', () => {
```
ujahio: ❓ : Is there a need for this test? It seems the next test is testing for both the "type" of the data returned from the helper and the actual formatted data.

web-client/src/presenter/computeds/TrialSession/getAllIrsPractitionersForSelectHelper.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export type FormattedIrsCalendarAdminInfo = {
+  label: string;
+  value: string;
+  email?: string;
+  name: string;
+  phone?: string;
+};
+
+export const getAllIrsPractitionersForSelectHelper = (get: Get): any => {
```
ujahio: nit: type return if possible.

shared/src/business/useCases/getAllUsersByRoleInteractor.ts
cruzjone-flexion: That would cause security risks. Moving validation to the front end where people/hackers can change the behavior as they please will allow people/hackers to gain access to private IRS Practitioners contact information. 

As of now this endpoint is only required to those that can access Create/Update a trial session and as an MVP that is the only Permission this endpoint is required until more features in the future require it we will expand that permission.

shared/src/business/useCases/trialSessions/generateTrialCalendarPdfInteractor.test.ts
cruzjone-flexion: True, will take a look at that. Outcome from copy and paste tests.

web-client/src/views/TrialSessions/SessionAssignmentsForm.tsx
```diff
@@ -8,18 +14,27 @@ export const SessionAssignmentsForm = connect(
...
   function SessionAssignmentsForm({
     form,
+    getAllIrsPractitionersForSelectHelper,
     judges,
     sessionAssignmentHelper,
     TRIAL_SESSION_PROCEEDING_TYPES,
+    updateScreenMetadataSequence,
     updateTrialSessionFormDataSequence,
   }) {
+    console.log(
```
ujahio: Remove log

web-client/src/presenter/computeds/TrialSession/getAllIrsPractitionersForSelectHelper.test.ts
cruzjone-flexion: My initial test to make sure everything is wired up correctly in the test file. Will remove since there is no benefits.

shared/src/business/useCases/getAllUsersByRoleInteractor.ts
ujahio: Thanks for explaining the security risks and I understand it even more. This is all backend so it makes sense to add it here. Thank you.

ujahio (approved):

PR #4399, opened 2023-12-08
DevEx 1231: Only index a record into ES if it was modified after a specific timestamp

jimlerza:
We are now setting a `DEPLOYMENT_TIMESTAMP` environment variable in the API lambdas at build time. This is the time that the Web-API Terraform deployment started, in epoch seconds.

In the processStreamRecords lambda, we are now comparing the creation time of a dynamodb stream event against the lambda's deployed timestamp and only indexing a record if it was written after the lambda was deployed.

Here's how this change will affect indexing in the three different types of deployments. Assume for clarity the current color is blue and the deploying color is green:

- non-migration
  - deployment_timestamp of blue is untouched; indexing continues on active color while deployment runs
  - deployment_timestamp of green is the time when terraform deployment started
  - green stream is turned on when the color switches; it starts consuming the stream but only indexes events after the deployment_timestamp
- migration
  - deployment_timestamp of blue is untouched; indexing continues on active color/db while deployment runs
  - deployment_timestamp of green is the time when terraform deployment started
  - data is copied from source db into destination db
  - green stream is turned on after the data is finished migrating; all migration writes will have timestamps after the deployment_timestamp
- glue job
  - we delete all dbs and clusters and create fresh empty alpha dbs + cluster with an “empty” deployment
  - after the color switch, green is the current color
  - we explicitly disable the green stream
  - we glue data from prod
  - after it finishes, we turn the green stream on
  - since all writes from glue job occured after the green stream’s deployment_timestamp, all of these writes will be indexed when the stream is enabled (provided it is enabled within 24 hours of the oldest glue write)


web-api/workflow-terraform/migration/main/lambdas/migration-segments.js
```diff
@@ -42,8 +42,6 @@ const scanTableSegment = async (
     await dynamoDbDocumentClient
       .scan({
         ExclusiveStartKey: lastKey,
-        ExpressionAttributeValues: { ':prefix': 'stream-event-id' },
-        FilterExpression: 'NOT begins_with(pk, :prefix)',
```
jimlerza: This code ensures that existing `stream-event-id` records are not migrated when we perform a migration. We don't want or need these records anymore. Should we wait to remove this code until we've done a migration in every environment?

web-api/src/lambdas/streams/processStreamRecordsLambda.ts
```diff
@@ -1,86 +1,42 @@
...
+    'dynamodb' in record &&
+    record.dynamodb &&
+    'ApproximateCreationDateTime' in record.dynamodb &&
+    typeof record.dynamodb.ApproximateCreationDateTime !== 'undefined'
+  ) {
+    // StreamRecord objects from local dynamodb have an ApproximateCreationDateTime with a type of Date
+    // StreamRecord objects from AWS dynamodb have an ApproximateCreationDateTime with a type of number (epoch seconds)
+    // here we'll handle both types and convert to epoch seconds if necessary
+    const approximateCreationDateTime: number =
+      Number(record.dynamodb.ApproximateCreationDateTime) > 9999999999
```
jimlerza: This is a pretty hacky way to determine if our number is epoch seconds or epoch millis. I am open to other suggestions.

jasonfu9966: it's not much better, but the only suggestion I have is using the length of the ACDateTime rather than the value

web-api/workflow-terraform/migration/main/lambdas/migration-segments.js
jasonfu9966: dum question: what's the difference between removing this code before we run the migrations vs. removing this code after?

jasonfu9966 (approved):
all in all, looks good

web-api/workflow-terraform/migration/main/lambdas/migration-segments.js
jimlerza: The purpose of this code is to _not migrate_ `stream-event-id` records.

Helpful context: for the first several months that the `stream-event-id` records were being generated, they were generated with TTL values in milliseconds (even though they're supposed to be seconds). We haven't run a migration in prod since then, so there are millions of `stream-event-id` records with TTLs that won't occur for 50,000 years.

So if we take this code out now, the `stream-event-id` records will continue to exist because without those lines, they'll get migrated.

Or we can wait to take this code out until we've run a migration in every environment.

codyseibert (approved):
lgtm

TomElliottFlexion (approved):

web-api/src/lambdas/streams/processStreamRecordsLambda.ts
mmarcotte: what about checking the types of what you're dealing with?

```
if (record?.dynamodb?.ApproximateCreationDateTime) {
  const {
    ApproximateCreationDateTime,
  }: { ApproximateCreationDateTime?: any } = record.dynamodb;

  let approximateCreationDateTime: number;

  if (ApproximateCreationDateTime instanceof Date) {
    approximateCreationDateTime = Math.floor(
      ApproximateCreationDateTime.getTime() / 1000,
    );
  } else if (typeof ApproximateCreationDateTime === 'number') {
    approximateCreationDateTime = ApproximateCreationDateTime;
  } else {
    throw new Error(
      `Could not handle ApproximateCreationDateTime: ${typeof ApproximateCreationDateTime}`,
    );
    // or return true; ? 
  }
  
  // ...

}
```

web-api/workflow-terraform/migration/main/lambdas/migration-segments.js
mmarcotte: I think it's safe, and important, to keep until we've run a migration.

web-api/src/lambdas/streams/processStreamRecordsLambda.ts
jimlerza: Love it, updated

mmarcotte (approved):

PR #4407, opened 2023-12-11
10185: IRS Calendar Administrator Lookup

cruzjone-flexion:
[ZenHub Card](https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/gh/flexion/ef-cms/10185)


shared/src/business/entities/trialSessions/IrsCalendarAdministratorInfo.ts
```diff
@@ -0,0 +1,32 @@
...
+  constructor(rawData: any) {
+    super('IrsCalendarAdministratorInfo');
+
+    this.name = rawData.name;
+    this.email = rawData.email;
+    this.phone = rawData.phone;
+  }
+
+  static VALIDATIONS = {
+    email: JoiValidationConstants.STRING.max(100).optional(),
```
mmarcotte, line 19: Does it make sense to use `JoiValidationConstants.EMAIL.` here?

cruzjone-flexion: The AC said they can modify the auto populated email input which use to be just a text area for all 3 inputs I wanted to make sure it continued to work as it did before. The legacy property in TrialSession.ts had only that string max 100 validation.

image

mmarcotte (approved):

nechama-krigsman (approved):

PR #4424, opened 2023-12-13
add usa-layout-grid to sass entrypoint

jasonfu9966:
[USWDS Version `3.7.1`](https://designsystem.digital.gov/about/releases/) removes the `usa-layout-grid` dependency from the `usa-footer` package, which reduces that package size but breaks things.

We fix this by re-importing it manually.


jimlerza (approved):

package.json
```diff
@@ -35,7 +35,7 @@
     "@fortawesome/react-fontawesome": "^0.2.0",
     "@joi/date": "^2.1.0",
     "@opensearch-project/opensearch": "^2.4.0",
-    "@uswds/uswds": "3.7.0",
+    "@uswds/uswds": "^3.7.1",
```
TomElliottFlexion: I think we should leave this locked (to 3.7.1). I have zero faith in USWDS maintaining correct semver now

TomElliottFlexion (approved):

package.json
jasonfu9966: I'm half with ya. I also no longer have faith in USDWS maintaining correct semver, but I'm a little disappointed this got through both manual and automated testing.

TomElliottFlexion: Splitting hairs here, but it only got through manual testing _because_ it wasn't version locked

TomElliottFlexion: But I do agree!

ujahio (approved):

mmarcotte (approved):

PR #4431, opened 2023-12-15
10205 - Auto-generate Entry of Appearance - Practitioner Flow

ujahio:


cypress/helpers/select-typeahead-input.ts
```diff
@@ -0,0 +1,7 @@
+export function selectTypeaheadInput(testId: string, value: string) {
+  // eslint-disable-next-line cypress/no-unnecessary-waiting
+  cy.wait(1000); // TODO: we shouldn't need this wait, but the app fails to type the document type without it
```
mmarcotte: Just curious, does this reliably pass with a wait time of less than a second?

mmarcotte (approved):

cypress/helpers/select-typeahead-input.ts
ujahio: Yeah I was able to test with a wait time of a tenth of a sec on edge and it passed reliably locally (10 tries) but in CI, it seems the anything less than a sec may not be enough.

mmarcotte (approved):

PR #4437, opened 2023-12-18
Fix flaky test when selecting status of matter

cruzjone-flexion:
Production code was updated to use a dictionary to avoid future migrations needed if update to label was needed.
Cypress test was not updated though, in test we randomly select a status of matter and select in the UI but when we asserted what was being shown in the label it does not match when a non single word status was selected.

i.e 
'AwaitingBriefs' !== 'Awaiting Briefs'


cypress/cypress-integration/integration/pending-motions-table-judge-and-chamber-dashboard.cy.ts
```diff
@@ -106,17 +112,19 @@ function generateRandomDate() {
 }
 
 function getRandomElementFromArray() {
```
TomElliottFlexion: This _might_ be worth a rename

TomElliottFlexion (approved):
Nice fix!

ujahio (approved):

nechama-krigsman (approved):

cypress/cypress-integration/integration/pending-motions-table-judge-and-chamber-dashboard.cy.ts
```diff
@@ -105,18 +113,20 @@ function generateRandomDate() {
...
+    AwaitingBriefs: 'Awaiting Briefs',
+    AwaitingConsideration: 'Awaiting Consideration',
+    AwaitingRelease: 'Awaiting Release',
+    Drafting: 'Drafting',
+    ReviewingDraft: 'Reviewing Draft',
+    RevisingDraft: 'Revising Draft',
+    Stayed: 'Stayed',
+    SubmittedToChiefJudge: 'Submitted to Chief Judge',
+    SubmittedToReporter: 'Submitted to Reporter',
+  };
```
mmarcotte: Can / should this reference the static property of the class itself?

`DocketEntryWorksheet.STATUS_OF_MATTER_OPTIONS_DICTIONARY` 

This way should that list change, we don't have to change it in two places?

cruzjone-flexion: Sadly if we import the entity into a cypress test that will require to update the configuration to allow entities to be bundled with cypress. Which will increase the size and import unnecessary files just for one dictionary.

cruzjone-flexion: Refactored test to not use dictionary and instead use the markup to get all available options and randomly select one.

cypress/cypress-integration/integration/pending-motions-table-judge-and-chamber-dashboard.cy.ts
```diff
@@ -113,20 +117,35 @@ function generateRandomDate() {
...
-    RevisingDraft: 'Revising Draft',
-    Stayed: 'Stayed',
-    SubmittedToChiefJudge: 'Submitted to Chief Judge',
-    SubmittedToReporter: 'Submitted to Reporter',
-  };
-
-  const statusOfMatterEntries = Object.entries(statusOfMatter);
-  const randomIndex = Math.floor(Math.random() * statusOfMatterEntries.length);
-  return statusOfMatterEntries[randomIndex];
+// function selectRandomStatusOfMatterEntry(): [string, string] {
```
ujahio: ❓ : Is this implementation still going to be necessary? Maybe, we can delete?

cypress/cypress-integration/integration/pending-motions-table-judge-and-chamber-dashboard.cy.ts
```diff
@@ -113,20 +117,35 @@ function generateRandomDate() {
...
+//     SubmittedToChiefJudge: 'Submitted to Chief Judge',
+//     SubmittedToReporter: 'Submitted to Reporter',
+//   };
+
+//   const statusOfMatterEntries = Object.entries(statusOfMatter);
+//   const randomIndex = Math.floor(Math.random() * statusOfMatterEntries.length);
+//   return statusOfMatterEntries[randomIndex];
+// }
+
+function selectRandomStatusOfMatterEntry(
```
ujahio, line 120: 👏🏿

cypress/cypress-integration/integration/pending-motions-table-judge-and-chamber-dashboard.cy.ts
cruzjone-flexion: Thnx. Totally forgot to do so

mmarcotte (approved):

PR #4441, opened 2023-12-19
10216: rework PublicCase.isPrivateDocument and accompanying tests

mmarcotte:


shared/src/business/entities/cases/PublicCase.isPrivateDocument.test.ts
```diff
@@ -1,155 +1,343 @@
...
+                  },
+                },
+                visibilityChangeDate,
+              );
+
+              expect(isPrivate).toEqual(false);
+            },
+          );
+
+          // there's currently no good way to determine this information
```
ujahio, line 220: ❓ : Are these test scenarios going to be addressed or can they be deleted?

shared/src/business/entities/cases/PublicCase.isPrivateDocument.test.ts
```diff
@@ -1,155 +1,343 @@
...
+                  docketEntryId: '123',
+                  eventCode,
+                  filedByRole,
+                  filingDate,
+                  isOnDocketRecord: true,
+                  isPaper: false,
+                  previousDocument: {
+                    docketEntryId: '2a389787-91d2-41ba-9c6e-2a13440a928b',
+                    documentType: 'Seriatim Answering Brief',
+                    // filedByRole: 'privatePractitioner'
```
ujahio: ❓ :  Is `filedByRole` still needed as a prop in previousDocument?

shared/src/business/entities/cases/PublicCase.isPrivateDocument.test.ts
mmarcotte: they are going to have to be addressed, but not in the scope of this pr

shared/src/business/entities/cases/PublicCase.isPrivateDocument.test.ts
mmarcotte: it is still needed, but it is not stored that way.

ujahio (approved):

PR #4450, opened 2023-12-21
10229: Bug: Printable practitioner case list

TomElliottFlexion:
https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/gh/flexion/ef-cms/10229


shared/src/business/useCases/generatePractitionerCaseListPdfInteractor.ts
```diff
@@ -34,13 +34,17 @@ export const generatePractitionerCaseListPdfInteractor = async (
...
-    .getCasesAssociatedWithUser({
+    .getDocketNumbersByUser({
       applicationContext,
       userId,
     });
 
+  const cases = await applicationContext
+    .getPersistenceGateway()
+    .getCasesByDocketNumbers({ applicationContext, docketNumbers });
+
```
TomElliottFlexion, line 47: Could use opensearch here, but trying to avoid it if not necessary

ujahio (approved):

PR #4454, opened 2023-12-21
10229: Bug: Printable case list

TomElliottFlexion:
Uses dynamo to get full cases instead of user-case records now; can be refactored to use OpenSearch if this is too slow

https://github.com/flexion/ef-cms/issues/10229


shared/src/business/useCases/generatePractitionerCaseListPdfInteractor.ts
```diff
@@ -34,13 +34,17 @@ export const generatePractitionerCaseListPdfInteractor = async (
...
-    .getCasesAssociatedWithUser({
+    .getDocketNumbersByUser({
       applicationContext,
       userId,
     });
 
+  const cases = await applicationContext
+    .getPersistenceGateway()
+    .getCasesByDocketNumbers({ applicationContext, docketNumbers });
+
```
TomElliottFlexion, line 47: Could use OpenSearch here, but trying to avoid it if not necessary

Absolutestunna: Hmm.. in my mind, it's 1 opensearch call vs 2 dynamodb calls to get a subset of a case. It's a nit for me but I prefer using opensearch for fetching cases in this instance.

ujahio (approved):

mmarcotte (approved):

PR #4465, opened 2023-12-27
10176: Refactor the migration script

cruzjone-flexion:


rachelschneiderman (approved):

web-api/workflow-terraform/migration/main/lambdas/migrations/throwAway.ts
mmarcotte, line 1: just highlighting i think you wish to remove this file at some point

web-api/workflow-terraform/migration/main/lambdas/migrations/0001-add-associated-judge-id.ts
```diff
@@ -18,62 +19,33 @@ const isRecordToUpdate = item => {
...
 ) => {
-  const judgeRecords = await getAllJudgeRecords(documentClient);
-
-  const judgesMap = judgeRecords.reduce((accumulator, judge) => {
-    accumulator[judge.name] = judge.userId;
-    return accumulator;
-  }, {});
-
   const itemsAfter: TDynamoRecord[] = [];
+  let judgesMap: { [key: string]: string } | null = null;
```
mmarcotte: with `judgesMap` scoped to just this function, it will be looked up every time this function is called, and this function could be called many times per segment as the lambda paginates through records. its value should not change throughout the course of this lambda running. does it make sense to lift it out of this function so it is only looked up once?

PR #4476, opened 2023-12-29
USTC 4372 reconciliation report enhancements

jtdevos:
This PR adds enhances the "reconciliation report" v2 API endpoint adding optional time start/end parameters.  These parameters allow the caller to further narrow the list of results to only include docket entries served to the IRS.


mmarcotte (approved):
you do have some linting errors, but we can clean those up before we go to staging.

PR #4480, opened 2024-01-01
10176: Provide Judge Id to Cases/CaseItems/WorkItems

cruzjone-flexion:
[Trello Card](https://trello.com/c/drLeqrUt/1167-associate-judge-with-a-case-using-userid)

Wherever we update/add/remove associatedJudge to Cases/CaseDeadlines/WorkItems we also update/add/remove to those entities. Now that we have the ID available we can start strangling out the usage of name and use the ID.

Migration: Our migration script will add the ID to all the entities already in the DB with the correct corresponding Judge ID

Feature Update: We also updated the Judge/Chambers Pending Motion Table to use the ID instead of the name


web-api/workflow-terraform/migration/bin/deploy-init.sh
```diff
@@ -9,23 +9,28 @@ ENVIRONMENT=$1
...
 echo "  - SOURCE_TABLE=${SOURCE_TABLE}"
 echo "  - DESTINATION_TABLE=${DESTINATION_TABLE}"
 echo "  - EFCMS_DOMAIN=${EFCMS_DOMAIN}"
 echo "  - DOCUMENTS_BUCKET_NAME=${DOCUMENTS_BUCKET_NAME}"
-
-export ENVIRONMENT="${ENVIRONMENT}"
-
-STREAM_ARN=$(aws dynamodbstreams list-streams --region us-east-1 --query "Streams[?TableName=='${SOURCE_TABLE}'].StreamArn | [0]" --output text)
+echo "  - STREAM_ARN=${STREAM_ARN}"
+echo "  - ELASTICSEARCH_ENDPOINT=${ELASTICSEARCH_ENDPOINT}"
```
mmarcotte: Hey can we remove these. I think they were here largely for our debugging, but there's no good reason to expose this information during CI. ty!

cruzjone-flexion: Sounds good will delete these logs

mmarcotte (approved):

nechama-krigsman (approved):

web-client/integration-tests-public/unauthedUserCaseSearchExactMatchMultipleFields.test.ts
```diff
@@ -21,7 +21,7 @@ const [firstName, lastName] = [
 
 const nameToSearchFor = `${firstName} ${lastName}`;
 
-const createdDocketNumbers = [];
+const createdDocketNumbers: any[] = [];
```
ujahio, line 24: small nit: I think it's safe to assume it'll be an array of strings in this instance. It may be more precise to type as such.

web-api/workflow-terraform/migration/main/lambdas/migrations/0001-add-associated-judge-id.ts
```diff
@@ -0,0 +1,61 @@
...
+
+const isWorkItem = item => {
+  return item.gsi1pk?.startsWith('work-item|');
+};
+
+const isRecordToUpdate = item => {
+  return isCaseRecord(item) || isCaseDeadline(item) || isWorkItem(item);
+};
+
+let judgesMap: { [key: string]: string } | null = null;
```
ujahio, line 23: small nit/ ❓ : Can this just be an empty object by default?

ujahio (approved):

jimlerza (approved):

PR #4482, opened 2024-01-01
New typescript check

cruzjone-flexion:
This new Github action will check the untouched files in your PR and see if the typescript error count has increased on any of them. This will catch potential bugs and catch any part of the code Devs did not manually check.

It is using the `tsc` CLI command on your PR and the destination branch (test/staging).
Create a map of files and their count.
Make an array of all the files you touched in your new PR 
Compare files not touched in your PR and check if the count has increased
Console log all the files 

Example:
image


TomElliottFlexion (approved):

.github/workflows/typescript-untouched-files-check.yml
```diff
@@ -0,0 +1,45 @@
+name: Typescript Untouched File Check
+
+on:
+  pull_request:
+  push:
+    branches:
+      - dependency-updates
```
ujahio: ❓ : Should this action only run on `dependency-updates`? It's my understanding it should run on any branch going to an upper envs.

ujahio (approved):

.github/workflows/typescript-untouched-files-check.yml
cruzjone-flexion: Good catch, will update. 
It seemed to have been working even with that hardcoded branch name. Even our Typescript count Github Action has dependency-updates hardcoded. Weird, will look into it. (Not too familiar with Github Action templates)

zachrog (approved):

mmarcotte (approved):

PR #4483, opened 2024-01-02
10210 bugfix

ujahio:


cruzjone-flexion (approved):

jasonfu9966 (approved):
just curious -- would this change create any weird interactions on narrower views, e.g. mobile?

ujahio:
> just curious -- would this change create any weird interactions on narrower views, e.g. mobile?

This component is for internal users. I'm not accounting for mobile view in this case.

PR #4490, opened 2024-01-02
10221 bug intermediate branch to test 1704238717

pixiwyn:


web-api/setup-elasticsearch-index.sh
```diff
@@ -20,29 +20,26 @@
```
pixiwyn, line 20: Not sure why this file is different, this isn't apart of my pr, but don't want to break test, please confirm this is OK before merging to test.

cruzjone-flexion: This is an update for our migration process, Mike worked on this

cruzjone-flexion (approved):

nechama-krigsman (approved):

PR #4496, opened 2024-01-03
10216 missing previous document

mmarcotte:


shared/src/business/entities/cases/PublicCase.ts
```diff
@@ -111,38 +111,56 @@ export class PublicCase extends JoiValidationEntity {
...
+      return false;
+    }
 
-    const hasPolicyDateImpactedEventCode =
-      POLICY_DATE_IMPACTED_EVENTCODES.includes(docketEntry.eventCode);
-    const isOrder = orderDocumentTypes.includes(docketEntry.documentType);
-    const isCourtIssuedDocument = COURT_ISSUED_EVENT_CODES.map(
-      ({ eventCode }) => eventCode,
-    ).includes(docketEntry.eventCode);
+    if (
```
zachrog, line 134: I like the return early pattern you have going here rather than nested ifs.

zachrog (approved):

PR #4530, opened 2024-01-08
10232 cases endpoint 504s

mmarcotte:
solves flexion#10232

simply removes the validation that was causing our Lambdas to labor for users with many cases. (1,000 - 10,000)

also includes refactoring to batchGet function to perform requests in parallel.


ujahio (approved):

jasonfu9966 (approved):

jimlerza (approved):

TomElliottFlexion (approved):

TomElliottFlexion:
Is there any concern about potentially delaying discovering broken cases? The frequency with which we find bad data via this mechanism gives me pause

mmarcotte:
> Is there any concern about potentially delaying discovering broken cases? The frequency with which we find bad data via this mechanism gives me pause

Tom, while I agree that this is an issue that needs addressing, I don't think this is the solution to that issue. [This issue](https://github.com/flexion/ef-cms/issues/10230) created by Jon and Nechama will be a much handier tool, I think.

PR #4531, opened 2024-01-08
10203: Add "ATP" event code;

cruzjone-flexion:
[ZenHub Card](https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/gh/flexion/ef-cms/10203)

Add new event code "ATP"


nechama-krigsman (approved):

ujahio (approved):

shared/src/business/useCaseHelper/serveDocumentAndGetPaperServicePdf.ts
```diff
@@ -15,12 +15,14 @@ export const serveDocumentAndGetPaperServicePdf = async ({
   applicationContext,
   caseEntities,
   docketEntryId,
+  electronicParties,
```
mmarcotte, line 18: does it make more sense for this just to default to 

```
electronicParties = []
```

that way something that calls it doesn't need to worry about passing in an empty array?

shared/src/business/useCases/docketEntry/addPaperFilingInteractor.ts
```diff
@@ -190,12 +190,19 @@ export const addPaperFiling = async (
   let paperServicePdfUrl;
 
   if (isReadyForService) {
+    const currentDocketEntry = caseEntities[0].getDocketEntryById({
+      docketEntryId,
+    });
+    const electronicParties =
+      currentDocketEntry.eventCode === 'ATP' ? [] : undefined;
+
```
mmarcotte, line 198: I'm trying to understand why this is necessary.

cruzjone-flexion: For event code ATP the method where we are sending in this variable requires a new process when sending email notification, and so to keep other interactors that use this method the same as before and not implement many if conditions we pass in this new electronicParties parameter and if it is undefined it works as usual but if we pass it in it uses it to override who gets an email notification.

mmarcotte (approved):

shared/src/business/useCaseHelper/serveDocumentAndGetPaperServicePdf.ts
cruzjone-flexion: If any other place calls this and does not define electronicParties it will work with the old behavior. Only when this is passed (truthy value) it will override and use new logic. By adding a default value we then need to add more if/else conditions into the method;

PR #4546, opened 2024-01-12
devex 1044: add terraform script that defines opensearch snapshot repository

jtdevos:
This imports existing resources into the terraform state [devex 1044](https://trello.com/c/eKQ7w2ZI/1044-define-the-elasticsearch-snapshot-repository-in-the-account-specific-terraform).

## Manual Deploy Steps:
 - [x] add a new `LOG_SNAPSHOT_BUCKET_NAME` variable to `account_deploy` secrets w/ the name of the existing log snapshots bucket in the given account
 - [x] run script to import existing resources into terraform state:
   ```
   cd iam/terraform/account-specific/main
   ../bin/deploy-opensearch-info.sh
   ```
 - [x] run account-specific terraform
   ```
   npm run deploy:account-specific
   ```


jtdevos:
fyi: we already ran this against the staging & prod USTC accounts.  It only needs to be run for the flexion account.

iam/terraform/account-specific/main/elasticsearch.tf
```diff
@@ -143,6 +143,17 @@ resource "aws_cognito_identity_pool_roles_attachment" "log_viewers" {
...
 }
 
+resource "opensearch_snapshot_repository" "archived-logs" {
+  name = "archived-logs"
+  type = "s3"
+  settings = {
+    bucket   = "${var.log_snapshot_bucket_name}"
+    region   = "us-east-1"
+    role_arn = "arn:aws:iam::${data.aws_caller_identity.current.account_id}:role/es-s3-snapshot-access"
+    # role_arn = aws_iam_role.es_s3_snapshot_access_role.arn
```
mmarcotte: is this still helpful? is this the one that should be using? looks incomplete

iam/terraform/account-specific/main/main.tf
```diff
@@ -12,11 +12,20 @@ provider "aws" {
   alias  = "us-west-1"
 }
 
+provider "opensearch" {
+  # url = aws_opensearch_domain.efcms-logs.url
```
mmarcotte: do we still need this comment? which one should we be using?

iam/terraform/account-specific/main/s3.tf
```diff
@@ -0,0 +1,51 @@
...
+      "Action": ["s3:PutObject", "s3:GetObject", "s3:DeleteObject"],
+      "Resource": "arn:aws:s3:::${var.log_snapshot_bucket_name}/*"
+    }
+  ]}
+EOF
+}
+
+resource "aws_s3_bucket" "ustc_log_snapshots_bucket" {
+  bucket = "${var.log_snapshot_bucket_name}"
+#   acl    = "private"
```
mmarcotte: should the acl be private? I think so...

iam/terraform/account-specific/main/variables.tf
```diff
@@ -34,3 +34,8 @@ variable "number_of_days_to_keep_info_logs" {
 variable "dawson_dev_trusted_role_arns" {
   type = set(string)
 }
+
+variable "log_snapshot_bucket_name" {
+  type = string
+  default = "ustc-log-snapshots"
+}
```
mmarcotte: new line please :)

iam/terraform/account-specific/main/elasticsearch.tf
jtdevos: Yeah I think this section needs to be fixed (as well as a section in `s3.tf`) but I'm not sure how.  

The commented variable `aws_iam_role.es_s3_snapshot_access_role.arn` resolves to `arn:aws:iam::{id goes here}:role/es_s3_snapshot_access_role`   Which is a resource that's defined in `s3.tf`.   

The not-commented line refers to `es-s3-snapshot-access`, which is referenced in `s3.tf` but not actually defined.    I assume we either need to choose one of these role names and then replace the references to the other, but I'm not sure which role name is "right".

iam/terraform/account-specific/main/s3.tf
jtdevos: Definitely should be private.  This was commented-out because the `acl` setting is deprecated for `aws_s3_bucket`.   Private is the default, however,  I added a `aws_s3_bucket_acl` block that sets the bucket to private just to be safe

iam/terraform/account-specific/main/main.tf
jtdevos: Comment no-longer needed.  The correct form is to use the `.endpoint` value.

jimlerza (approved):

iam/terraform/account-specific/main/elasticsearch.tf
jtdevos: Okay I think this section is truly fixed this time in the latest commit. There were two issues:
- we were referring to the wrong ARN name (should be `es_s3_snapshot_access_role` not `es-s3-snapshot-access`
- we used string interpolation to specify the `role_arn` value instead of using terraforms expression syntax (`aws_iam_role.es_s3_snapshot_access_role.arn`).   Originally me and Jim thought this line was causing an error, but upon a further look it turns out not to be the case.

mmarcotte (approved):

PR #4585, opened 2024-01-24
Devex health migration script

jtdevos:
Updated the `health-status` script to work when the clusters have different index names. This happens when there is a mapping change and a database migration in same deployment.


zachrog (approved):
Nice! If you are feeling up to it would you mind updating this file to be a Typescript file? If not no worries.

jimlerza:
> Nice! If you are feeling up to it would you mind updating this file to be a Typescript file? If not no worries.

[Typescriptified](d2b3fa269d69d3d1c3f40afff9ba825c4ba93cf2)!

mmarcotte (approved):

PR #4591, opened 2024-01-24
10219 upload failures to s3 intermediate branch to test 1706135786

codyseibert:
catching any errors involved with a user's browser uploading a file to our s3 bucket and send the error to our own api so we can log the error


codyseibert:
closing for now since we are looking into using sentry and it already catches our errors

PR #4632, opened 2024-02-07
Opex: List Trial Sessions by Year

jimlerza:
Added a script to output some trial session information by year.


scripts/reports/trial-sessions-by-year.ts
```diff
@@ -0,0 +1,46 @@
+// usage: npx ts-node --transpile-only ./scripts/reports/trial-sessions-by-year.ts 2023 > ~/Desktop/2023-trial-sessions.csv
+
+import {
+  FORMATS,
+  formatDateString,
+} from '@shared/business/utilities/DateHandler';
+import { createApplicationContext } from '@web-api/applicationContext';
+
+const year = process.argv[2] || '2022';
```
nechama-krigsman, line 9: Just curious, why is the default 2022?
Also, should the start time be midnight?

nechama-krigsman (approved):

scripts/reports/trial-sessions-by-year.ts
jimlerza: We defaulted to 2022 because that's the year they originally wanted these stats for!

As for the time, it actually is midnight Eastern Time, which is the Court's timezone.

nechama-krigsman: 👍

ujahio (approved):

mmarcotte (approved):

PR #4634, opened 2024-02-08
10257 speed up serve to irs to test

jtdevos:
This introduces a smoketest that tries to do more things while the petitions clerk is creating a paper petition. It also checks the boxes to enable the maximum number of auto-generated orders to stress-test pdf generation.

Also changes the `/serve-to-irs/` endpoint to be async, and parallelizes `serveCaseToIrsInteractor.ts`.


mmarcotte:
we inadvertently created this branch based off of test. recreating with one based off of staging

PR #4638, opened 2024-02-12
10225: BUGFIX: If you have multiple tabs open when serving the 30 day Notice of Trial, the other tabs display a 404 error

cruzjone-flexion:
The issue here was there was no "clientConnectionId" being passed to the interactor, so the websocket response was sent to all the browser tabs opened by this user.

The handler of the websocket response would redirected to a Page that required things/state to be ready (which only the tab that requested the async data has). Which is why all the other tabs would 404.

SOLUTION: Pass the "clientConnectionId" from state to the interactor;


shared/src/proxies/trialSessions/serveThirtyDayNoticeProxy.ts
```diff
@@ -2,11 +2,11 @@ import { post } from '../requests';
 
 export const serveThirtyDayNoticeInteractor = (
   applicationContext,
-  { trialSessionId },
+  { clientConnectionId, trialSessionId },
```
Absolutestunna, line 5: small nit: add type

ujahio (approved):

mmarcotte (approved):

PR #4657, opened 2024-02-15
10257 serve case to irs fixes

mmarcotte:
solves flexion#10257

makes serve to irs endpoint async


jimlerza (approved):

cypress/helpers/create-and-serve-paper-petition.ts
```diff
@@ -52,6 +77,55 @@ export function createAndServePaperPetition(
...
+        { eventCode: 'RQT', index: 4, servedTo: 'R' },
+        { eventCode: 'NOTR', index: 5, servedTo: 'P' },
+      ];
+
+      expectedDocuments.forEach(({ eventCode, index, servedTo }) => {
+        cy.get(
+          `[data-test-document-index="${index}"] > [data-test-document-meta="eventCode"]`,
+        ).should('have.text', eventCode);
+        cy.get(
+          `[data-test-document-index="${index}"] > [data-test-document-meta="servedPartiesCode"]`,
```
zachrog: For cypress I just wanted to ask about a couple of the different selectors I see. 
Why did you opt for data-test-document-index, data-test-document-meta as opposed to data-testid?

cypress/helpers/search-by-docket-number-in-header.ts
```diff
@@ -1,4 +1,5 @@
 export function searchByDocketNumberInHeader(docketNumber: string) {
+  cy.get('[data-testid="docket-number-search-input"]').scrollIntoView();
```
zachrog, line 2: image
Just some information you might like, there is a way to specify in cypress for each describe how you want cypress to scroll. It has helped me sooooo much and makes it so that everytime you select something on the page the browser will center it in the screen.

web-client/src/presenter/actions/StartCaseInternal/serveCaseToIrsAction.ts
```diff
@@ -16,16 +16,20 @@ export const serveCaseToIrsAction = async ({
   props,
 }: ActionProps) => {
   const docketNumber = props.docketNumber || get(state.caseDetail.docketNumber);
+  const clientConnectionId = get(state.clientConnectionId);
 
-  const pdfUrl = await applicationContext
-    .getUseCases()
-    .serveCaseToIrsInteractor(applicationContext, {
-      docketNumber,
+  try {
```
zachrog: Now that this is async and the real work is happening behind the scenes with the lambda I don't think that we need the try catch around this interactor call because all of our async calls respond back that they work initially. It also looks like you handled the async error in the socketRouter too so niiiiiiice.

zachrog (approved):

cypress/helpers/create-and-serve-paper-petition.ts
mmarcotte: Yeah, you bet.

1. `data-testid` was already present on the row we were querying. And it felt like a daunting challenge to determine the `docketEntryId` for each docket entry, which is what it would have taken during the process of filing and serving the petition to the IRS with all of the attachments. 
2. Thought this was much more representative of how Tenille would test it 
3. Thought it would be more maintainable in the long run.
4. `data-testid` is not descriptive. Now when I read the test as a human, I have a sense of what I'm looking for. 
5. Other tests can now more easily check the row for a specific docket record index rather than having to know the `docketEntryId`.

web-client/src/presenter/actions/StartCaseInternal/serveCaseToIrsAction.ts
mmarcotte: 🤔 Hmm, I'm a little confused. How else would I report to the client when an error occurs in the async lambda? 

The async lambdas report immediately to the client, and then they rely on a websocket message to relay successes, failures, or progress updates as necessary.

zachrog: In an async lambda the api gateway immediately responds successfully before the lambda's interactor has even started. So the try/catch here can only really catch if the user's computer does not have internet access or if our API is down which can happen but is just unlikely.

Like you said it is in the websocket messages that all of the success, errors, or progress is communicated. So to answer your question the websockets is where you would report to the client that an error has occurred in the async lambda.

cypress/helpers/create-and-serve-paper-petition.ts
zachrog: Those make sense. I think wanting more descriptive selectors is a good thing. My suggestion would be to make the value descriptive, not the html attribute descriptive to achieve that. I suggest that for two reasons:

1. consistency in selectors
2. Cypress specifically looks for `data-testid` (among others) when determining a good selector to use with the new record feature of cypress. 
image
https://docs.cypress.io/api/cypress-api/selector-playground-api#Default-Selector-Priority

My suggestion would be to do something like data-testid=`document-index-{index}` as opposed to data-test-document-index=`${index}`

web-client/src/presenter/actions/StartCaseInternal/serveCaseToIrsAction.ts
mmarcotte: Yes, I'm sorry. I get it now. For some odd reason I thought this was referring to the try ... catch block in the Interactor, not the Action. 

Yes, I agree this would really only catch whether or not the user's computer has internet access. And should that happen, I think there's probably a better strategy to employ to handle failed network requests.

cypress/helpers/create-and-serve-paper-petition.ts
zachrog: Nice job!

PR #4669, opened 2024-02-21
10216 - adding another isFileAttached for stamped motions

codyseibert:


mmarcotte (approved):

rachelschneiderman (commented):
Can you please add to `stamp-order.cy.ts` so this bug would have been caught in CI?

PR #4670, opened 2024-02-21
Devex set is file attached

mmarcotte:


rachelschneiderman:
I didn't get my comment added in time to [this PR](https://github.com/ustaxcourt/ef-cms/pull/4669), is it possible to add to `stamp-order.cy.ts` to catch the bug that was found on ustc/test yesterday?

PR #4675, opened 2024-02-23
10270 zip bug

zachrog:
# What error is occurring?
- On the public api when downloading the printable docket record the lambda explodes with the below error:
```
Cannot find module './572.js'
Require stack:
- /var/task/api-public.js
- /var/runtime/index.mjs
```

## Why is this error ocurring?
This error is occurring because terraform is not zipping all of the build files that are generated by webpack. When webpack bundles our code it outputs multiple *.js files for the entry-points but it also outputs files that are not related to any entry points. If you run a lambda build by running `npm run build:lambda:api` and look in the output directory `web-api/terraform/template/lambdas/dist` You will see these files:
 - 10.js
 - 136.js
 - 232.js
 - 304.js
 - 360.js
 - 572.js
 - 656.js
 - 660.js
 - 776.js
 - 876.js
 - api-public.js

Looking at these files you can see 572.js file which is the file that the public-api error is mentioning. Terraform is not bundling api-public.js + 572.js into the same .zip file when it creates the zip and uploads it to AWS. The reason that Terraform is not bundling 572.js is because we are only telling terraform to create a zip of api-public.js instead of the whole directory. This is occurring on line 229 inside file `web-api/terraform/template/main-east.tf`

## Why does webpack produce multiple files for a single entry point?
Webpack produces multiple files for a single entry point when our app or our dependencies utilize dynamic imports. A dynamic import is an import that happens in-line instead of the top level of a file and is utilized to help reduce load times caused by imports, and only to import a file when it is needed. A dynamic import looks like this: `const { default: chromium } = await import('@sparticuz/chromium');`.

## Why does it only occur on public?
This occurs only on public because the public api terraform creates a zip out of a single file while the regular app makes a zip out of the whole directory

## Why did this start happening with an upgrade to @aws-sdk/client-lambda?
This likely occurred because @aws-sdk/client-lambda lambda started to utilize dynamic imports somewhere in their codebase, causing webpack to code split and make multiple output files.

Successful deploy: https://app.circleci.com/pipelines/github/flexion/ef-cms/51044/workflows/df8fe717-a363-4550-9aa8-f65a6e6c563d

### DIFF
--- a/cypress/cypress-readonly/integration/public/public-docket-record.cy.ts
+++ b/cypress/cypress-readonly/integration/public/public-docket-record.cy.ts
@@ -1,0 +1,12 @@
+describe('Public Docket Record', () => {
+  it('should allow the user to generate and download a PDF of the docket record', () => {
+    cy.visit('/');
+    cy.get('input#docket-number').type('104-20');
+    cy.get('button#docket-search-button').click();
+    cy.get('[data-testid="header-public-case-detail"]').contains(
+      'Docket Number: 104-20',
+    );
+    cy.get('[data-testid="print-public-docket-record-button"]').click();
+    cy.get('[data-testid="modal-header"]');
+  });
+});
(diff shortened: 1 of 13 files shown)


jtdevos (approved):
Deployed to ustc-exp1 and confirmed bug no longer present.   Great work, many thanks to you both!

jtdevos:
Also, I really appreciate the thorough breakdown & discussion of the underlying problem.  Great stuff.

ujahio (approved):

jimlerza (approved):

TomElliottFlexion (approved):

PR #4676, opened 2024-02-23
reverting certain aws 3.513.0 deps to 3.490.0

jtdevos:
(note, i should have checked the existing pull requests before adding this one - I believe Tom's PR also may fix this.  if so, pls disregard!)

Through trial-and-error, I was able to get the print docket record action working again by reverting some of the `@aws-sdk/` dependencies from `3.513` to `3.490`.   I see that @TomElliottFlexion has a narrower set of changes (d'oh!) but I haven't yet had a chance to deploy to `ustc-exp1` to see if it works.


jtdevos:
seems like Rachel + Zack's PR is the correct approach. closing.

PR #4681, opened 2024-02-26
10217 + 10218 - Attachments to Petition (Modify E-filing and Paper Flow)

ujahio:
[#10217](https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/gh/flexion/ef-cms/10217) and [#10218](https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/gh/flexion/ef-cms/10218)

This story gives petitioners the option to upload an Attachment to Petition if desired. It is included as part of the petitioners e-filing flow. Also included are text changes.

#### Other things accomplished
- Renamed `petitionerCreatesEletronicCase` cypress test to have the correct spelling, renamed others like it
- Extracted the `setupPercentDone` function into its own action, refactored all usages. It had been exported and used in five or so places.


jasonfu9966 (approved):

web-client/src/presenter/actions/setProgressForFileUploadAction.ts
```diff
@@ -0,0 +1,79 @@
...
+  files: {
+    size: string;
+    name: string;
+  };
+}>): {
+  fileUploadProgressMap: FileUploadProgressMapType;
+} => {
+  const { files } = props;
+  const loadedAmounts: Record<string, number> = {};
+  const startTime = new Date();
```
mmarcotte, line 18: can you just add the linting rule to this line, too.

mmarcotte (changes requested):
great work! just one nitpick change

web-client/src/presenter/actions/setProgressForFileUploadAction.ts
jasonfu9966: sure! though I was wondering, since this is prohibited, should we use DateHandler instead? I was wondering why it's "O.K. to use Date constructor for calculating time duration."

mmarcotte: Fair question. While I think this implementation is simple and straightforward enough to use `Date.getTime()`, which is why I suggested putting the linting exception in place here, perhaps it is best to stay consistent with the currently adopted approach and use a combindation of `createISODateString` and `dateStringsCompared`

jasonfu9966: upon further review/investigation/whatnot I think it'd be better to add the linting rule to the line instead of making the change. thanks!

mmarcotte (approved):

PR #4694, opened 2024-03-01
10217 story to test 

jasonfu9966:
fixed an issue where the "Invalid PDF" modal was not showing properly.


nechama-krigsman (approved):
can we create a test that would catch this?

jasonfu9966:
> can we create a test that would catch this?

thank you for the suggestion! I added a test for this.

PR #4701, opened 2024-03-04
Opex: trial sessions by fiscal year report

jimlerza:
Refactored the `trial-sessions-by-year` report to work for either calendar or fiscal year, and to optionally output aggregated statistics instead of raw data.


ujahio (approved):

package.json
```diff
@@ -183,6 +183,8 @@
     "reindex:elasticsearch": "./reindex-elasticsearch.sh",
     "report:event-codes-by-year": "npx ts-node --transpile-only scripts/reports/event-codes-by-year.ts",
     "report:petition-counts": "npx ts-node --transpile-only scripts/reports/petition-counts.ts",
+    "report:trial-sessions-by-fiscal-year": "npx ts-node --transpile-only scripts/reports/trial-sessions-by-fiscal-year.ts",
```
zachrog, line 186: Totally recognize that people may feel differently, but for me I don't see the need to add individual scripts to the package.json if they are only a wrapper for calling a node script. If it is in the scripts directory(thank you btw) then I will run it as a script.

I know that is our pattern right now and I am okay with it, but wanted to bring it up as I see our package.json getting larger and larger.

zachrog (approved):

PR #4702, opened 2024-03-05
10216: deprecate isMinuteEntry 

mmarcotte:
… since its real source of truth is the eventCode. it was redundant


jasonfu9966 (approved):
love it

PR #4707, opened 2024-03-06
10270: BUG: Email sent verification

TomElliottFlexion:
- Uses existing efcms domain and mx records to allow SES email receiving
- Adds a new email forwarding ruleset, rule, and associated s3 bucket + policy to receive smoketest emails
    - Forwarding rule and s3 bucket + policy are not deployed to prod
- Adds a standalone Cypress smoketest that updates a petitioner's email address and confirms that the verification email was received
- Will live in test for at least a week to ensure stability


jasonfu9966 (approved):

cypress/cypress-smoketests/integration/verify-email-receipt.cy.ts
```diff
@@ -0,0 +1,67 @@
...
+  });
+} else {
+  describe('we do not want this test to run locally, so we mock out a test to make it skip', () => {
+    it('should skip', () => {
+      expect(true).to.equal(true);
+    });
+  });
+}
+
+const readS3InboxAndRetry = ({
```
cruzjone-flexion: Can we use the existing retry method Cody created?

cruzjone-flexion (approved):

cypress/cypress-smoketests/support/email-receipt.ts
```diff
@@ -0,0 +1,110 @@
...
+  if (retries) {
+    try {
+      return await deleteAllItemsInEmailBucket(bucketName);
+    } catch (error) {
+      console.error(
+        'Error with deleting items in email bucket. Retrying. Error: ',
+        error,
+      );
+
+      setTimeout(() => {}, 2000);
```
nechama-krigsman: do we need to make sure it waits before retrying? Should we wrap the timeout in a promise to make sure it waits?

nechama-krigsman (approved):

cypress/cypress-smoketests/support/email-receipt.ts
TomElliottFlexion: Just saw this, agreed!

cypress/cypress-smoketests/integration/verify-email-receipt.cy.ts
TomElliottFlexion: We did initially refactor to use Cody's retry, but then realized we don't need to be retrying in Cypress and moved the logic to the `read` and `delete` tasks

zachrog: Loved that this moved into the task! nice move.

PR #4711, opened 2024-03-06
Devex 1258: convert `migration` and `migration-segments` lambdas to typescript

jimlerza:


web-api/utilities/scanFull.ts
```diff
@@ -1,9 +1,10 @@
 import { Key } from 'aws-sdk/clients/dynamodb';
 import { TDynamoRecord } from '@web-api/persistence/dynamo/dynamoTypes';
+import type { DynamoDBDocument } from '@aws-sdk/lib-dynamodb';
```
zachrog, line 3: 🚀

web-api/workflow-terraform/migration/main/lambdas/migration.ts
```diff
@@ -0,0 +1,114 @@
...
+    migrateRecords: migrationsCallback;
+  },
+): Promise<void> => {
+  const promises: Promise<PutCommandOutput>[] = [];
+
+  items = await migrateRecords(applicationContext, { documentClient, items });
+
+  for (const item of items) {
+    promises.push(
+      documentClient.put({
```
zachrog: If this is for the migration lambda I imagine the performance is quite important. If we are trying to do many puts or many deletes I think it is important that we use batchWrite() instead of individual puts or individual deletes. 

There are a couple of gotchas with batchWrite() like it can only handle 25 commands at a time, but let me know if I can provide any help on how to implement or point to examples.

zachrog (changes requested):
Blocking for request to switch to batchwrite instead of individual put or individual delete.

Overall love that you are tackling this migration to typescript. Thankyou!

web-api/workflow-terraform/migration/main/lambdas/migration.ts
```diff
@@ -0,0 +1,126 @@
...
+      items,
+      migrateRecords: migrations,
+    });
+  }
+
+  requests = [...requests, ...generateDeleteRequests(event)];
+
+  const requestChunks = chunk(requests, 25);
+  const commands: BatchWriteCommand[] = [];
+  for (const requestChunk of requestChunks) {
```
zachrog, line 110: Nice

zachrog (approved):

web-api/workflow-terraform/migration/main/lambdas/migration-segments.ts
```diff
@@ -1,41 +1,50 @@
...
-const { createLogger } = require('../../../../src/createLogger');
-const { migrationsToRun } = require('./migrationsToRun');
+} from '@shared/business/utilities/DateHandler';
+import { createLogger } from '@web-api/createLogger';
+import { migrationsToRun } from './migrationsToRun';
+import { migrateItems as validationMigration } from './migrations/0000-validate-all-items';
+import promiseRetry from 'promise-retry';
+import type { Context, Handler, SQSEvent } from 'aws-lambda';
 
+// import type { EfcmsEntity } from './migration';
```
mmarcotte: nit: don't think you mean to have left this hanging around

mmarcotte (changes requested):
just one nit

web-api/send-migration-segment-messages.ts
```diff
@@ -0,0 +1,68 @@
...
+    new Array(totalSegments).fill(null).map((_v, i) => ({
+      segment: i,
+      totalSegments,
+    })),
+  );
+
+  const { failed } = await addToQueue({ QueueUrl, messages: segments });
+  if (failed.length) {
+    // TODO: bail?
+    // process.exit(1);
```
TomElliottFlexion: Should this exit here?

web-api/workflow-terraform/migration/main/lambdas/migration-segments.test.ts
```diff
@@ -0,0 +1,150 @@
...
+import { mockClient } from 'aws-sdk-client-mock';
+import type { Context } from 'aws-lambda';
+
+const ddbMock = mockClient(DynamoDBDocumentClient);
+const sqsMock = mockClient(SQSClient);
+
+beforeEach(() => {
+  ddbMock.reset();
+  sqsMock.reset();
+  // jest.resetAllMocks();
```
TomElliottFlexion, line 20: I believe these should be reset automatically in jest-unit.config.ts

TomElliottFlexion (approved):

web-api/send-migration-segment-messages.ts
jimlerza: Thanks for mentioning this.

We weren't previously detecting failures, but we as you can see, we are doing so now. It doesn't look like we can easily retry the failed segment messages, though; the `BatchResultErrorEntry` errors do not include the original message contents.

I think we should exit. If we don't exit here, we could potentially switch over to a destination table with incomplete data and then delete the source table in the cleanup step. That would be real bad.

mmarcotte (approved):

PR #4742, opened 2024-03-14
10068 story 

ujahio:
Story: [10068](https://github.com/flexion/ef-cms/issues/10068)


jasonfu9966 (approved):

TomElliottFlexion:
May I suggest either pulling test into this branch, or closing and re-opening the PR? That way we wouldn't see the extra changes

web-client/src/presenter/computeds/formattedDocketEntries.ts
```diff
@@ -343,12 +338,56 @@ export const formattedDocketEntries = (
...
       });
+    })
+    .map(docketEntry => {
+      const computeMinuteEntry = (rawDocketEntry: RawDocketEntry): boolean => {
+        const MINUTE_ENTRIES_EVENT_CODES = Object.keys(
+          MINUTE_ENTRIES_WITHOUT_DOCS_MAP,
+        ).map(key => MINUTE_ENTRIES_WITHOUT_DOCS_MAP[key].eventCode);
+
+        return MINUTE_ENTRIES_EVENT_CODES.includes(rawDocketEntry.eventCode);
+      };
```
jasonfu9966, line 349: We're adding this compute here, but we noticed that there is similar work going into the `DocketEntry` entity through a PR that is currently awaiting approvals to go to `test`

We would prefer it if we could use the new work going to the `DocketEntry` entity but we can't realistically get that work until it's merged into staging.

Open to thoughts or suggestions from reviewers on how to handle this situation.

web-client/src/presenter/sequences/batchDownloadDocketEntriesSequence.ts
```diff
@@ -0,0 +1,21 @@
+import { addPrintableDocketRecordCheckAction } from '@web-client/presenter/actions/addPrintableDocketRecordCheckAction';
+import { batchDownloadDocketEntriesAction } from '@web-client/presenter/actions/batchDownloadDocketEntriesAction';
+import { clearModalAction } from '../actions/clearModalAction';
+import { generateDocketRecordPdfUrlAction } from '@web-client/presenter/actions/generateDocketRecordPdfUrlAction';
+import { setShowModalAction } from '../actions/setShowModalAction';
+
+export const batchDownloadDocketEntriesSequence = [
+  clearModalAction,
+  addPrintableDocketRecordCheckAction,
+  {
```
TomElliottFlexion, line 10: Nit: I'm not sure we need an entire cerebral conditional to check this variable: we could combine `addPrintableDocketRecordCheckAction` and `generateDocketRecordPdfUrlAction`

web-client/src/views/DocketRecord/DocketRecord.tsx
```diff
@@ -9,27 +9,42 @@ import { UnsealDocketEntryModal } from './UnsealDocketEntryModal';
...
+    useEffect(() => {
+      if (!docketRecordHelper.showBatchDownloadControls) return;
+
+      const documentsSelectorHeaderInput = window.document.getElementById(
+        'all-selectable-docket-entries-checkbox',
+      ) as HTMLInputElement;
+      if (formattedDocketEntriesHelper.someDocumentsSelectedForDownload) {
+        documentsSelectorHeaderInput.indeterminate = true;
+      } else documentsSelectorHeaderInput.indeterminate = false;
+    }, [formattedDocketEntriesHelper.someDocumentsSelectedForDownload]);
```
TomElliottFlexion: Not super important: since we don't have a React testing framework, ideally most of the display logic would be done in a computed. I don't know if it's possible to bind documentsSelectorHeaderInput.indeterminate to state, but that would allow this logic to be tested

TomElliottFlexion (approved):

shared/src/proxies/documents/batchDownloadDocketEntriesProxy.ts
```diff
@@ -0,0 +1,13 @@
+import { DownloadDocketEntryRequestType } from '@shared/business/useCases/document/batchDownloadDocketEntriesInteractor';
+import { get } from '../requests';
+
+export const batchDownloadDocketEntriesInteractor = (
+  applicationContext,
+  params: DownloadDocketEntryRequestType,
+) => {
+  return get({
+    applicationContext,
+    endpoint: '/async/documents/batch-download',
```
nechama-krigsman, line 10: Should this endpoint specify it's downloading docket entries?

shared/src/business/useCases/document/batchDownloadDocketEntriesInteractor.ts
```diff
@@ -0,0 +1,204 @@
...
+  };
+
+  const onProgress = progressData => {
+    applicationContext.getNotificationGateway().sendNotificationToUser({
+      applicationContext,
+      clientConnectionId,
+      message: {
+        action: 'batch_download_progress',
+        ...progressData,
+        numberOfDocketRecordsToGenerate: 0,
```
nechama-krigsman, line 126: is this number always set to zero?

cypress/cypress-integration/integration/batch-download-documents.cy.ts
nechama-krigsman, line 1: There are a couple of todo comments and some code that is commented out in this file. were those addressed or can they be removed?

nechama-krigsman (approved):

shared/src/business/useCases/document/batchDownloadDocketEntriesInteractor.ts
ujahio: The batch download operation is written to download cases and documents associated to those cases. For this story, we are only downloading documents on a case, hence, 0 for case downloads.

shared/src/proxies/documents/batchDownloadDocketEntriesProxy.ts
ujahio: We are specifically downloading the documents "associated" with the docket entries. We aren't downloading the docket entries's data itself.

PR #4745, opened 2024-03-15
10216 missing previous document refactor2

mmarcotte:
flexion#10216


TomElliottFlexion (approved):

shared/src/business/entities/cases/Case.ts
```diff
@@ -60,10 +60,10 @@ import { JoiValidationConstants } from '../JoiValidationConstants';
 import { JoiValidationEntity } from '../JoiValidationEntity';
 import { Petitioner } from '../contacts/Petitioner';
 import { PrivatePractitioner } from '../PrivatePractitioner';
+import { type RawUser, User } from '../User';
```
jtdevos, line 63: Didn't know you could mix import / import type on the same line.  Neat!

shared/src/business/entities/DocketEntry.ts
```diff
@@ -554,24 +781,53 @@ export class DocketEntry extends JoiValidationEntity {
...
+    isFileAttached,
+  }: {
+    eventCode: string;
+    isFileAttached?: boolean;
+  }): boolean {
+    if (eventCode === 'RQT') {
+      return !isFileAttached;
+    }
+
     const MINUTE_ENTRIES_EVENT_CODES = Object.keys(MINUTE_ENTRIES_MAP).map(
```
jtdevos: I think this could also be 
```suggestion
    const MINUTE_ENTRIES_EVENT_CODES = Object.values(MINUTE_ENTRIES_MAP).map(v => v.eventCode);
```

jimlerza (approved):

PR #4753, opened 2024-03-18
10315: Forgot password ignore leading or trailing spaces on code

pixiwyn:


TomElliottFlexion (approved):

jimlerza (approved):

mmarcotte (approved):

mmarcotte:
Could a better user experience be that the user can not put any spaces into the code field? Seems like an onChange could simply remove any spaces (or non numeric characters) for that matter. Just my $0.02.

PR #4757, opened 2024-03-19
Fix Recurring Issue with Cognito Local File Formatting

pixiwyn:
When starting the API, Cognito local modifies the format of config.json and clients.json, while this isn't our correct format, this PR will hopefully address this so it doesn't keep showing up as a change.


rachelschneiderman (approved):
Thank you!

TomElliottFlexion (approved):

TomElliottFlexion:
Nice fix

PR #4759, opened 2024-03-19
10249 export custom case report

nechama-krigsman:


web-api/src/business/useCases/customCaseReport/createCsvCustomCaseReportFileInteractor.ts
```diff
@@ -0,0 +1,161 @@
...
+    caseTypes,
+    clientConnectionId,
+    endDate,
+    filingMethod,
+    highPriority,
+    judges,
+    preferredTrialCities,
+    procedureType,
+    startDate,
+    totalCount,
```
cruzjone-flexion, line 33: We can fetch count instead of passing it through the front end

web-client/src/presenter/state.ts
```diff
@@ -575,6 +575,7 @@ export const baseState = {
     zipInProgress?: boolean;
     totalFiles?: number;
     fileCount?: number;
+    title: string;
```
cruzjone-flexion: Make optional, we just introduced this property. So only this feature is setting it

cruzjone-flexion (approved):

shared/src/proxies/customCaseReport/createCsvCustomCaseReportFileProxy.ts
```diff
@@ -0,0 +1,13 @@
...
+import { post } from '../requests';
+
+export const createCsvCustomCaseReportFileInteractor = (
+  applicationContext,
+  data: CustomCaseReportCsvRequest,
+) => {
+  return post({
+    applicationContext,
+    body: data,
+    endpoint: '/async/reports/csv/custom-case-report',
```
mmarcotte: somewhat nitpicky, but could this be rearranged please?

`/async/reports/custom-case-report/csv`

I'd also consider adding format=csv into a request variable

web-api/src/app.ts
```diff
@@ -800,6 +801,10 @@ app.get(
     '/judge-activity-report/closed-cases',
     lambdaWrapper(getCasesClosedByJudgeLambda),
   );
+  app.post(
+    '/async/reports/csv/custom-case-report',
+    lambdaWrapper(createCsvCustomCaseReportFileLambda, { isAsync: true }),
+  );
```
mmarcotte: somewhat nitpicky, but could this be rearranged please?

`/async/reports/custom-case-report/csv`

I'd also consider adding format=csv into a request variable

mmarcotte (changes requested):

web-api/src/app.ts
cruzjone-flexion: I also think we should probably add the keyword "export" in the path .
Maybe 
`/async/export/reports/custom-case-report/csv`

Using a request variable can probably be done when we start supporting multiple formats, not now.

mmarcotte (approved):

PR #4770, opened 2024-03-20
10249 fix pre epoch cases intermediate branch to test 1710967520

mmarcotte:
This PR allows cases with a `receivedAt` of before the UNIX date time epoch to show up in the Custom Case Report as well as the downloadable custom case report. For the first page of the Custom Case Report, we will simply not include a `searchAfter` parameter to the Opensearch query instead of the default `receivedAt` of `0`. This way `receivedAt` with negative values will be included.

Screenshot 2024-03-20 at 4 41 49 PM


jimlerza (approved):

rachelschneiderman (changes requested):
Is there a cypress test that could be written that would have caught this? Maybe paper file a case so you can set the received at date to something pre-1970?

mmarcotte:
> Is there a cypress test that could be written that would have caught this? Maybe paper file a case so you can set the received at date to something pre-1970?

Sure thing. I added some logic to `cypress/cypress-integration/integration/custom-case-report-csv-export.cy.ts` test.

rachelschneiderman (approved):

PR #4775, opened 2024-03-21
10068 story

ujahio:
Story: [10068](https://github.com/flexion/ef-cms/issues/10068)

Enables internal users to batch-download docket entries (attached documents).


shared/src/business/entities/cases/Case.ts
```diff
@@ -808,7 +808,7 @@ export class Case extends JoiValidationEntity {
   }
 
   hasPrivatePractitioners() {
-    return this.privatePractitioners.length > 0;
+    return this.privatePractitioners!.length > 0;
```
mmarcotte: is this correct? 

It would seem that `privatePractitioners` is optional, and so we might have to check for its existence rather than assume it exists. Why not a `?` instead of a `!`?

shared/src/business/entities/cases/Case.ts
```diff
@@ -1910,7 +1916,7 @@ export class Case extends JoiValidationEntity {
    * @returns {Case} this case entity
    */
   fileCorrespondence(correspondenceEntity) {
-    this.correspondence = [...this.correspondence, correspondenceEntity];
+    this.correspondence = [...this.correspondence!, correspondenceEntity];
```
mmarcotte: Is this correct?

Perhaps we should remove the `?` on the property of the Case Class that suggests it's optional?

web-client/src/presenter/presenter.ts
```diff
@@ -1197,6 +1202,7 @@ export const presenterSequences = {
     setSelectedAddressOnFormSequence as unknown as Function,
   setSelectedBatchIndexSequence:
     setSelectedBatchIndexSequence as unknown as Function,
+  setSelectedDocumentsForDownloadSequence,
```
mmarcotte, line 1205: Why don't some of these have `as unknown as Function`?

ujahio: Good question. `setSelectedDocumentsForDownloadSequence` is typed already in its declaration. `web-client/src/presenter/sequences/setSelectedDocumentsForDownloadSequence.ts`

shared/src/business/utilities/getCaseDocumentsIdsFilteredByDocumentType.ts
```diff
@@ -0,0 +1,29 @@
...
+    docketEntries: RawDocketEntry[];
+    docketRecordFilter: string;
+    documentsToProcess: { docketEntryId: string }[];
+  },
+): string[] => {
+  const formattedDocketEntries = documentsToProcess.map(docSelected => {
+    return docketEntries.find(
+      docEntry => docEntry.docketEntryId === docSelected.docketEntryId,
+    );
+  });
```
mmarcotte: I'm having a hard time understanding the utility of this block and function. Can you help me understand it?

What's the difference between documentsToProcess and docketEntries?

jasonfu9966: documentsToProcess is the list of documents selected for download. docketEntries is the list of all the documents on the docket record. We'll change the names to be more clear.

mmarcotte (approved):

shared/src/business/utilities/getCaseDocumentsIdsFilteredByDocumentType.ts
```diff
@@ -0,0 +1,29 @@
+export const getCaseDocumentsIdsFilteredByDocumentType = (
+  applicationContext,
```
cruzjone-flexion: Type AppContext

shared/src/business/utilities/getDocketEntriesByFilter.test.ts
```diff
@@ -0,0 +1,66 @@
+import { DOCKET_RECORD_FILTER_OPTIONS } from '@shared/business/entities/EntityConstants';
```
cruzjone-flexion, line 1: Fix Typing errors

web-client/src/presenter/actions/batchDownloadDocketEntriesAction.ts
```diff
@@ -0,0 +1,38 @@
...
+  fileId: string;
+}>) => {
+  const docIdsSelectedForDownload = get(state.documentsSelectedForDownload);
+  const { docketEntries, docketNumber } = get(state.caseDetail);
+  const clientConnectionId = get(state.clientConnectionId);
+  const { docketRecordFilter } = get(state.sessionMetadata);
+
+  const filteredDocumentsIds = applicationContext
+    .getUtilities()
+    .getCaseDocumentsIdsFilteredByDocumentType(applicationContext, {
```
cruzjone-flexion, line 18: Fix Type error

web-client/src/presenter/computeds/docketRecordHelper.ts
```diff
@@ -18,7 +29,17 @@ export const docketRecordHelper = (get: Get): any => {
 
   const sortLabelTextMobile = sortLabelsMobile[sortOrder];
 
+  const documentsIdsForDownload = applicationContext
+    .getUtilities()
+    .getCaseDocumentsIdsFilteredByDocumentType(applicationContext, {
```
cruzjone-flexion, line 34: Fix Type error

cruzjone-flexion (approved):
Looks good, only suggestion I have is fix the Typing errors. Since yall have the latest knowledge about the code might be easier to fix now rather than later. But will still approve.

mmarcotte (approved):

PR #4779, opened 2024-03-25
10322: BUG: Icon spacing

TomElliottFlexion:
USWDS 3.8.0 introduced a new `$theme-button-icon-gap` which adds default spacing for buttons with icons
https://github.com/uswds/uswds/pull/5398
Screenshot 2024-03-25 at 11 31 17 AM

### DIFF
--- a/web-client/src/ustc-ui/Button/Button.tsx
+++ b/web-client/src/ustc-ui/Button/Button.tsx
@@ -31,7 +31,6 @@ export const Button = props => {
   const classes = classNames(
     className,
     'usa-button',
-    tooltip && 'usa-tooltip',
     !overrideMargin && `margin-${marginDirection}-205`,
     overrideMargin,
     icon && !shouldWrapText && 'no-wrap',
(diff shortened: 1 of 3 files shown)


web-client/src/uswds/_index.scss
```diff
@@ -2,7 +2,8 @@
   $theme-show-notifications: false,
   $theme-font-path: '../../node_modules/@uswds/uswds/dist/fonts',
   $theme-image-path: '../../node_modules/@uswds/uswds/dist/img',
-  $theme-font-weight-semibold: 600
+  $theme-font-weight-semibold: 600,
+  $theme-button-icon-gap: 0,
```
TomElliottFlexion, line 6: Setting the default icon gap to the previous value, 0

web-client/src/ustc-ui/Button/Button.tsx
```diff
@@ -31,7 +31,6 @@ export const Button = props => {
   const classes = classNames(
     className,
     'usa-button',
-    tooltip && 'usa-tooltip',
```
TomElliottFlexion, line 34: `usa-tooltip` class is incompatible with `usa-button` now, since the former uses `inline-block` display styling and the latter uses `inline-flex`

cruzjone-flexion (commented):
Thanks for updating DepUpdate Doc.

cruzjone-flexion (approved):

jasonfu9966 (approved):

TomElliottFlexion:
On hold until USWDS is upgraded to 3.8.0

PR #4781, opened 2024-03-25
10260: Update Terraform template to pass ID to Lambda

cruzjone-flexion:


web-api/terraform/api/api-async.tf
```diff
@@ -202,6 +202,7 @@ resource "aws_api_gateway_integration" "api_async_integration_get" {
   "httpMethod" : "$context.httpMethod",
   "headers" : {
     "Authorization": "$input.params('Authorization')",
+    "asyncsyncid": "$input.params('Asyncsyncid')",
```
TomElliottFlexion, line 205: Out of curiosity only: why is this id in the headers? (As opposed to the body of the request for example)

TomElliottFlexion (approved):

zachrog (approved):

web-api/terraform/api/api-async.tf
cruzjone-flexion: Its for generic purposes, not just specific endpoints.

PR #4792, opened 2024-03-28
10162 lift stay proceedings intermediate branch to test 1711636813

jtdevos:


cypress/cypress-integration/integration/petitioner-files-motion-to-stay-proceedings.cy.ts
```diff
@@ -0,0 +1,111 @@
...
+
+      loginAsDocketClerk1();
+      searchByDocketNumberInHeader(docketNumber);
+
+      cy.get('[data-testid="case-detail-menu-button"]').click();
+      cy.get('[data-testid="menu-button-add-paper-filing"]').click();
+      cy.get('[data-testid="primary-document-type-search"]').type(
+        'MLSP{enter}',
+      );
+      // cy.get('#react-select-2-option-0').click();
```
TomElliottFlexion: Nit: still necessary?

TomElliottFlexion (approved):
Looks good! I'm not sure how up-to-date we keep the .csv files in `shared/src/tools/`, but this could be added there as well

cypress/cypress-integration/integration/petitioner-files-motion-to-stay-proceedings.cy.ts
jtdevos: It is not - good catch!

PR #4813, opened 2024-04-02
DD 10283: Change Country Dropdown to Radio Buttons - to test 

jasonfu9966:
only minor refactoring happened in this PR


TomElliottFlexion (approved):

web-client/src/views/StartCase/Country.tsx
```diff
@@ -56,7 +56,7 @@ export const Country = connect(
               className="usa-radio__label"
               data-testid="international-country-btn"
```
nechama-krigsman: Should the data-testid change here as well?

web-client/src/views/StartCase/Country.tsx
```diff
@@ -84,7 +84,7 @@ export const Country = connect(
             />
             <label
               className="usa-radio__label"
-              data-testid="international-country-btn"
+              data-testid="domestic-country-btn"
```
nechama-krigsman: Is this testId correct?

web-client/src/views/StartCase/Country.tsx
nechama-krigsman: Never mind...i was looking at old commit.

jasonfu9966: not a bad catch though!

web-client/src/views/StartCase/Country.tsx
nechama-krigsman: same here!

nechama-krigsman (approved):

PR #4822, opened 2024-04-09
10260: Fix 504 when uploading large PDF files;

cruzjone-flexion:
THE PROBLEM:
When a user uploads a large file there is a potential that the HTTP endpoint we are using is synchronous (which has a 29 second timeout). This means that if the processing of that PDF takes longer than 29 seconds it timeouts throwing a 504 error.

THE SOLUTION:
We wanted to leverage our ASYNC lambda that has a timeout of 15 minutes instead, but we did not want to convert all the endpoints to async and restructure the front end code to go through the Socket router. **ALSO, IRS users cannot use web sockets**. So we decided to use ASYNC endpoint and updated lambda wrapper to check the asyncsyncid flag in the HTTP 'headers' and if that id is available we store the results in the DB. Which then we start polling and checking if the results are available in the DB and once it is we continue our process synchronously  

🟢  Add Cypress smoke test test around "Complete Document QC" Feature


shared/src/notifications/saveRequestResponse.ts
```diff
@@ -0,0 +1,25 @@
+import { FORMATS, formatNow } from '@shared/business/utilities/DateHandler';
+import { put } from '@web-api/persistence/dynamodbClientService';
+
+export const saveRequestResponse = async ({
```
zachrog: If this is meant to only e called from the web-api would you mind moving it out of shared/ and into web-api/?

shared/src/proxies/requests.ts
```diff
@@ -122,6 +125,43 @@ export const post = async ({
...
+  request,
+  asyncSyncId = applicationContext.getUniqueId(),
+) => {
+  getMemoized.clear();
+
+  return new Promise((resolve, reject) => {
+    const callback = results => {
+      applicationContext
+        .getAsynSyncUtil()
+        .removeAsyncSyncCompleter(asyncSyncId);
```
zachrog: Love the move to polling as a way of getting around websockets. I talked with Jon a little and I think it would be possible to achieve polling without using the asyncSyncCompleter Dict. I am suggesting not using the dictionary to reduce some of the steps in making polling happen and so we do not have an object in applicationContext that is holding many callback functions and growing in size accidentally.

nechama-krigsman (approved):

zachrog (approved):

mmarcotte (approved):
Looks great, I'd like a walkthrough if possible of the asyncsync polling you built. Thanks for the hard work!

PR #4827, opened 2024-04-10
10208: Ignore Username Case Sensitivity On Login

rachelschneiderman:
flexion#10208

- Lowercase all email inputs when interacting with Cognito (getUserGateway).
- Requires a cognito migration where we lowercase all emails in cognito user pool.
- Updated how we seed users in lower environments to be a script instead of an API endpoint.
- Removed unnecessary shell scripts which only called through to typescript files.
- Removed ClamAV/virus scan infrastructure and code.

## Deployment Steps:
- Use the environment switcher to point to the deploying environment.
- Run an account specific deploy:
`npm run deploy:account-specific`
- Run the migration script (Runs at a rate or 10 users/sec):
NOTE: If you would like to run as a "dry-run" first, change `dryRun = true` at the top of the file. When you are ready to run and update users in the environment, set `dryRun = false` at the top of the file.
`npx ts-node --transpile-only scripts/run-once-scripts/cognitoUserIdAndEmailMigration.ts`
- Kick off Circle deploy


web-api/src/app.ts
```diff
@@ -1008,7 +1001,6 @@ app.get(
   );
   app.get('/users-by-role', lambdaWrapper(getAllUsersByRoleLambda));
   app.get('/users', lambdaWrapper(getUserLambda));
-  app.post('/users', lambdaWrapper(createUserLambda));
```
mmarcotte, line 1011: We use this API to create Court users today. Why did case-sensitivity necessitate the removal of this endpoint?

zachrog: Good question. 
Short answer: It was not required, but felt like a good idea. Creating users was changed from hitting an endpoint to running a script.

Longer answer: We saw that our pipeline was flakey when creating judges and creating court users. It would just randomly fail, and sometimes it would fail and then log the password in the error. When re-run with no changes it would pass. When we investigated what was happening we thought it was strange that we had a public endpoint that basically no one was allowed to hit but the dev team. If that was the case why have it be an endpoint? Why not make it a script?

So we decided to delete that endpoint, interactor, and code because it felt strange to have a public route for creating a user that was not meant to be hit by the public, we had multiple paths for creating a user which needed to retain context of lowercasing, remove methods off of appContext that were not meant to be used by the rest of the application, reduced pipeline complexity as we do not need to activate a user to to then have that user create more users we can just run a script, and (subjectively) make the code simpler for what was happening when creating judgeUsers and testUsers. 

Totally open to input though.

shared/admin-tools/util.ts
```diff
@@ -4,10 +4,12 @@ import { DynamoDBDocument } from '@aws-sdk/lib-dynamodb';
 import readline from 'readline';
 
 const { ENV } = process.env;
-const UserPoolCache = {};
+const UserPoolCache: Record<string, string> = {};
 
-// Look up the current version so that we can perform searches against it
-export const getVersion = async (): Promise<string> => {
```
jimlerza, line 10: This method is also used in the `become-user` script (as well as a pair of one-off scripts), which I don't see updated in this PR.

Rather than breaking scripts that use it, perhaps we could still export a `getVersion` method that calls this method and returns the `version` string?

scripts/circleci/judge/bulkImportJudgeUsers.helpers.ts
```diff
@@ -1,11 +1,15 @@
...
 import {
-  getServices,
-  getToken,
-  readCsvFile,
-} from '../../../web-api/importHelpers';
+  getDestinationTableInfo,
+  getUserPoolId,
+} from '../../../shared/admin-tools/util';
 import { parse } from 'csv-parse';
-import axios from 'axios';
```
jimlerza, line 8: I (for one) LOVE this change! ;)

zachrog (approved):

shared/admin-tools/util.ts
```diff
@@ -88,57 +112,6 @@ export const getClientId = async (
   return undefined;
 };
 
-// Generate a strong password that makes Cognito happy
-export const generatePassword = (numChars: number): string => {
```
jimlerza, line 92: We implemented this because it gave us control over which special characters to include. Before implementing this, we kept getting passwords with characters that our various bash scripts had trouble with.

jimlerza (approved):

PR #4845, opened 2024-04-18
10281 story

pixiwyn:
[Zenhub Card](https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/gh/flexion/ef-cms/10281)

Improves the Practitioner Search results by adding a few extra columns and replacing the "load more" button with pagination.


jasonfu9966 (approved):

jasonfu9966 (approved):

jimlerza (approved):

web-api/src/business/useCases/practitioner/getPractitionersByNameInteractor.ts
```diff
@@ -0,0 +1,61 @@
...
+      name: string;
+      practiceType: string;
+      practitionerType: string;
+    }[];
+    total: number;
+  };
+};
+
+export const getPractitionersByNameInteractor = async (
+  applicationContext: IApplicationContext,
```
rachelschneiderman: Thank you for moving this interactor to web-api!

Nit: This can be ServerApplicationContext

web-api/src/persistence/elasticsearch/getPractitionersByName.ts
```diff
@@ -1,19 +1,27 @@
 import { IS_PRACTITIONER } from './helpers/searchClauses';
-import { MAX_SEARCH_CLIENT_RESULTS } from '../../../../shared/src/business/entities/EntityConstants';
+import { PRACTITIONER_SEARCH_PAGE_SIZE } from '@shared/business/entities/EntityConstants';
+import { PractitionerSearchResultType } from '@web-client/presenter/computeds/AdvancedSearch/practitionerSearchHelper';
```
rachelschneiderman: Hmmm are we sure we want presenter types imported here in the backend?

I am totally open to the idea that I'm wrong, just don't want to tightly couple our persistence to our presenter.

web-client/integration-tests/journey/admissionsClerkSearchesForPractitionersByName.ts
rachelschneiderman, line 1: YAY 👏

web-client/src/views/AdvancedSearch/PractitionerSearchResults.tsx
```diff
@@ -1,73 +1,129 @@
...
+                    }}
+                  />
+                )}
+            </div>
+            <div className="text-right margin-bottom-2">
+              <span className="text-bold" id="custom-case-result-count">
+                Count: &nbsp;
+              </span>
+              <span data-testid="practitioner-search-result-count">
+                {formatPositiveNumber(
```
rachelschneiderman: I think any formatting should be done in the helper so we can test for it and to separate concerns a bit.

web-client/src/views/CaseDetail/AddIrsPractitionerModal.tsx
```diff
@@ -34,6 +34,7 @@ export const AddIrsPractitionerModal = connect(
         className="counsel-modal"
         confirmLabel="Add to Case"
         confirmSequence={confirmSequence}
+        dataTestId="irs-practitioner-modal"
```
rachelschneiderman, line 37: I think this should be `data-test-id`

web-client/src/views/CaseDetail/AddPrivatePractitionerModal.tsx
```diff
@@ -36,6 +36,7 @@ export const AddPrivatePractitionerModal = connect(
         className="counsel-modal"
         confirmLabel="Add to Case"
         confirmSequence={confirmSequence}
+        dataTestId="private-practitioner-modal"
```
rachelschneiderman, line 39: I think this should be `data-test-id`

web-client/src/views/CaseDetail/AddIrsPractitionerModal.tsx
pixiwyn: it's a prop on Modal which is why it's like that.

web-client/src/views/CaseDetail/AddPrivatePractitionerModal.tsx
pixiwyn: it's a prop on Modal which is why it's like that.

web-client/src/views/AdvancedSearch/PractitionerSearchResults.tsx
pixiwyn: Good call, will adjust.

web-client/src/views/CaseDetail/AddIrsPractitionerModal.tsx
rachelschneiderman: Ahhh, thanks. I missed that!

web-api/src/persistence/elasticsearch/getPractitionersByName.ts
jimlerza: Is there already a place for shared types? Nothing jumping out at me in the `shared` directory.

Where's a better place for this type to live?

rachelschneiderman: I think coming up with a shared place is a good idea, I don't want to dismiss that at all 😄  

I think my original point was more that the type the presenter is using to render the view is, _in most cases_, different than what persistence is returning anyways. For example, the type here isn't accurate. `getPractitionersByName` is typed as returning a list of `PractitionerSearchResultType` - which includes the property `formattedAdmissionsDate` which is computed in the helper. To me, it seems like this type should be what is included in `_source`, since that is what is being pulled for each matching record in OpenSearch

jasonfu9966: I created a new type in `getPractitionersByName` and changed the name of the old `PractitionerSearchResultType`. let me know what you think!

jimlerza (approved):

PR #4855, opened 2024-04-23
Devex 10334 - AWS v2 -> v3 conversion: APIGateway

jimlerza:
See flexion#10334


web-api/smoke-tests.ts
```diff
@@ -1,11 +1,11 @@
 import { CognitoIdentityProvider } from '@aws-sdk/client-cognito-identity-provider';
-import AWS from 'aws-sdk';
+import { getServices } from './importHelpers';
 import assert from 'assert';
 import axios from 'axios';
 
 const ENV = process.argv[2];
 const REGION = process.argv[3];
-const DEPLOYING_COLOR = process.argv[4];
+// const DEPLOYING_COLOR = process.argv[4];
```
mmarcotte: Looks like this was borrowed from a bash script once upon a time. And this variable was used on line 13. Maybe line 13 can be fixed to use backticks to be a template string. Then again, it doesn't appear that `DEPLOYING_COLOR` is actually required.

mmarcotte (changes requested):

web-api/smoke-tests.ts
jimlerza: Good catch! I think we actually want to keep this: it allows us to run these tests against any color by passing a color in.

The most recent commit restores this functionality.

mmarcotte (approved):

TomElliottFlexion (approved):

PR #4859, opened 2024-04-23
Return stringified version of the json response;

cruzjone-flexion:
🟢  [CircleCi Run](https://app.circleci.com/pipelines/github/flexion/ef-cms/52577/workflows/608ef40a-8772-4354-9167-9505f31e0643)

### DIFF
--- a/web-api/src/lambdaWrapper.test.ts
+++ b/web-api/src/lambdaWrapper.test.ts
@@ -30,7 +30,9 @@ describe('lambdaWrapper', () => {
         send: jest.fn(),
       }),
     };
-    JSON.parse = jest.fn();
+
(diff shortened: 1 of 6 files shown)


web-api/src/lambdaWrapper.test.ts
```diff
@@ -30,7 +30,9 @@ describe('lambdaWrapper', () => {
         send: jest.fn(),
       }),
     };
-    JSON.parse = jest.fn();
+
+    const orignalParse = JSON.parse;
```
TomElliottFlexion: If we care about resetting JSON.parse, we need `JSON.parse = originalParse` in an afterAll

TomElliottFlexion (approved):

PR #4863, opened 2024-04-24
10358/10359/10360 - BUGFIXES

cruzjone-flexion:
---**10358 Docket Clerks are getting a spinning wheel when Completing QC**---
**BUG** - Docket Clerks are getting a spinning wheel when Completing QC. When a docket clerk goes to their QC inbox, selects a document to QC and either clicks "Complete" or "Complete and Send Message", they get stuck on the spinning wheel. If they refresh their screen, they will get the "Document already handled" message.

**REASON** - With the introduction of the new AsyncSyncHandler architecture, we store responses from the interactor in DynamoDB. The front end then begins polling for these responses, with the spinner persisting until the polling mechanism retrieves the response from the database and relays it to the front end. However, we overlooked the maximum size limit of a database item (400KB). Certain interactors return the Case entity, which can become substantially large. Consequently, attempts to save a large Case entity result in an error due to exceeding the database's storage limit, preventing the item from being saved. As a result, the front end continues to display the spinner, waiting indefinitely for results that will never be available.

**SOLUTION** - Instead of saving the response into one Item/Row in the DB we now chunk up the response and save each chunk as a single Item/Row with an index and # of total records we expect. Our polling mechanism now checks if all records are available and once it is it reconstructs the response and sends it to the front end.

---**10359/10360 Actions being reverted after a while (serving orders, signing orders)**---
**BUG** - The actions taken by the user were being reversed after a period of time. For instance, in issue 10360, an order was created and signed. Subsequently, the user was redirected to the "Drafts" tab of the case and can see the Order is there and signed. However, after waiting a few minutes and refreshing the page, the newly created and signed Order was displayed with a status indicating that the order needed to be signed.

**REASON** - In the Lambda wrapper we had added a ternary condition to not stringify the body of our HTTP response. This was decided because there was some code in the Lambda Wrapper that suggested that the HTTP response is being responded too immediately and gave us the impression that it would never respond to the request at the end of the process. This is not the case in a deployed environment. So it was trying to send a JSON as a response of the HTTP request. This threw an error and would retry the process even though it completed the AsyncSyncHandler Promise in the front end. So the User would take more actions on the newly created order while in AWS the action of creating the order would rerun 3 time overriding any modified properties with the default values (in 10360's case saying it needs to be signed).

**SOLUTION** - Fix the type of the response, keep it as string and only parse it out in our logic for the AsyncSync Handler.

---**CircleCi Runs**---
🔴 [Cypress test against code without a fix](https://app.circleci.com/pipelines/github/flexion/ef-cms/52596/workflows/89a5896f-c3a1-4b65-91d8-5c07aa71ccc0)
image

🟢  [Cypress test against code with a fix](https://app.circleci.com/pipelines/github/flexion/ef-cms/52598/workflows/7535e72f-058d-4ed8-a237-49e6ea3b0e7f)
image


TomElliottFlexion (approved):

web-api/src/lambdaWrapper.ts
```diff
@@ -97,3 +114,16 @@ export const lambdaWrapper = (
     }
   };
 };
+
+function chunkString(str) {
+  const CHUNK_SIZE = 2000; //TODO: change it to more realistic size
```
mmarcotte, line 119: Are there follow up stories for addressing this?

mmarcotte (approved):

PR #4868, opened 2024-04-25
Devex 10297 - AWS v2 -> v3 conversion: Stragglers

jimlerza:
Rounding up some stragglers that didn't warrant their own devex cards:

* `SNSClient` in `switch-bounce-handler-colors`
* `CloudFrontClient` and `Route53Client` in `switch-ui-colors` and `switch-public-ui-colors`

See flexion#10297


web-client/switch-api-colors.helpers.ts
```diff
@@ -0,0 +1,63 @@
...
+              domainName,
+              restApiId: apiGatewayRecord.id,
+              stage: environmentName,
+            }),
+          );
+          if (createBasePathMappingResult) {
+            console.log(createBasePathMappingResult);
+          }
+        }
+      } catch (err: any) {
```
zachrog: As a heads up this catch statement will log the error but swallow the error and have the script continue running. This means that the node script might error but not exit with a non zero status code. This is a bug we ran into in develop. This has been fixed in staging

jtdevos (approved):
Zach had a good point re: swallowing an error.  Otherwise looks good to me 👍

web-client/switch-api-colors.helpers.ts
jimlerza: [Resolved](https://github.com/ustaxcourt/ef-cms/pull/4868/commits/8b3d51c353573dee8de35dd290ab7a06c1eaa2aa)

mmarcotte (approved):

PR #4874, opened 2024-04-29
10358: add unit test to ensure multiple chunks over chunk size

En-8:


pixiwyn (approved):

web-api/src/lambdaWrapper.ts
```diff
@@ -11,6 +11,8 @@ export const headerOverride = {
   'X-Content-Type-Options': 'nosniff',
 };
 
+export const CHUNK_SIZE = 2000;
```
mmarcotte: Just curious why you've settled on 2k. The database can hold 400K, right? Would something higher result in less reads and writes? Or even 4K as that's one [read capacity unit](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.ReadWriteCapacityMode.html#HowItWorks.requests)?

pixiwyn: Not sure, it looks like 2k was what was pushed out. We just exported the value so we could reuse it. We could consider increasing this if you like.

pixiwyn: If we do increase it, I'd recommend push this to test and re-testing everything.

jimlerza: Bumping it up to 4K sounds right to me - if I understand read capacity units (admittedly a big "if"), that should halve the cost of these retrievals.

En-8: I think 4k makes a lot of sense. We'll update it. 
I tried exploring string encoding and size per chunk etc and didn't find anything crazy helpful -- I did confirm at least with some basic testing that these are a single byte per character.

jimlerza (approved):

PR #4883, opened 2024-05-03
10284: Make all state select (dropdowns) use full name for us states and other states.

pixiwyn:
Petitioner: case info - edit contact info
Petitioner: create petition Petitioner info et al
Petitioner: Create petition Petitioner and spouse info
Docket Clerk: Add Petitioner to case
Practitioner: Edit contact info
Practitioner: create petition Petitioner info et al
-Admissions clerk: Add practitioner to Dawson


zachrog (approved):

jtdevos (approved):

jimlerza:
This is probably a dumb question, but the title of this card mentions "all state select (dropdowns)" - does that mean the only state dropdown we have is in the StartCase view?

jimlerza (approved):

pixiwyn:
@jimlerza posted all the locations that use this in the description. Seems like we use that same state dropdown everywhere. Thanks for you merging in staging just logged in and was checking my prs!!

PR #4889, opened 2024-05-06
Devex: Change user index name in the setup-glued-judges script

jimlerza:


jimlerza:
After looking into this further, I think we actually don't want this change.

This script runs _before_ the reindex occurs on a non-migration reindex deployment, so it needs to be querying the "old" index. With this code in place, it will be trying to query the "new" index, which won't exist yet in a non-migration reindex deployment.

PR #4896, opened 2024-05-08
Reduce throttling when creating test users. Create serially instead of parallel.

zachrog:


TomElliottFlexion (approved):

scripts/user/setup-test-users.ts
```diff
@@ -93,74 +93,68 @@ const setupPetitioners = async (
...
+    { barNumber: 'PT7777', role: 'privatePractitioner' },
+    { barNumber: 'PT8888', role: 'privatePractitioner' },
+  ];
+
+  for (let i = 0; i < allPractitioners.length; i++) {
+    const { barNumber, role } = allPractitioners[i];
 
-    await Promise.all(promises);
+    const practiceType = role === 'privatePractitioner' ? 'Private' : 'IRS';
+    const email = `${role}${i + 1}@example.com`;
```
jimlerza: We are now doing all of the practitioners in the same loop.

This means we're creating irsPractitioner1 through irsPractitioner10 and privatePractitioner11 through privatePractitioner20.

Are we gonna clean up the old privatePractitioner1 through privatePractitioner10 records?

Are we gonna remove all instances of privatePractitioner < 11 in various tests?

jimlerza (changes requested):

scripts/user/setup-test-users.ts
zachrog: Great catch! Fixed it!

jimlerza (approved):

rachelschneiderman (approved):

PR #4924, opened 2024-05-16
5512: Petitioner: DAWSON Petition Generation

TomElliottFlexion:
## 5512

flexion#5512

### Manual steps

- [ ] Setup the feature flag:
   ```bash
   scripts/dynamo/setup-updated-petition-flow-flag.sh
   ```


cypress/helpers/cypressTasks/dynamo/dynamo-helpers.ts
```diff
@@ -51,6 +51,49 @@ export const expireUserConfirmationCode = async (
   return null;
 };
 
+export const toggleFeatureFlag = async ({
+  flag,
+  flagValue,
+}: {
+  flag: string;
+  flagValue: any;
```
zachrog, line 59: Aren't our feature flag values booleans?

cypress/local-only/tests/integration/dashboard/private-practitioner-dashboard.cy.ts
```diff
@@ -2,6 +2,22 @@ import { loginAsPrivatePractitioner } from '../../../../helpers/authentication/l
 import { practitionerCreatesElectronicCase } from '../../../../helpers/fileAPetition/practitioner-creates-electronic-case';
 
 describe('Private practitioner views dashboard', () => {
+  before(() => {
```
zachrog: Wouldn't we want our cypress tests to be using the new petitioner flow? Why are we using the old flow for these tests?

cypress/local-only/tests/integration/fileAPetitionUpdated/file-a-petition-step-1-business--partnership-tax-matters-partner.cy.ts
```diff
@@ -0,0 +1,371 @@
+import { InputFillType, selectInput, textInput } from './petition-helper';
+import { loginAsPetitioner } from '../../../../helpers/authentication/login-as-helpers';
+
```
zachrog, line 3: LOVE all the cypress tests you all have written. AWESOME.

shared/src/business/entities/contacts/ContactFactoryUpdated.ts
```diff
@@ -0,0 +1,113 @@
+import { BusinessContact } from './BusinessContact';
+import { CONTACT_TYPES, PARTY_TYPES } from '../EntityConstants';
+import { ContactUpdated } from '@shared/business/entities/contacts/ContactUpdated';
+import { DeceasedSpouseContact } from '@shared/business/entities/contacts/DeceasedSpouseContact';
+import { OtherContact } from '@shared/business/entities/contacts/OtherContact';
+import { SpouseContact } from '@shared/business/entities/contacts/SpouseContact';
+
+export function ContactFactoryUpdated({
```
zachrog, line 8: ContactFactory_FinalV2.docx

cypress/local-only/tests/integration/dashboard/private-practitioner-dashboard.cy.ts
TomElliottFlexion: Indeed! Got most of them, working on the rest

cypress/helpers/cypressTasks/dynamo/dynamo-helpers.ts
jimlerza: > Aren't our feature flag values booleans?

I believe the Clerk of Court and Chief Judge values are formatted like feature flags (that is, their value is stored in the `current` field).

cypress/local-only/tests/integration/dashboard/private-practitioner-dashboard.cy.ts
TomElliottFlexion: Donezo

Mwindo:
I'm looking over this PR currently--thank you for all of your great work on this!

I will probably have a few notes, but one thing I noticed quickly while testing locally is the volatility of form state. A couple of examples:

- In Step 1, if I choose to file for "Myself" and enter in all of my information and then think, "Oh, duh, I was supposed to file for my spouse too!" I will have to re-enter that information because it is cleared upon selecting "Myself and my spouse." (And vice-versa.) I'm not sure it is worth fixing in this PR, but I think it's something worth addressing as a follow-up.
- If a user refreshes or (navigates to another page and goes back in their browser history), their progress is cleared and they start back at step 1. I suspect this would be an uncommon occurrence, but users who encounter the behavior are likely to be frustrated. I wonder if we can trigger an alert on refresh/redirection? (E.g., "You will lose your progress if you refresh or navigate away from this page. Do you want to continue?") This would have to be separate from the modal. I think this probably _is_ worth addressing in the current PR.

TomElliottFlexion:
> I'm looking over this PR currently--thank you for all of your great work on this!
> 
> I will probably have a few notes, but one thing I noticed quickly while testing locally is the volatility of form state. A couple of examples:
> 
> * In Step 1, if I choose to file for "Myself" and enter in all of my information and then think, "Oh, duh, I was supposed to file for my spouse too!" I will have to re-enter that information because it is cleared upon selecting "Myself and my spouse." (And vice-versa.) I'm not sure it is worth fixing in this PR, but I think it's something worth addressing as a follow-up.
> * If a user refreshes or (navigates to another page and goes back in their browser history), their progress is cleared and they start back at step 1. I suspect this would be an uncommon occurrence, but users who encounter the behavior are likely to be frustrated. I wonder if we can trigger an alert on refresh/redirection? (E.g., "You will lose your progress if you refresh or navigate away from this page. Do you want to continue?") This would have to be separate from the modal. I think this probably _is_ worth addressing in the current PR.

Thank you! I don't disagree with any of these points, however I think you'll have to take this up with Chris/Tenille/UX, since they're functionality changes

shared/src/business/entities/cases/PaperPetition.ts
```diff
@@ -145,6 +145,9 @@ export class PaperPetition extends JoiValidationEntity {
     });
     this.petitioners = [contacts.primary];
     if (contacts.secondary) {
+      if (!contacts.secondary.phone) {
+        contacts.secondary.phone = 'N/A';
```
Mwindo, line 149: Maybe worth pulling N/A out as a constant since it is used elsewhere.

web-client/src/views/StepIndicator.tsx
```diff
@@ -0,0 +1,121 @@
...
+    stepIndicatorInfo: state.stepIndicatorInfo,
+    updateStepIndicatorSequence: sequences.updateStepIndicatorSequence,
+  },
+  function StepIndicator({ stepIndicatorInfo, updateStepIndicatorSequence }) {
+    const { currentStep, steps } = stepIndicatorInfo;
+
+    if (PREVIOUS_STEP !== currentStep) {
+      PREVIOUS_STEP = currentStep;
+      window.scrollTo(0, 0);
+    }
```
Mwindo, line 20: I think we can get rid of this global variable and just use:

```
    useEffect(() => {
      window.scrollTo(0, 0);
    }, [currentStep]);
```

instead of 

```
    if (PREVIOUS_STEP !== currentStep) {
      PREVIOUS_STEP = currentStep;
      window.scrollTo(0, 0);
    }
```

web-client/src/views/StartCaseUpdated/UpdatedFilePetitionStep1.tsx
```diff
@@ -0,0 +1,721 @@
...
+import { StateDrivenFileInput } from '@web-client/views/FileDocument/StateDrivenFileInput';
+import { UpdatedFilePetitionButtons } from '@web-client/views/StartCaseUpdated/UpdatedFilePetitionButtons';
+import { WarningNotificationComponent } from '@web-client/views/WarningNotification';
+import { connect } from '@web-client/presenter/shared.cerebral';
+import { sequences, state } from '@web-client/presenter/app.cerebral';
+import { useValidationFocus } from '@web-client/views/UseValidationFocus';
+import React from 'react';
+import classNames from 'classnames';
+
+/* eslint-disable max-lines */
```
Mwindo, line 21: Any reason not to split out components into separate files in a FilePetitionStep1 folder?

web-client/src/views/StartCase/CountryUpdated.tsx
```diff
@@ -0,0 +1,103 @@
...
+  function Country({
+    addressInfo,
+    handleBlur,
+    handleChange,
+    registerRef,
+    type,
+    validationErrors,
+  }) {
+    return (
+      <React.Fragment>
```
Mwindo, line 35: Any reason not to nix the `React` import use <>, which we seem to use more elsewhere?

web-client/src/presenter/sequences/petitionGenerationLiveValidationSequence.ts
```diff
@@ -0,0 +1,48 @@
...
+  error: [setSingleValidationErrorAction],
+  success: [],
+};
+
+export const petitionGenerationLiveValidationSequence = [
+  startShowValidationAction,
+  determineStepPathAction,
+  {
+    step1: [
+      getStep1DataAction,
```
Mwindo, line 28: Is there a reason this is not something like `getPetitionStep1DataAction`?

shared/src/business/entities/contacts/ContactUpdated.ts
```diff
@@ -0,0 +1,133 @@
...
+    name: JoiValidationConstants.STRING.max(100)
+      .required()
+      .messages({ '*': 'Enter name' }),
+    phone: JoiValidationConstants.STRING.max(100)
+      .when('contactType', {
+        is: CONTACT_TYPES.secondary,
+        otherwise: joi.required(),
+        then: joi.optional(),
+      })
+      .messages({ '*': 'Enter phone number' }),
```
Mwindo, line 84: This validation doesn't seem specific to this PR (we do similar things elsewhere), but any reason not to do validation that the entry is a valid phone number (of some kind) or N/A?

Mwindo (commented):
So far this looks good to me! I have included a few minor comments/questions.

shared/src/business/entities/cases/ElectronicPetition.ts
```diff
@@ -74,6 +82,9 @@ export class ElectronicPetition extends JoiValidationEntity {
     this.petitioners = [contacts.primary];
 
     if (contacts.secondary) {
```
akuny, line 84: Could avoid the nested if statement with something like:

```ts
if (contacts.secondary) {
  contacts.secondary.phone = contacts.secondary.phone || 'N/A';
  this.petitioners.push(contacts.secondary);
}
```

cypress/local-only/tests/integration/fileAPetitionUpdated/file-a-petition-step-1-business--partnership-tax-matters-partner.cy.ts
Mwindo: I agree! Although I wonder if we should add accessibility tests for the new screens in a follow-up PR.

web-client/src/presenter/sequences/updatePetitionFormValueSequence.ts
```diff
@@ -0,0 +1,3 @@
+import { setPetitionFormValueAction } from '@web-client/presenter/actions/setPetitionFormValueAction';
+//TODO: type here instead of presenter.ts
```
akuny, line 2: To this point, some sequences are typed in the sequence's file itself, while others are typed in presenter.ts. May be good to consistently type these in the sequence files.

web-client/src/views/StartCase/PlaceOfLegalResidenceSelect.tsx
```diff
@@ -0,0 +1,57 @@
...
+  data,
+  handleBlur,
+  handleChange,
+  refProp,
+  type,
+}) => {
+  return (
+    <select
+      className={className ? classNames(className, 'usa-select') : 'usa-select'}
+      data-testid={`${type}.placeOfLegalResidence`}
```
akuny, line 19: Might be better to use a separating character other than a period since they can screw up DOM queries.

web-client/src/views/StartCaseUpdated/PetitionFormResponse.tsx
```diff
@@ -0,0 +1,141 @@
...
+import { FormGroup } from '@web-client/ustc-ui/FormGroup/FormGroup';
+import { Mobile, NonMobile } from '@web-client/ustc-ui/Responsive/Responsive';
+import { props as cerebralProps } from 'cerebral';
+import { connect } from '@web-client/presenter/shared.cerebral';
+import { sequences } from '@web-client/presenter/app.cerebral';
+import { state } from '@web-client/presenter/app.cerebral';
+import React from 'react';
+
+const props = cerebralProps as unknown as {
+  count: number;
```
akuny, line 11: A more explicit name than "count" might help make it more obvious what is being counted in the conditional renders below.

web-client/src/views/UseValidationFocus.tsx
```diff
@@ -0,0 +1,58 @@
+import { MutableRefObject, useEffect, useRef } from 'react';
+
+export const useValidationFocus = (
```
akuny, line 3: There may be enough conditional logic here to merit a test suite for this file, or to move the logic out of the presentational component and test it in isolation.

web-client/src/presenter/actions/StartCase/setDefaultContactStateAction.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  const TYPES = [
+    'filingType',
+    'businessType',
+    'otherType',
+    'estateType',
+    'minorIncompetentType',
+  ];
+
+  if (TYPES.includes(props.key)) {
```
akuny, line 34: The early return here is a bit sneaky. Might be worth considering refactoring lines 36-52 to address that and reduce duplicated code.

web-client/src/presenter/actions/StartCase/updatePartyTypeActionUpdated.ts
```diff
@@ -0,0 +1,97 @@
...
+    store.unset(state.form.estateType);
+    store.unset(state.form.minorIncompetentType);
+    store.unset(state.form.hasSpouseConsent);
+  };
+
+  const updatePartyType = (newPartyType: string) => {
+    store.set(state.form.partyType, newPartyType);
+  };
+
+  const handleFilingType = () => {
```
akuny, line 28: Some of these handlers use destructuring syntax to access the props' value property, while others don't. It might be good to be consistent.

web-client/src/presenter/actions/addAnotherIrsNoticeToFormAction.ts
```diff
@@ -0,0 +1,16 @@
+import { FORMATS } from '@shared/business/utilities/DateHandler';
+import { state } from '@web-client/presenter/app.cerebral';
+
+export const addAnotherIrsNoticeToFormAction = ({
+  applicationContext,
+  get,
+  store,
+}: ActionProps) => {
+  const data = get(state.irsNoticeUploadFormInfo);
+  if (data.length >= 5) return;
```
akuny, line 10: Could use a variable to help make this magic number more meaningful.

web-client/src/presenter/actions/generatePetitionPreviewPdfUrlAction.ts
```diff
@@ -0,0 +1,80 @@
+import { state } from '@web-client/presenter/app.cerebral';
+
+export const generatePetitionPreviewPdfUrlAction = async ({
+  applicationContext,
+  get,
+  store,
+}: ActionProps) => {
+  const petitionFormatted = get(state.petitionFormatted);
```
akuny, line 8: Could refactor to remove some duplication in this action.

web-client/src/presenter/actions/setPetitionFormValueAction.ts
```diff
@@ -0,0 +1,17 @@
+import { state } from '@web-client/presenter/app.cerebral';
+
+export const setPetitionFormValueAction = ({
```
akuny, line 3: This action is really similar to `setFormValueAction.ts` above. Maybe they could be consolidated with some adjustments.

web-client/src/presenter/actions/validateUploadPetitionStep2Action.ts
```diff
@@ -0,0 +1,32 @@
...
+    step2Data.petitionFacts.filter(r => r.length >= 1),
+  );
+
+  store.set(state.form.petitionReasons, step2Data.petitionReasons);
+  store.set(state.form.petitionFacts, step2Data.petitionFacts);
+
+  let errors = new UploadPetitionStep2(
+    step2Data,
+  ).getFormattedValidationErrors();
+  if (errors) {
```
akuny, line 26: Here we evaluate if `errors` is truthy, but in `updatedValidatePetitionAction.ts` we check for falsy. Might be a good idea to use consistent logic across these related actions.

web-client/src/presenter/computeds/showContactsHelperUpdated.ts
```diff
@@ -0,0 +1,51 @@
...
+    filingType === 'Myself and my spouse' ||
+    filingType === 'Myself'
+  );
+}
+function getShowContactSecondary(partyType, PARTY_TYPES, props) {
+  const isContactSecondaryPartyType = [
+    PARTY_TYPES.petitionerDeceasedSpouse,
+    PARTY_TYPES.petitionerSpouse,
+  ].includes(partyType);
+  if (!isContactSecondaryPartyType) return false;
```
akuny, line 47: This might be able to work as a single if statement.

akuny (commented):
This feature is cool! 😎

shared/src/business/entities/contacts/ContactUpdated.ts
nechama-krigsman: There is no specific validation, since we support international and domestic phone numbers. We just have a max of 100 characters. (legacy decision)

cruzjone-flexion (approved):

nechama-krigsman (approved):

jimlerza (approved):

PR #4925, opened 2024-05-16
10350: Enable IRS Practitioners to file EA as First IRS Document on a case

En-8:
https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/gh/flexion/ef-cms/10350

This PR adds Entry of Appearance as an external filing event, but limits its availability as a filing option to only IRS practitioners and only if they're filing the first IRS document on the case.


web-client/src/views/FileDocument/FileDocumentReview.tsx
```diff
@@ -192,222 +195,239 @@ export const FileDocumentReview = connect(
...
+        {fileDocumentHelper.isAutoGenerated ? (
+          <FileEntryOfAppearanceReview
+            documentTitle={form.documentTitle}
+            pdfPreviewUrl={pdfPreviewUrl}
+            showModal={showModal}
+            onBack={() => navigateBackSequence()}
+            onCancel={() => formCancelToggleCancelSequence()}
+            onSubmit={() => submitExternalDocumentSequence()}
+          />
+        ) : (
```
En-8, line 207: This is really what was added. The rest of the changes in this file are the result of nesting what was there before inside the 'else' of this ternary.

cypress/local-only/integration/caseDetail/docketRecord/electronicFiling/practitioner-files-external-document.cy.ts
rachelschneiderman, line 1: Nit: Could this test file be `private-practitioner-files-external-document`? Just so it's clear the user that's the test subject. Otherwise looks great!

shared/src/business/entities/externalDocument/ExternalDocumentInformationFactory.ts
```diff
@@ -38,6 +39,7 @@ export class ExternalDocumentInformationFactory extends JoiValidationEntity {
   public secondarySupportingDocuments?: object[];
   public selectedCases?: string[];
   public supportingDocuments?: object[];
+  public generationType?: string;
```
rachelschneiderman: Is there anyway for this to not be an optional property?

web-client/src/presenter/computeds/completeDocumentTypeSectionHelper.ts
```diff
@@ -8,6 +8,24 @@ import { state } from '@web-client/presenter/app.cerebral';
 
 import { ClientApplicationContext } from '@web-client/applicationContext';
 import { Get } from 'cerebral';
+import { PublicCase } from '@shared/business/entities/cases/PublicCase';
+
+export const isFirstIrsFiling = (
```
rachelschneiderman: Should this be a static method off of an entity? It seems like it's enforcing some business rules.

 I love that it's extracted out, super easy to read and understand!

web-client/src/views/FileDocument/FileEntryOfAppearanceReview.tsx
```diff
@@ -0,0 +1,152 @@
+import { Button } from '../../ustc-ui/Button/Button';
```
rachelschneiderman, line 1: I know we are removing pa11y right now, so not a great time to ask. Is this new view covered by an accessibility test?

web-client/src/views/FileDocument/PrimaryDocumentGeneratedTypeForm.tsx
```diff
@@ -0,0 +1,230 @@
+import { Button } from '@web-client/ustc-ui/Button/Button';
```
rachelschneiderman, line 1: Same here, does this have an accessibility test?

web-client/src/views/FileDocument/StateDrivenFileInput.tsx
```diff
@@ -8,25 +8,36 @@ import { sequences } from '@web-client/presenter/app.cerebral';
...
-    ariaDescribedBy: props.ariaDescribedBy,
-    constants: state.constants,
-    file: props.file,
-    fileInputName: props.name,
-    form: state.form,
-    id: props.id,
-    updateFormValueSequence: sequences[props.updateFormValueSequence],
-    validationSequence: sequences[props.validationSequence],
-  },
+type StateDriveFileInputProps = {
```
rachelschneiderman, line 11: 👏 👏 👏 Nice!!!

cypress/local-only/integration/caseDetail/docketRecord/electronicFiling/irs-practitioner-files-entry-of-appearance-as-first-irs-doc.cy.ts
jimlerza, line 1: Do we need a test that ensures that subsequent IRS users can't also file EAs after one has already been filed?

web-client/src/views/FileDocument/FileEntryOfAppearanceReview.tsx
rachelschneiderman: Nate and I will write an axe test together in a new PR.

web-client/src/views/FileDocument/PrimaryDocumentGeneratedTypeForm.tsx
rachelschneiderman: Nate and I will write an axe test together in a new PR.

cypress/local-only/integration/caseDetail/docketRecord/electronicFiling/irs-practitioner-files-entry-of-appearance-as-first-irs-doc.cy.ts
En-8: I'm not quite sure I follow. Based on regression tests outlined in the card, it appears that IRS practitioners _other_ than those who are already associated with a case should still be able to file EA in the "request access to case" workflow:

> IRS Practitioner accesses a case where there are already IRS Practitioners associated, Request Access to Case button is available, and the Practitioner can file an EA as per usual.

However, an IRS practitioner that _is already associated_ with a case should _not_ be able to file an EA. I'm not sure we have that case covered, so I can add something for it.

Let me know if any of that isn't making sense and we can chat.

web-client/src/presenter/computeds/completeDocumentTypeSectionHelper.ts
En-8: Done 👍🏻 made it a static method on the case entity. Seemed natural given that everything in there was based on the case that was being passed in.

rachelschneiderman (approved):

shared/src/business/entities/externalDocument/ExternalDocumentInformationFactory.ts
En-8: Done 👍🏻

jimlerza (approved):

PR #4928, opened 2024-05-16
10385: allow irs superuser to download served STIN

jtdevos:


jimlerza (approved):

jimlerza:
You could add `it()` blocks to `web-api/hostedEnvironmentTests/irsSuperUser.test.ts` to do this for realsies:

```typescript
it('can download a STIN that has been served', async () => {
  // petitioner creates new case
  // petition is served
  // create user in IRS POOL
  // get reconciliation report
  // get document-download-url for STIN from case
});
```

TomElliottFlexion (approved):

rachelschneiderman (commented):
Looks good to me, I would echo Jim's comment about added a test that hits a real endpoint. Let me know if you would like assistance since I just added a test for that with Zach!

PR #4930, opened 2024-05-16
(OpEx) Report: Total cases open on a specific day of the year in each of the last 5 years

jimlerza:
Added a report to get us approximate, best-we-can-do statistics on the total cases open before the US Tax Court on a given calendar day in each of the last 5 years.

It's approximate because we don't have a complete case status history for all cases.

Therefore, this report can't account for cases closed before the target date and opened up again after the target date. Those cases ideally wouldn't count towards the aggregate totals, but do.

This report's aggregate totals also completely discard any cases that went to appeal. Cases that went to appeal before the target date are rightfully ignored, as these cases were already on appeal and not pending before the US Tax Court. Cases that went to appeal after the target date could have been closed on the target date, or they could have been open. There is no way to reliably determine this programmatically, so this report outputs a CSV containing the docket numbers of all cases that were opened before the target date and went to appeals after the target date, with the expectation that a human will examine these cases and manually add them to the aggregate totals if they are determined to have been pending before the US Tax Court on the target date.


scripts/reports/cases-open-on-date.ts
```diff
@@ -0,0 +1,212 @@
...
+  let totalsOutput = '"Date","Cases Open"';
+  for (const year of Object.keys(totals)) {
+    const targetDate = validateDateAndCreateISO({ day, month, year })!;
+    const date = targetDate.split('T')[0];
+    totalsOutput += `\n"${date}","${totals[year]}"`;
+  }
+  const monthAndDay = validateDateAndCreateISO({
+    day,
+    month,
+    year: String(thisYear),
```
jtdevos: nit: `new String(thisYear)` or `` `${thisYear}` ``

jtdevos (approved):

jasonfu9966 (approved):

scripts/reports/cases-open-on-date.ts
jimlerza: Fixed!

PR #4946, opened 2024-05-20
10386 - Practitioner Count Bugfix

jasonfu9966:
moved formatting the count to the helper, and introduced a new prop to toggle when to show the paginator. this prop is computed in the helper as well.


web-client/src/presenter/computeds/AdvancedSearch/practitionerSearchHelper.ts
```diff
@@ -78,7 +79,8 @@ export const practitionerSearchHelper = (
     result = {
       ...result,
       formattedSearchResults: paginatedResults,
-      numberOfResults: Number(formatPositiveNumber(searchResults.total)),
+      numOfResultsDisplay: formatPositiveNumber(searchResults.total),
+      numberOfResults: searchResults.total,
```
pixiwyn: I'm not a fan of returning the same value twice, once as a number, and the other as a string. Can you please revert this so we format the number in the view as that's a better solution than this IMO.

web-client/src/presenter/computeds/AdvancedSearch/practitionerSearchHelper.test.ts
```diff
@@ -130,6 +146,7 @@ describe('practitionerSearchHelper', () => {
       },
     });
     expect(result).toMatchObject({
+      numOfResultsDisplay: '1',
```
pixiwyn: Since the issue was numbers with commas, the test if it exists should look at a number with a comma in it.

pixiwyn (changes requested):
See feedback. Also please take your time with this, there's no rush, since we won't deploy for another few weeks with the holiday.

web-client/src/presenter/computeds/AdvancedSearch/practitionerSearchHelper.ts
pixiwyn: Or look at solving this differently? I might have sometime later this afternoon to help if needed.

web-client/src/presenter/computeds/AdvancedSearch/practitionerSearchHelper.ts
```diff
@@ -29,9 +29,11 @@ type PractitionerSearchHelperResult = {
   showStateSelect?: boolean;
   activePage?: number;
   formattedSearchResults?: FormattedPractitionerSearchResultType[];
-  numberOfResults?: number;
+  numberOfResults?: string;
+  numOfResultsDisplay?: string;
```
pixiwyn: Don't forget to remove this.

web-client/src/presenter/computeds/AdvancedSearch/practitionerSearchHelper.test.ts
```diff
@@ -111,6 +111,44 @@ describe('practitionerSearchHelper', () => {
...
+        },
+        advancedSearchTab: 'practitioner',
+        searchResults: {},
+      },
+    });
+    expect(result).toMatchObject({
+      activePage: undefined,
+      showPractitionerSearch: true,
+      showStateSelect: false,
+    });
```
pixiwyn: Would expect showPaginator: false ?

web-client/src/presenter/computeds/AdvancedSearch/practitionerSearchHelper.ts
jasonfu9966: good catch!

web-client/src/presenter/computeds/AdvancedSearch/practitionerSearchHelper.ts
```diff
@@ -78,10 +80,11 @@ export const practitionerSearchHelper = (
     result = {
       ...result,
       formattedSearchResults: paginatedResults,
-      numberOfResults: Number(formatPositiveNumber(searchResults.total)),
+      numberOfResults: formatPositiveNumber(searchResults.total),
```
pixiwyn, line 81: this is still going to give nan which is why we set it to 0 before. I get the idea here and think this is better though.

web-client/src/presenter/computeds/AdvancedSearch/practitionerSearchHelper.ts
```diff
@@ -78,10 +80,11 @@ export const practitionerSearchHelper = (
...
       pageCount,
       pageSize: PRACTITIONER_SEARCH_PAGE_SIZE,
       showNoMatches: false,
+      showPaginator: searchResults.total > PRACTITIONER_SEARCH_PAGE_SIZE,
```
pixiwyn: Nit pick but if this is undefined doesn't that error?

pixiwyn: My idea while on lunch was checking if set, or page count.

web-client/src/presenter/computeds/AdvancedSearch/practitionerSearchHelper.ts
```diff
@@ -29,9 +29,11 @@ type PractitionerSearchHelperResult = {
   showStateSelect?: boolean;
   activePage?: number;
   formattedSearchResults?: FormattedPractitionerSearchResultType[];
-  numberOfResults?: number;
+  numberOfResults?: string;
+  numOfResultsDisplay?: string;
   pageCount?: number;
   pageSize?: number;
+  showPaginator?: boolean;
```
pixiwyn: Nit pick, not crazy about all the optional fields as a best practice.

web-client/src/presenter/computeds/AdvancedSearch/practitionerSearchHelper.ts
jasonfu9966: I don't think it's going to show NaN unless `searchResults.total` is undefined or doesn't exist, but this line is in a block that only executes if it's defined. see line 66.

pixiwyn (approved):

web-client/src/presenter/computeds/AdvancedSearch/practitionerSearchHelper.ts
```diff
@@ -29,9 +29,10 @@ type PractitionerSearchHelperResult = {
   showStateSelect?: boolean;
```
zachrog: Hey I see why these are optional parameters but I think, it is much easier to understand if a function always returns a consistent shape rather than having to null check every value. Could you refactor this helper to have no optional return values, so that we know the properties will always exist?

For reference: web-client/src/presenter/computeds/customCaseReportHelper.ts

zachrog (changes requested):
Requesting that the helper always returns a constant shape so that the properties can be relied on existing without having to null check or default values.

web-client/src/presenter/computeds/AdvancedSearch/practitionerSearchHelper.ts
jasonfu9966: [Let's consolidate these two conversations here!](https://github.com/ustaxcourt/ef-cms/pull/4951#pullrequestreview-2069139559)

zachrog (approved):

web-client/src/presenter/computeds/AdvancedSearch/practitionerSearchHelper.ts
```diff
@@ -24,39 +23,40 @@ export type FormattedPractitionerSearchResultType = {
...
-    state.advancedSearchForm.caseSearchByName.countryType,
-  );
   const searchResults = get(state.searchResults['practitioner']);
   const activePage: number = get(
     state.advancedSearchForm.practitionerSearchByName.pageNum,
   );
 
   let result: PractitionerSearchHelperResult = {
     activePage,
-    showPractitionerSearch: permissions?.MANAGE_PRACTITIONER_USERS,
```
rachelschneiderman, line 53: This is still being referenced and needs to be updated.

web-client/src/views/AdvancedSearch/PractitionerSearchResults.tsx
```diff
@@ -25,23 +24,21 @@ export const PractitionerSearchResults = connect(
...
-                        selectedPage: pageChange.selected,
-                      });
-                      focusPaginatorTop(paginatorTop);
-                    }}
-                  />
-                )}
+              {practitionerSearchHelper.showPaginator && (
+                <Paginator
+                  breakClassName="hide"
+                  forcePage={practitionerSearchHelper.activePage}
```
rachelschneiderman, line 30: I believe these should be defaulted to 0 in the helper.

rachelschneiderman (changes requested):

web-client/src/views/AdvancedSearch/PractitionerSearchResults.tsx
jasonfu9966: it's being defaulted in the helper in the block starting on line 45, except `activePage`, which is defaulted in `defaultAdvancedSearchFormAction` because it's shared across a few components. would it make sense to default it another way?

web-client/src/presenter/computeds/AdvancedSearch/practitionerSearchHelper.ts
jasonfu9966: thanks

web-client/src/views/AdvancedSearch/PractitionerSearchResults.tsx
rachelschneiderman: Hmmm I only see it referenced in the helper and the tsx, are there other places I am missing? To me, I don't see a reason it needs to be a value on state.

jasonfu9966: makes sense to me!

jasonfu9966: I feel I should do *some* explaining:
You were right that `activePage` doesn't need to be on state, so I removed it. but when it changes, it reads the value from `pageNum` in state, which is set when the "search" button or any of the paginator buttons is clicked. The "search" button is on a different component than the paginator is on.

jtdevos (approved):

rachelschneiderman (approved):

PR #4950, opened 2024-05-20
10263: Allow Case Services Supervisor to see STIN during QC

En-8:


shared/src/business/entities/DocketEntry.ts
```diff
@@ -41,6 +41,29 @@ import {
   createISODateString,
 } from '../utilities/DateHandler';
 
+const canViewSTIN = (entry, petitionDocketEntry, user) => {
```
zachrog: Typically we attach these business functions directly to the entity, I am wondering if there was a reason you did not want it to be a static method off of DocketEntry?

shared/src/business/entities/DocketEntry.ts
```diff
@@ -670,22 +693,10 @@ export class DocketEntry extends JoiValidationEntity {
...
-        return true;
-      } else if (
-        user.role === ROLES.irsSuperuser &&
-        DocketEntry.isServed(entry) &&
-        DocketEntry.isServed(petitionDocketEntry)
-      ) {
-        return true;
-      }
-      return false;
+      return canViewSTIN(entry, petitionDocketEntry, user);
```
zachrog: niccccceeeeee

web-client/src/presenter/computeds/formattedWorkQueue.ts
```diff
@@ -272,9 +272,15 @@ export const getWorkItemDocumentLink = ({
   const baseDocumentLink = `/case-detail/${workItem.docketNumber}/documents/${workItem.docketEntry.docketEntryId}`;
   const documentViewLink = `/case-detail/${workItem.docketNumber}/document-view?docketEntryId=${workItem.docketEntry.docketEntryId}`;
 
+  const { USER_ROLES } = applicationContext.getConstants();
   let editLink = documentViewLink;
   if (showDocumentEditLink) {
-    if (permissions.DOCKET_ENTRY) {
+    if (
```
zachrog, line 285: It feels strange to me that the web-client needs to do so much computation to understand if a link is editable. It feels as if the function to determine if a link is editable should be based off of permissions and come from our entities. As in we should have a shared function between frontend and backend around if a document is editable.

En-8: We think that what we're modifying here is view logic and not really business logic that makes sense to share between back/front -- we're not computing whether or not the given user can edit, only providing a 'shortcut' in the view to get to the edit screen. But we're open to further discussion.

shared/src/business/entities/DocketEntry.ts
En-8: We really only pulled this out into into its own function because eslint was complaining about the complexity of `isDownloadable`. The thinking was that it's essentially a private helper function containing some of the details of `isDownloadable` that are specific to the STIN document type.

Cody and I are leaning toward putting this back inline in `isDownloadable` and just silencing the eslint warning, but are curious to hear thoughts on this.

zachrog: That makes sense to me. I realize now that you did not export this constant, so it is just for this file. That makes sense and I take back what I said. I like the extraction, but my only ask would be if you are going to extract it, then type the function.

jtdevos (approved):

web-client/src/presenter/computeds/formattedWorkQueue.ts
zachrog: I see what you are saying. I thought the editability of this was being derived based off of permissions and that is incorrect. My only ask is that you add types for getDocketEntryEditLink() function.

zachrog (approved):

shared/src/business/entities/DocketEntry.ts
En-8: Types, coming right up!

web-client/src/presenter/computeds/formattedWorkQueue.ts
En-8: Types, coming right up!

cypress/local-only/integration/documentQC/stin-visible-during-docket-qc.cy.ts
```diff
@@ -0,0 +1,33 @@
+import { goToCase } from '../../../helpers/caseDetail/go-to-case';
```
rachelschneiderman, line 1: Y'all NAILED the organization of this test file, IMO 👏

rachelschneiderman (approved):

jimlerza (approved):

nechama-krigsman (approved):

PR #4951, opened 2024-05-20
10386 practitioner count bugfix

jasonfu9966:
moved formatting the count to the helper, and introduced a new prop to toggle when to show the paginator. this prop is computed in the helper as well.


jtdevos (approved):

TomElliottFlexion (approved):

web-client/src/presenter/computeds/AdvancedSearch/practitionerSearchHelper.ts
```diff
@@ -29,9 +29,10 @@ type PractitionerSearchHelperResult = {
   showStateSelect?: boolean;
   activePage?: number;
   formattedSearchResults?: FormattedPractitionerSearchResultType[];
-  numberOfResults?: number;
+  numberOfResults?: string;
```
rachelschneiderman: Is there a way that these can be required instead of optional? It makes sense to me that if no results were returned `numberOfResults` would be "0" or `showPaginator` would be "false"

web-client/src/views/AdvancedSearch/PractitionerSearchResults.tsx
```diff
@@ -25,23 +24,21 @@ export const PractitionerSearchResults = connect(
...
-                        selectedPage: pageChange.selected,
-                      });
-                      focusPaginatorTop(paginatorTop);
-                    }}
-                  />
-                )}
+              {practitionerSearchHelper.showPaginator && (
+                <Paginator
+                  breakClassName="hide"
+                  forcePage={practitionerSearchHelper.activePage || 0}
```
rachelschneiderman: Same thing here, if these aren't optional and are defaulted in the helper we can avoid doing any logic in the view component.

rachelschneiderman (changes requested):
One thing that I noticed that Github won't allow me to comment on. practitionerSearchHelper exports `showPractitionerSearch` as a variable which is the same value as `advancedSearchHelper.showPractitionerSearch`. Can that variable be removed from practitionerSearchHelper since it's only used in one place where advancedSearchHelper is already being used?

web-client/src/presenter/computeds/AdvancedSearch/practitionerSearchHelper.ts
jasonfu9966: It's a possibility. Here's the thought with this:
- If there are results, an entire `PractitionerSearchHelperResult` will be returned
- If there are no results, it will only return `showNoMatches` and `showSearchResults`, and "only return what it needs"
- If the search hasn't been run yet, it will omit `showNoMatches` and `showSearchResults`, as well as everything else. the absence of these two props is how we know the search hasn't been run yet.

I will concede it's an iffy way of keeping track of these three states, and I'm open to suggestions on making this particular thing better. Should we use separate types for the different returns? 

Edit: let me see if I can figure something out.

jasonfu9966: Never mind, it just works. updated. let me know what you think.

rachelschneiderman (approved):

web-client/src/presenter/computeds/AdvancedSearch/practitionerSearchHelper.ts
```diff
@@ -24,39 +23,40 @@ export type FormattedPractitionerSearchResultType = {
...
-  pageSize?: number;
-  showPractitionerSearch?: boolean;
+  showNoMatches: boolean;
+  showSearchResults: boolean;
+  activePage: number;
+  formattedSearchResults: FormattedPractitionerSearchResultType[];
+  numberOfResults: string;
+  pageCount: number;
+  pageSize: number;
+  showPaginator: boolean;
```
zachrog, line 33: Looking much better, nice.

web-client/src/views/AdvancedSearch/PractitionerSearchResults.tsx
zachrog: Agreed.

web-client/src/views/AdvancedSearch/PractitionerSearchResults.tsx
```diff
@@ -27,9 +27,9 @@ export const PractitionerSearchResults = connect(
               {practitionerSearchHelper.showPaginator && (
                 <Paginator
                   breakClassName="hide"
-                  forcePage={practitionerSearchHelper.activePage || 0}
+                  forcePage={practitionerSearchHelper.activePage}
```
rachelschneiderman, line 30: I think these still need to be defaulted in the helper.

web-client/src/presenter/computeds/AdvancedSearch/practitionerSearchHelper.ts
```diff
@@ -24,40 +23,40 @@ export type FormattedPractitionerSearchResultType = {
...
-    state.advancedSearchForm.caseSearchByName.countryType,
-  );
   const searchResults = get(state.searchResults['practitioner']);
   const activePage: number = get(
     state.advancedSearchForm.practitionerSearchByName.pageNum,
   );
 
   let result: PractitionerSearchHelperResult = {
     activePage,
-    showPractitionerSearch: permissions?.MANAGE_PRACTITIONER_USERS,
```
rachelschneiderman, line 53: This is still being used somewhere and needs to be updated. Just FYI.

web-client/src/views/AdvancedSearch/PractitionerSearchResults.tsx
jasonfu9966: it's being defaulted in the helper in the block starting on line 45, except `activePage`, which is defaulted in `defaultAdvancedSearchFormAction` because it's shared across a few components. would it make sense to default it another way?

web-client/src/presenter/computeds/AdvancedSearch/practitionerSearchHelper.ts
jasonfu9966: thanks

web-client/src/views/AdvancedSearch/PractitionerSearchResults.tsx
rachelschneiderman: Hmmm I only see it referenced in the helper and the tsx. To me, I don't see a reason it needs to be a value on state.

PR #4967, opened 2024-05-28
Move files from shared/ to web-api. (No Functional Changes)

zachrog:
Dropping coverage requirements in shared, because no files were modified, only moved.


shared/src/business/test/createTestApplicationContext.ts
jimlerza, line 1: Can we standardize on `@web-api` over `../../../../web-api/src` in this file?

shared/src/business/test/createTestApplicationContext.ts
```diff
@@ -25,8 +25,8 @@ import {
 } from '../../sharedAppContext';
 import { User } from '../entities/User';
 import { abbreviateState } from '../utilities/abbreviateState';
-import { acquireLock } from '@shared/business/useCaseHelper/acquireLock';
-import { addDocketEntryForSystemGeneratedOrder } from '../useCaseHelper/addDocketEntryForSystemGeneratedOrder';
+import { acquireLock } from '@web-api/business/useCaseHelper/acquireLock';
+import { addDocketEntryForSystemGeneratedOrder } from '../../../../web-api/src/business/useCaseHelper/addDocketEntryForSystemGeneratedOrder';
```
jtdevos: Hey Zach - would a package scope import work here?  In the context of a no-functional-changes change, would such a change not be as "safe" updating the relative path?

web-api/src/business/useCaseHelper/acquireLock.ts
```diff
@@ -7,7 +8,7 @@ export const checkLock = async ({
   onLockError,
   options = {},
 }: {
-  applicationContext: IApplicationContext;
+  applicationContext: ServerApplicationContext;
```
jtdevos, line 11: Is there a rule-of-thumb for when to use IApplicationContext vs. ServerApplicationContext?   Do the references to IApplicationContext not work here after the move from `shared` to `web-api`?

jimlerza (approved):

shared/src/business/test/createTestApplicationContext.ts
jimlerza: Looks like we might be able to set VSCode to always generate import paths using our aliases:

```"typescript.preferences.importModuleSpecifier": "non-relative"```

TomElliottFlexion (approved):

web-api/src/business/useCaseHelper/acquireLock.ts
rachelschneiderman: My understanding of the rule of thumb is `ServerApplicationContext` is applicable to functions in `web-api/src` and ClientApplicationContext is for functions in `web-client/src`. `IApplicationContext` is for the `shared/src` fuzzy middle ground functions 😄

rachelschneiderman (approved):

shared/src/business/test/createTestApplicationContext.ts
zachrog: The short answer is absolutely. The long answer is that when moving these files to web-api I am relying on vscode to update all of the imports of functions so if it was relative then it stays relative and if it was absolute it stays absolute. So if it was using "@web-api/" before then the import will continue to be "@web-api", and if it was using "../../shared/src" then it will update it to "../../web-api/src".

My preference is the @web-api/ syntax but it is just a preference. Changing all 200+ files and all of their imports to this syntax felt separate from what I was trying to achieve with this PR. Let me know if you disagree or would like to mass update all imports.

web-api/src/business/useCaseHelper/acquireLock.ts
zachrog: Yes! `ServerApplicationContext` is meant for functions we know are going to be run on the server(interactors, persistence...) and `IApplicationContext` is meant for functions that could be run on the client or on the server. So like Rachel said generally use `IApplicationContext` in shared/ and `ServerApplicationContext` in web-api/. But make sure it really only gets used on the server before using it!

shared/src/business/test/createTestApplicationContext.ts
zachrog: Yes we can! Yeah when we switched over to typescript I asked everyone to change that import setting to non-relative!

The unfortunate part is, when moving files/functions vscode will automatically update all imports(yay) but it will not update the path from relative to absolute (darn). Let me update this file though.

PR #4971, opened 2024-05-28
10374 - fixing issue with indeterminate related to undefined dom element

codyseibert:


jimlerza (approved):

rachelschneiderman (approved):
Is there a test that can be added that would have caught this bug? I am OK if it's in another PR since we are reaching cutoff time for release

PR #4978, opened 2024-05-29
10387 irs superuser test 

jasonfu9966:
does not need UI testing


cypress/helpers/cypressTasks/cognito/cognito-helpers.ts
```diff
@@ -61,20 +70,44 @@ const getUserPoolId = async (): Promise<string> => {
...
+  await getCognito().adminCreateUser({
+    TemporaryPassword: password,
+    UserAttributes: [
+      {
+        Name: 'custom:role',
+        Value: role,
+      },
+      {
+        Name: 'email',
+        Value: userName,
```
rachelschneiderman: This needs to be .toLowerCase()-d to avoid any login issues.

cypress/deployed-and-local/integration/irsSuperUser/irs-super-user-can-download-stin.cy.ts
```diff
@@ -0,0 +1,79 @@
...
+/**
+ *
+ * This function asserts that the IRS Reconciliation report is accessible and that it contains
+ * at least one downloadable STIN record. Method fails test if either of these conditions is
+ * not met.
+ *
+ * @param bearerToken bearer token for IRS superuser session (see tasks::getBearerToken)
+ */
+const verifyReconciliationReportAndStinUrl = (bearerToken: string) => {
+  const url = `https://api-${Cypress.env('DEPLOYING_COLOR')}.${getCypressEnv().env}.ef-cms.ustaxcourt.gov/v2/reconciliation-report/today`;
```
rachelschneiderman, line 55: I'm curious why we need to hit the reconciliation report endpoint here. Would it be possible to intercept the serve case response instead to get the `docketEntryId` for the stin so the test doesn't have to do more work?

cypress/deployed-and-local/integration/irsSuperUser/irs-super-user-can-download-stin.cy.ts
```diff
@@ -0,0 +1,79 @@
+import { ROLES } from '../../../../shared/src/business/entities/EntityConstants';
+import { getCypressEnv } from '../../../helpers/env/cypressEnvironment';
+import { loginAsPetitioner } from '../../../helpers/authentication/login-as-helpers';
+import { petitionerCreatesElectronicCase } from '../../../helpers/fileAPetition/petitioner-creates-electronic-case';
+import { petitionsClerkQcsAndServesElectronicCase } from '../../../helpers/documentQC/petitions-clerk-qcs-and-serves-electronic-case';
+
+describe('irs superuser integration', () => {
+  //if local or prod, skip this test
+  if (getCypressEnv().env === 'local' || getCypressEnv().env === 'prod') return;
```
rachelschneiderman: In other tests that can only run on hosted environments, I believe the describe block is wrapped with the conditional check, would it be possible to do that here as well for consistency? For example, see `verify-email-receipt.cy.ts`. Also, since this test is located in the deployed-and-local folder it should never run on prod as prod only runs readonly cypress tests anyways.

cypress/helpers/cypressTasks/cognito/cognito-helpers.ts
```diff
@@ -180,3 +213,59 @@ export const deleteAllCypressTestAccounts = async (): Promise<null> => {
...
+  const irsUserPoolId = await getUserPoolId(true);
+  if (!irsUserPoolId) return null;
+  const irsAccounts = await getAllCypressTestAccounts(irsUserPoolId);
+  await Promise.all(
+    irsAccounts.map(user => deleteAccount(user, irsUserPoolId)),
+  );
+  return null;
+};
+
+export async function getBearerToken({
```
rachelschneiderman: This function logs errors specific to the IRS Superuser, that's fine but then it doesn't make sense to me to have `isIrsEnv` be a variable since I am not sure the function is set up to handle or is even relevant to other user types.

cypress/helpers/cypressTasks/cognito/cognito-helpers.ts
```diff
@@ -61,20 +70,44 @@ const getUserPoolId = async (): Promise<string> => {
...
-  )?.Value!;
-
-  return userId;
+  await getCognito().adminSetUserPassword({
+    Password: password,
+    Permanent: true,
+    UserPoolId: userPoolId,
+    Username: userName,
+  });
+  return `created user ${userName}`;
```
rachelschneiderman: Is this consumed anywhere?

cypress/deployed-and-local/integration/irsSuperUser/irs-super-user-can-download-stin.cy.ts
jtdevos: We do this because it roughly mirrors what the IRS does - first they trawl through the reconciliation report to get a listing of docket entries, and with that list they download some subset of that list.   So in that respect I feel the extra work in the test is worth it.

cypress/helpers/cypressTasks/cognito/cognito-helpers.ts
jasonfu9966: it's not, but cypress tasks need to return *something*. would it be better to consume it in an assert? maybe change the return to a void?
edit: for now I added an assert to expect the function to return this string, but it doesn't feel quite right because account creation isn't really the point of this test. open to suggestions.

cypress/deployed-and-local/integration/irsSuperUser/irs-super-user-can-download-stin.cy.ts
jasonfu9966: thanks! we weren't super sure about whether this would run in prod because of the cypress reorg changes. we'll get this fixed.

cypress/helpers/cypressTasks/cognito/cognito-helpers.ts
rachelschneiderman: I believe the tasks can also return null!

cypress/helpers/cypressTasks/cognito/cognito-helpers.ts
```diff
@@ -180,3 +213,57 @@ export const deleteAllCypressTestAccounts = async (): Promise<null> => {
...
+  password: string;
+  userName: string;
+}): Promise<string> {
+  const userPoolId = await getUserPoolId(true);
+  const clientId = await getClientId(userPoolId);
+  const initiateAuthResult = await getCognito().adminInitiateAuth({
+    AuthFlow: AuthFlowType.ADMIN_USER_PASSWORD_AUTH,
+    AuthParameters: {
+      PASSWORD: password,
+      USERNAME: userName,
```
rachelschneiderman: Oops! Just a couple more emails to lowercase!

rachelschneiderman (approved):

PR #4987, opened 2024-05-31
10368 Prevent use of deploying color for real users

En-8:
https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/gh/flexion/ef-cms/10368

This PR disables any HTTP traffic to the deploying color other than traffic originating from the test suite. We achieved this through a combination of a lambda environment variable noting whether or not traffic is disabled and a header allowing test traffic through this check.

In the event that real traffic hits the passive or deploying color, we send back a 500 response with a `X-Force-Refresh` header, and the UI sees that header and prompts the user via modal to reload their page. This is true for both internal and public users.

Since we have this new way to limit traffic to the passive color, we removed the CI step for deleting API mappings, since it's no longer needed.

We also added a new Error page for the UI to display in the event of 500 errors, and fixed an error handling infinite loop that would occur if the request to the feature flag endpoint resulted in a 500.


jtdevos (approved):

web-api/terraform/applyables/blue/blue.tf
```diff
@@ -102,6 +102,7 @@ module "api-east-blue" {
     DYNAMODB_TABLE_NAME    = var.blue_table_name
     ELASTICSEARCH_ENDPOINT = length(regexall(".*beta.*", var.blue_elasticsearch_domain)) > 0 ? data.terraform_remote_state.remote.outputs.elasticsearch_endpoint_beta : data.terraform_remote_state.remote.outputs.elasticsearch_endpoint_alpha
     REGION                 = "us-east-1"
+    DISABLE_HTTP_TRAFFIC   = "true"
```
jimlerza, line 105: nit: I think these were alphabetized

jimlerza (approved):

web-client/src/presenter/state-public.ts
```diff
@@ -49,6 +49,7 @@ export const baseState = {
   },
   isPublic: true,
   isTerminalUser: false,
+  modal: {},
```
TomElliottFlexion, line 52: Does the modal work on mobile?

TomElliottFlexion (approved):

web-client/src/presenter/state-public.ts
En-8: Indeed! Tenille was able to see & use it on a mobile device on test: https://flexion.slack.com/archives/CD6DEU37Z/p1718387128439669

TomElliottFlexion: Sweetness

PR #4994, opened 2024-06-03
Move files from shared/ to web-api/ (No Functional Changes)

zachrog:
Dropping coverage requirements in shared, because no files were modified, only moved.


shared/src/business/useCases/public/getPublicCaseInteractor.ts
```diff
@@ -11,7 +12,7 @@ import { formatPublicCase } from '../../useCaseHelper/consolidatedCases/formatPu
  * @returns {object} the case data
  */
 export const getPublicCaseInteractor = async (
```
rachelschneiderman, line 14: Should this be moved to web-api/ if appContext is going to be typed that way?

web-api/src/notifications/retrySendNotificationToConnections.ts
```diff
@@ -1,4 +1,6 @@
 // eslint-disable-next-line spellcheck/spell-checker
+import { Connection } from '@web-api/notifications/sendNotificationToConnection';
```
rachelschneiderman, line 2: Just curious, why were the notification functions moved to `web-api/notifications` instead of `web-api/src/notifications`?

rachelschneiderman (approved):

web-api/src/notifications/retrySendNotificationToConnections.ts
zachrog: Those functions are in web-api/src. I feel like maybe we should change it to be less confusing eventually.

The alias @web-api/ is short for ./web-api/src/

rachelschneiderman: Oh I was reading the file explorer in the pull request view wrong. Thanks!

shared/src/business/useCases/public/getPublicCaseInteractor.ts
zachrog: Moved!

PR #4995, opened 2024-06-03
Devex 10382: Add Cypress Axe-Core Tests

pixiwyn:


cypress/local-only/support/commands.ts
```diff
@@ -112,6 +115,22 @@ Cypress.Commands.add('waitUntilSettled', (maxTries = 20) => {
   waitAndSee(0);
 });
 
+Cypress.Commands.add('runA11y', () => {
```
zachrog: Rather than injecting this into the global cypress can this be extracted and used like a function?

I typically want to avoid adding anything to the cypress object as it makes it harder to trace and has more complex syntax than a typical Javascript function. https://docs.cypress.io/api/cypress-api/custom-commands#Best-Practices

zachrog (changes requested):
I cannot believe how much work you all did on this and I love it. This is really incredible.

My only ask is to pull the a11y command from the global cypress and make it a plain javascript function. Let me know if you disagree!

rachelschneiderman (approved):

zachrog (approved):

jtdevos (approved):

jimlerza (approved):

jimlerza (changes requested):
It appears the github actions are still looking for some output from some tests that have been removed:

Screenshot 2024-06-14 at 9 16 32 AM

jimlerza (approved):

PR #4997, opened 2024-06-04
Opex: Report of the closed date of cases opened in a given year

jimlerza:
Retrieves all cases from opensearch and outputs a CSV file to `~/Documents`.


jtdevos (approved):
how dare you delete my report

TomElliottFlexion (approved):

jimlerza:
This is presumably for the purposes of calculating either "number of days between each case's open date and closed date" or "number of days between each case's closed date and today" and in either case, we can make that calculation programmatically.

PR #5001, opened 2024-06-05
Devex 10338 - AWS v2 -> v3 conversion: SES

jtdevos:
flexion#10338


web-api/src/persistence/ses/getSesStatus.ts
```diff
@@ -1,15 +1,21 @@
...
-  const { SendDataPoints } = await SES.getSendStatistics({}).promise();
+  const cmd = new GetSendStatisticsCommand();
+  const { SendDataPoints } = await sesClient.send(cmd);
   const numberOfDataPoints = HOURS_TO_MONITOR * 4; // each data point is a 15 minute increment
-  const sesHealth = SendDataPoints.slice(0, numberOfDataPoints).every(
+  const sesHealth = SendDataPoints?.slice(0, numberOfDataPoints).every(
     ({ Rejects }) => Rejects === 0,
   );
-
+  //todo: what is the meaning of SendDataPoints being undefined.  Can we assume success? Failure?
```
zachrog: Is this todo meant to be here?

zachrog (approved):

web-api/src/persistence/ses/getSesStatus.ts
jtdevos: Mostly no, but kinda yes?  It was a note to self until I figured out the answer, but I forgot to remove it. That said, I still don't know the answer.

web-api/src/applicationContext.ts
```diff
@@ -180,41 +184,16 @@ export const createApplicationContext = (
     getEmailClient: () => {
```
rachelschneiderman: As we have been going through these upgrades, Zach encouraged me to pull these functions out into their own files to reduce the file size of applicationContext and make it a bit easier to see what's going on. I'd love to see that happen here too if possible 😄 `'@web-api/persistence/cognito/getCognito'` I think is a good example of one we did.

web-api/src/dispatchers/ses/sendBulkTemplatedEmail.ts
```diff
@@ -15,77 +22,85 @@ import { backOff } from '../../../../shared/src/tools/helpers';
...
- * @param {string} providers.templateName name of the SES template
- * @returns {void}
  */
 export const sendBulkTemplatedEmail = async ({
   applicationContext,
   defaultTemplateData,
   destinations,
   templateName,
-}) => {
+}: {
```
rachelschneiderman, line 32: 👏 👏 👏 👏 👏

web-api/src/dispatchers/ses/sendBulkTemplatedEmail.ts
```diff
@@ -15,77 +22,85 @@ import { backOff } from '../../../../shared/src/tools/helpers';
...
+    const params: SendBulkTemplatedEmailCommandInput = {
+      DefaultTemplateData: JSON.stringify(defaultTemplateData),
+      Destinations: destinations.map(destination => ({
+        Destination: {
+          ToAddresses: [destination.email],
+        },
+        ReplacementTemplateData: JSON.stringify(destination.templateData),
+      })),
+      ReturnPath:
+        process.env.BOUNCED_EMAIL_RECIPIENT ||
```
rachelschneiderman: Nit: Another thing we have been doing is moving these references to `process.env` to `environment.ts`, just to try and standardize where we keep our environment variables. Just a nit but if you have time, that would be awesome too!

jtdevos: It looks like the convention for `environment` is for props to have a default.  So I think I'll play it safe and update it here since the default is obvious but leave the other process.env refs alone where I'm less sure (e.g. `getEmailClient`)

rachelschneiderman (approved):

PR #5005, opened 2024-06-06
10387: make irs superuser test compatible with all envs

TomElliottFlexion:


jtdevos (approved):
good catch, thank you!

TomElliottFlexion (commented):
Robocop is complaining because staging hasn't yet been backmerged into test

jasonfu9966 (approved):

PR #5014, opened 2024-06-07
Devex 10337 - AWS v2 -> v3 conversion: SQS

jtdevos:
Mostly straightforward changes to update to v3 api. However, I deleted `@web-api/persistence/sqs/getMessages.ts` because I found zero references to it and because it's implementation seems broken.

flexion#10337


shared/src/business/useCases/users/generateChangeOfAddress.ts
```diff
@@ -88,33 +89,29 @@ const generateChangeOfAddressForPractitioner = async ({
...
+          requestUserId,
+          updatedEmail,
+          updatedName,
+          user,
+          websocketMessagePrefix,
+        }),
+        QueueUrl: `https://sqs.${process.env.REGION}.amazonaws.com/${process.env.AWS_ACCOUNT_ID}/change_of_address_queue_${process.env.STAGE}_${process.env.CURRENT_COLOR}`,
+      });
+    });
+    await Promise.all(cmds.map(cmd => sqs.send(cmd)));
```
jimlerza: Perhaps we could do a batch here? A la `shared/admin-tools/aws/sqsHelper.ts::addToQueue`?

jimlerza (commented):
Couple of thoughts at this stage:

- I think we should pursue your idea of a single client, like we did for OpenSearch.
- The email-related files modified in this PR will likely have conflicts with our SES branch. It might be easier to merge that branch into this one now and just make them depend on each other.

shared/src/business/useCases/users/generateChangeOfAddress.ts
jtdevos: I agree that should do batching where possible, my only reservation is doing so in the context of this v2->v3 upgrade task.  Let's give it a shot. If it works, great, and if it ends up breaking things I feel we should revert and punt the idea to a separate devex task.

TomElliottFlexion (approved):

zachrog (approved):

PR #5025, opened 2024-06-10
10330 to test

akuny:
- Add validation logic for deadline due dates to ensure that dates prior to the current date are not accepted
- Refine validation messages for CourtIssuedDocumentTypeX entities that have min or max date validation logic


jtdevos:
My understanding is that Joi validation rules are applied both on the client side, and also on the backend during migrations.   In the migration scenario, will this mean that older records eventually fail validation during a migration?

PR #5027, opened 2024-06-11
Move files from shared/ to web-api 2(No Functional Changes)

zachrog:


jimlerza (approved):

shared/src/business/useCases/externalDocument/generateExternalDocumentTitle.ts
```diff
@@ -10,7 +11,7 @@ import { cloneDeep } from 'lodash';
  * @returns {string|void} document title
  */
 export const generateExternalDocumentTitle = (
-  applicationContext: IApplicationContext,
+  applicationContext: ClientApplicationContext,
```
rachelschneiderman, line 14: Should this move to `web-client/` if it is going to be typed this way?

zachrog: Done!

TomElliottFlexion (approved):

PR #5034, opened 2024-06-12
5512 - Switch Petition and Petitioner Information step

cruzjone-flexion:
**Feedback**
- 🟢 Switch Petition and Petitioner Information step in new flow
- 🟢 Reorder the Type of Notice/Case for when user has IRS Notice
- 🟢 Add "Deficiency" as an option for when user does not have IRS Notice

**Refactor**
- 🟡 Start renaming files (do not use step# pattern)


web-client/src/presenter/sequences/petitionGenerationLiveValidationSequence.ts
```diff
@@ -19,15 +19,15 @@ export const petitionGenerationLiveValidationSequence = [
   startShowValidationAction,
   determineStepPathAction,
   {
-    //can remove step 1?
     step1: [
-      getStep1DataAction,
-      validateUploadPetitionStep1Action,
+      getStep2DataAction,
```
TomElliottFlexion: Perhaps we should rename these right away

TomElliottFlexion (approved):

PR #5040, opened 2024-06-13
10400 - Bug Fix

pixiwyn:


cypress/local-only/integration/fileAPetition/file-a-petition.cy.ts
```diff
@@ -183,9 +183,24 @@ describe('creation form', () => {
...
+    });
+    cy.get('button#petition-preview-button').click();
+    cy.get('@windowOpen').should(
+      'be.calledWithMatch',
+      new RegExp('^blob:', 'm'),
+      '_blank',
+    );
   });
 
   it('submits forms and redirects to the file petition success page', () => {
```
pixiwyn: viewport will reset to desktop automatically here ^

web-client/src/views/PDFPreviewButton.tsx
```diff
@@ -63,7 +61,8 @@ export const PDFPreviewButton = connect<
           <Button
             {...buttonProps}
             onClick={() => {
-              return openPdfInNewTabSequence({ file });
+              const url = window.URL.createObjectURL(file);
+              window.open(url, '_blank');
```
pixiwyn, line 65: This was the simplest solution to allow PDFs to open on iOS. With cerebral, the timing is off, and this helps reduce our dependency on cerebral (as an added bonus).

zachrog (approved):

jimlerza (approved):

PR #5048, opened 2024-06-14
10351 cold cases

En-8:
https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/gh/flexion/ef-cms/10351

flexion#10351


shared/src/business/entities/cases/Case.ts
```diff
@@ -221,20 +221,24 @@ export class Case extends JoiValidationEntity {
...
+    const aSplit = docketNumberA.split('-');
+    const bSplit = docketNumberB.split('-');
+
+    if (aSplit[1] !== bSplit[1]) {
+      // compare years if they aren't the same;
+      // compare as strings, because they *might* have suffix
+      return aSplit[1].localeCompare(bSplit[1]);
+    } else {
+      // compare index if years are the same, compare as integers
+      return +aSplit[0] - +bSplit[0];
```
jtdevos, line 238: Curious about the `+` operator use on this line.  Does this implicitly cast the two operands as numbers, or is there another purpose?

web-client/src/views/ColdCaseReport/ColdCaseReport.tsx
```diff
@@ -0,0 +1,36 @@
...
+        <BigHeader text="Reports" />
+        <section className="usa-section grid-container">
+          <SuccessNotification />
+          <ErrorNotification />
+          <div className="title">
+            <h1>Cold Case Report</h1>
+          </div>
+
+          <div className="margin-bottom-2">
+            Showing cases not at issue with no activity in the last 120 days and
```
jtdevos: Nit: Instead of hard-coding 120 days, it might be good to have this value be a property of the report that can be exposed to the components and interpolated in this text instead.

jtdevos (changes requested):

jimlerza (approved):

shared/src/business/entities/cases/Case.ts
En-8: We didn't add this line, only moved it, so I had to look this one up. It does seem to be only for converting the operand to a number. 

til: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Unary_plus

I'm personally unsure if there's any other context surrounding its use.

web-client/src/views/ColdCaseReport/ColdCaseReport.tsx
En-8: How do you feel about the constant-based approach I just pushed?

akuny (approved):

web-client/src/views/ColdCaseReport/ColdCaseReport.tsx
jtdevos: Perfect - thank you sir!

jtdevos (approved):

PR #5075, opened 2024-06-19
10346 - Large PDF Errors

codyseibert:
flexion#10346


web-api/terraform/modules/api/alarms.tf
```diff
@@ -0,0 +1,25 @@
...
+  log_group_name = "/aws/lambda/api_async_${var.environment}_${var.current_color}"
+  pattern        = "\"Failed to parse PDF\""
+
+  metric_transformation {
+    name      = "PDFParseErrors"
+    namespace = "EFCMS"
+    value     = "1"
+  }
+}
+
```
zachrog, line 13: Have you been able to confirm that this alarm is stable for a minimum of a week in test? 

A noisy alarm can be worse than no alarm.

web-client/src/presenter/presenter.ts
```diff
@@ -1490,6 +1491,7 @@ export const presenterSequences = {
     validateChangeLoginAndServiceEmailSequence as unknown as Function,
   validateCourtIssuedDocketEntrySequence:
     validateCourtIssuedDocketEntrySequence as unknown as Function,
+  validateCourtOrderSequence: validateCourtOrderSequence as unknown as Function,
```
zachrog: We should only be doing as unknown as Function for old Sequences. All new sequences should be typed to indicate what their props are.

docs/typesOnTSX.md

zachrog (approved):

web-api/terraform/modules/api/alarms.tf
En-8: It's been in test since 6/27, so nearly a week, and seems to not be noisy. Totally agree that a noisy alarms can do more damage than having none 👍🏻

web-api/src/app-local.ts
```diff
@@ -86,6 +86,10 @@ localStreamsApp.get('/isDone', (req, res) => {
 
 const processChunks = async () => {
   for (const chunk of chunks) {
+    // uncomment this if you want to try and fix
```
TomElliottFlexion, line 89: TY!

web-client/src/views/CreateOrder/CreateOrder.tsx
```diff
@@ -89,12 +94,26 @@ export const CreateOrder = connect(
...
+                <FormGroup
+                  grow
+                  omitFormGroupClass
+                  errorText={validationErrors.documentContents}
+                >
+                  <TextEditor
+                    defaultValue={richText}
+                    editorDelta={editorDelta}
+                    updateFormValueSequence={v => {
+                      console.log(v);
```
TomElliottFlexion: Nit

TomElliottFlexion (approved):

zachrog (approved):

jimlerza (approved):

PR #5093, opened 2024-06-21
Devex 10415: AWS v2 -> v3 conversion: ApiGatewayManagementApiClient 

jtdevos:
flexion#10415


web-api/src/notifications/sendNotificationToConnection.ts
```diff
@@ -1,20 +1,18 @@
+import {
+  type ApiGatewayManagementApiClient,
+  PostToConnectionCommand,
+} from '@aws-sdk/client-apigatewaymanagementapi';
 import { ServerApplicationContext } from '@web-api/applicationContext';
 
 export type Connection = {
+  clientConnectionId: string;
```
TomElliottFlexion, line 8: I _think_ this should be optional, since we sometimes set up a connection with no client id to broadcast to all tabs

TomElliottFlexion (approved):

jimlerza (approved):

web-api/src/notifications/sendNotificationToConnection.ts
TomElliottFlexion: After looking closer, I was being dumb and assuming this type encompassed all user connections rather than a single connection, which definitely does need a client id. Please ignore!

TomElliottFlexion (approved):

PR #5107, opened 2024-06-25
10379: Trigger computed when user logs out and logs back in;

cruzjone-flexion:


En-8 (approved):

web-client/src/presenter/computeds/myAccountHelper.ts
```diff
@@ -1,10 +1,12 @@
 import { ClientApplicationContext } from '@web-client/applicationContext';
 import { Get } from 'cerebral';
+import { state } from '@web-client/presenter/app.cerebral';
+
 export const myAccountHelper = (
   get: Get,
   applicationContext: ClientApplicationContext,
 ): any => {
-  const user = applicationContext.getCurrentUser();
+  const user = get(state.user);
```
zachrog, line 9: So you are telling me that we are storing the user in appContext and it is not accurate! Nice find.

cruzjone-flexion: No, user is saved in appContext correctly but fetching from appContext was not re-triggering the computed.

PR #5110, opened 2024-06-26
10379: BUG: The My Account information is displaying incorrect information

cruzjone-flexion:
Well i was let go today, here is the PR for 10379
Nice working with yall 
- Jonathan Cruz


zachrog (approved):

cypress/local-only/tests/integration/myAccount/bug-switching-accounts-state-cache.cy.ts
jimlerza, line 1: This test makes the implicit assumption that petitioners' my account pages will never get a contact information card. Should we instead assert something that is specific to each user or something?

jimlerza:
Is this problem more pervasive than just this view?

Are we not clearing state upon logout?

zachrog:
> Is this problem more pervasive than just this view?
> 
> Are we not clearing state upon logout?



> Is this problem more pervasive than just this view?
> 
> Are we not clearing state upon logout?

This particular problem is not a clearing state issue (which is a problem in other areas). The problem is that when we use helpers we are generating computed state and computed state is only re-computed if the underlying state changes. Well in this instance the underlying state is stored not in cerebral but in applicationContext, so the computed state does not re-compute.

Is this more pervasive than just this? Yes. Everywhere that we call applicationContext.getCurrentUser() in a helper can have this problem. I wanted to talk about this next week as it relates to our other applicationContext conversation.

jimlerza (approved):

PR #5122, opened 2024-07-04
5512 to test - feedback

nechama-krigsman:


shared/src/business/entities/startCase/UploadPetitionStep2.ts
```diff
@@ -1,184 +1,103 @@
...
-  public otherType: string;
-  public hasSpouseConsent: boolean;
-  public estateType: string;
-  public minorIncompetentType: string;
-  public contactPrimary?: {};
-  public contactSecondary?: {};
+  public petitionRedactionAcknowledgement?: boolean;
+  public petitionFile?: File;
+  public petitionFileSize?: number;
+  public petitionType: (typeof PETITION_TYPES)[keyof typeof PETITION_TYPES];
```
zachrog, line 11: YAY!

web-client/src/presenter/presenter.ts
```diff
@@ -1444,6 +1441,7 @@ export const presenterSequences = {
     updateDocketNumberSearchFormSequence as unknown as Function,
   updateFileDocumentWizardFormValueSequence:
     updateFileDocumentWizardFormValueSequence as unknown as Function,
+  updateFilePetitionSequence: updateFilePetitionSequence as unknown as Function,
```
zachrog: as unknown as Function is for legacy sequences. Would you mind updating the type so that we know if this sequence needs any props? If you need examples take a look at 

docs/typesOnTSX.md

zachrog (changes requested):
Looking good. Just requesting changes so that the sequences are typed.

PR #5142, opened 2024-07-12
10425 - 10417 Remove Get CurrentUser minimum changes

zachrog:
flexion#10425
flexion#10417


web-api/src/lambdas/v1/getDocumentDownloadUrlLambda.test.ts
jimlerza, line 1: Great refactor here. I like this pattern a lot.

jimlerza (approved):

JayFlexy (commented):
How serious are we adherring to the GitHub Action Lint warnings? There are several that complain about the use of Date object or missing await keywords. Otherwise, lgtm!

En-8 (approved):

PR #5156, opened 2024-07-16
10323: Preserving initial tab in order

Mwindo:
This does not run into the same issues as https://github.com/ustaxcourt/ef-cms/pull/1408.


web-client/src/views/CreateOrder/TextEditor.tsx
```diff
@@ -19,13 +19,29 @@ const ReactQuill = React.lazy(async () => {
...
+const formatTextToMaintainInitialTab = (text: string) => {
+  if (text) {
+    // This will render an "invisible" space at the beginning of an initial tabbed line,
+    // which is not ideal, but which is better than not rendering the tab.
+    // eslint-disable-next-line no-control-regex
+    text = text.replace(/^<p>(?<!\u200B)\t/, '<p>&#x200B;\t');
+  }
+  console.log('defaultValue after applying', text);
+  return text;
+};
```
Mwindo: I don't love this solution. That said, I considered several alternatives, many involving changing how the quill component renders tabs (e.g., use 12 spaces as we do in `createOrderAction`). All of these options suffered either from 1) being a sweeping change to the editor that could have unexpected consequences, 2) inconsistent behavior with tab spacing (for instance, inconsistencies with how existing tabs vs. new are displayed in the editor), and/or 3) requiring the user to backspace several times to eliminate a tab.

The solution here seemed like the best compromise between 1) getting the bug fixed, 2) simplicity, and 3) making the minimal change possible to the editor (i.e., doing nothing at all unless the first line has a tab).

web-client/src/views/CreateOrder/TextEditor.tsx
```diff
@@ -82,6 +82,12 @@ export const TextEditor = ({
             const fullDelta = editor.getContents();
             const documentContents = editor.getText();
             const converter = new QuillDeltaToHtmlConverter(fullDelta.ops, {
+              // We add custom CSS styles to maintain tabbing across save and reload.
+              // (Note that preserveWhitespace does not work: https://github.com/ustaxcourt/ef-cms/pull/1408.)
+              // TODO: preserve toolbar-generated indentation across save and reload.
```
Mwindo: I was hoping that maintaining the indentation level--which the user can define by using the toolbar--would be easy enough to restore upon edit. Unfortunately, it isn't, and I don't want to spend too much time on it since this is already an improvement.

web-client/src/views/CreateOrder/TextEditor.tsx
```diff
@@ -19,13 +21,40 @@ const ReactQuill = React.lazy(async () => {
   return { default: reactQuill };
 });
 
+// Quill does two things to render indentations:
+// 1) Adds padding-left so that the richText (and therefore the pdf) has indentations.
+// 2) Adds custom ql-indent-{level} classes to display the indentation in the Quill editor itself.
+// As part of the richText, 1 is automatically preserved across save and reload.
+// As part of the client-side Quill text rendering, 2 is *not* preserved across save and reload,
+// so even though the indentations are "there" in the richText, the user won't see them in the editor.
+// We call addQuillIndentationClasses on loading the text into the editor in order to correct that.
```
Mwindo, line 30: I'm not a fan of long-winded comments, but I spent enough time mucking through this that I think it would be more helpful than hurtful to other developers.

jimlerza: I appreciate the detail!

jimlerza (approved):

TomElliottFlexion (approved):

PR #5157, opened 2024-07-16
10417 minimum changes intermediate branch to test 1721159320

En-8:


JayFlexy (approved):
lgtm

web-api/src/applicationContext.ts
```diff
@@ -99,7 +99,7 @@ const entitiesByName = {
 };
 
 export const createApplicationContext = (
-  appContextUser,
+  appContextUser = {},
```
TomElliottFlexion, line 102: Not sure if this is important, but why are we defaulting this? It seems like we would want this to be undefined for the conditional on 107

TomElliottFlexion: After reading further this makes more sense

PR #5159, opened 2024-07-17
10372: complete message from inbox

TomElliottFlexion:


web-client/src/presenter/sequences/setSelectedMessagesSequence.ts
```diff
@@ -0,0 +1,7 @@
+import { setSelectedMessagesAction } from '../actions/Messages/setSelectedMessagesAction';
+
+export const setSelectedMessagesSequence = [
+  setSelectedMessagesAction,
+] as unknown as (props: {
+  messages: { messageId: string; parentMessageId: string }[];
+}) => void;
```
jtdevos, line 7: Not a criticism, but I don't quite understand what's going on here.  It looks like you are casting an array as a function that returns void... is that right?  Can you maybe explain it to me when you have time?

jtdevos (approved):
Was hoping we could have a chat about that aforementioned piece that I don't understand, otherwise looks good to me 👍

web-client/src/presenter/sequences/setSelectedMessagesSequence.ts
TomElliottFlexion: Let's chat! Short version: this is how we trick Typescript into thinking that Cerebral is type-safe. We haven't finished converting all sequence to use this pattern, which is why `sequences` in `presenter.ts` has `as unknown as Function` all over the place. There are a few dozen sequences that use the "better" pattern, such as `gotoEditOrderSequence` or `formatAndUpdateDateFromDatePickerSequence`

pixiwyn (approved):

jimlerza (approved):

PR #5161, opened 2024-07-17
10413: DOJ Practitioners cannot "Represent a Party" unless case is "On Appeal"

cruzjone-flexion:
07/18/2024 - DO NOT MERGE TO TEST UNTIL 12:30pm MDT!!!


web-client/src/presenter/computeds/caseDetailHeaderHelper.ts
```diff
@@ -1,8 +1,46 @@
...
+const isUserADojPractitioner = (
+  user: RawUser | RawPractitioner | RawIrsPractitioner,
+): boolean => {
+  if (user.role !== ROLES.irsPractitioner) return false;
+  const irsPractitioner: RawPractitioner = user as RawPractitioner;
+  if (irsPractitioner.practiceType !== PRACTICE_TYPE.DOJ) return false;
+  return true;
+};
+
+const shouldShowRepresentAPartButton = (
```
En-8: Missing a 'y' on 'Party' here 😄

web-client/src/presenter/computeds/caseDetailHeaderHelper.ts
```diff
@@ -1,8 +1,46 @@
...
+  caseDetail: RawCase,
+  user: RawUser | RawPractitioner | RawIrsPractitioner,
+  caseHasRespondent: boolean,
+  isRepresentAPartyForm: boolean,
+  isCaseSealed: boolean,
+  isCurrentPageFilePetitionSuccess: boolean,
+): boolean => {
+  const isDojPractitioner = isUserADojPractitioner(user);
+
+  const hideRepresentAPartyToDojUser =
```
En-8: I'm getting a little confused by this variable name, but it may just be a me thing -- when `canDojPractitionerRepresentParty` is `true`, we set this 'hideRepresentParty' thing to `true`. Then, when `hide...` is true (as long as all the others eval to true), the function `shouldShow...()` returns `true`? If I'm following, would it be clearer to name this `show...` instead of `hide...`, or am I misreading this?

En-8 (approved):

web-client/src/presenter/computeds/caseDetailHeaderHelper.ts
cruzjone-flexion: Hmmm I think there was a reason 'Y' I did that..... I will fix it.

web-client/src/presenter/computeds/caseDetailHeaderHelper.ts
cruzjone-flexion: You are correct, will update name. (Think I started implementing it as when to hide and then implemented when to show)

PR #5165, opened 2024-07-18
10421: Add ", et al." to Docket Number in Page Headers When Generating Orders in Consolidated Cases

Mwindo:
ZenHub Issue: https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/gh/flexion/ef-cms/10421


shared/src/business/utilities/documentGenerators/order.test.ts
```diff
@@ -220,6 +220,74 @@ describe('orders and notices', () => {
...
+    testDescription: 'generates a Notice with added docket entries',
+  });
+
+  describe('non-first page docket headers', () => {
+    const orderContentWithMultiplePages =
+      SYSTEM_GENERATED_DOCUMENT_TYPES.orderForAmendedPetitionAndFilingFee.content.repeat(
+        3,
+      );
+    // A full-image comparison is not granular enough to detect the difference we are trying
+    // to test, so we focus the comparison on the part of the image that matters.
```
Mwindo, line 232: I'm not sure how well this will get along with puppeteer updates. Another testing option is to create PageMetaHeaderDocket.test.ts that checks the rendered React contains/does not contain the appropriate text, although 1) there are no other tests like this in the project and 2) this is a proxy (even if tightly coupled) for testing the pdf itself.

TomElliottFlexion: This is a nice solution; do we have reason to believe puppeteer will have issues with it?

Mwindo: The solution is similar to lowering (i.e., making stricter) the threshold for acceptable differences between the expected and actual images in this test. (Since the threshold is percentage-based and we will be checking fewer pixels.) So I suspect this check will be more sensitive to puppeteer updates if puppeteer decides to render things slightly differently. But take this with a grain of salt since I have not been around for any puppeteer updates yet!

TomElliottFlexion: Gotcha, thanks! These tests are mostly a smoketest against unnoticed/unintended changes to pdf generation, so if we have to regenerate after a puppeteer update, we can use our discretion in determining whether or not a change is acceptable. It's not _too_ difficult to regenerate pdfs, so I don't have many concerns with this approach

TomElliottFlexion: Assuming this consistently passes _without_ puppeteer updates, that is

TomElliottFlexion (approved):
Nice

shared/src/business/utilities/documentGenerators/order.test.ts
Mwindo: Cool! Thank you for the explanation! The test is passing consistently (and failed appropriately when I swapped the two relevant expected image files).

Mwindo: Also, thanks to @codyseibert for the recommendation on this approach!

shared/src/business/utilities/pdfGenerator/components/PageMetaHeaderDocket.tsx
```diff
@@ -1,30 +1,38 @@
 import React from 'react';
 
 export const PageMetaHeaderDocket = ({
+  addedDocketNumbers = [],
   docketNumber,
   // do not set this to true unless the content of the PDF also is rendering the century schoolbook font; there is a puppeteer documented here https://github.com/puppeteer/puppeteer/issues/422#issuecomment-759424240
```
Mwindo: @Mwindo Make this comment grammatical and resolve after (waiting to do this for now in case there are other changes to make)

akuny (approved):

jimlerza (approved):

PR #5168, opened 2024-07-19
10396-story: Fix a11y nested-interaction violations

Mwindo:
ZenHub ticket: https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/gh/flexion/ef-cms/10396


web-client/src/ustc-ui/Tabs/Tabs.tsx
```diff
@@ -45,17 +45,18 @@ const renderTabFactory = ({
...
     };
 
     const buttonProps = {
+      'aria-controls': tabContentId,
+      'aria-selected': isActiveTab,
       className: childClassName,
       disabled,
       id: buttonId,
       onClick: () => setTab(tabName),
+      role: 'tab',
```
Mwindo, line 52: In our DOM tree, we were generating something like the following:
```
<ul role="tablist">
    <button>Blah</button>
  </li>
  ...
</ul>
```

The issue here is that `button` is inherently interactive, and it was being nested within an `li` that was rendered interactive because of its role as a tab child element. I fixed this by moving the `role`, `aria-control`, and `aria-selected` properties to the `button` element and signaling that the `li` element is non-semantic (in other words, just used for display) with `role='presentation'`.

PR #5169, opened 2024-07-19
5512 test - feedback

nechama-krigsman:


cypress/local-only/tests/integration/caseDetail/docketRecord/electronicFiling/logged-in-user-can-see-un-served-petition-document.cy.ts
```diff
@@ -9,6 +9,22 @@ import {
 } from '../../../../../../helpers/fileAPetition/petitioner-creates-electronic-case';
 
 describe('Logged In User Can See Un-Served Petition Document', () => {
+  before(() => {
+    cy.task('toggleFeatureFlag', {
```
TomElliottFlexion, line 13: Let's get rid of this pattern when we can; we need to update all tests to be compatible with the new flow anyway

TomElliottFlexion (approved):
TY!

cruzjone-flexion (approved):

cypress/local-only/tests/integration/caseDetail/docketRecord/electronicFiling/logged-in-user-can-see-un-served-petition-document.cy.ts
nechama-krigsman: 👍

PR #5176, opened 2024-07-23
10349 Bug: Idle Logout Issues (with automated tests)

Mwindo:
See [ticket](https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/gh/flexion/ef-cms/10349) for bug details.

There ended up being several issues related to idle sign-out behavior. Broadly, they were the following:

- With multiple tabs open, idle logout behavior was not always consistent.
- Message alerts (e.g., "Success" or "Failure" boxes) remained on the sign in screen after being logged out.
- The app could attempt to sign out an idle user when the client needed to refresh, which would cause confusing front-end behavior (e.g., infinite loading icon).

This PR enforces the following behavior:

1. New tabs start tracking the idle behavior of the user on page load rather than on interaction, which ensures tabs' idle timeouts are synced.
2. New tabs dismiss "Are you still there?" modals in other (possibly older, forgotten) tabs to prevent potential surprise logouts in the current tab.
3. Multiple tabs signing out at the same time all end up at the Idle Logout screen. (Before this PR, some tabs ended up at the Idle Logout screen, some ended up at the Login screen, and some ended up with an infinite-until-refresh loading icon).
4. We broadcast the "you need to close and refresh" event to all tabs to keep them consistent.
5. We do not track idle behavior or sign the user out if the client needs to be refreshed.
6. We reset the idle time tracking when the user signs out to prevent the "Are you still there?" modal from appearing in some cases immediately upon signing in.


web-client/src/AppInstanceManager.tsx
```diff
@@ -18,13 +18,15 @@ export const AppInstanceManager = connect(
     appInstanceManagerHelper: state.appInstanceManagerHelper,
     confirmStayLoggedInSequence: sequences.confirmStayLoggedInSequence,
     resetIdleTimerSequence: sequences.resetIdleTimerSequence,
-    signOutSequence: sequences.signOutSequence,
+    signOutIdleSequence: sequences.signOutIdleSequence,
+    signOutUserInitiatedSequence: sequences.signOutUserInitiatedSequence,
```
Mwindo, line 25: `signOutUserInitiatedSequence` is more or less just a wrapper for `signOutSequence` for consistency in nomenclature/conceptual boundaries given the new `signOutUserInitiatedSequence`. However, if we prefer to just have `signOutSequence` instead, I'm happy to make that change.

web-client/src/AppInstanceManager.tsx
```diff
@@ -36,8 +38,13 @@ export const AppInstanceManager = connect(
         case 'stayLoggedIn':
           confirmStayLoggedInSequence();
           break;
-        case 'logout':
-          signOutSequence({
+        case 'idleLogout':
+          signOutIdleSequence({
+            skipBroadcast: true,
+          });
+          break;
```
Mwindo, line 45: This separate message helps keep behavior between "regular sign out" and "idle sign out" distinct.

web-client/src/presenter/actions/clearIdleTimerAction.ts
```diff
@@ -0,0 +1,9 @@
+import { state } from '@web-client/presenter/app.cerebral';
+
+export const clearIdleTimerAction = ({ store }: ActionProps) => {
+  store.unset(state.lastIdleAction);
+  store.set(state.idleLogoutState, {
+    logoutAt: undefined,
+    state: 'INITIAL',
+  });
+};
```
Mwindo, line 9: We were never clearing this on sign out. This led to a minor (and humorous!) bug in which a user who waited on the screen for 55 minutes after signing out would see the "Are you still there?" modal.

web-client/src/presenter/sequences/signOutIdleSequence.ts
```diff
@@ -0,0 +1,21 @@
...
+    // To avoid race conditions, we ignore redundant calls to sign out that arise when
+    // multiple tabs broadcast the idle sign out event.
+    no: [],
+    yes: [
+      setLogoutTypeAction('idleLogout'),
+      signOutSequence,
+      navigateToLoginSequence,
+      setupCurrentPageAction('IdleLogout'),
+    ],
+  },
```
Mwindo, line 20: For idle sign out, every tab will both sign out itself and broadcast other tabs to sign out. This led to issues: sometimes, for instance, the loading screen would occur at the end (which led to infinite spinning wheel).

For regular sign out, we don't run into this issue.

web-client/src/views/IdleActivityMonitor.tsx
```diff
@@ -46,6 +46,9 @@ export const IdleActivityMonitor = connect(
         handleIdleLogoutSequence();
       }, 1000);
 
+      // Make sure we broadcast activity as soon as a new tab loads
+      broadcastIdleStatusActiveSequence();
```
Mwindo, line 50: Otherwise, a user could sign in, refresh, do nothing for 55 minutes, and never see the "Are you still there?" modal.

web-client/src/views/IdleLogout.tsx
```diff
@@ -16,7 +19,14 @@ export const IdleLogout = connect(
           United States Tax Court website for information on court services and
           contact information.
         </p>
-        <Button onClick={() => navigateToLoginSequence()}>Log In</Button>
+        <Button
+          onClick={() => {
+            navigateToLoginSequence();
+            gotoLoginSequence();
```
Mwindo: We need `gotoLoginSequence` (which sets up the page action) in addition to `navigateToLoginSequence` so that the page action isn't stuck on `IdleLogout`, which otherwise renders this button non-functional (since we navigate to /login ... but we might already be there, looking at the `IdleLogout` screen!).

web-client/src/presenter/sequences/signOutIdleSequence.ts
```diff
@@ -0,0 +1,22 @@
...
+  {
+    // To avoid race conditions, we ignore redundant calls to sign out that arise when
+    // multiple tabs broadcast the idle sign out event.
+    no: [],
+    yes: [
+      setLogoutTypeAction(LOGOUT_BROADCAST_MESSAGES.idleLogout),
+      signOutSequence,
+      navigateToLoginSequence,
+      setupCurrentPageAction('IdleLogout'),
+    ],
```
Mwindo, line 20: For idle sign out, every tab will both sign out itself and broadcast other tabs to sign out. This led to issues: sometimes, for instance, the loading screen would end up being set at the wrong time (which led to an infinite spinning wheel).

For regular sign out, we don't run into this issue.

web-client/src/presenter/actions/clearIdleTimerAction.ts
```diff
@@ -0,0 +1,10 @@
+import { IDLE_LOGOUT_STATES } from '@shared/business/entities/EntityConstants';
+import { state } from '@web-client/presenter/app.cerebral';
+
+export const clearIdleTimerAction = ({ store }: ActionProps) => {
+  store.unset(state.lastIdleAction);
+  store.set(state.idleLogoutState, {
+    logoutAt: undefined,
+    state: IDLE_LOGOUT_STATES.INITIAL,
+  });
+};
```
Mwindo: We were never clearing this on sign out. This led to a minor (and humorous!) bug in which a user who waited on the screen for 55 minutes after signing out would see the "Are you still there?" modal.

web-client/src/presenter/state.ts
```diff
@@ -1,6 +1,10 @@
 /* eslint-disable max-lines */
 import { FormattedPendingMotionWithWorksheet } from '@web-api/business/useCases/pendingMotion/getPendingMotionDocketEntriesForCurrentJudgeInteractor';
 import { GetCasesByStatusAndByJudgeResponse } from '@web-api/business/useCases/judgeActivityReport/getCaseWorksheetsByJudgeInteractor';
+import {
+  IDLE_LOGOUT_STATES,
+  IdleLogoutStateType,
+} from '@shared/business/entities/EntityConstants';
```
Mwindo, line 7: This file is using `getConstants` already. However, given a recent discussion about eliminating `getConstants`, I imported directly from `EntityConstants` instead.

web-client/src/views/IdleLogout.tsx
Mwindo: This was an artifact of my having changed the route on idle logout. In fact, the current app behavior does *not* do that, so I have undone this change, which makes this comment obsolete.

web-client/src/presenter/sequences/signOutIdleSequence.ts
```diff
@@ -0,0 +1,20 @@
...
+  isLoggedInAction,
+  {
+    // To avoid race conditions, we ignore redundant calls to sign out that arise when
+    // multiple tabs broadcast the idle sign out event.
+    no: [],
+    yes: [
+      setLogoutTypeAction(LOGOUT_BROADCAST_MESSAGES.idleLogout),
+      signOutSequence,
+      setupCurrentPageAction('IdleLogout'),
+    ],
```
Mwindo: The issue: all tabs/windows will time out at the same time and send a broadcast message to the others to log out as well, which can lead to race issues (e.g., an infinite loading screen).

If someone has a better solution, I'm all ears! The other solutions I considered are either insufficient or more involved:

1) We use `window.document.visibilityState` to only broadcast from visible tabs. However, this doesn't solve the issue since the user might have multiple visible windows. Moreover, I am afraid of something like the following occurring: user starts to log out but something hangs; tab is minimized in the meantime; no broadcast happens, and therefore sign out fails in other tabs.
2) Do something similar, but set some local state to keep track of the last focused tab and only broadcast from it. This might be a good solution but requires more work and seems more prone to error (now we have to keep track of tabs--e.g., what happens when the last active tab is closed?).

Redundant calls to sign out aren't ideal, but they are already happening _without_ being accounted for.

web-client/src/presenter/actions/isLoggedInAction.ts
```diff
@@ -0,0 +1,8 @@
+import { state } from '@web-client/presenter/app.cerebral';
+
+export const isLoggedInAction = ({ get, path }: ActionProps) => {
+  if (get(state.user)) {
```
Mwindo: Maybe token is better to check?

cypress/local-only/tests/integration/logoutBehavior/idle-logout.cy.ts
```diff
@@ -0,0 +1,69 @@
+import { IWindowWithCerebralExposed } from '../../../../../types/IWindow';
+import { loginAsColvin } from '../../../../helpers/authentication/login-as-helpers';
+import { overrideIdleTimeouts } from '../../../support/idleLogoutHelpers';
+import { retry } from '../../../../helpers/retry';
+
+describe('Idle Logout Behavior', () => {
+  const DEFAULT_IDLE_TIMEOUT = 500;
+  it('should automatically log user out after refresh with option to log back in', () => {
+    loginAsColvin();
+    cy.reload(); // Refresh ensures we track idle time even without interaction on the page
```
Mwindo, line 10: This was the source of a bug, hence the test.

Mwindo:
EDIT: After discussion, having the automated tests seemed better than not having them. If they ever become flaky or problematic to maintain, we can ditch them.

Original comment below

------

@jimlerza @zacharyrogers1 Could you take a look at the new testing paradigm here (don't worry about the other stuff for now) and see what you think? I'm happy to get rid of it all if you guys don't think it is worthwhile, but "it's working on my machine."

The basic idea:
1) Expose the cerebral controller on the window object in non-production environments (see `app.tsx`).
2) With that exposed, we can override any constants in our cypress tests--in my case, the idle timeouts. See `idleLogoutHelpers.ts`.
3) Native cypress cannot open multiple tabs, and no tests I did to "fake" multiple tabs were effective. I therefore used the [cypress puppeteer plugin](https://github.com/cypress-io/cypress/blob/develop/npm/puppeteer/README.md), which is apparently in public beta but is listed on the "official" [plugins page](https://docs.cypress.io/plugins). This configuration is mostly in `cypress.config.ts`, with the tests in `idle-logout.cy.ts`.

As far as I can tell, the pros:
- We can test things involving multiple tabs.
- We have specific tests for cross-tab logout behavior. (One or two tests TBD, if I continue down this route.)
- The tests aren't beautiful, but they aren't terrible.

The cons:
- The puppeteer plugin is a new dependency.
- Exposing the cerebral controller isn't my _favorite_ thing in the world.
- The tests represent a small degree of new complexity to manage.
- If we decide this is a reasonable approach, I will have to determine why the checks are failing in GA despite succeeding consistently locally. EDIT: They are working. Funnily enough, this was the most challenging portion of the PR.

web-client/src/app.tsx
```diff
@@ -271,13 +277,11 @@ const app = {
 
     root.render(
       <Container app={cerebralApp}>
-        {!process.env.CI && (
-          <>
-            <IdleActivityMonitor />
-            <AppInstanceManager />
-            <GlobalModalWrapper />
-          </>
-        )}
```
Mwindo, line 280: This check appears unnecessary--tests are not failing with it removed. It was originally added years ago, apparently to prevent accessibility checks from failing at that time. I think it is best to remove because 1) as said, it appears to be unnecessary and 2) it is a "gotcha" for anyone trying to write automated tests depending on the components within this block (as I was!).

In fact, I think this gotcha already _did_ trip us up. See [my comment below on `broadcastLogoutAction.js`](https://github.com/ustaxcourt/ef-cms/pull/5176/files#r1696026167).

web-client/src/presenter/actions/broadcastLogoutAction.ts
```diff
@@ -6,11 +8,14 @@
  */
 export const broadcastLogoutAction = async ({
   applicationContext,
+  get,
   props,
 }: ActionProps) => {
   // for some reason this causes jest integration tests to never finish, so don't run in CI
-  if (!process.env.CI && !props.skipBroadcast) {
+  if (!props.skipBroadcast) {
```
Mwindo, line 15: See my above comment: https://github.com/ustaxcourt/ef-cms/pull/5176/files#r1695792563. I highly suspect that this was failing because of our !process.env.CI check in app.tsx.

Mwindo: Aaaand I forgot to move the "for some reason" comment lol. Will do on next push.

web-client/src/presenter/actions/broadcastLogoutAction.test.ts
```diff
@@ -6,25 +6,7 @@ import { runAction } from '@web-client/presenter/test.cerebral';
...
-      },
-    });
-
-    expect(
-      applicationContext.getBroadcastGateway().postMessage,
-    ).not.toHaveBeenCalled();
-  });
-
-  it('does not broadcast an event if skipBroadcast is true and CI is defined', async () => {
-    process.env.CI = 'true';
```
Mwindo, line 27: See [my comment on `broadCastLogoutAction` below](https://github.com/ustaxcourt/ef-cms/pull/5176/files#r1696026167). We should no longer need to test anything to do with `process.env.CI` being defined/undefined here.

web-client/src/presenter/actions/broadcastLogoutAction.ts
```diff
@@ -6,11 +8,13 @@
  */
 export const broadcastLogoutAction = async ({
   applicationContext,
+  get,
   props,
 }: ActionProps) => {
-  // for some reason this causes jest integration tests to never finish, so don't run in CI
-  if (!process.env.CI && !props.skipBroadcast) {
+  if (!props.skipBroadcast) {
```
Mwindo, line 14: See comments above: https://github.com/ustaxcourt/ef-cms/pull/5176/files#r1695792563,  https://github.com/ustaxcourt/ef-cms/pull/5176/files#r1695823068

zachrog (approved):

zachrog (approved):

web-client/src/presenter/actions/setDawsonHasUpdatedAction.ts
```diff
@@ -0,0 +1,24 @@
...
+  get,
+  props,
+  store,
+}: ActionProps) => {
+  store.set(state.dawsonHasUpdated, true);
+
+  // Clear the refresh token interval since all subsequent requests until refresh will fail
+  const oldInterval = get(state.refreshTokenInterval);
+  clearInterval(oldInterval);
+  store.unset(state.refreshTokenInterval);
```
Mwindo: I'm debating on whether to pull this block out as a separate action. I like that it is closely coupled to this action. I don't like that it is essentially a subaction.

web-client/src/presenter/sequences/signOutIdleSequence.ts
```diff
@@ -0,0 +1,32 @@
...
+        clientDoesNotNeedToRefresh: [
+          setLogoutTypeAction(BROADCAST_MESSAGES.idleLogout),
+          signOutSequence,
+          setupCurrentPageAction('IdleLogout'),
+        ],
+        // If the client needs to refresh, the sign-out will fail,
+        // and this can lead to inconsistent front-end behavior.
+        clientNeedsToRefresh: [],
+      },
+    ],
```
Mwindo, line 27: Although we clear the idle activity interval in IdleActivityMonitor, there is still be a brief period in which this could be called before the interval has been cleared.

web-client/src/presenter/actions/handleIdleLogoutAction.ts
```diff
@@ -1,35 +1,48 @@
...
   const lastIdleAction = get(state.lastIdleAction);
   const idleLogoutState = get(state.idleLogoutState);
   const isUploading = get(state.fileUploadProgress.isUploading);
-  const user = get(state.user);
+  const userIsLoggedIn = !!get(state.token);
+  const clientNeedsToRefresh = get(state.clientNeedsToRefresh);
 
-  if (user && !isUploading && idleLogoutState.state === 'INITIAL') {
+  // Short-circuit here to prevent showing the "Are you still there?" modal, etc. when inappropriate
+  if (!userIsLoggedIn || clientNeedsToRefresh) {
```
Mwindo, line 13: Originally, the "user is logged in?" check was short-circuiting, just like I have it here. Looking at the Git history, the check was accidentally moved into the first `if` block. After moving it back out, I also refactored the `if/else if` into multiple independent `if` blocks for easier readability since this was an easy mistake to make.

web-client/src/AppInstanceManager.tsx
jimlerza: Your instincts are correct here! I like that they are both explicit.

web-client/src/presenter/actions/broadcastLogoutAction.ts
jimlerza: Nice catch!

jimlerza (approved):

PR #5181, opened 2024-07-23
10417 remove get current user

zachrog:
# In this PR
 - Removal of .getCurrentUser() off of applicationContext. The current user is passed into all functions.
 - Creation of a new AuthUser entity.
# Manual Deploy Steps
- Run account specific (Added permission to ci/cd to allow cognito add custom attributes)
- Run script to give petitioner role to all cognito users without role(We expect this to take ~2hrs)
  - Run environment switcher
  - `npx ts-node --transpile-only scripts/run-once-scripts/add-petioner-role-to-roleless-user.ts`
    - [x] ustc-dev
    - [x] ustc-exp1
    - [x] ustc-irs
    - [x] ustc-stg
    - [x] ustc-test
- Deploy as normal
- Manually add all required attributes to all users in the IRS user pool. (custom:userId, email, name, custom:role)
  - [x] ustc-dev
  - [x] ustc-exp1
  - [x] ustc-irs
  - [x] ustc-stg
  - [x] ustc-test


JayFlexy (approved):
LGTM

shared/src/business/entities/JoiValidationEntity.ts
```diff
@@ -192,14 +192,18 @@ export abstract class JoiValidationEntity {
...
       rawEntity =>
-        new this(rawEntity, args).validate().toRawObject() as ExcludeMethods<T>,
+        new this(rawEntity, ...args).validate().toRawObject() as ExcludeMethods<
+          InstanceType<ClassType>
+        >,
     );
   }
 }
+
+type Tail<T extends any[]> = T extends [any, ...infer Rest] ? Rest : never;
```
jtdevos, line 209: Based on the commit comment I assume that this is merely a typing change and not a substantive change. My Typescript-fu is weak: could someone explain the effect that this change has and/or how it improves the developer experience?

web-api/src/business/useCases/document/batchDownloadDocketEntriesInteractor.ts
```diff
@@ -19,14 +20,16 @@ export type DownloadDocketEntryRequestType = {
...
 ) => {
   try {
     await batchDownloadDocketEntriesHelper(
       applicationContext,
       downloadDocketEntryRequestInfo,
+      authorizedUser,
     );
   } catch (error: any) {
-    const { userId } = applicationContext.getCurrentUser();
+    const { userId } = authorizedUser;
```
jtdevos, line 32: Should there be a check that the user is authorized (or at least logged in?)

shared/src/business/entities/JoiValidationEntity.ts
zachrog: Ugh.... This function. So there is one minor change to the functionality of this function but everything else is type changes. The functional change is that now validateRawCollection spreads args (...args) so that a constructor can accept more than two arguments.

The typing for this function is, and let's be honest, crazy. It is just crazy. Before we made changes this function was not typed and just accepted any's which meant that if you did not pass in the right constructor arguments you would not be warned. We updated the types and now Typescript will complain if you do not pass in the correct arguments. Hooray! 

The typing to achieve that is pretty ugly but it is honest. Personally I do not think this function should exist because the typing is complex and the functionality is not that helpful. So we only improved the typing because we did not think it was a good idea to undertake a refactoring of validateRawCollection in such a massive PR.

web-api/src/business/useCases/document/batchDownloadDocketEntriesInteractor.ts
zachrog: Good question. For this one we kept the same functionality as existed before but would be happy to add an if check and a test.

shared/src/business/entities/JoiValidationEntity.ts
jtdevos: Thanks, Zach! Appreciate the explanation.

web-api/src/business/useCases/document/batchDownloadDocketEntriesInteractor.ts
jtdevos: Thanks, Zach.  Let's punt in the interest of minimizing functionality changes for now...

jtdevos (approved):

jimlerza (approved):

PR #5193, opened 2024-07-26
5512: Handle Feedback

cruzjone-flexion:


shared/src/business/entities/cases/ElectronicPetition.ts
```diff
@@ -178,11 +174,6 @@ export class ElectronicPetition extends JoiValidationEntity {
         '*': 'Your Petition file size is empty',
         'number.max': `Your Petition file size is too big. The maximum file size is ${MAX_FILE_SIZE_MB}MB.`,
       }),
-    petitionRedactionAcknowledgement: joi.boolean().when('petitionType', {
```
TomElliottFlexion, line 181: I think we need this?

nechama-krigsman: Yes, this is going to to be included as a followup story or as part of 10339

TomElliottFlexion (approved):

nechama-krigsman (approved):

PR #5195, opened 2024-07-29
10349 Bug: Idle Logout Issues (without automated tests)

Mwindo:
See [ticket](https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/gh/flexion/ef-cms/10349) for bug details.

TL;DR: With multiple tabs, idle logout behavior was not always consistent.

This PR enforces the following behavior:

1. New tabs start tracking the idle behavior of the user on page load rather than on page interaction, which ensures tabs' idle timeouts are synced. (Otherwise, if the user opened a new tab and did nothing, no idle activity was tracked.)
2. New tabs dismiss "Are you still there?" modals in other (possibly older, forgotten) tabs to prevent potential surprise logouts in the current tab.
3. Multiple tabs signing out at the same time all end up at the Idle Logout screen. (Before this PR, some tabs ended up at the Idle Logout screen, some ended up at the Login screen, and some ended up with an infinite-until-refresh loading icon.)

A good portion of the git diff is just refactoring out magic strings.

EDIT to below: I have automated tests working [here](https://github.com/ustaxcourt/ef-cms/pull/5176).
--------
Automated testing for the enforced behavior outlined above is, unfortunately, difficult: the test requires changing our idle session constants as well as opening (or pretending to open) multiple tabs (and Cypress does not natively support multiple tabs). To make a long story short: I had everything working locally but could not get the tests to pass in CI (see storied commit history). Based on previous discussion with Chris H., Jim L., and Zach about the costs/benefits of automated testing for this bugfix, I have decided to punt on the tests. I can at least provide the testers with scripts they can run in console to override the 55-minute session idle time.

To test manually, you can either 1) edit `getConstants.ts`' `SESSION_MODAL_TIMEOUT` and `SESSION_TIMEOUT` or 2)  run the following command (replacing 3000 ms with any other timeout you want) in the browser console on any DAWSON window you open:

```
window.cerebral.getModel().set(['constants'], {
    ...window.cerebral.getState().constants,
    SESSION_MODAL_TIMEOUT: 3000,
    SESSION_TIMEOUT: 3000,
```


web-client/src/app.tsx
```diff
@@ -271,13 +271,13 @@ const app = {
 
     root.render(
       <Container app={cerebralApp}>
-        {!process.env.CI && (
+        {
```
Mwindo, line 280: This check appears unnecessary--tests are not failing with it removed. It was originally added years ago, apparently to prevent accessibility checks from failing at that time. I think it is best to remove because 1) as said, it appears to be unnecessary and 2) it is a "gotcha" for anyone trying to write automated tests depending on the components within this block (as I was!).

NB: I will removed the unnecessary curly braces that I accidentally left in.

web-client/src/presenter/state.ts
```diff
@@ -1,6 +1,10 @@
 /* eslint-disable max-lines */
 import { FormattedPendingMotionWithWorksheet } from '@web-api/business/useCases/pendingMotion/getPendingMotionDocketEntriesForCurrentJudgeInteractor';
 import { GetCasesByStatusAndByJudgeResponse } from '@web-api/business/useCases/judgeActivityReport/getCaseWorksheetsByJudgeInteractor';
+import {
+  IDLE_LOGOUT_STATES,
+  IdleLogoutStateType,
+} from '@shared/business/entities/EntityConstants';
```
Mwindo, line 7: This file is using getConstants already. However, given a recent discussion about eliminating getConstants, I imported directly from EntityConstants instead.

web-client/src/presenter/sequences/signOutIdleSequence.ts
```diff
@@ -0,0 +1,20 @@
...
+  isLoggedInAction,
+  {
+    // To avoid race conditions, we ignore redundant calls to sign out that arise when
+    // multiple tabs broadcast the idle sign out event.
+    no: [],
+    yes: [
+      setLogoutTypeAction(LOGOUT_BROADCAST_MESSAGES.idleLogout),
+      signOutSequence,
+      setupCurrentPageAction('IdleLogout'),
+    ],
```
Mwindo, line 18: The issue: all tabs/windows will time out at the same time and send a broadcast message to the others to log out as well, which can lead to race issues (e.g., an infinite loading screen).

If someone has a better solution, I'm all ears! The other solutions I considered are either insufficient or more involved:

1) We use `window.document.visibilityState` to only broadcast from visible tabs. However, this doesn't solve the issue since the user might have multiple visible windows. Moreover, I am afraid of something like the following occurring: user starts to log out but something hangs; tab is minimized in the meantime; no broadcast happens, and therefore sign out fails in other tabs.
2) Do something similar, but set some local state to keep track of the last focused tab and only broadcast from it. This might be a good solution but requires more work and seems more prone to error (now we have to keep track of tabs--e.g., what happens when the last active tab is closed?).

Redundant calls to sign out aren't ideal, but they are already happening _without_ being accounted for.

web-client/src/presenter/actions/isLoggedInAction.ts
```diff
@@ -0,0 +1,8 @@
+import { state } from '@web-client/presenter/app.cerebral';
+
+export const isLoggedInAction = ({ get, path }: ActionProps) => {
+  if (get(state.user)) {
+    return path.yes();
+  }
+  return path.no();
```
Mwindo, line 7: Maybe token is better to check?

web-client/src/views/IdleLogout.tsx
```diff
@@ -16,7 +18,14 @@ export const IdleLogout = connect(
           United States Tax Court website for information on court services and
           contact information.
         </p>
-        <Button onClick={() => navigateToLoginSequence()}>Log In</Button>
+        <Button
+          data-testid="idle-logout-login-button"
+          onClick={() => {
+            navigateToLoginSequence();
+          }}
+        >
```
Mwindo: Unneeded now, will remove.

web-client/src/presenter/computeds/showAppTimeoutModalHelper.test.ts
```diff
@@ -18,6 +19,9 @@ describe('showAppTimeoutModalHelper', () => {
   it('does not show the modal due to no user', () => {
     const result = runCompute(showAppTimeoutModalHelper, {
       state: {
+        idleLogoutState: {
+          state: IDLE_LOGOUT_STATES.COUNTDOWN,
+        },
```
Mwindo, line 24: This test was actually not working (it wasn't setting the idleLogoutState), nor was the behavior it was testing in some cases.

web-client/src/app.tsx
```diff
@@ -262,6 +263,11 @@ const app = {
       returnSequencePromise: true,
     });
 
+    // Expose Cerebral for testing
+    if (process.env.NODE_ENV !== 'production') {
+      (window as unknown as ITestableWindow).cerebral = cerebralApp;
+    }
```
Mwindo, line 269: Exposing this so that manual tests (and, if I can get them to work in the future, automated tests) can be done more easily.

cypress/CYPRESS-README.md
```diff
@@ -1,7 +1,7 @@
 
 # Best Practices
 
-In order to write a realiable cypress test suites, there are some best practices we should follow that our outlined in the cypress documentation and also some best practices we have learned from trying to write realiable tests.
+In order to write a reliable cypress test suites, there are some best practices we should follow that our outlined in the cypress documentation and also some best practices we have learned from trying to write reliable tests.
```
TomElliottFlexion, line 4: Thank you

web-client/src/app.tsx
TomElliottFlexion: Well this makes me a little nervous, but if we're not doing it in prod I reckon it's fine

web-client/src/presenter/actions/broadcastIdleStatusActiveAction.test.ts
```diff
@@ -19,7 +20,7 @@ describe('broadcastIdleStatusActiveAction', () => {
     expect(
       applicationContext.getBroadcastGateway().postMessage.mock.calls[0][0],
     ).toMatchObject({
-      subject: 'idleStatusActive',
+      subject: LOGOUT_BROADCAST_MESSAGES.idleStatusActive,
```
TomElliottFlexion, line 23: Nitpick only because of coverage benchmarks; the other branch of the ternary could use a test

TomElliottFlexion (approved):
Thank you, this is a big improvement

JayFlexy (approved):

types/ITestableWindow.d.ts
```diff
@@ -0,0 +1,6 @@
+export interface ITestableWindow {
```
zachrog, line 1: The types/ folder should be considered deprecated as we are trying to avoid global types. Could you move this interface somewhere else?

web-client/src/presenter/presenter.ts
```diff
@@ -1291,7 +1291,10 @@ export const presenterSequences = {
     showThirtyDayNoticeModalSequence as unknown as Function,
   showViewPetitionerCounselModalSequence:
     showViewPetitionerCounselModalSequence as unknown as Function,
+  signOutIdleSequence: signOutIdleSequence as unknown as Function,
```
zachrog, line 1294: Sequences should not be typed with "as unknown as Function" as that is only for legacy functions. 

You can look at web-client/src/presenter/sequences/sortTableSequence.ts for an example on how to type sequences.

web-client/src/presenter/computeds/showAppTimeoutModalHelper.ts
```diff
@@ -7,6 +8,6 @@ export const showAppTimeoutModalHelper = (get: Get): any => {
 
   return {
     currentUser,
-    showModal: modalState === 'COUNTDOWN',
+    showModal: !!currentUser && modalState === IDLE_LOGOUT_STATES.COUNTDOWN,
```
zachrog, line 11: We are making changes to some of the user state. Could you look for something more specific on the user object like the userId? 

Something like !!currentUser?.userId

I am suggesting this because I think we are moving to always having a user object in state, even if a user is not signed in. The user object will have empty values but will always exist

zachrog (changes requested):
This is awesome. You did some great investigation.

Only requesting changes for the types on the sequences.

web-client/src/presenter/computeds/showAppTimeoutModalHelper.ts
Mwindo: Thank you, this makes sense!

Mwindo:
Thank you for the reviews! @zachrog, I've addressed your feedback in https://github.com/ustaxcourt/ef-cms/pull/5176 (where I have automated tests working as well). If we are happy (enough) with the test setup there, I will probably close this PR.

zachrog (approved):

Mwindo:
Closing this and continuing over at https://github.com/ustaxcourt/ef-cms/pull/5176

PR #5198, opened 2024-07-31
10349 bug intermediate branch to test 1722427896

Mwindo:
See https://github.com/ustaxcourt/ef-cms/pull/5176


web-client/src/presenter/presenter.ts
```diff
@@ -1321,7 +1321,10 @@ export const presenterSequences = {
     showThirtyDayNoticeModalSequence as unknown as Function,
   showViewPetitionerCounselModalSequence:
     showViewPetitionerCounselModalSequence as unknown as Function,
+  signOutIdleSequence,
   signOutSequence: signOutSequence as unknown as Function,
+  signOutUserInitiatedSequence:
+    signOutUserInitiatedSequence as unknown as Function,
```
nechama-krigsman, line 1327: Can this sequence be typed instead of `as unknown as Function`?

nechama-krigsman (approved):

web-client/src/presenter/presenter.ts
Mwindo: Ah, yes, thank you! I typed the other sequence but forgot to type this one.

PR #5202, opened 2024-07-31
10323 story intermediate branch to test 1722463985

Mwindo:
Note that Tenille has tested this change in experimental1 ~~(EDIT: minus the indentation fix!)~~, so the behavior seems to be working as intended and does not run into the same issues as https://github.com/ustaxcourt/ef-cms/pull/1408.


TomElliottFlexion (approved):

web-client/src/views/CreateOrder/TextEditor.tsx
```diff
@@ -19,13 +21,40 @@ const ReactQuill = React.lazy(async () => {
...
+};
+
 export const TextEditor = ({
   defaultValue,
   editorDelta,
   updateFormValueSequence,
   updateScreenMetadataSequence,
 }) => {
   const quillEscapeRef = useRef(null);
+  defaultValue = addQuillIndentationClasses(defaultValue);
```
akuny: Instead of reassigning the `defaultValue` that is passed in as an argument, could we instead make a new variable with a descriptive name and then use that new variable in the JSX?

akuny (approved):
Looks good--just one minor comment.

PR #5207, opened 2024-08-02
Devex rebuild cognito

zachrog:
We can no longer destroy and recreate our Cognito due to an incorrect configuration.

Dropping OAuth(which is unused) allows us to rebuild cognito


En-8 (approved):

JayFlexy (approved):

jimlerza:
Added the "Do Not Merge" label because we need to verify with the IRS that this doesn't break eTCS before we merge this.

TomElliottFlexion (approved):

PR #5208, opened 2024-08-02
10396 story intermediate branch to test 1722630752

Mwindo:
**The issue**

In our DOM tree, we were generating something like the following:

```
<ul role="tablist">
    <button>Blah</button>
  </li>
  ...
</ul>
```
The issue here is that button is inherently interactive, and it was being nested within an li that was rendered interactive because of its role as a tab child element. I fixed this by moving the role, aria-control, and aria-selected properties to the button element and signaling that the li element is non-semantic (in other words, just used for display) with role='presentation'. **This is already in test.**

**Related issue**

Originally, we conditionally put h2 headings into our tabs to fix an issue in which we rendered h1 and h3 but not h2 on certain pages (see https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/gh/flexion/ef-cms/8106). However, this solution does not work: everything nested within a role="tab" element is kept out of the accessibility tree (see [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role#all_descendants_are_presentational)), so adding an h2 within the "tab" element was no better than not adding one at all. Unfortunately, surrounding the "tab" element in a heading (rather than nesting it within)--which was my first approach to correcting this--leads to other accessibility violations.

From what I can tell after some thinking and sleuthing, the best way to handle things is to conditionally put a non-visual heading within the tab panel _content_ (i.e., not within the "tab" at all) that is still accessible in the accessibility tree. This solution ensures the following:

1. We have no accessibility violations.
2. As before, we can programmatically tell tabs when they need to render a heading to avoid skipping heading levels.
3. The heading is actually part of the accessibility tree.
4. While separate from the tab button, the heading does not appear on screen, which would look strange (since it would just display the same name as the tab itself).


web-client/src/styles/custom.scss
```diff
@@ -2325,6 +2325,18 @@ button.change-scanner-button {
   text-wrap: balance;
 }
 
+// For content that should be visually hidden but present in the accessibility tree
+.sr-only {
```
Mwindo, line 2329: See https://stackoverflow.com/questions/19758598/what-is-sr-only-in-bootstrap-3

web-client/src/ustc-ui/Tabs/Tabs.tsx
```diff
@@ -62,8 +55,9 @@ const renderTabFactory = ({
 
     return (
       <li {...tabProps}>
+        {' '}
         <button {...buttonProps} data-testid={child.props['data-testid']}>
-          <HeadingElement className="button-text">{title}</HeadingElement>{' '}
+          <span className="button-text">{title}</span> {icon}
```
Mwindo: One side effect of this change: the slight difference in styles between the tabs on, e.g., the advanced search page vs. the message page will be removed. I discussed this and got approval from UX.

web-client/src/styles/tabs.scss
```diff
@@ -34,18 +34,22 @@
     li {
       display: inline-block;
 
+      .tab-header {
+        display: inline;
+      }
```
Mwindo: Oops, this is a testing artifact. I will remove it.

akuny (approved):

PR #5212, opened 2024-08-05
5512 Handle feedback

nechama-krigsman:


cruzjone-flexion (approved):

shared/src/business/entities/cases/ElectronicPetition.ts
```diff
@@ -178,11 +174,6 @@ export class ElectronicPetition extends JoiValidationEntity {
         '*': 'Your Petition file size is empty',
         'number.max': `Your Petition file size is too big. The maximum file size is ${MAX_FILE_SIZE_MB}MB.`,
       }),
-    petitionRedactionAcknowledgement: joi.boolean().when('petitionType', {
```
TomElliottFlexion, line 181: Is there any harm in leaving this in since it'll be added back with 10339? Not a whole lot of point in validating something we don't save here, but also not a whole lot of point in removing it

shared/src/business/entities/EntityConstants.ts
```diff
@@ -1622,6 +1623,11 @@ export const JUDGE_TITLES = [
 ] as const;
 export type JudgeTitle = (typeof JUDGE_TITLES)[number];
 
+export const PETITION_TYPES = {
```
TomElliottFlexion, line 1632: TY!

TomElliottFlexion (approved):

jimlerza (approved):

PR #5213, opened 2024-08-05
Devex: Remove AWS SDK v2 from package.json

jimlerza:
flexion#10297

Removing the last of AWS v2:

- Instances of `S3` and `CloudWatchLogsClient` in `env-destroy-utils`
- Instances of `DynamoDBClient` and `DynamoDBStreamsClient` in `app-local`


jtdevos (approved):
Only note is that it's preferred to do `git mv` when renaming files because git (or vscode) is sometimes not smart enough to detect a rename and ends up wiping out the revision history for a file.

jimlerza:
> Only note is that it's preferred to do `git mv` when renaming files because git (or vscode) is sometimes not smart enough to detect a rename and ends up wiping out the revision history for a file.

I always do. In my experience, if the total diff is over a certain percentage of the file, the diff viewer renders it like a delete & create. But if you pull up the renamed file, the untouched lines are still attributed to their original authors.

TomElliottFlexion (approved):

En-8 (approved):

package.json
```diff
@@ -48,7 +49,6 @@
     "@sparticuz/chromium": "123.0.1",
     "@uswds/uswds": "3.7.1",
     "aws-lambda": "1.0.7",
-    "aws-sdk": "2.1663.0",
```
En-8, line 51: 🎉 🎉 🎉 🎉 🎉 🎉

PR #5215, opened 2024-08-05
10102 story to test

akuny:
This PR addresses feedback from initial stakeholder review. The bulk of the 70 files affected by this PR were changed to align naming conventions used in the codebase with business terminology (that is, refer to the specific type of order this story adds as "Status Report Order" instead of "Status Report Order Response".

The following files were refactored to make the similarities in business logic between the two more apparent:

- `web-client/src/presenter/computeds/draftDocumentViewerHelper.ts`
- `web-client/src/presenter/computeds/messageDocumentHelper.ts`


web-client/src/presenter/computeds/messageDocumentHelper.ts
```diff
@@ -47,6 +48,8 @@ export const messageDocumentHelper = (
       useArchived: true,
     }) || {};
 
+  console.log('caseDocument', caseDocument);
```
nechama-krigsman: Can we remove the logs in this file?

nechama-krigsman: Never mind...i was looking at a specific commit

nechama-krigsman (approved):

PR #5216, opened 2024-08-06
Remove deprecated lines from husky pre-commit

Mwindo:
We currently get a warning from husky when running pre-commit hooks:

Screenshot 2024-08-06 at 10 10 08 AM

According to the [documentation ("How to Migrate")](https://github.com/typicode/husky/releases/tag/v9.0.1), we should remove two lines from our pre-commit hook file. That is what this PR does. Afterwards, husky runs and the warning goes away:

Screenshot 2024-08-06 at 10 14 26 AM


jimlerza:
I think we probably need to keep the shebang, and maybe even set it to `/bin/bash`?

I believe we need shebangs in all of our scripts, because without them, we'll have discrepancies when we run things locally (scripts without a shebang will run in `zsh` on macOS) and inside our docker containers (scripts without a shebang will run in `bash` in debian).

Mwindo:
> I think we probably need to keep the shebang, and maybe even set it to /bin/bash?

According to the documentation linked above:

> Note: sh will be used to run hooks, even if a shebang is set.

So I think the shebang won't do anything anyway. If this turns out to be a problem, we will need to figure out a different solution!

jimlerza (approved):

zachrog (approved):

nechama-krigsman (approved):

PR #5217, opened 2024-08-06
10441-story: Adding Judge Way and updating Judge Landy

Mwindo:
Usage : 
`npx ts-node --transpile-only scripts/run-once-scripts/add-new-judge.ts` 

`npx ts-node --transpile-only scripts/run-once-scripts/promote-special-trial-judge.ts stjudge.landy@example.com judge.landy@example.com`

Modeled off of https://github.com/ustaxcourt/ef-cms/pull/3953


web-client/src/business/chambers/getJudgesChambers.ts
```diff
@@ -203,6 +205,12 @@ const JUDGES_CHAMBERS: Record<string, JudgeInfo> = omitBy(
       phoneNumber: '(202) 521-0778',
       section: 'vasquezsChambers',
     },
+    WAYS_CHAMBERS_SECTION: {
+      judgeFullName: 'Kashi Way',
+      label: 'Way’s Chambers',
+      phoneNumber: '', // TODO
```
Mwindo, line 211: This is unknown for the time being and should be filled in once we have the information.

scripts/user/add-user.ts
```diff
@@ -100,6 +100,7 @@ const checkParams = params => {
           'torosChambers',
           'urdasChambers',
           'vasquezsChambers',
+          'waysChambers',
```
Mwindo, line 103: Note to self and future devs: this isn't necessary for adding the judge, which is done through Zendesk automation and python scripts that call the `/users` DAWSON endpoint to eventually hit the `createUserLambda`. This is for ensuring future added users can be validated appropriately via the new chamber section.

TomElliottFlexion: Update: turns out we haven't been using the Zendesk automation for judges specifically; it has been done manually

TomElliottFlexion (approved):

cruzjone-flexion (approved):

jimlerza (approved):

PR #5221, opened 2024-08-07
10396-story: Fix a11y nested-interactive issue

Mwindo:
**The issue**

In our DOM tree, we were generating something like the following:

```
<ul role="tablist">
    <button>Blah</button>
  </li>
  ...
</ul>
```
The issue here is that button is inherently interactive, and it was being nested within an li that was rendered interactive because of its role as a tab child element. I fixed this by moving the role, aria-control, and aria-selected properties to the button element and signaling that the li element is non-semantic (in other words, just used for display) with role='presentation'. **This is already in test.**

**Related issue**

Originally, we conditionally put h2 headings into our tabs to fix an issue in which we rendered h1 and h3 but not h2 on certain pages (see https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/gh/flexion/ef-cms/8106). However, this solution does not work: everything nested within a role="tab" element is kept out of the accessibility tree (see [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role#all_descendants_are_presentational)), so adding an h2 within the "tab" element was no better than not adding one at all. Unfortunately, surrounding the "tab" element in a heading (rather than nesting it within)--which was my first approach to correcting this--leads to other accessibility violations.

From what I can tell after some thinking and sleuthing, the best way to handle things is to conditionally put a non-visual heading within the tab panel _content_ (i.e., not within the "tab" at all) that is still accessible in the accessibility tree. This solution ensures the following:

1. We have no accessibility violations.
2. As before, we can programmatically tell tabs when they need to render a heading to avoid skipping heading levels.
3. The heading is actually part of the accessibility tree.
4. While separate from the tab button, the heading does not appear on screen, which would look strange (since it would just display the same name as the tab itself).

### DIFF
--- a/cypress/local-only/support/generalCommands/checkA11y.ts
+++ b/cypress/local-only/support/generalCommands/checkA11y.ts
@@ -9,9 +9,6 @@ export function checkA11y(): void {
     {
       includedImpacts: impactLevel,
       retries: 3,
-      rules: {
-        'nested-interactive': { enabled: false }, // https://github.com/flexion/ef-cms/issues/10396
-      },
     },
     terminalLog,
   );
--- a/cypress/local-only/support/pages/create-paper-petition.ts
+++ b/cypress/local-only/support/pages/create-paper-petition.ts
@@ -8,7 +8,7 @@ export const createPaperPetition = () => {
   navigateToDocumentQC('petitionsclerk');
 
   getCreateACaseButton().click();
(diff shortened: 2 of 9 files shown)


jimlerza (approved):

jtdevos (approved):
Good stuff - and definitely appreciate the walk-through of the approaches and thought-process.

PR #5223, opened 2024-08-07
using pqueue instead of chunk+Promise.all() for setTrialSessionCalendarInteractor

jtdevos:


web-api/src/business/useCases/trialSessions/setTrialSessionCalendarInteractor.ts
```diff
@@ -144,10 +145,8 @@ export const setTrialSessionCalendarInteractor = async (
     // Story: 10422
```
akuny, line 145: If this comment proved useful, it may be worth updating it to include some information about why we're using the `p-queue` module here as opposed to the previous approach.

akuny (approved):

jimlerza (approved):

PR #5234, opened 2024-08-09
Dep updates 08/05/24

TomElliottFlexion:
- Removed jwkToPem in favor of node crypto
- **New ECR image**
- [Successful migration ](https://app.circleci.com/pipelines/github/flexion/ef-cms/55655/workflows/686ec600-4582-4fb0-a962-34d1978106d8)


.eslintrc.js
```diff
@@ -58,7 +58,10 @@ module.exports = {
     {
       files: ['**/*.ts', '**/*.tsx'],
       rules: {
-        '@typescript-eslint/no-unused-vars': 'error',
+        '@typescript-eslint/no-unused-vars': [
+          'error',
+          { caughtErrors: 'none' },
```
Mwindo, line 63: What is the reasoning behind this change?

package.json
```diff
@@ -69,18 +69,17 @@
     "image-blob-reduce": "4.1.0",
     "joi": "17.13.3",
     "jsonwebtoken": "9.0.2",
-    "jwk-to-pem": "2.0.5",
```
Mwindo, line 72: 🚀

web-api/src/lambdas/cognitoAuthorizer/createAuthorizer.ts
```diff
@@ -1,13 +1,15 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
Mwindo: Is this line necessary?

Mwindo (approved):
I left a couple of comments on eslint lines, but otherwise this looks good to me!

.eslintrc.js
TomElliottFlexion: An update to the default rules in `@typescript-eslint/no-unused-vars` caught about 50 instances in our application of `catch (e)` or `catch(err)` and treated them as unused vars. This setting undoes that

web-api/src/lambdas/cognitoAuthorizer/createAuthorizer.ts
TomElliottFlexion: Not anymore! Good catch

.eslintrc.js
Mwindo: Could we instead replace the instances of `catch (e) { ... }` and `catch(err) { ... }` with `catch { ... }` (i.e., `catch` with no parameter)? I feel like that is cleaner than disabling this eslint rule completely. Maybe others have opinions here, too.

TomElliottFlexion: To clarify, this setting is confusingly named: `caughtErrors: none` refers to this specific try-catch pattern, and doesn't disable the rule entirely. This just changes the defaults back to what we had originally instituted. I'm certainly not opposed to refactoring away from the `catch (e)` pattern, but I don't think this eslint rule defaults update warrants it

jimlerza (approved):

.eslintrc.js
Mwindo: Gotcha!

PR #5235, opened 2024-08-09
10349 Bug: Idle Logout Issues, Part 2 (Interactions with "DAWSON has been updated")

Mwindo:
After my previous [PR to test](https://github.com/ustaxcourt/ef-cms/pull/5204), Tenille observed some weird client-side behavior in test. This seems to have occurred because there was a deploy to test during her idle activity. When a user idles for too long, we will still attempt to sign the user out even if they need to refresh, but this attempt will fail and can produce weird behavior on the front end (for example, clearing the modals so everything looks ok and then getting an infinite loading bar).

This PR addresses that issue and related issues by enforcing the following behaviors:

1. We broadcast the "you need to close and refresh" event to all tabs to keep them consistent.
2. We do not track idle behavior or sign the user out if the client needs to be refreshed.
3. We do not continue to send refresh token requests if the client needs to be refreshed.


web-client/src/presenter/sequences/signOutIdleSequence.ts
```diff
@@ -12,9 +13,17 @@ export const signOutIdleSequence = [
...
+      {
+        clientDoesNotNeedToRefresh: [
+          setLogoutTypeAction(BROADCAST_MESSAGES.idleLogout),
+          signOutSequence,
+          setupCurrentPageAction('IdleLogout'),
+        ],
+        // If the client needs to refresh, the sign-out will fail,
+        // and this can lead to inconsistent front-end behavior.
+        clientNeedsToRefresh: [],
+      },
```
Mwindo, line 26: Although we clear the idle activity interval in IdleActivityMonitor (see comment below), there is still a brief period in which this could be called before the interval has been cleared.

web-client/src/views/IdleActivityMonitor.tsx
```diff
@@ -49,6 +51,11 @@ export const IdleActivityMonitor = connect(
     }, []);
 
     useEffect(() => {
+      // The user needs to refresh, so stop tracking idle timeout
+      if (clientNeedsToRefresh) {
+        return;
+      }
```
Mwindo, line 57: This will prevent us from showing the "Are you still there?" modal and trying to log out.

En-8 (approved):

En-8 (approved):

web-client/src/presenter/actions/handleIdleLogoutAction.ts
```diff
@@ -7,20 +7,25 @@ export const handleIdleLogoutAction = ({ get, path, store }: ActionProps) => {
...
-    userIsLoggedIn &&
-    !isUploading &&
-    idleLogoutState.state === IDLE_LOGOUT_STATES.INITIAL
-  ) {
+  // Short-circuit here to prevent showing the "Are you still there?" modal, etc. when inappropriate
+  if (!userIsLoggedIn || clientNeedsToRefresh) {
+    return path.continue();
+  }
+
+  if (!isUploading && idleLogoutState.state === IDLE_LOGOUT_STATES.INITIAL) {
```
Mwindo, line 17: Looking at the short history of this file, it appears this `!userIsLoggedIn` check only being in the initial `if` block was an unintentional artifact of a refactor. Since this was an easy mistake to make--the nested `if/else if` makes this action's logic difficult to understand at a glance--I've pulled each condition out into a separate `if/return` block.

PR #5236, opened 2024-08-10
10458-devex: fix favicon pattern matching

Mwindo:
The favicon is displaying inconsistently. I suspect this has more to do with caching than with browsers: we were not correctly copying favicon files over.

Before this change, here is the build output in dist:

Screenshot 2024-08-10 at 10 46 52 AM

After this change:

Screenshot 2024-08-10 at 10 46 22 AM

Doing a hard refresh after the change got the favicon displaying in my browser.


TomElliottFlexion (approved):

TomElliottFlexion (approved):
Nice catch!

jimlerza (approved):

PR #5237, opened 2024-08-10
10458 devex: fix favicon pattern matching

Mwindo:
The favicon is displaying inconsistently. I suspect this has more to do with caching than with browsers: we were not correctly copying favicon files over.

Before this change, here is the (truncated) build output in dist (notice no favicons):

Screenshot 2024-08-10 at 10 46 52 AM

After this change:

Screenshot 2024-08-10 at 10 46 22 AM

Doing a hard refresh after the change got the favicon displaying in my browser.


esbuildHelper.mjs
```diff
@@ -136,7 +136,7 @@ export default async function ({
       copy({
         assets: [
           {
-            from: ['web-client/src/favicons'],
+            from: ['web-client/src/favicons/**/*'],
```
Mwindo, line 139: Because this is a directory, we need to specify pattern matching for files.

En-8 (approved):

PR #5239, opened 2024-08-10
10429 Design Debt: Align Iteration Field Error 

Mwindo:


web-client/src/views/FileDocument/NonstandardForm.tsx
```diff
@@ -189,136 +260,8 @@ export const NonstandardForm = connect(
         )}
         {helper[level].ordinalField && (
           <>
-            <NonMobile>
-              <FormGroup className="grid-row grid-gap">
```
Mwindo, line 193: This FormGroup--specifically the grid-row class--was the issue. (I did some testing without it being present at all, and it seemed to be perfectly fine. So I just nixed it.) I also refactored to avoid more or less duplicating 30-some lines of code.

pixiwyn (approved):

TomElliottFlexion (approved):

PR #5240, opened 2024-08-12
10102 Status Report Order Updates (To Test)

Mwindo:
This PR includes the following changes:

1. The Clear All button now works in the Status Report Order form when the user reloads the order.
2. The language around lead cases has been modified.

This _does not_ include the change to automatically attach to the message thread a Status Report Order created directly from a Status Report message attachment.


web-client/src/presenter/actions/StatusReportOrder/clearStatusReportOrderFormAction.ts
```diff
@@ -1,22 +1,15 @@
 import { STATUS_REPORT_ORDER_OPTIONS } from '@shared/business/entities/EntityConstants';
-import { isEmpty } from 'lodash';
 import { state } from '@web-client/presenter/app.cerebral';
 
-export const clearStatusReportOrderFormAction = ({
-  get,
-  store,
-}: ActionProps) => {
-  const documentToEdit = get(state.documentToEdit);
-  if (isEmpty(documentToEdit)) {
```
Mwindo, line 10: This check was there to prevent loading an existing Status Report Order from clearing the existing data from the form. However, we no longer need this check because we have since switched to using separate routes for creating and editing a Status Report Order.

Mwindo:
Closing. See https://github.com/ustaxcourt/ef-cms/pull/5247 instead.

PR #5241, opened 2024-08-12
Remove duplicate, outdated fileCourtIssuedOrderInteractor files

Mwindo:
After a renaming event, we currently have two `fileCourtIssuedOrderInteractor.ts` files and associated test files, one in `useCases/courtIssuedOrder`, the other in `useCases/courtIssuedDocument`. Based on their histories (the former [here](https://github.com/ustaxcourt/ef-cms/commits/staging/web-api/src/business/useCases/courtIssuedOrder/fileCourtIssuedOrderInteractor.ts), the latter [here](https://github.com/ustaxcourt/ef-cms/commits/staging/web-api/src/business/useCases/courtIssuedDocument/fileCourtIssuedOrderInteractor.ts)), it looks like the former is the most recent version that we want to keep. Likewise--and thankfully--I cannot find any usages of the latter.


TomElliottFlexion (approved):

Mwindo:
This has been in test for a couple weeks now. It should be safe to merge.

jimlerza (approved):

PR #5242, opened 2024-08-12
10291: Remove close link and cancel upload button from FileUploadProgressModal

En-8:


web-client/src/views/FileUploadStatusModal.tsx
```diff
@@ -42,19 +39,6 @@ export const FileUploadStatusModal = connect(
             />
           </div>
           {helper.isHavingSystemIssues && <div>Still processing...</div>}
-          {helper.isCancelable && (
-            <div className="cancel">
-              <Button
-                link
-                onClick={e => {
-                  e.stopPropagation();
-                  cancelUploadSequence();
```
Mwindo, line 51: I think we can get rid of this sequence, the underlying action, and the related test since it doesn't appear to be used anywhere else.

web-client/src/views/DocketEntryQc.tsx
```diff
@@ -125,7 +124,6 @@ export const DocketEntryQc = connect(
             onConfirmSequence={completeDocketEntryQCAndSendMessageSequence}
           />
         )}
-        {showModal === 'FileUploadStatusModal' && <FileUploadStatusModal />}
```
Mwindo, line 128: Is this being removed because it isn't pertinent to Docket Entry QC? (Sorry if this is an obvious question!)

En-8: I believe so, yeah. As far as I can tell, the only place the modal state is set to this value is within the "openFileUploadStatusModalAction", and that doesn't appear to be used in any of the sequences on which this component depends.

web-client/src/views/FileUploadStatusModal.tsx
En-8: Nice catch, removal incoming 👍🏻

web-client/src/views/DocketEntryQc.tsx
Mwindo: Cool, thank you!

Mwindo (approved):
Looks good to me--thank you!

PR #5246, opened 2024-08-13
10339 story

TomElliottFlexion:


shared/src/business/entities/contacts/ContactFactoryUpdated.ts
```diff
@@ -8,12 +8,14 @@ import { SpouseContact } from '@shared/business/entities/contacts/SpouseContact'
...
   contactInfoPrimary,
   contactInfoSecondary,
+  filingType,
   hasSpouseConsent,
   partyType,
   petitionType,
 }: {
   contactInfoPrimary: {};
   contactInfoSecondary: {};
+  filingType: string;
```
zachrog: Type can be updated to FilingType

web-client/src/presenter/computeds/updatedFilePetitionHelper.ts
```diff
@@ -36,22 +39,50 @@ export const updatedFilePetitionHelper = (
...
     filingOptions,
+    isPetitioner,
+    isPractitioner,
     otherContactNameLabel,
     showContactInformationForOtherPartyType:
       getShowContactInformationForOtherPartyType(partyType, PARTY_TYPES),
   };
 };
 
+function formatFilingTypes(filingOptions) {
```
zachrog: Add FilingType type here I think?

cruzjone-flexion (approved):

cypress/local-only/tests/integration/fileAPetitionUpdated/file-a-petition-step-1-myself.cy.ts
```diff
@@ -25,7 +25,7 @@ describe('File a petition', () => {
       });
     });
 
-    it('should display a validaiton error message if user does not select filing type', () => {
+    it('should display a validation error message if user does not select filing type', () => {
```
TomElliottFlexion, line 28: TY!!

web-client/src/presenter/sequences/gotoDashboardSequence.ts
```diff
@@ -115,7 +115,7 @@ export const gotoDashboardSequence = [
           setDefaultCaseTypeToDisplayAction,
           getOpenAndClosedCasesForUserAction,
           setCasesAction,
-          setupCurrentPageAction('DashboardPractitioner'),
+          setupCurrentPageAction('DashboardPetitioner'),
```
TomElliottFlexion: Out of curiosity, why did this change? The tests that reference it are still called "practitioner"

web-client/src/styles/custom.scss
```diff
@@ -2395,3 +2395,16 @@ button.change-scanner-button {
 .no-wrap-white-space {
   white-space: normal;
 }
+
+.align-items-center {
+  align-items: center;
```
TomElliottFlexion, line 2400: 👍👍👍

web-client/src/presenter/sequences/gotoDashboardSequence.ts
TomElliottFlexion: After reading further, I see that we now only have one dashboard for petitioners and practitioners, nice! How do you feel about giving the page a more generic name?

TomElliottFlexion (commented):
Nice job Nechama!

web-client/src/presenter/sequences/gotoDashboardSequence.ts
nechama-krigsman: That is a good idea. I'll update. Thanks!

nechama-krigsman (approved):

web-client/src/presenter/computeds/updatedFilePetitionHelper.ts
```diff
@@ -20,9 +21,14 @@ interface IOtherContactNameLabel {
 }
 
 type UpdatedFilePetitionHelper = {
-  filingOptions: string[];
+  filingOptions: { label: string; value: string }[];
+  getLetterByIndex: (index: number) => string;
+  isPetitioner: boolean;
+  isPractitioner: boolean;
```
jtdevos, line 27: Is it possible for for both `isPetitioner` and `isPractitioner` to both be true (or both be false)?  If not, would it be better to just have one of these vars?

web-client/src/presenter/computeds/dashboardExternalHelper.ts
```diff
@@ -1,44 +1,55 @@
...
     showFilingFee,
     showPetitionWelcomePage: cases.length === 0,
     showStartButton,
+    welcomeMessage,
+    welcomeMessageTitle,
   };
 };
+
+const messages = {
+  [ROLES.privatePractitioner]: {
```
jtdevos, line 44: TIL about computed property names.  Neat!

web-client/src/presenter/computeds/updatedFilePetitionHelper.ts
nechama-krigsman: It’s not possible for both isPetitioner and isPractitioner to be true or both false at the same time, so we could simplify this by having just one property and negating it when necessary. Originally, I used both properties for readability, but I do see that we are negating in some places. Are you okay with this change going in a followup PR along with the cleanup?

jtdevos: Thanks Nechama, yeah I think a followup PR is fine.

jtdevos: Oh and I do like the readability argument, my only concern is us winding in a situation where we wind up in an invalid state.  If practical, I think it would be fine to have both props, but have one be a computed prop that is the negation of the other.

jtdevos (approved):

jimlerza (approved):

PR #5247, opened 2024-08-13
10102 Story: Status Report Order

Mwindo:
This PR includes the following changes:

1. The Clear All button now works in the Status Report Order form when the user reloads the order.
2. The language around lead cases has been modified.
3. Status Report Orders created from the messages view attach the order to the message thread (without duplicates).


cypress/local-only/tests/integration/statusReportOrder/edit-status-report-order.cy.ts
```diff
@@ -165,6 +165,10 @@ describe('edit status report order', () => {
       cy.contains('a', messages.testStatusReportOrderSigned.name).click();
       cy.get('[data-testid="edit-signed-document-button"]').click();
       cy.get('[data-testid="modal-button-confirm"]').click();
+      cy.get('#status-report-due-date-picker').clear();
+      cy.get('#status-report-due-date-picker').type(
+        formatNow(FORMATS.MMDDYYYY),
+      );
```
Mwindo, line 171: This test was secretly relying on other tests to run first and update the date. These lines fix that.

web-client/src/presenter/actions/StatusReportOrder/clearStatusReportOrderFormAction.ts
```diff
@@ -1,22 +1,15 @@
 import { STATUS_REPORT_ORDER_OPTIONS } from '@shared/business/entities/EntityConstants';
-import { isEmpty } from 'lodash';
 import { state } from '@web-client/presenter/app.cerebral';
 
-export const clearStatusReportOrderFormAction = ({
-  get,
-  store,
-}: ActionProps) => {
-  const documentToEdit = get(state.documentToEdit);
-  if (isEmpty(documentToEdit)) {
```
Mwindo, line 10: This check was there to prevent loading an existing Status Report Order from clearing the existing data from the form. However, we no longer need this check because we have since switched to using separate routes for creating and editing a Status Report Order.

pixiwyn (approved):

PR #5248, opened 2024-08-13
Expand untouched file check to check ALL files for increased type error count

En-8:
Expanding the TS untouched file check script to report on ALL files that increased in type error count in a PR relative to staging, instead of only untouched files.


Mwindo (approved):
Thank you!

TomElliottFlexion (approved):

jimlerza (approved):

PR #5251, opened 2024-08-14
10432: Performance Measurement - TO TEST

cruzjone-flexion:


scripts/dynamo/setup-sequence-performance-minimum-time.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+# creates the entry for minimum time for sequence performance logging in the dynamo deploy table
```
Mwindo, line 3: I haven't looked over things carefully, and I'm not sure how likely this is, but one thing we may want to be mindful of when analyzing the data is that a sequence could be running fast 95% of the time (and therefore not logged) but very slow 5% of the time (and therefore logged), skewing the apparent distribution.

cruzjone-flexion: Actually in the process of refactoring to use a black list instead.

PR #5252, opened 2024-08-14
10353 Bug: Duplicate Cognito Accounts

Mwindo:
Cognito will throw an error with the message `Alias entry already exists for a different username` when we attempt to create two accounts with the same email nearly simultaneously. Unfortunately, Cognito still creates the duplicate account. We sometimes run into the error when a user quickly hits submit multiple times in `CreatePetitionerAccount/CreatePetitionerAccountForm.tsx`.

This PR fixes the issue on the front end by adding a debounce and deactivating the signup button. For all practical purposes, that should solve the issue.

I thought about how we might enforce no duplicate Cognito users on the back end, too, which would be ideal. I think the solution would go something like this: On getting the duplicate error message, we delete the duplicate account. This is easier said than done: because the request "fails," we don't get the duplicate Cognito user id with which to key off of. Instead, then, we would need to kick off some asynchronous task to get all users associated with the email and delete those without `custom:userId` or something (since our custom attributes--which we set after successfully signing up a user in Cognito--won't have been set for the failed Cognito request).

Frankly, I don't think it is worth it. I am happy enough with the front-end fix and pointing out that the issue is, after all, due to a bug with Cognito. That said, if reviewers think that this back-end solution is worthwhile, I can set it up.


cypress/deployed-and-local/integration/authentication/petitioner-account-creation.cy.ts
```diff
@@ -87,11 +87,37 @@ describe('Petitioner Account Creation', () => {
   });
 
   describe('Create Petitioner Account and login', () => {
-    const TEST_EMAIL = `cypress_test_account+success_${GUID}@example.com`;
     const TEST_NAME = 'Cypress Test';
     const TEST_PASSWORD = generatePassword(VALID_PASSWORD_CONFIG);
 
+    it('should prevent multiple submissions', () => {
```
Mwindo, line 93: This fails (at least locally) without my changes in place.

cypress/deployed-and-local/integration/authentication/petitioner-account-creation.cy.ts
```diff
@@ -87,11 +87,37 @@ describe('Petitioner Account Creation', () => {
...
+      );
+      cy.get(
+        '[data-testid="petitioner-account-creation-confirm-password"]',
+      ).type(TEST_PASSWORD);
+      cy.intercept('POST', '/auth/account/create').as('accountCreationRequest');
+
+      // eslint-disable-next-line cypress/unsafe-to-chain-command
+      cy.get('[data-testid="petitioner-account-creation-submit-button"]')
+        .click({ force: true })
+        .click({ force: true });
```
Mwindo, line 113: I've confirmed that this test fails as expected without the changes made in this PR.

web-api/src/business/useCases/auth/signUpUserInteractor.ts
```diff
@@ -28,6 +28,11 @@ export const signUpUserInteractor = async (
       email: user.email,
     });
 
+  // Note that this check can fail to catch two (nearly) simultaneous requests,
+  // and Cognito can therefore create accounts with the same email.
+  // (See https://stackoverflow.com/questions/50730759/user-pool-allows-two-users-with-same-email-despite-configuration)
+  // On the frontend, we can prevent multiple submissions; if we wanted to enforce non-duplicates on the back-end,
```
Mwindo, line 34: It bothers me that I used both `frontend` (unhyphenated) and `back-end` (hyphenated). I'd fix this before merging to staging, if this comment is even still around.

jimlerza (approved):

web-client/src/views/CreatePetitionerAccount/CreatePetitionerAccountForm.tsx
```diff
@@ -30,6 +35,18 @@ export const CreatePetitionerAccountForm = connect(
...
+    // Re-enabled submit button if submission was unsuccessful
+    useEffect(() => {
+      if (alertError || alertWarning) {
+        setSubmitDisabled(false);
+      }
+    }, [alertError, alertWarning]);
+
+    const submitFunction = debounce(() => {
+      submitCreatePetitionerAccountFormSequence();
+    }, 500);
```
jimlerza: Just curious - how'd we land on 500ms for this? If I understand correctly, this is the amount of time that we'll wait before issuing the click? Is this delay noticeable to end users?

jimlerza (approved):

web-client/src/views/CreatePetitionerAccount/CreatePetitionerAccountForm.tsx
Mwindo: Good question. My research online (read: Google/StackOverflow) suggested that 500 ms is a good starting point (see, e.g., https://ux.stackexchange.com/questions/95336/how-long-should-the-debounce-timeout-be), if on the high end; I chose a higher value to make sure there is virtually no chance to trigger the error. We can probably reduce this if need be.

That said, the user _does_ get "immediate" feedback: the button deactivates as soon as they click. So the 500 ms delay is lumped into the apparent network submission time, if that makes sense, so it shouldn't be too noticeable.

jtdevos: My only nit is a suggestion to replace the magic number w/ a constant e.g. DEBOUNCE_TIME_MS or somesuch.

jtdevos (changes requested):
left a comment re: magic numbers - let me know if you disagree

Mwindo:
> left a comment re: magic numbers - let me know if you disagree

Sure, I can pull this out into a constant.

jtdevos (approved):

PR #5254, opened 2024-08-15
10432: Performance Measurement - TO TEST

cruzjone-flexion:
This PR might not get merged.
It is working but using CloudWatch Metrics will be too expensive.


web-api/src/business/useCases/system/logUserPerformanceDataInteractor.test.ts
```diff
@@ -0,0 +1,65 @@
...
+    userId: 'SOME_TEST_USER_ID',
+  };
+
+  beforeEach(() => {
+    applicationContext
+      .getPersistenceGateway()
+      .saveSystemPerformanceData.mockImplementation(() => {});
+  });
+
+  it('should call persistence method with correct data', async () => {
```
Mwindo: Perhaps `it('should not allow unauthorized call to log data')`

web-client/src/presenter/actions/performanceMeasurementEndAction.ts
```diff
@@ -0,0 +1,34 @@
...
+    actionPerformanceArray: { actionName: string; duration: number }[];
+  } = {
+    actionPerformanceArray,
+    duration: durationInSeconds,
+    sequenceName,
+  };
+
+  applicationContext
+    .getUseCases()
+    .logUserPerformanceDataInteractor(applicationContext, RESULTS);
```
Mwindo, line 33: Just glancing over things, so I might be missing something--what if this call fails? Especially since we are using this "everywhere," I'm thinking we probably want to have better exception handling so the app works even if our performance monitoring doesn't.

cypress/deployed-and-local/integration/documentQC/complete-docket-qc.cy.ts
```diff
@@ -183,9 +183,14 @@ describe('Document QC Complete', () => {
       });
 
       cy.visit('/document-qc/my/inbox');
-      cy.get(
-        `[data-testid="message-queue-docket-number-${unservedDocketNumber}"]`,
-      ).should('be.visible');
+
+      retry(() => {
+        cy.reload(true);
```
Mwindo, line 188: Why the reload? (Just making sure I understand!)

scripts/env/environments/00-common
```diff
@@ -46,6 +46,7 @@ IRS_CLIENT_ID=$(aws cognito-idp list-user-pool-clients \
   echo "DAWSON_ENV='${ENV}'"
   echo "DYNAMODB_TABLE_NAME='${SOURCE_TABLE}'"
   echo "ELASTICSEARCH_ENDPOINT='${ELASTICSEARCH_ENDPOINT}'"
+  echo "ELASTICSEARCH_INFO_ENDPOINT='${ELASTICSEARCH_INFO_ENDPOINT}'"
```
Mwindo, line 49: Nitpick, and probably not worth doing if it involves changing infrastructure, but I wonder if we should give this a clearer name? `Info` can mean just about anything.

web-client/src/presenter/actions/performanceMeasurementEndAction.ts
Mwindo: Nevermind, ignore.

web-api/src/app.ts
```diff
@@ -1072,6 +1073,16 @@ app.delete(
   app.put('/work-items', lambdaWrapper(assignWorkItemsLambda));
 }
 
+/**
+ * logging
+ */
+{
+  app.post(
```
Mwindo, line 1080: Could this contribute to high server load if we end up logging a lot?

web-client/src/presenter/presenter.ts
```diff
@@ -1646,6 +1649,59 @@ export const presenterSequences = {
     validateUserContactSequence as unknown as Function,
 };
 
+function wrapActionsWithPerformanceLogic(structure) {
```
Mwindo, line 1652: I'm not sure I fully understand the behavior of nested Sequences. For instance, in the following screen shot (going to case 107-19 and editing an existing Status Report Order), I get two different time values for the same action `'getStatusReportOrderPdfUrlAction2'`, which should have only been run once:

Screenshot 2024-08-15 at 3 30 06 PM

I'm not sure how much this will skew other sequences.

web-client/src/presenter/presenter.ts
```diff
@@ -1661,7 +1717,33 @@ export const presenter = {
     [ActionError, setCurrentPageErrorSequence], // generic error handler
   ],
   providers: {} as { applicationContext: ClientApplicationContext; router: {} },
-  sequences: presenterSequences,
+  sequences: Object.entries(presenterSequences).reduce(
+    (acc, [sequenceName, sequence]) => {
+      const SYSTEM_PERFORMANCE_BLACKLIST: string[] = [
```
Mwindo, line 1722: 👍

Mwindo (approved):
This is super cool! I left a few comments. I would want to understand things a little more clearly before merging to a non-test environment (mostly around server load implications and ~~making sure nested sequences are giving accurate data~~ EDIT: Nevermind! I thought we might be logging something wrong, but it turns out the logging caught something that _was_ wrong!), but I think it's good to test.

web-client/src/presenter/presenter.ts
Mwindo: Actually, I'm double-checking we aren't running this twice on accident. Stay tuned!

Mwindo: We are! It looks like the logging might help us find instances where we are making redundant calls--it's already helped once!

PR #5256, opened 2024-08-15
10464 Bug: Invalid documents in scenario dropdowns

Mwindo:
We do not want to show unserved, undocketed, or stricken documents from "What Document are You Filing?" dropdowns. This PR enforces these checks and updates tests accordingly.

### DIFF
--- a/web-client/src/presenter/computeds/selectDocumentTypeHelper.test.ts
+++ b/web-client/src/presenter/computeds/selectDocumentTypeHelper.test.ts
@@ -3,9 +3,10 @@ import {
   MAX_TITLE_LENGTH,
   getOptionsForCategory,
   getOrdinalValuesForUploadIteration,
-  getPreviouslyFiledDocuments,
+  getValidPreviouslyFiledDocuments,
 } from './selectDocumentTypeHelper';
 import { MOCK_CASE } from '../../../../shared/src/test/mockCase';
+import { MOCK_DOCUMENTS } from '@shared/test/mockDocketEntry';
 import { applicationContextForClient as applicationContext } from '@web-client/test/createClientTestApplicationContext';
 
 describe('selectDocumentTypeHelper', () => {
@@ -48,16 +49,28 @@ describe('selectDocumentTypeHelper', () => {
         scenario: 'Nonstandard A',
       };
 
+      const mockDocuments = MOCK_DOCUMENTS.map(d => {
+        return {
+          ...d,
+          isOnDocketRecord: true,
+          isStricken: false,
+          servedAt: '2019-08-25T05:00:00.000Z',
+        };
+      });
+
       const result = getOptionsForCategory({
         applicationContext,
-        caseDetail: MOCK_CASE,
+        caseDetail: {
+          ...MOCK_CASE,
+          docketEntries: mockDocuments,
+        },
         categoryInformation: mockCategoryInformation,
         selectedDocketEntryId: mockSelectedDocketEntryId,
       });
 
-      expect(result).toEqual({
+      expect(result).toMatchObject({
         previousDocumentSelectLabel: 'Which document are you objecting to?',
-        previouslyFiledDocuments: MOCK_CASE.docketEntries.filter(
+        previouslyFiledDocuments: mockDocuments.filter(
           d => d.eventCode !== INITIAL_DOCUMENT_TYPES.stin.eventCode,
         ),
         showNonstandardForm: true,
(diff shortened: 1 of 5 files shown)


web-client/src/presenter/computeds/selectDocumentTypeHelper.test.ts
```diff
@@ -48,16 +49,28 @@ describe('selectDocumentTypeHelper', () => {
...
       };
 
+      const mockDocuments = MOCK_DOCUMENTS.map(d => {
+        return {
+          ...d,
+          isOnDocketRecord: true,
+          isStricken: false,
+          servedAt: '2019-08-25T05:00:00.000Z',
+        };
+      });
```
Mwindo, line 59: `MOCK_DOCUMENTS` is used in a lot of places. I modified them here in the tests as needed to avoid affecting other tests.

web-client/src/presenter/computeds/selectDocumentTypeHelper.ts
```diff
@@ -149,11 +149,23 @@ export const getPreviouslyFiledDocuments = (
     return { ...doc, documentTitle };
   };
 
-  return caseDetail.docketEntries
+  const formattedCaseDetail = applicationContext
+    .getUtilities()
+    .getFormattedCaseDetail({
+      applicationContext,
+      caseDetail,
+    });
```
Mwindo, line 157: We already have logic on this function to determine if a document has been served or not.

web-client/src/presenter/state.ts
```diff
@@ -733,6 +733,7 @@ export const baseState = {
     pdfsAppended: 0,
     totalPdfs: 0,
   },
+  parentMessageId: undefined,
```
Mwindo, line 736: These were last-minute additions to 10102, in staging but not in test.

web-client/src/presenter/sequences/editUnsignedDraftDocumentSequence.ts
```diff
@@ -17,7 +16,6 @@ export const editUnsignedDraftDocumentSequence = [
         isStatusReportOrder: [
           setEditStatusReportOrderFormAction,
           navigateToPathAction,
-          statusReportOrderPdfPreviewSequence,
```
Mwindo, line 20: Thanks to https://github.com/ustaxcourt/ef-cms/pull/5254, I discovered this is a redundant call. I'm slipping it into this PR to avoid a whole new branch.

web-client/src/presenter/sequences/StatusReportOrder/updateStatusReportOrderFormValueSequence.ts
```diff
@@ -3,4 +3,4 @@ import { setFormValueAction } from '../../actions/setFormValueAction';
 export const updateStatusReportOrderFormValueSequence = [
   setFormValueAction,
   clearJurisdictionRadioAction,
-] as unknown as (props: { key: string; value: any }) => void;
+] as unknown as (props: { key: string; value: string | boolean }) => void;
```
Mwindo, line 6: This is a last-minute addition to 10102, in staging but not in test.

web-client/src/presenter/sequences/editUnsignedDraftDocumentSequence.ts
En-8: As long as you tell me you have high confidence that our test suite covers the functionality this would impact on the slim chance something did go wrong with this, then I think it's reasonable 😄

En-8 (approved):

web-client/src/presenter/sequences/editUnsignedDraftDocumentSequence.ts
Mwindo: In addition to having tested manually, I've just double-checked that `edit-status-report-order.cy.ts` does check this (and fails successfully when I modify things further). Also, I've walked through the logic just to make sure I'm not lying to myself and taking advantage of your trust 🤣 

We call `statusReportOrderPdfPreviewSequence.ts` in three places:

1. `navigateToEditOrderSequence.ts` -- to show an existing order pdf if the user is editing
2. `gotoStatusReportOrderSequence.ts` -- to show an existing order pdf if the user is editing
3. `StatusReportOrder.tsx` -- to show the pdf when the user clicks "Preview"

In `navigateToEditOrderSequence.ts`, in the case of a status report order, we navigate to one of two "status-report-order-edit" routes, which both call `gotoStatusReportOrderSequence.ts`. However, in these edit cases, `gotoStatusReportOrderSequence.ts` itself calls `statusReportOrderPdfPreviewSequence.ts`. So any call to `statusReportOrderPdfPreviewSequence.ts` in `navigateToEditOrderSequence.ts` will always precede the call to `statusReportOrderPdfPreviewSequence.ts` in `gotoStatusReportOrderSequence.ts` and can thus be eliminated.

web-client/src/presenter/state.ts
```diff
@@ -870,3 +871,10 @@ export type CreateCaseIrsForm = {
   irsNoticeFileUrl?: string;
   cityAndStateIssuingOffice?: string;
 };
+
+export type ViewerDocument = {
+  docketEntryId: string;
+  documentTitle?: string; // Should this be required?
```
TomElliottFlexion, line 877: I don't believe we can guarantee that there will be a documentTitle here, so I think this is correct

TomElliottFlexion (approved):

jimlerza (approved):

PR #5263, opened 2024-08-19
10463 Bug: update telephone input elements

akuny:
Change type attribute from `tel` to `text` for phone number input elements so that users can write "N/A" as instructed on mobile, if necessary.

Using `type="tel"` [does not introduce validation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/tel) as with `email` or `url`. As far as I can tell, the key difference between a `tel` input element and a `text` one is that mobile users are directed to a numbers-only keyboard in the case of the former. Since this is the behavior we want to change, `type="text"` should work.

### DIFF
--- a/web-client/src/styles/forms.scss
+++ b/web-client/src/styles/forms.scss
@@ -126,7 +126,6 @@ legend {
 
 input[type='email'],
 input[type='number'],
-input[type='tel'],
 input[type='text'],
 textarea,
 select {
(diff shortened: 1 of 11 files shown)


Mwindo (approved):
Is it worth getting rid of the styling [in forms.scss](https://github.com/ustaxcourt/ef-cms/blob/f58728da92350dc04581fc9587d82dab8d75a709/web-client/src/styles/forms.scss#L129)? Anyway, this fix makes sense to me. Thank you!

akuny:
> Is it worth getting rid of the styling [in forms.scss](https://github.com/ustaxcourt/ef-cms/blob/f58728da92350dc04581fc9587d82dab8d75a709/web-client/src/styles/forms.scss#L129)? Anyway, this fix makes sense to me. Thank you!

Thank you for catching that: there are no longer any `type="tel"` input elements, the styling is indeed no longer necessary.

nechama-krigsman (approved):

PR #5264, opened 2024-08-19
10455 Story: Add/Remove/Modify Judge Users

Mwindo:
## Purpose

We hard code chambers data in [getJudgesChambers.ts](https://github.com/ustaxcourt/ef-cms/blob/staging/web-client/src/business/chambers/getJudgesChambers.ts) and in [add-user.ts](https://github.com/ustaxcourt/ef-cms/blob/16fcac2764f3a588ca768016b838dc753450d73d/scripts/user/add-user.ts#L76-L109). This requires us to deploy code in order to add/remove/update a judge. The goal of this PR is to decouple modifications to judges and code deployments in test/prod and to provide clear steps for onboarding/updating a judge. (#5217 demonstrated to us that this process was _not_ clear/well-documented.)

Note that this PR attempts to do the minimal-ish work possible to accomplish these goals.

### 1. Hard-coded data
Functionally, we use hard-coded chambers data in a few spots:

**getJudgesChambers.ts**

- When calculating the trial session form in [computeTrialSessionFormDataAction.ts](https://github.com/ustaxcourt/ef-cms/blob/f58728da92350dc04581fc9587d82dab8d75a709/web-client/src/presenter/actions/TrialSession/computeTrialSessionFormDataAction.ts#L108-L114).
- When getting the information to display the available chambers sections in the message modal dropdown [here](https://github.com/ustaxcourt/ef-cms/blob/f58728da92350dc04581fc9587d82dab8d75a709/web-client/src/presenter/computeds/messageModalHelper.ts#L65-L67).
- In tests. E.g., here in [mockUsers.ts](https://github.com/ustaxcourt/ef-cms/blob/16fcac2764f3a588ca768016b838dc753450d73d/shared/src/test/mockUsers.ts#L17).

**add-user.ts**

- When validating the addition of a new, non-judge user.

This PR removes the above hard-coded logic and instead creates a new endpoint (and associated lambda, proxy, and actions) to get/set judge chambers information from our database. It also makes slight updates for loading judges in non-test/prod environments. (See Environment-Specific Considerations below.)

### 2. Unclear processes for adding/updating a judge

Our processes for onboarding/updating a judge are not well-documented and are ad-hoc. (See [this PR](https://github.com/ustaxcourt/ef-cms/pull/5217).) This PR adds a few (admittedly miserable) scripts as a slight improvement.

## Broad Overview of the Implementation

Most of the hard-coded data we want to decouple from is already available in the database. The one exception is judge chambers' phone numbers; they were only ever in `getJudgesChambers.ts`. Maybe the most conceptually appropriate way to get them into non-hard-coded persistence would be to add the phone number to the chambers record (and maybe even create a chambers entity) since, technically, the phone number is for each judge's chambers section and not for the judge themselves.

However, our chambers records are currently very sparse; in fact, the only attributes they have are the pk (e.g., `section|soAndSoChambers`) and the sk (which is just the userId of the judge). Given the added complexity of the "chambers-first" approach (which would require stitching together some records)--and given the fact that the only non-pk attribute for chambers' records is just a reference to the relevant judge anyway--I decided on a "judges-first" approach in which we add the phone number to the judge record. This is easier because 1) we can get all the information we need from one set of records rather than stitching data together and 2) a judge is already a `User` entity with a `contact` attribute.

### Environment-Specific Considerations

We have to update judge phone numbers in three separate ways depending on the environment:
- For local data, we need to add the phone number (when applicable, i.e., when it exists in `getJudgesChambers.ts`) to the judge users in `efcms-local.json`. (Otherwise, tests relying on this, if any, will fail.)
- For data in deployed environments other than test/prod, we need to update [judge_users.csv](https://github.com/ustaxcourt/ef-cms/blob/staging/scripts/circleci/judge/judge_users.csv) and `bulkImportJudgeUsers.helpers.ts` similarly.
- For data in test/prod, we need to "manually" add the information. I've provided a one-off script, `add-phone-numbers-to-judge-records.ts`, to update current judges. I've also provided `add-judge.ts` and `update-judge.ts` so that new judges can be onboarded/updated more efficiently.

## Manual Steps

We need to run `scripts/run-once-scripts/add-phone-numbers-to-judge-records.ts` before deploying.

## Fast-follow tasks:

After `scripts/run-once-scripts/add-phone-numbers-to-judge-records.ts` is run, we should pare down and maybe remove entirely `JUDGES_CHAMBERS`. While still used in some tests, I believe we may only strictly need it for the one-time script add-phone-numbers-to-judge-records.ts.

We can also remove `scripts/run-once-scripts/add-phone-numbers-to-judge-records.ts`.

Issues:
- [x] Fix legacy judges appearing in exp1
- [x] Fix User contact validation errors (and/or tests)

Tests:
- [x] Test `scripts/run-once-scripts/add-phone-numbers-to-judge-records.ts` in experimental
- [x] Test `scripts/user/add-judge.ts` in experimental
- [x] Test `scripts/user/update-judge.ts` in experimental
- [x] Make sure `scripts/user/add-user.ts` get the chambers sections dynamically in experimental
- [x] Make sure that messages dropdown works for chambers selection in experimental
- [x] Make sure that phone numbers appear in the trial session information in experimental
- [x] Test `scripts/run-once-scripts/add-phone-numbers-to-judge-records.ts` in test
- [x] Test `scripts/user/add-judge.ts` in test
- [x] Test `scripts/user/update-judge.ts` in test
- [x] Make sure `scripts/user/add-user.ts` get the chambers sections dynamically in test
- [x] Make sure that messages dropdown works for chambers selection in test
- [x] Make sure that phone numbers appear in trial session information  in test
- [x] What, exactly, do we need the test Judges chambers for now? Can we at least remove those phone numbers? A: We need this 1) to add the phone numbers to Dynamo in `add-phone-numbers...` and 2) something like this, at least, for test data like in messageModalHelper.test.ts.
- [x] Update offboard-judge-user.ts
- [x] Update documentation for onboarding/updating/offboarding
- [x] Test that judge_users.csv is loaded correctly on experimental environment
- [x] Test above scripts and add-user.ts (updated to get chambers dynamically) on experimental
- [x] Add the manual deploy steps needed for this PR. (And clarify what they are/what order. E.g., we might need to modify the validation code first to avoid judge users being "invalid" if they have contact: phone.)
- [x] Check validation works. Clean up console logs, poor typing, etc.
- [x] Make sure there is no getCurrentUser (merge in 10417)

Fast-follow tasks:
- [ ] Pare down and maybe remove entirely `JUDGES_CHAMBERS`. While still used in some tests, I believe we may only strictly need it for the one-time script `add-phone-numbers-to-judge-records.ts`.


web-api/src/app.ts
```diff
@@ -879,6 +880,7 @@ app.delete(
     lambdaWrapper(getDocumentQCInboxForSectionLambda),
   );
   app.get('/sections/:section/judge', lambdaWrapper(getJudgeInSectionLambda));
+  app.get('/sections/chambers', lambdaWrapper(getJudgesChambersLambda));
```
Mwindo: We _could_ call `/sections/judge/users`, a route a few lines up, and use that to construct the chambers data on the frontend, which would prevent us from needing another interactor etc. However, this would also mean that the frontend would be coupled to the particular way in which we store data (i.e., that we fetch user data to build chambers data). I have therefore preferred to create a new endpoint and construct the chambers data on the backend so that the frontend can be agnostic about the details.

docs/operations/runbooks/add-or-update-judge-user.md
```diff
@@ -0,0 +1,21 @@
+# Adding or Updating a Judge User
+
+## Description
+This runbook describes the process of onboarding or updating a Judge user. DO NOT use this script to offboard a judge. Instead, see `offboard-judge-user.md`.
```
Mwindo: Change script to runbook

scripts/user/update-judge.ts
```diff
@@ -0,0 +1,301 @@
+import * as client from '../../web-api/src/persistence/dynamodbClientService';
```
Mwindo, line 2: Frankly, this script stinks. It is long, and it uses some direct calls to Dynamo instead of persistence abstractions.

That said, our current ad-hoc process of updating a judge is worse, and this script is meant to be a temporary bandaid that has at least been tested without pressing time constraints.

I have preferred to keep the script all in one place rather than breaking it apart because I think locality here will make it easier to improve later (ideally being merged into the DAWSON application logic itself).

In the cases where I used direct calls to Dynamo, no existing persistence methods seemed quite right. They either did too much (e.g., tried to update a Cognito password) or too little (e.g., did not update the section record) and/or were unclear and therefore dangerous to use. (For instance, web-api/src/persistence/dynamo/users/updateUserRecords.ts seems to expect a user with a bar number, but there is nothing else about this function that would suggest a particular user type.) Rather than tie this to persistence methods that were never designed for updating a judge, and rather than create new persistence methods that might be used for unintended purposes in the application, I have preferred direct calls to Dynamo as necessary to keep this script self-contained.

If there are easy things to improve, I am happy to do so; but I think trying to make this rock-solid--especially if our plan is to move to Postgres--is probably not worth the effort.

shared/src/business/entities/User.ts
```diff
@@ -13,6 +13,21 @@ import { JoiValidationEntity } from '@shared/business/entities/JoiValidationEnti
 import { formatPhoneNumber } from '../utilities/formatPhoneNumber';
 import joi from 'joi';
 
+export interface JudgeContact {
```
Mwindo: I would love to refactor User/JudgeUser, but this PR is not the place for that.

cypress/local-only/tests/integration/trialSession/trial-session-paper-pdf.cy.ts
```diff
@@ -120,6 +120,10 @@ describe('Trial Session Paper Pdf', { scrollBehavior: 'center' }, () => {
           );
           cy.visit(`/edit-trial-session/${createdTrialSessionId}`);
           cy.get('[data-testid="trial-session-judge"]').select('Colvin');
+          cy.get('[data-testid="trial-session-judge"] option:selected').should(
```
Mwindo, line 123: Sometimes the selection hadn't actually finished being loaded in before the test continued.

scripts/user/add-user.ts
```diff
@@ -133,20 +117,27 @@ const sendWelcomeEmail = async ({ email }) => {
 
 // eslint-disable-next-line @typescript-eslint/no-floating-promises
 (async () => {
-  const params = {
+  const { tableName } = await getDestinationTableInfo();
+  environment.dynamoDbTableName = tableName;
+
+  const judgesChambers: JudgeChambersInfo[] = await applicationContext
+    .getUseCases()
+    .getJudgesChambersInteractor(applicationContext, mockJudgeUser);
```
Mwindo: Using the `mockJudgeUser` for permissions. Sort of silly. Happy to hear better options.

shared/src/business/entities/User.ts
```diff
@@ -118,8 +133,17 @@ export class User extends JoiValidationEntity {
...
   };
 
+  static CONTACT_SCHEMA = joi.when('role', {
+    is: joi.invalid(ROLES.judge, ROLES.legacyJudge),
+    otherwise: joi
+      .object()
+      .keys({ phone: User.USER_CONTACT_VALIDATION_RULES.phone })
+      .optional(),
+    then: joi.object().keys(User.USER_CONTACT_VALIDATION_RULES).optional(),
+  });
```
Mwindo: I would prefer to use `joi.valid`. That said, for reasons I do not fully understand, the approximately six trillion different ways I tried `joi.valid` did not work properly for me, so here we are.

shared/src/test/mockJudgesChambers.ts
```diff
@@ -0,0 +1,237 @@
+import { JudgeChambersInfo } from '@shared/proxies/users/getJudgesChambersProxy';
+
+const JUDGES_CHAMBERS: Record<string, JudgeChambersInfo> = {
```
Mwindo, line 3: To be pared down or removed after we run `scripts/run-once-scripts/add-phone-numbers-to-judge-records.ts`

web-api/src/business/useCases/user/getUsersInSectionInteractor.test.ts
```diff
@@ -30,7 +30,7 @@ describe('Get users in section', () => {
     {
       ...mockJudgeUser,
       isSeniorJudge: false,
-      judgeFullName: 'Test Judge 1',
+      judgeFullName: 'Test Judge 2',
```
Mwindo, line 33: Judge 1 is above

web-client/src/presenter/actions/getJudgesChambersAction.ts
```diff
@@ -0,0 +1,21 @@
+import { JudgeChambersInfo } from '@shared/proxies/users/getJudgesChambersProxy';
+import { isEmpty, sortBy } from 'lodash';
+import { state } from '@web-client/presenter/app.cerebral';
+
+export const getJudgesChambersAction = async ({
+  applicationContext,
+  get,
+}: ActionProps) => {
+  const judgesChambersCached = get(state.judgesChambers);
```
Mwindo, line 23: Updates to chambers are relatively infrequent, so I think it is best to cache.

jtdevos (approved):

jimlerza (approved):

TomElliottFlexion (approved):

PR #5266, opened 2024-08-19
10464 Bug: Invalid documents in scenario dropdowns

Mwindo:
We do not want to show unserved, undocketed, or stricken documents from "What Document are You Filing?" dropdowns. This PR enforces these checks and updates tests accordingly.

Additional notes on the to-test PR [here](https://github.com/ustaxcourt/ef-cms/pull/5256).


jimlerza (approved):

web-client/src/presenter/computeds/selectDocumentTypeHelper.test.ts
```diff
@@ -48,16 +49,28 @@ describe('selectDocumentTypeHelper', () => {
         scenario: 'Nonstandard A',
       };
 
+      const mockDocuments = MOCK_DOCUMENTS.map(d => {
```
akuny: Could this variable be extracted somehow so that it doesn't have to be defined repeatedly in multiple test cases?

akuny (approved):
One pedantic comment re minor code duplication in a test case, otherwise LGTM!

web-client/src/presenter/computeds/selectDocumentTypeHelper.test.ts
Mwindo: I debated on whether or not to do that myself--happy to do so :)

PR #5273, opened 2024-08-21
10459 Bug: Fix cognito users missing attributes

jimlerza:
In prod, there are 37 users in cognito that don't have the `custom:userId` and `custom:role` attributes.

For all 37 users, let's:
* update their cognito records:
  * set their `custom:userId` to be the same as their `sub`
  * set their `custom:role` attribute to "petitioner"

For the users that have already confirmed their email addresses, let's add a user entity in dynamodb.

To the users that have not confirmed their email addresses, let's send another confirmation email.

To execute:
```bash
npx ts-node --transpile-only scripts/run-once-scripts/fix-cognito-users-without-ids.ts
```

### DIFF
--- a/scripts/run-once-scripts/fix-cognito-users-without-ids-constants.ts
+++ b/scripts/run-once-scripts/fix-cognito-users-without-ids-constants.ts
@@ -1,0 +1,33 @@
+/* eslint-disable max-lines */
+// import { DateTime } from 'luxon';
+import { UserType } from '@aws-sdk/client-cognito-identity-provider';
+
+export const usersWithoutUserIds: UserType[] = [
+  // {
+  //   Attributes: [
+  //     {
+  //       Name: 'sub',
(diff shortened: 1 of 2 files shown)


scripts/run-once-scripts/fix-cognito-users-without-ids-constants.ts
jimlerza, line 1: I have a version of this file populated with the real users we need to fix. This is just an example. The UUID in the comment is not a real user id.

Mwindo (approved):
Looks reasonable to me :) Thank you!

JayFlexy (approved):

jtdevos (approved):

zachrog (approved):
Everything looks pretty good here, nice job Jim!

jtdevos (approved):

PR #5274, opened 2024-08-22
10439: Docket Clerk - Remove Ability to Create Regular/Small/Hybrid/H-S Trial Sessions

JayFlexy:
This PR introduces additional checks to restrict docket clerk users from adding or editing Trial Sessions, except for those categorized as Special or Motion/Hearing. Despite these restrictions, docket clerk users will still have the ability to view all other types of Trial Sessions.
[ZenHub](https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/gh/flexion/ef-cms/10439)

This PR also includes fixes for cypress test race conditions.

Currently failing [here](https://github.com/ustaxcourt/ef-cms/blob/2b75d8e1c2b25e5ac6727a1c380e3fe6139ad6db/cypress/helpers/fileAPetition/petitioner-creates-electronic-case.ts#L228) in integration tests. Other failures are fixed


web-client/src/presenter/computeds/TrialSession/addTrialSessionInformationHelper.ts
```diff
@@ -31,12 +32,18 @@ export const addTrialSessionInformationHelper = (
     });
   }
 
+  // NOTE: what happens if isStandaloneSession is true and user is docketclerk?
```
JayFlexy, line 44: Unsure, of the business usecase here or if this state is even possible. Thought I'd mention it to get some feedback or context from the team.

web-client/src/presenter/computeds/formattedTrialSessionDetails.ts
```diff
@@ -88,15 +88,22 @@ export const formattedTrialSessionDetails = (
...
+      return editableSessionTypes.includes(sessionType);
+    };
 
     canDelete = trialDateInFuture && !formattedTrialSession.isCalendared;
     canEdit =
       trialDateInFuture &&
       formattedTrialSession.sessionStatus !== SESSION_STATUS_GROUPS.closed &&
       !isChambersUser;
 
+    if (user.role === USER_ROLES.docketClerk && canEdit) {
```
JayFlexy, line 103: Kept it simple here for now but this could always change if we find that it is necessary to have additional checks for other roles

web-client/src/presenter/computeds/TrialSession/addTrialSessionInformationHelper.ts
```diff
@@ -31,12 +32,18 @@ export const addTrialSessionInformationHelper = (
...
+  const getSessionTypes = (user: AuthUser) => {
+    return user.role === 'docketclerk'
+      ? ['Special', 'Motion/Hearing']
```
akuny: Since the array `['Special', 'Motion/Hearing']` is used twice in this computed on lines 31 and 38, could it be extracted as a variable and given a descriptive name (as in the `docketClerkCanEditCheck` function included with this PR)?

akuny (approved):

web-client/src/presenter/computeds/TrialSession/addTrialSessionInformationHelper.ts
JayFlexy: Yep, good call

JayFlexy: As a follow-up, we seem to be checking for those trial session types for different reasons. If I were to combine them into a single variable would that dilute readability of the code? I am all for DRY but only if whatever is being assigned as a consistent use afterwards @akuny . Thoughts?

akuny: That's a great point! In that case, my instinct would be to create two separate variables, one in `getSessionTypes` and the other in the anonymous function passed to `filter` on line 30, each with a descriptive name. But that might be overkill.

JayFlexy: Ah, nice. I can do that! I do like adding code as docs

zachrog (approved):

JayFlexy:
~~Currently failing [here](https://github.com/ustaxcourt/ef-cms/blob/2b75d8e1c2b25e5ac6727a1c380e3fe6139ad6db/cypress/helpers/fileAPetition/petitioner-creates-electronic-case.ts#L228) in integration tests. Other failures are fixed~~

jtdevos (approved):

jimlerza (approved):

PR #5108, opened 2024-06-25
10102-Story: Judge/CA/ADC: Generate Response To Status Report

pixiwyn:

PR #5279, opened 2024-08-23
10399 Bug: direct users editing a PDF upload to the correct form

akuny:
PR [2795](https://github.com/ustaxcourt/ef-cms/pull/2795/) let users create a draft document using the "Upload  PDF" form (*/upload-court-issued/*) without a default `documentType`, instead leaving that property undefined. When the document finalized in the "Add Docket Entry" form, a document type is then assigned.

Changes in PR [5108](https://github.com/ustaxcourt/ef-cms/pull/5108) introduced a [bug](https://github.com/ustaxcourt/ef-cms/pull/5108/files#r1681766044) wherein users who tried to edit a document that they uploaded using the "Upload PDF" form were directed to the wrong form.

This PR ensures that any document in draft state with an undefined `documentType` property redirects to the edit uploaded court issued pdf (_/edit-upload-court-issued/_) form, not the edit order (_/edit-order/_) form as before.

### DIFF
--- a/cypress/local-only/tests/integration/caseDetail/docketRecord/courtIssuedFiling/docket-clerk-uploads-and-edits-court-issued-docket-entry.cy.ts
+++ b/cypress/local-only/tests/integration/caseDetail/docketRecord/courtIssuedFiling/docket-clerk-uploads-and-edits-court-issued-docket-entry.cy.ts
@@ -1,0 +1,60 @@
+import { faker } from '@faker-js/faker';
+import { goToCase } from '../../../../../../helpers/caseDetail/go-to-case';
+import { loginAsDocketClerk1 } from '../../../../../../helpers/authentication/login-as-helpers';
+
+describe('Docket clerk uploads and edits court-issued docket entries', () => {
+  it('should let a docket clerk upload a court-issued docket entry and the edit it while it is unsigned', () => {
+    const leadCase = '111-19';
+
+    loginAsDocketClerk1();
+    goToCase(leadCase);
+
+    // Arrange: create the docket entry
+    cy.get('[data-testid="case-detail-menu-button"]').click();
+    cy.get('[data-testid="menu-button-upload-pdf"]').click();
+    const title = `${faker.word.adjective()} ${faker.word.noun()}`;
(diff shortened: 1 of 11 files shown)


Mwindo:
EDIT: I see you just marked this as a draft! (Sorry! I started looking at it just before!) Feel free to ignore these comments accordingly.

Thank you for tackling this @akuny ! This is a great start. In addition, I think we need to handle signed drafts. Right now, there are two additional places where editing the uploaded order will bring you to the generic editor rather than the upload page (since this is an error that was perpetuated by PR 5108, but which existed before too):

1. Editing a signed order from the drafts view.
2. Editing a signed order from the messages view if the order is attached to a message.

I haven't looked into this too much further, but I believe they both rely on `getDocumentEditUrlAsPathAction`. Also, could we look to see if [this other usage of the === 'Miscellaneous'](https://github.com/ustaxcourt/ef-cms/blob/f63ddafe019ce47dae1c22631e94c3b591e81a00/shared/src/business/utilities/getFormattedCaseDetail.ts#L251-L263) is actually used anywhere? If not, can we remove it? If so, can we update it similarly?

akuny:
> EDIT: I see you just marked this as a draft! (Sorry! I started looking at it just before!) Feel free to ignore these comments accordingly.
> 
> Thank you for tackling this @akuny ! This is a great start. In addition, I think we need to handle signed drafts. Right now, there are two additional places where editing the uploaded order will bring you to the generic editor rather than the upload page (since this is an error that was perpetuated by PR 5108, but which existed before too):
> 
> 1. Editing a signed order from the drafts view.
> 2. Editing a signed order from the messages view if the order is attached to a message.
> 
> I haven't looked into this too much further, but I believe they both rely on `getDocumentEditUrlAsPathAction`. Also, could we look to see if [this other usage of the === 'Miscellaneous'](https://github.com/ustaxcourt/ef-cms/blob/f63ddafe019ce47dae1c22631e94c3b591e81a00/shared/src/business/utilities/getFormattedCaseDetail.ts#L251-L263) is actually used anywhere? If not, can we remove it? If so, can we update it similarly?

This is very helpful, thank you!

PR #5284, opened 2024-08-26
10397: Fix List Item Accessibility Violation

TomElliottFlexion:
- Replaces `react-paginate` with a custom paginator
- Uses same USWDS styling as old paginator


web-client/src/ustc-ui/Pagination/Paginator.tsx
```diff
@@ -1,51 +1,60 @@
...
-  breakClassName?: string;
-  marginPagesDisplayed?: number;
-  pageCount: number;
-  id?: string;
-  pageRangeDisplayed: number;
-  onPageChange: (selectedItem: { selected: number }) => void;
-  forcePage: number;
+  totalPages: number;
+  currentPageIndex: number;
+  onPageChange: (currentPage) => void;
```
akuny: Could we specify the expected type of the `currentPage` parameter in the `onPageChange` function?

web-client/src/ustc-ui/Pagination/Paginator.tsx
```diff
@@ -1,51 +1,60 @@
...
-        />
-      </nav>
-    </>
+    <nav
+      aria-label="Pagination"
+      className="usa-pagination margin-bottom-0 margin-top-0"
+      role="navigation"
+    >
+      <ul className="usa-pagination__list">
+        {currentPage > 1 && (
```
akuny: Could we extract the comparisons that determine which section renders out into functions with descriptive names? (Something like `shouldRenderPrevious` and `shouldRenderNext`.)

Although, as a counterpoint, one could argue that all such comparisons in this component could be extracted like this, which in turn could be overkill.

akuny (approved):
This looks good! I do think the new `Paginator` component could benefit from being unit tested, though.

cypress/local-only/tests/accessibility/customCaseReport/docket-clerk.cy.ts
```diff
@@ -24,7 +24,7 @@ describe('Custom Case Report - Docket Clerk Accessibility', () => {
       {
         includedImpacts: impactLevel,
         retries: 3,
-        rules: { listitem: { enabled: false } }, // https://github.com/flexion/ef-cms/issues/10397
+        rules: { listitem: { enabled: true } },
```
Mwindo: Do we need this line? I think enabled is the default.

web-client/src/ustc-ui/Pagination/Paginator.tsx
```diff
@@ -1,51 +1,60 @@
...
-          pageLinkClassName="usa-pagination__button cursor-pointer"
-          pageRangeDisplayed={pageRangeDisplayed}
-          previousClassName="usa-pagination__item"
-          previousLinkClassName="usa-pagination__link usa-pagination__previous-page cursor-pointer"
-          renderOnZeroPageCount={null}
-          onPageChange={onPageChange}
-        />
-      </nav>
-    </>
+    <nav
```
Mwindo, line 20: I'm looking through the differences in the DOM structure to make sure we aren't overlooking anything related to aria-labels, etc.

Before:

Screenshot 2024-08-27 at 10 06 58 AM

After this PR:

Screenshot 2024-08-27 at 10 14 25 AM



I think it would be good to include the `aria-label` and `aria-current` attributes for the current page.

web-client/src/ustc-ui/Pagination/Paginator.tsx
```diff
@@ -1,51 +1,60 @@
...
+      role="navigation"
+    >
+      <ul className="usa-pagination__list">
+        {currentPage > 1 && (
+          <li className="usa-pagination__item">
+            <Button
+              link
+              aria-label="Previous"
+              className="usa-pagination__link usa-pagination__previous-page cursor-pointer"
+              disabled={currentPage === 1}
```
Mwindo: Do we need this if we are only rendering it for `currentPage > 1`? Similar question for the next button below.

Mwindo: Actually, another thing that react-paginate seems to do is keep these buttons around but disable/enable them. Note the `hide` class, `aria-disabled`, `disabled`, `tabindex=-1`, etc. If I recall correctly, it is better for accessibility to keep the prev/next buttons but hide/disable them.

Screenshot 2024-08-27 at 10 26 52 AM

It might be better to mimic react-paginate's behavior.

Mwindo (commented):
Thanks for tackling this @TomElliottFlexion! I left a few comments about accessibility features we might want to maintain as we switch off of react-paginate, i.e., keeping the disabled buttons around and adding a few more attributes.

cypress/local-only/tests/accessibility/customCaseReport/docket-clerk.cy.ts
TomElliottFlexion: I wanted to make it explicit, but you're right, this is just cruft

web-client/src/ustc-ui/Pagination/Paginator.tsx
TomElliottFlexion: Good call, done!

TomElliottFlexion:
> This looks good! I do think the new `Paginator` component could benefit from being unit tested, though.

I'm afraid we currently don't use a React testing library, and therefore don't test React components. This is intentional, and why we strive to do most front-end logic in computeds and actions. It's possible we should reconsider this paradigm, but no unit tests on React components for now

web-client/src/ustc-ui/Pagination/Paginator.tsx
TomElliottFlexion: Thank you, added!

web-client/src/ustc-ui/Pagination/Paginator.tsx
TomElliottFlexion: This is a good idea, fixed! I'd love to hear more about this accessibility recommendation, since I think we are violating it dozens of times throughout the app

web-client/src/ustc-ui/Pagination/Paginator.tsx
```diff
@@ -1,51 +1,74 @@
...
+      role="navigation"
+    >
+      <ul className="usa-pagination__list">
+        <li className="usa-pagination__item">
+          <Button
+            link
+            aria-disabled={previousDisabled}
+            aria-label="Previous"
+            className={classNames(
+              `${previousDisabled && 'hide'}`,
```
Mwindo, line 32: Funnily enough, looking at `hide` (which just does `display: none`), I'm not sure it is any better than conditional rendering. I'm not an expert on accessibility by any means, but I think we might want to use [the sr-only class](https://github.com/ustaxcourt/ef-cms/blob/97ec44dfb6354a0e5c619b0181fd8b57e2bdd25d/web-client/src/styles/custom.scss#L2326C1-L2336C2) that I added for a related ticket. See, e.g., https://piccalil.li/blog/visually-hidden/ and https://stackoverflow.com/questions/19758598/what-is-sr-only-in-bootstrap-3.

In any case, I think the code is at least better-structured this way, and I'll let you decide if you want to use `sr-only` or not.

(NB: It looks like we use the same `display: none` approach for `usa-sr-only`, which is probably incorrect and which we will probably want to change at some point. But that's outside the scope of this ticket.)

Mwindo (approved):
Thanks @TomElliottFlexion!

JayFlexy (approved):
lgtm

web-client/src/ustc-ui/Pagination/Paginator.tsx
TomElliottFlexion: Let's fix this!

PR #5288, opened 2024-08-27
10365 Bug: Remove leading/trailing spaces from user-entered passwords

Mwindo:
We prevent passwords having leading/trailing spaces from being saved. In this PR, we also remove leading/trailing spaces from the password field if users are trying to log in.

See notes on [the ticket ](https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/gh/flexion/ef-cms/10365) about justifications for the approach in this PR.


akuny (approved):

web-client/src/presenter/actions/Login/submitLoginAction.tsx
```diff
@@ -12,17 +12,22 @@ export const submitLoginAction = async ({
 }> => {
   const { email, password } = get(state.authentication.form);
 
+  const cleanedPassword = password.trim();
```
zachrog, line 15: We are preventing all NEW passwords from having leading or trailing spaces, but this was only introduced 8 months ago. For everyone else they could have a leading space and we would be blocking them from logging in. In general although we are trying to help the user I think any modification of the user's input for the password is probably not a great idea, as it is not what they typed.

I am going to block this PR until we talk about it as I want to make sure we understand the consequences of this story.

zachrog (changes requested):

web-client/src/presenter/actions/Login/submitLoginAction.tsx
Mwindo: Thank you!

> but this was only introduced 8 months ago

This is great to know. I picked this ticket up, realized it might be tricky for reasons similar to what you outlined, put it back, got a comment from Tenille, and picked it back up. Maybe we can instead flag the user, similar to a "caps lock" warning icon.

zachrog: I am going to remove the block on this PR to allow you to continue forward! Thanks for chatting with everyone!

zachrog (commented):
Removing block to allow you to continue forward. I trust you to have the right conversations with the right people!

zachrog (approved):

PR #5291, opened 2024-08-27
10377: Avoid running sequence multiple times at once TO TEST

cruzjone-flexion:


web-client/src/presenter/utilities/singletonSequenceDecorator.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const singletonSequenceDecorator = (
+  sequenceName: string,
+  actionsList: any[],
+): any[] => {
+  return [
+    () => ({ sequenceName }),
+    setSingletonSequenceNameAction,
+    {
+      alreadyRunning: [],
```
Mwindo, line 12: Is it worth logging a warning message here to flag devs that the same sequence is being run in cases they didn't expect?

Mwindo (commented):
I think this is a clever solution! That said, I have two concerns:

1) If one of the actions in `singletonSequenceDecorator`'s `actionsList` fails, `unsetSingleSequenceNameAction` will not be called. This means the sequence cannot be run again unless the user refreshes. (I tested this locally to confirm.) We will need to handle errors to avoid corrupting global singleton state.
2) This ties us more closely to Cerebral. A debounce/deactivate solution can be done in pure React.

PR #5298, opened 2024-08-28
10377: Avoid running sequence multiple times at once TO TEST

cruzjone-flexion:


Mwindo (approved):
Thanks @cruzjone-flexion!

I had to implement something similar [here](https://github.com/ustaxcourt/ef-cms/pull/5252/files). Based on research from that, I think a debounce--in addition to the button deactivation--would be a good idea since I suspect a user _could_ actually trigger multiple clicks even with the deactivate strategy. (For instance, a JS script to click the button multiple times gets several clicks through before the deactivate, but it can't get around the debounce.) Maybe there is even a way to extract this into a component like DebouncedButton, but idk. I'll leave that all up to you.

cruzjone-flexion:
@Mwindo 
Yeah already working on implementing on the shared Button component. At first I thought debounce might be overkill but guess there is no harm in adding that it too.

zachrog (approved):

web-client/src/ustc-ui/Button/Button.tsx
```diff
@@ -44,8 +49,31 @@ export const Button = props => {
...
 
+  const debouncedWrapper = debounce(async () => {
+    const results = onClick();
+    if (!(results instanceof Promise))
+      throw new Error('Convert onClick method to async');
+
+    await results.finally(() => {
+      setDisableButton(false);
+    });
+  }, 500);
```
Mwindo: Sorry to comment on a draft PR--I just saw the update and was excited about having a debounced button component, then I noticed my previous comment might have misguided you slightly! I'm guessing you took the 500 ms from the PR I linked in my comment, but there is certainly nothing sacred about 500 ms, and it might even make certain actions feel a little sluggish. (In my use case, a 500 ms hit wasn't a big deal because it was a one-time account creation button.)

cruzjone-flexion: 500 ms should be enough for the UI to disable the button;

PR #5300, opened 2024-08-28
Fix migration queue empty flag

TomElliottFlexion:
- Our `setup-for-blue-green-migration.sh` script (which runs as part of the migration infrastructure in Circle) sets a `migration-queue-empty` flag in the deploy table to false to indicate that a migration is yet to complete
- The `wait-for-migration` step is meant to look for this flag to determine when a migration is finished processing, but was instead looking for `migration-queue-is-empty`, causing incomplete migrations when `migration-queue-is-empty` was not reset to true


Mwindo (approved):
Thanks for following up on this @TomElliottFlexion!

jimlerza (approved):

PR #5305, opened 2024-08-29
Make explicit unified logger

zachrog:
## What Is In This Change
- Create a singleton logger for web-api that all can import from without going through appContext
- Replaces the singleton logger which was created in express middleware and passed around through appContext
- This DOES NOT change what we are logging or how. The interface and information is purposefully kept the same. Refactoring can be done for what info we are logging but that is not the goal of this PR.

## Why This Change
- Whenever we use applicationContext.logger.info it includes data about what authenticated user is currently running the function. This feature however is tied up inside of applicationContext.
- This change preserves the ability to attach extra context to every logger.info() call without the need to go through applicationContext.
- This is the last piece of stateful information inside web-api applicationContext and allows us to freely move away from appContext if we choose.


web-api/src/utilities/logger/getLogger.ts
jtdevos, line 1: If you anticipate that addContext/addUser will only be called once at-most in the objects lifecycle, might it be better to have this data be optional arguments to the constructor, and simply replacing the singleton with a new logger as the context/user info becomes available?   But this works too, so your call.

Regardless - this is good stuff!  I definitely like the idea of pulling out logger from appcontext as there have been times where I wanted to log something in a consistent way but didn't otherwise need the app context (e.g. reports, one-off scripts).

zachrog: I think that makes sense however, we do not have all of the information all at once to be able to create a new Logger. We gain context of the request first, do some logging, then at a later time we gain context of who the user is. I like the idea of making it clear what context we want with the logger by making it part of the constructor but, I think it limits the use of the logger because we may not know all of that information all at once.

jimlerza (approved):

web-api/src/applicationContext.ts
```diff
@@ -97,20 +96,8 @@ const entitiesByName = {
...
-  if (process.env.NODE_ENV === 'production') {
-    const authenticated = user && Object.keys(user).length;
-    logger.defaultMeta = logger.defaultMeta || {};
-    logger.defaultMeta.user = authenticated
-      ? user
-      : { role: 'unauthenticated' };
-  }
-
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export const createApplicationContext = (appContextUser = {}) => {
```
Mwindo, line 100: Is there a good reason not to get rid of `appContextUser` besides a slightly larger git diff?

Mwindo (approved):
LGTM--thanks @zachrog!

web-api/src/applicationContext.ts
zachrog: Good Question. No that is the only reason.

If I removed it I would need to update 50 files and I wanted this PR to really show what was changing for the logger.

PR #5306, opened 2024-08-30
10437 Story - Blocked Cases Report Enhancements

zachrog:
- Adds filters for Case Status + Reason to blocked cases report.
- Allows user to export report as csv.


cruzjone-flexion (approved):

web-client/src/applicationContext.ts
```diff
@@ -624,6 +624,22 @@ const applicationContext = {
   convertBlobToUInt8Array: async blob => {
     return new Uint8Array(await new Response(blob).arrayBuffer());
   },
+  createCsvString: (
+    data: any[],
+    config: { displayLabel: string; key: string }[],
+  ) => {
+    const headers = config.map(c => `"${c.displayLabel}"`).join();
```
akuny, line 631: It may be worth escaping certain characters that could break the CSV format like quotation marks inside the functions passed to the map method calls here.

web-client/src/presenter/actions/Reports/BlockedCaseReport/exportCsvBlockedCaseReportAction.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  const formatString = (s: string | undefined) =>
+    (s || '').split('\n').join(' ').trim();
+
+  const formattedBlockedCases = blockedCases.map(c => ({
+    ...c,
+    allReasons: `${formatString(c.blockedReason)} ${formatString(c.automaticBlockedReason)}`,
+  }));
+
+  const csvString = applicationContext.createCsvString(
```
akuny: It may be worth considering extracting the formatting logic that prepares the formattedBlockedCases constant into its own function on applicationContext as well so that it can be tested in isolation apart from Cerebral.

web-client/src/presenter/actions/setFormValueAction.ts
akuny, line 1: For what it's worth, do we definitely want to use this action to set non-form state properties? I'm sure making a separate action called `setBlockedCaseReportFilterAction` would be tedious, but allowing this particular action to ignore the form prop on state entirely may make things tricky down the road.

If so, the term "root" could be made more explicit by being called "stateProp" since from what I can tell the user is specifying which prop on the state tree to update.

.eslintrc.js
```diff
@@ -44,6 +44,7 @@ module.exports = {
         'no-underscore-dangle': 'off',
         'promise/always-return': 'off',
         'promise/catch-or-return': 'off',
+        'promise/no-nesting': 'off',
```
akuny, line 47: Do we definitely want to turn this off by default, or are nested promises used sparingly enough that adding a comment to ignore this rule on a case by case basis would work?

akuny (approved):
This looks great! I left a few questions, but none of them are pressing.

.eslintrc.js
jimlerza: We utilize nested promises pretty extensively in Cypress tests. I believe this is a limitation of Cypress itself?

zachrog: Yeah in the context of the whole application we do not like nesting promises. In cypress the context changes as we HAVE to nest promises as that is how their chainables work, so we modified our lint to ignore those rules in cypress.

web-client/src/applicationContext.ts
zachrog: Good Point.

Right now we have 3 csv libraries and none of them can be ran in jest with the js-dom environment which is frustrating, so we decided to just write our own. Looking back I don't think that is a good idea either. I think we can convert to one of our other patterns (which is not testing it) but using a standard library to achieve this.

Ultimately I would like to see us converge on a single csv library that works on frontend/backend and can be run in the context of a test to reduce this nonsense.

web-client/src/presenter/actions/Reports/BlockedCaseReport/exportCsvBlockedCaseReportAction.ts
zachrog: Because this is such a specific use case of how two strings should be combined(for display purposes) I do not see much utility in extracting it out into a function as I don't know who else would want to use this? But if you have an idea of where else this might be useful I would be down to extract!

cruzjone-flexion (approved):

cypress/helpers/fileAPetition/create-and-serve-paper-petition.ts
```diff
@@ -1,8 +1,29 @@
...
 import { loginAsPetitionsClerk1 } from '../authentication/login-as-helpers';
 
 export function createAndServePaperPetition(
-  options = { yearReceived: '2020' },
-) {
+  {
+    procedureType = 'Regular',
+    trialLocation = 'Birmingham, Alabama',
+    yearReceived = '2020',
+  }: Partial<{
```
jtdevos, line 9: TIL about Partial types!

zachrog: Everything is optional!

jimlerza (approved):

PR #5312, opened 2024-09-03
10001 (and 10002 and 10003) Design Debt: More Granular File Validation Errors

Mwindo:
TL;DR: When something goes wrong with a (typically PDF) file upload, the user doesn't know until we try to parse it on the backend, at which point we only display a generic error. This means 1) the error is often not specific enough to be helpful and 2) the error might come at the end of a long form rather than as soon as the user selects the file. This PR addresses the issue.

This should free up the court's time addressing frequent help tickets (multiple per day) regarding failed file uploads. It will also prevent corrupted/invalid files from being uploaded to S3.

**N.B.: The PR diff is not as intimidating as it looks. A lot of the diff is refactoring.**

**Manual step: the usataxcourt website FAQ should be updated at https://ustaxcourt.gov/dawson_faqs_case_management.html#FileUpload**

---

Tickets: [10001](https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/gh/flexion/ef-cms/10001) (password-protected PDFs), [10002](https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/gh/flexion/ef-cms/10002) (invalid files types), [10003](https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/gh/flexion/ef-cms/10003) (invalid/corrupted PDFs). (But note that this PR is best summarized as "implement client-side file validation." )

Figma [here](https://www.figma.com/design/pBCkGpOvCG8e72iNnDB2a0/USTC-%7C-File-Upload%3A-Helpful-Error-Messaging-%7C-10001%2C-10002%2C-10003-(Updated)?node-id=0-1&node-type=canvas&t=i60mADqkwUn4qiSd-0).

### The Approach in This PR

In order to provide the user immediate feedback (rather than waiting until they try to submit), we use pdf-js client-side on file selection to check for issues. If issues _are_ discovered--e.g., the PDF is password-protected or the wrong file type--we show a modal with an error message and clear the file selection. (While validating--which is usually very quick--we display a loading overlay.) This sidesteps difficulties with joi document validation.

In addition, we send an error message to the backend for logging. This error message is slightly decoupled from what is shown to the user in order for us to have finer-grained control over what is logged. (E.g., while we may display the same generic error message for multiple types of file parsing errors, we log them more granularly for debugging/prod-support purposes.)

While the front-end messaging has been overhauled, this PR does not change our back-end PDF validation implementation at all. In other words, the client-side validation is there for convenience to filter out bad documents, but we still validate on the back end.

The validation has been added to (almost) everywhere we upload files. (I was told to ignore `PractitionerAddEditDocumentation`, which might be changed soon, and `PetitionQcScanBatchPreview`, which, despite some code for uploading, doesn't seem to actually allow uploads.)

### Examples of New Error Messages

https://github.com/user-attachments/assets/9eb7cdc7-67b5-4da5-87a0-08ad36ccd7b7


web-client/src/presenter/actions/updateValidationErrorsAction.ts
```diff
@@ -0,0 +1,46 @@
+import { state } from '@web-client/presenter/app.cerebral';
+
+// TODO: Test more thoroughly
+function recursiveMerge(obj1, obj2) {
```
Mwindo, line 4: This seems

shared/src/proxies/logErrorProxy.ts
```diff
@@ -0,0 +1,15 @@
+import { ClientApplicationContext } from '@web-client/applicationContext';
+import { post } from './requests';
+
+export const logErrorInteractor = (
```
Mwindo, line 4: The ticket specifies that they still want the errors logged to Opensearch.

Mwindo: TODO: Is it necessary to add an authorization check so arbitrary users cannot log errors by spamming the endpoint?

web-client/src/views/FileHandlingHelpers/cloneFile.error.test.ts
```diff
@@ -29,6 +29,7 @@ describe('cloneFile', () => {
   it('should fail when attempting to clone something other than a file', async () => {
     let error;
     try {
+      // @ts-ignore
```
Mwindo, line 32: Not even sure if this test is worth it since TypeScript should enforce the type?

cypress/helpers/fileAPetition/petitioner-creates-electronic-case.ts
```diff
@@ -293,40 +294,11 @@ export function petitionerAttemptsToUploadCorruptPdf() {
...
-  cy.get('[data-testid="stin-preview-button"]').should('exist');
-  cy.get('[data-testid="petition-preview-button"]').should('exist');
-  cy.get('[data-testid="atp-preview-button"]').should('exist');
-
-  cy.get('[data-testid="file-petition"]').click();
-
-  cy.get('[data-testid="modal-dialog"]').should('exist');
-  cy.get('[data-testid="modal-dialog-header"]').should(
-    'contain.text',
-    'Your Request Was Not Completed',
```
Mwindo, line 337: We don't need to go to the end of the process to get the error anymore.

cypress/helpers/file/upload-file.ts
```diff
@@ -3,13 +3,29 @@
...
     'validated',
   );
   cy.get(`[data-testid="${testId}"]`).attachFile(
     '../../helpers/file/sample.pdf',
   );
-  cy.get('[data-testid="upload-file-success"]').should('exist');
+  cy.get('[data-testid^="upload-file-success"]').should('exist');
+}
+
+export function attachFile({
```
Mwindo, line 17: Because validating on file selection introduces a slight delay and causes tests to fail, I created this helper function to ensure our Cypress tests wait for success appropriately. A good amount of the git diff in this PR is just refactoring cy.get().attachFile to use this helper and renaming `uploadFile` to `attachSamplePdfFile` to avoid confusion with `attachFile`.

cypress/local-only/tests/integration/fileUpload/upload-court-issued-pdf-validation.cy.ts
```diff
@@ -0,0 +1,86 @@
...
+      selector: '[data-testid="primary-document-file"]',
+    });
+
+    cy.get('[data-testid="file-upload-error-modal"]').contains(
+      'The file is encrypted or password protected. Remove encryption or password protection and try again.',
+    );
+  });
+
+  it('should display error modal when a file larger than the limit is selected', () => {
+    // Create a 300MB Blob (300 * 1024 * 1024 bytes)
```
Mwindo: Remove this comment

shared/src/business/entities/EntityConstants.ts
```diff
@@ -1723,3 +1723,8 @@ export const STATUS_REPORT_ORDER_OPTIONS = {
     restored: 'restoredToGeneralDocket',
   },
 };
+
+export const TROUBLESHOOTING_INFO = {
+  APP_SUPPORT_EMAIL: 'dawson.support@ustaxcourt.gov',
+  FILE_UPLOAD_TROUBLESHOOTING_LINK: '', // TODO 10001
```
Mwindo: @jtdevos Once we know what the DAWSON file troubleshooting URL will be, I will put it here :)

shared/src/business/entities/EntityConstants.ts
```diff
@@ -1723,3 +1723,8 @@ export const STATUS_REPORT_ORDER_OPTIONS = {
     restored: 'restoredToGeneralDocket',
   },
 };
+
+export const TROUBLESHOOTING_INFO = {
+  APP_SUPPORT_EMAIL: 'dawson.support@ustaxcourt.gov',
```
Mwindo, line 1738: Pulling this out as a constant since I see it being used all over the place. This small change constitutes another non-trivial portion of the diff.

web-client/src/presenter/actions/validateFileSizeAction.ts
```diff
@@ -1,4 +1,4 @@
-import { limitFileSize } from '../../views/limitFileSize';
+import { limitFileSize } from '../../views/FileHandlingHelpers/limitFileSize';
```
Mwindo: TODO: Is limitFileSize and this action even needed anymore?

web-client/src/views/FileDocument/StateDrivenFileInput.tsx
```diff
@@ -48,6 +61,61 @@ export const StateDrivenFileInput = connect<
 
     const fileOnForm = file || form[fileInputName] || form.existingFileName;
 
+    // Setting the filename here so that we can display it before validation
+    // finishes, otherwise a slow machine might have slight lag and allow the user
+    // to "choose file" again.
```
Mwindo, line 64: Maybe this is overkill?

web-client/src/views/FileDocument/StateDrivenFileInput.tsx
```diff
@@ -48,6 +61,61 @@ export const StateDrivenFileInput = connect<
...
+    const [selectedFilename, setSelectedFilename] = useState('');
+
+    const onFileSelectionChange = async (
+      e: React.ChangeEvent<HTMLInputElement>,
+    ) => {
+      setSelectedFilename(e.target?.files?.[0]?.name || '');
+      setIsLoadingSequence();
```
Mwindo, line 71: TODO: Maybe everything following this should be an action/sequence.

web-client/src/views/FileDocument/StateDrivenFileInput.tsx
Mwindo: Here is an example of the issue I can get when throttling my browser _without_ `selectedFilename`. (Slight delay between valid selection and updated UI leads to "Choose file" being able to be selected again.)


https://github.com/user-attachments/assets/a5574ccc-160e-41aa-83ad-21c6a6daa205

web-client/src/views/AppComponent.tsx
```diff
@@ -304,6 +305,7 @@ export const AppComponent = connect(
         {showModal === 'WebSocketErrorModal' && <WebSocketErrorModal />}
         {showModal === 'AppMaintenanceModal' && <AppMaintenanceModal />}
         {showModal === 'GenericErrorModal' && <GenericErrorModal />}
+        {showModal === 'FileUploadErrorModal' && <FileUploadErrorModal />}
```
Mwindo, line 306: This was being called in many different components, often like `<FileUploadErrorModal confirmSeqence={something} />`. However, 1) now this modal is used more often, so it makes sense to centralize it by placing it in the React hierarchy tree here (in fact, I think most modals are best done that way to avoid multiple components potentially rendering the same modal) and 2) `confirmSeqence={something}` was not actually doing anything, so getting rid of it in the components that used it was perfectly safe.

web-client/src/views/FileDocument/StateDrivenFileInput.tsx
```diff
@@ -48,6 +57,53 @@ export const StateDrivenFileInput = connect<
...
 
+    // Setting the filename here so that we can display it before validation
+    // finishes, otherwise a slow machine might have slight lag and allow the user
+    // to "choose file" again.
+    const [selectedFilename, setSelectedFilename] = useState('');
+
+    const onFileSelectionChange = async (
+      e: React.ChangeEvent<HTMLInputElement>,
+    ) => {
+      setSelectedFilename(e.target?.files?.[0]?.name || '');
```
Mwindo, line 70: There are a lot of similarities to the implementation in `ScanBatchPreview.tsx`, and therefore there might be a way to refactor. That said, that was true before this PR, too, and this PR is already large, so I think further refactoring is best reserved for later.

This isn't a very Cerebral-specific way to do things--i.e., it is not calling a sequence directly but instead passing sequences into functions. Here are my thoughts:

1. The original file validation here was also being done in a "non-Cerebral" way, so this follows the existing precedent.
2. There are benefits to pulling into a sequence (consistency) as well as benefits to keeping it distinct (modularity). I can't decide which is a better tradeoff, but the former takes more work.

Therefore, I have left things as is.

Mwindo: EDIT: Trying to do a lightweight refactor to have the best of both worlds

cypress/helpers/file/upload-file.ts
```diff
@@ -3,13 +3,29 @@
...
+  selector,
+  selectorToAwaitOnSuccess,
+}: {
+  selector: string;
+  filePath: string;
+  selectorToAwaitOnSuccess?: string;
+}) {
+  cy.get(selector).attachFile(filePath);
+  if (selectorToAwaitOnSuccess) {
+    cy.get('loading-overlay').should('not.exist');
```
akuny: Should this `cy.get` call look for `'[data-testid="loading-overlay"]'` instead?

web-client/src/views/FileHandlingHelpers/fileValidation.ts
```diff
@@ -0,0 +1,204 @@
...
+      messageToDisplay: errorInformation?.errorMessageToDisplay!,
+      messageToLog: errorInformation?.errorMessageToLog,
+    });
+    target.value = '';
+    return;
+  }
+  onSuccess({ selectedFile });
+};
+
+export const validateFile = async ({
```
akuny, line 144: Is this function called anywhere apart from `validateFileOnSelect` in this file? In that case, do we necessarily need to export it and put it under test? Just a thought if we want to trim down the test suite for this file.

akuny (approved):
@Mwindo Just a couple of minor comments, thanks!

cypress/helpers/file/upload-file.ts
Mwindo: 🤦 Thanks @akuny!

web-client/src/views/FileHandlingHelpers/fileValidation.ts
Mwindo: Good point--I think you're correct! Thank you :)

Mwindo: Actually, I remember why I did this now. My thinking basically boils down to the following: `validateFile` (even if not used anywhere else currently) is the underlying logic, independent of any file selection event. `validateFileOnSelect` is the way we happen to use it in our UI, and this method (and its tests) alone is what we would replace if we opted to go a more "Cerebral" route; `validateFile` (and its tests) could be left alone. So this separation decouples the UI implementation from the validation logic. 

What do you think about this idea?

shared/src/business/entities/EntityConstants.ts
jtdevos: Thanks Chris - will follow-up when Tenille and I settle on a name.

jtdevos (approved):

shared/src/business/entities/EntityConstants.ts
jtdevos: Looks like we're gonna go with `https://ustaxcourt.gov/dawson_faqs_case_management.html#FileUpload`. Note that this page hasn't been updated yet, but it will be in place prior to deployment.

cypress/local-only/tests/integration/fileUpload/upload-court-issued-pdf-validation.cy.ts
```diff
@@ -0,0 +1,119 @@
+import { goToCase } from '../../../../helpers/caseDetail/go-to-case';
+
+import { MAX_FILE_SIZE_MB } from '../../../../../shared/src/business/entities/EntityConstants';
+import { attachFile } from '../../../../helpers/file/upload-file';
+import { loginAsDocketClerk1 } from '../../../../helpers/authentication/login-as-helpers';
+
+describe('upload court issued document validations', () => {
+  const docketNumber = '102-67'; // Any existing docket number works
+  const encoding = 'binary'; // Some Cypress tests were not mimicking real app behavior without manually setting the encoding to binary
```
Mwindo, line 9: The reason (I think) we need to manually set the encoding: corrupt/encrypted files make it difficult for cypress with [cypress-file-upload](https://www.npmjs.com/package/cypress-file-upload) to determine the right encoding. E.g., we want a corrupt pdf to be treated as a binary file, but because it is corrupt cypress thinks it is something else.

jimlerza (approved):

PR #5316, opened 2024-09-03
10401: BUG: Printable Case Inventory Report times out; 

cruzjone-flexion:


web-api/terraform/modules/api/pdf-generation.tf
```diff
@@ -7,8 +7,8 @@ module "pdf_generation_lambda" {
   lambda_name    = "pdf_generator_${var.environment}_${var.current_color}"
   role           = var.lambda_role_arn
   environment    = var.lambda_environment
-  timeout        = "29"
-  memory_size    = "3008"
+  timeout        = "900"
+  memory_size    = "6508"
```
Mwindo, line 11: Where does this number come from?

web-api/storage/scripts/loadTest/loadTestPdfGeneration.ts
```diff
@@ -10,22 +11,55 @@ import axios from 'axios';
...
+      data: {
+        response: any;
+      };
+    } = await axios.get(URL, {
+      headers,
+    });
+
+    if (!results) return await checkPDFComplete(asyncSyncId);
+    const { response } = results.data;
+    if (!response) return await checkPDFComplete(asyncSyncId);
```
Mwindo, line 38: Is there any issue constantly calling checkPDFComplete rather than assigning a brief wait time between calls?

Mwindo (approved):
I left a couple of questions, but this looks good to me. That said, it probably wouldn't hurt to have another set of eyes. Thanks @cruzjone-flexion!

web-api/terraform/modules/api/pdf-generation.tf
cruzjone-flexion: 900 is the same timeout for async lambdas
6GB is the Memory that made this lambda capable of handling the 720 page PDF so added an extra half GB

web-api/storage/scripts/loadTest/loadTestPdfGeneration.ts
cruzjone-flexion: There is a brief 1 second wait time in the beginning of the function 
image

Mwindo: 🤦 Thanks!

tsconfig.json
```diff
@@ -17,7 +17,7 @@
       "jest",
       "node"
     ],
-    "target": "es6",
```
jtdevos: Does this the change to ES2020 target affect the client-side code?  If so do the major browsers all support ES2020?

jtdevos (changes requested):
Not requesting any changes per se, but holding pending my question re: ES2020 transpilation.

web-api/storage/scripts/loadTest/loadTestPdfGeneration.ts
```diff
@@ -10,22 +11,55 @@ import axios from 'axios';
...
+
+    const responseObj = JSON.parse(response);
+    if (+responseObj.statusCode === 200) return;
```
jimlerza, line 41: Is this shorthand for converting the `statusCode` string to a number?

whoa.gif

jimlerza (approved):

jtdevos (approved):

PR #5327, opened 2024-09-04
Devex Make explicit unified logger

zachrog:
## What Is In This Change
- Create a singleton logger for web-api that all can import from without going through appContext
- Replaces the singleton logger which was created in express middleware and passed around through appContext
- This DOES NOT change what we are logging or how. The interface and information is purposefully kept the same. Refactoring can be done for what info we are logging but that is not the goal of this PR.

## Why This Change
- Whenever we use applicationContext.logger.info it includes data about what authenticated user is currently running the function. This feature however is tied up inside of applicationContext.
- This change preserves the ability to attach extra context to every logger.info() call without the need to go through applicationContext.
- This is the last piece of stateful information inside web-api applicationContext and allows us to freely move away from appContext if we choose.


web-api/storage/scripts/createUsers.ts
```diff
@@ -6,15 +6,22 @@ import {
...
 
-  const applicationContext = createApplicationContext({
-    role: ROLES.admin,
+  const applicationContext = createApplicationContext();
+  getLogger().addUser({
+    user: {
+      email: 'system@ustc.gov',
+      name: 'ustc automated system',
+      role: ROLES.admin,
+      userId: 'N/A',
```
Mwindo, line 23: 👍

web-api/src/createLogger.ts
```diff
@@ -37,12 +38,17 @@ export const removeDuplicateLogInformation = format(logEntry => {
   return copy;
 });
 
-export const createLogger = (opts = {}) => {
+export const createLogger = (opts: LoggerOptions = {}) => {
   const options = {
     defaultMeta: {},
-    level: opts.logLevel || process.env.LOG_LEVEL || 'debug',
+    level: opts.level || process.env.LOG_LEVEL || 'debug',
```
Mwindo, line 44: 👍

Mwindo (approved):
Looks good for testing to me--thanks for tackling this @zachrog!

TomElliottFlexion (approved):

PR #5328, opened 2024-09-05
10282: design debt

akuny:
Slightly bump up the minimum width of the `.usa-date-picker` from 245px to 250px class so that "September 2024" appears on a single line.

### DIFF
--- a/web-client/src/styles/forms.scss
+++ b/web-client/src/styles/forms.scss
@@ -668,7 +668,7 @@ label.ustc-upload {
 }
 
 .usa-date-picker {
-  min-width: 245px;
+  min-width: 250px;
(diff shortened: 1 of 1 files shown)


Mwindo (approved):
LGTM! This is probably the simplest fix given that we already use a min-width, and I doubt this will affect the layout much. In the spirit of "walk multiple paths," here is another option, which looks ever so slightly different: 

```
.usa-date-picker__calendar__month-label {
  display: flex;
  justify-content: center;
}
```

Thanks @akuny!

TomElliottFlexion (approved):

nechama-krigsman (approved):

PR #5332, opened 2024-09-05
10455 Story: Add/Remove/Modify Judge Users Updates

Mwindo:
This represents a few fixes after more testing:

- I was sending not one, but two emails on adding a new judge.
- I forgot to make the password temporary (because `add-user.ts` doesn't do so ... but that's because add-user isn't used in production).
- Phone numbers were cleared in update-judge if the phone parameter wasn't set.
- I created a new `getNewPasswordForEnvironment` function to randomize the temporary password in production.


scripts/user/add-judge.ts
```diff
@@ -110,14 +110,11 @@ requireEnvVars(['ENV']);
 
   console.log('Adding user information to Dynamo and Cognito ... ');
   const { userId } = await createOrUpdateUser(applicationContext, {
-    password: environment.defaultAccountPass,
-    setPasswordAsPermanent: true,
+    password: getNewPasswordForEnvironment(),
+    setPasswordAsPermanent: false,
```
Mwindo, line 114: Make the new password temporary.

scripts/user/add-user.ts
```diff
@@ -104,6 +104,18 @@ const checkParams = ({
...
+    await applicationContext.getCognito().adminCreateUser({
+      MessageAction: 'RESEND',
+      UserPoolId: environment.userPoolId,
+      Username: email.toLowerCase(),
+    });
+  } catch (err) {
+    console.error('Error sending welcome email', err);
+  }
+};
+
```
Mwindo, line 118: This, along with the deletion of email-helpers.ts, is reverting an unnecessary change I made.

scripts/user/make-new-password.ts
```diff
@@ -0,0 +1,37 @@
+import { environment } from '@web-api/environment';
+import { shuffle } from 'lodash';
+
+export const makeNewPassword = (): string => {
```
Mwindo, line 4: Moved from rotate-environments-secrets.ts

scripts/user/make-new-password.ts
```diff
@@ -0,0 +1,37 @@
...
+  // build the password
+  for (let i = result.length; i <= passwordLength; i++) {
+    result += getRandomChar(allCharacters);
+  }
+
+  // shuffle the password
+  return shuffle(result.split('')).join('');
+};
+
+export const getNewPasswordForEnvironment = (): string => {
```
Mwindo, line 32: New function to get a temporary password

scripts/user/update-judge.ts
```diff
@@ -145,7 +145,7 @@ const updateDynamoJudgeUserRecord = async ({
   dynamoUser.name = updates.name || dynamoUser.name;
   dynamoUser.judgePhoneNumber = updates.phone
     ? updates.phone
-    : dynamoUser.phone;
+    : dynamoUser.judgePhoneNumber;
```
Mwindo, line 148: Whoops

scripts/user/add-user.ts
```diff
@@ -126,9 +138,9 @@ const checkParams = ({
   checkParams({ params, validChambersSections });
   environment.userPoolId = await getUserPoolId();
   await createOrUpdateUser(applicationContext, {
-    password: environment.defaultAccountPass,
+    password: getNewPasswordForEnvironment(),
     setPasswordAsPermanent: true,
```
Mwindo, line 142: Is setPasswordAsPermanent right? I just noticed this while testing, and I think this should be false.

jimlerza (approved):

PR #5340, opened 2024-09-09
Opex: fix setup-glued-judges

jimlerza:
Don't bail if we couldn't find a cognito user for a newly-glued judge, just create one.


TomElliottFlexion (approved):

Mwindo (approved):
Thanks @jimlerza!

zachrog (approved):

PR #5344, opened 2024-09-09
10427 Adjust padding of inline links in petition generation flow

En-8:
flexion#10427


web-client/src/ustc-ui/InlineLink/InlineLink.tsx
```diff
@@ -0,0 +1,41 @@
...
+  icon,
+  iconSize = '1x',
+  tooltip,
+}: {
+  children: string | React.ReactNode;
+  className?: string;
+  icon?: IconProp;
+  href: string;
+  iconSize?: '1x' | '2x' | '3x' | '4x' | '5x';
+  tooltip?: string;
```
Mwindo, line 19: Might be worth moving out as a separate interface for readability, but no big deal either way.

Mwindo (approved):
LGTM! Thanks, @En-8 :)

web-client/src/ustc-ui/InlineLink/InlineLink.tsx
En-8: I'm personally not all that opinionated on whether inline types or an extracted type is more readable in this file. Extracting the type though does mean you lose the more granular type information on hover where this component is referenced. E.g. if I make this "InlineLinkProps" then hover over a use of InlineLink, all I see is...
Screenshot 2024-09-18 at 8 34 51 AM

But if I keep it inline I get...
Screenshot 2024-09-18 at 8 34 25 AM

I could see how this becomes a potential downside for a component with many props, but for one with relatively few it seems a decent QoL improvement. Happy to chat more if you have other feels or if I'm missing anything 😄

Mwindo: I'm happy either way, and you have good reasons here :)

jtdevos (approved):

PR #5350, opened 2024-09-10
Update phone number for Judge Way now that we have it 

Mwindo:
I was hoping to put this in 10455, but it likely won't be ready in time for deploy. In the meantime, let's get Judge Way's phone number into the system.


TomElliottFlexion (approved):

Mwindo:
I've just realized this change will conflict with 10455 changes in test. I will deploy to an experimental environment instead.

PR #5353, opened 2024-09-10
10427 story v2 intermediate branch to test 1725995155

En-8:
Allow text wrapping for InlineLinks, and a couple very minor tweaks to force inline links onto their own lines in some spots.

I just removed the text wrapping configurability on the component entirely since it's not needed at the moment -- if we get to a point where it is needed we can pretty easily add it back in.


akuny (approved):

TomElliottFlexion (approved):

web-client/src/views/Dashboards/FilingFeeOptions.tsx
```diff
@@ -51,6 +51,7 @@ export const FilingFeeOptions = function () {
             <br />
             <br />
             <strong>Can’t afford to pay the filing fee?</strong>
+            <br />
             <InlineLink
```
nechama-krigsman, line 55: Should the prop be removed here?

En-8: Ah yes, good catch. Sorry I didn't see this before I merged! I've made the corresponding change in the staging PR.

nechama-krigsman: 👍 Didnt realize it was merged already!

PR #5365, opened 2024-09-12
 10001 (and 10002 and 10003) Design Debt: More Granular File Validation Errors

Mwindo:
**TL;DR:** When something goes wrong with a (typically PDF) file upload, the user doesn't know until we try to parse it on the backend, at which point we only display a generic error. This means 1) the error is often not specific enough to be helpful and 2) the error might come at the end of a long form rather than as soon as the user selects the file. This PR addresses the issue.

This should free up the court's time addressing frequent help tickets (multiple per day) regarding failed file uploads. It will also prevent corrupted/invalid files from being uploaded to S3.

---

**N.B.: The PR diff is not as intimidating as it looks. A lot of the diff is refactoring.**

Tickets: [10001](https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/gh/flexion/ef-cms/10001) (password-protected PDFs), [10002](https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/gh/flexion/ef-cms/10002) (invalid files types), [10003](https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/gh/flexion/ef-cms/10003) (invalid/corrupted PDFs). (But note that this PR is best summarized as "implement client-side file validation." )

Figma [here](https://www.figma.com/design/pBCkGpOvCG8e72iNnDB2a0/USTC-%7C-File-Upload%3A-Helpful-Error-Messaging-%7C-10001%2C-10002%2C-10003-(Updated)?node-id=0-1&node-type=canvas&t=i60mADqkwUn4qiSd-0).

### The Approach in This PR

In order to provide the user immediate feedback (rather than waiting until they try to submit), we use pdf-js client-side on file selection to check for issues. If issues _are_ discovered--e.g., the PDF is password-protected or the wrong file type--we show a modal with an error message and clear the file selection. (While validating--which is usually very quick--we display a loading overlay.) This sidesteps difficulties with joi document validation.

In addition, we send an error message to the backend for logging. This error message is slightly decoupled from what is shown to the user in order for us to have finer-grained control over what is logged. (E.g., while we may display the same generic error message for multiple types of file parsing errors, we log them more granularly for debugging/prod-support purposes.)

While the front-end messaging has been overhauled, this PR does not change our back-end PDF validation implementation at all. In other words, the client-side validation is there for convenience to filter out bad documents, but we still validate on the back end.

The validation has been added to (almost) everywhere we upload files. (I was told to ignore `PractitionerAddEditDocumentation`, which might be changed soon, and `PetitionQcScanBatchPreview`, which, despite some code for uploading, doesn't seem to actually allow uploads.)

### Work left to do

File selection components can be refactored. I think this is probably worth attacking in a separate unit of work.

### Examples of New Error Messages

https://github.com/user-attachments/assets/9eb7cdc7-67b5-4da5-87a0-08ad36ccd7b7


web-client/src/views/FileDocument/StateDrivenFileInput.tsx
```diff
@@ -48,6 +57,53 @@ export const StateDrivenFileInput = connect<
 
     const fileOnForm = file || form[fileInputName] || form.existingFileName;
 
+    // Setting the filename here so that we can display it before validation
+    // finishes, otherwise a slow machine might have slight lag and allow the user
+    // to "choose file" again.
+    const [selectedFilename, setSelectedFilename] = useState('');
```
Mwindo, line 63: See https://github.com/ustaxcourt/ef-cms/pull/5312/files#r1752579003 for why this is here. Happy to see alternatives.

web-client/src/views/FileDocument/StateDrivenFileInput.tsx
```diff
@@ -48,6 +57,53 @@ export const StateDrivenFileInput = connect<
...
+
+    const onFileSelectionChange = async (
```
Mwindo, line 65: **TL;DR:** I don't think there is a good way to implement this without refactoring the entire file (along with `ScanBatchPreview.tsx` and `PetitionQxScanBatchPreview.tsx`). I have opted to take the "least-resistance" approach and [open a ticket to refactor](https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/gh/flexion/ef-cms/10486).

**More details:** 

The main things I dislike with this approach: 1) the `setIsLoadingSequence` and `setIsNotLoadingSequence`; 2) the mixture of Cerebral and DOM operations (which is nonetheless almost inevitable, given we have to set `e.target.value = ''` somehow); 3) `selectedFilename`, which is basically a hack to make the view nicer for slower machines.

This `onFileSelectionChange` is not a very "Cerebral" way to do things, which would involve using a sequence (although we would still need to do a little DOM manipulation to unset `e.target.value` on error.) This setup is just using a (n almost!) pure TS function with `onSuccess` and `onError` handles vaguely reminiscent of something like `useQuery`.

I have heard mixed opinions about making this more Cerebral-esque. I spent a few hours going that direction to see what it would look like. Right now, my overall feeling is that the advantages of moving things into Cerebral in this case (consistency, extensibility of sequences, doing more outside of the component rather than within it, being able to use the loading decorator) do not necessarily outweigh the disadvantages (opacity of types, potential over-abstraction, relying on state management library when we don't have to, slightly more time to finish). If enough people strongly disagree, I can switch to my Cerebral-work branch and continue that direction.

There are certainly ways to improve this: my aim in this unit of work was to keep existing logic mostly intact and just add the validation.

web-client/src/views/FileHandlingHelpers/cloneFile.error.test.ts
```diff
@@ -29,6 +29,7 @@ describe('cloneFile', () => {
   it('should fail when attempting to clone something other than a file', async () => {
     let error;
     try {
+      // @ts-ignore
```
Mwindo, line 32: I can probably just get rid of this test given the typing should enforce the intended behavior.

cypress/deployed-and-local/integration/advancedSearch/search.cy.ts
```diff
@@ -104,10 +105,11 @@ describe('Advanced Search', () => {
...
       cy.get('[data-testid="upload-description"]').type(opinionTitle);
-      cy.get('[data-testid="primary-document-file"]').attachFile(
-        '../../helpers/file/sample.pdf',
-      );
-      cy.get('[data-testid="upload-file-success"]').should('exist');
+      attachFile({
+        filePath: '../../helpers/file/sample.pdf',
+        selector: '[data-testid="primary-document-file"]',
+        selectorToAwaitOnSuccess: '[data-testid^="upload-file-success"]',
+      });
```
Mwindo, line 112: Because validating on file selection introduces a slight delay and causes tests to fail, I created this helper function to ensure our Cypress tests wait for success appropriately. A good amount of the git diff in this PR is just refactoring `cy.get(...).attachFile` to use this helper and renaming `uploadFile` to `attachSamplePdfFile` to avoid confusion with `attachFile`.

shared/src/business/entities/EntityConstants.ts
```diff
@@ -1733,3 +1733,8 @@ export const STATUS_REPORT_ORDER_OPTIONS = {
     restored: 'restoredToGeneralDocket',
   },
 };
+
+export const TROUBLESHOOTING_INFO = {
+  APP_SUPPORT_EMAIL: 'dawson.support@ustaxcourt.gov',
```
Mwindo, line 1738: Pulling this out as a constant since I see it being used all over the place. This small change constitutes another non-trivial portion of the diff.

web-client/src/views/AppComponent.tsx
```diff
@@ -302,6 +303,7 @@ export const AppComponent = connect(
         {showModal === 'WebSocketErrorModal' && <WebSocketErrorModal />}
         {showModal === 'AppMaintenanceModal' && <AppMaintenanceModal />}
         {showModal === 'GenericErrorModal' && <GenericErrorModal />}
+        {showModal === 'FileUploadErrorModal' && <FileUploadErrorModal />}
```
Mwindo, line 306: This was being called in many different components, often like `<FileUploadErrorModal confirmSeqence={something} />`. However, 1) now this modal is used more often, so it makes sense to centralize it by placing it in the React hierarchy tree here (in fact, I think most modals are best done that way to avoid multiple components potentially rendering the same modal) and 2) `confirmSeqence={something}` was not actually doing anything, so getting rid of it in the components that used it was perfectly safe.

cypress/local-only/support/pages/start-a-case.ts
```diff
@@ -7,7 +7,7 @@ export const fillInAndSubmitForm = () => {
 
   // wizard step 1
   cy.get('input#stin-file').should('be.enabled').attachFile(w3Dummy);
-  cy.get('[data-testid="upload-file-success"]');
+  cy.get('[data-testid^="upload-file-success"]');
```
jimlerza, line 10: This comment was originally

> Is this a stray character?

but then I learned about the "starts with" CSS selector! Neat!

cypress/local-only/tests/integration/fileUpload/upload-court-issued-pdf-validation.cy.ts
```diff
@@ -0,0 +1,98 @@
...
+  });
+
+  it('should display error modal when a file larger than the limit is selected', () => {
+    cy.intercept('POST', '/logError').as('logErrorRequest');
+    const largeFile = new Blob(
+      [new ArrayBuffer((MAX_FILE_SIZE_MB + 1) * 1024 * 1024)],
+      {
+        type: 'application/pdf',
+      },
+    );
```
jimlerza, line 71: I definitely appreciate that we're not adding a 251MB file into the repo for this!

akuny (approved):

web-client/src/views/FileHandlingHelpers/fileValidation.ts
```diff
@@ -0,0 +1,204 @@
...
+  }
+  const correctFileValidation = validateCorrectFileType({
+    allowedFileExtensions,
+    file,
+  });
+  if (!correctFileValidation.isValid) {
+    return correctFileValidation;
+  }
+  if (file.type === 'application/pdf') {
+    return await validatePdf({ file });
```
jimlerza, line 162: I have a couple of questions on this piece -

* Can we trust this to always be populated by the built-in nodejs `fs` package? (false positives)
* Are there any legitimate PDF files (hi-res, interactive, etc) that don't get this type? (false negatives)

jimlerza (approved):
This looks great!

web-client/src/views/FileHandlingHelpers/fileValidation.ts
jimlerza: To clarify - I think I'm asking "is it detecting the actual mimetype of the file blob or is it just reading a header or something?" on that first point but I can most certainly research that on my own!

Mwindo: Good question!

In this case, I don't think node's `fs` is involved. The file type is being determined by the browser, usually via the `File` API, which typically uses the file extension but may also infer the MIME type from the file's header. This detection _isn't_ always perfect. A legitimate PDF saved with a wrong extension (e.g., .csv) will likely not be recognized as application/pdf (and my earlier file extension check will return invalid even before this line). Conversely, a file mislabeled as a PDF (e.g., a .csv file renamed to .pdf) might pass the MIME type check, but then the PDF validation logic would catch the mismatch and flag it as invalid.

Regarding different types of PDFs (such as PDF/A or PDF/E), our current validation logic doesn't support them, so files with those specific formats may fail validation, even though they are legitimate PDFs. (But this is already the case!)

Does that make sense and answer your questions?

jimlerza: Yes it does, thanks!

PR #5371, opened 2024-09-13
10074 bug: Auto Scroll Issues on Petitions Clerk Create Case Form

Mwindo:
On the form for petitions clerk creating a case, we validated on blur, on change, etc., and this validation would automatically (and disruptively) scroll to the top to show the validation errors message box. This PR adds a `preventAutoScroll` flag (or rather revises an existing flag that was not fully implemented) in order to prevent this scrolling behavior.

---

A couple of notes:

1. It would be very easy to get carried away refactoring. I tried to find a middle ground: I updated onBlur and onChange to take in functions rather than strings as needed, updated a few components that caused typing issues as needed, but otherwise kept things the same.
2. I decided that `preventAutoScroll: false` would be the default even if it turns out that we need `preventAutoScroll: true` in most instances. (The only time it makes sense to scroll, I think, is on submission.) This was to keep behavior the same so that developers are not surprised about new behavior.

### DIFF
--- a/web-client/src/presenter/computeds/alertHelper.ts
+++ b/web-client/src/presenter/computeds/alertHelper.ts
@@ -15,7 +15,6 @@ export const alertHelper = (get: Get): any => {
 
   return {
     messagesDeduped: uniq(alertError.messages).filter(Boolean),
-    preventAutoScroll: false,
     responseCode: alertError.responseCode,
     showErrorAlert:
       !!alertError.title || !!alertError.message || !!alertError.messages,
--- a/web-client/src/views/CaseDetail/PetitionPaymentForm.tsx
+++ b/web-client/src/views/CaseDetail/PetitionPaymentForm.tsx
@@ -6,6 +6,171 @@ import { sequences, state } from '@web-client/presenter/app.cerebral';
 import React from 'react';
 import classNames from 'classnames';
 
+interface ComponentProps {
+  onUpdate: (args: Record<string, any>) => void;
+  onValidate: (args: Record<string, any>) => void;
+}
+
+// It would be better to use ComponentProps & CerebralProps, but
+// typing won't work nicely with CerebralProps. Hence the & any.
+const PetitionPaymentFormComponent: React.FC<ComponentProps & any> = ({
+  bind,
+  DATE_FORMATS,
+  formatAndUpdateDateFromDatePickerSequence,
+  onUpdate,
+  onValidate,
+  paymentStatus,
+  validationErrors,
+}) => {
+  return (
+    <>
+      <h4 className="margin-bottom-2">Petition Fee</h4>
+      <FormGroup errorText={validationErrors.petitionPaymentStatus}>
+        <fieldset className="usa-fieldset">
+          <legend className="usa-legend" id="fee-paid-legend">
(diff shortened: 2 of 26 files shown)


web-client/src/presenter/computeds/alertHelper.ts
```diff
@@ -15,7 +15,6 @@ export const alertHelper = (get: Get): any => {
 
   return {
     messagesDeduped: uniq(alertError.messages).filter(Boolean),
-    preventAutoScroll: false,
```
Mwindo, line 18: Not terribly helpful since it was always false!

web-client/src/views/CaseDetailEdit/PartyInformation.tsx
```diff
@@ -87,7 +89,7 @@ export const PartyInformation = connect(
               showPrimaryContact={caseDetailEditHelper.showPrimaryContact}
               showSecondaryContact={caseDetailEditHelper.showSecondaryContact}
               useSameAsPrimary={true}
-              onBlur="validateCaseDetailSequence"
+              onBlur={validateCaseDetailSequence}
```
Mwindo: Todo: Probably should have preventAutoScroll: true

web-client/src/views/StartCase/CaseTypeSelect.tsx
```diff
@@ -69,7 +69,7 @@ export const CaseTypeSelect = connect(
                   value: e.target.value,
                 });
                 if (onChangePreValidation) onChangePreValidation();
-                if (validation) validation();
+                if (validationFunction) validationFunction();
```
Mwindo: Todo: Probably should have preventAutoScroll: true

web-client/src/views/StartCase/ContactPrimary.tsx
```diff
@@ -13,7 +13,7 @@ import React from 'react';
 const props = cerebralProps as unknown as {
   contactsHelper: string;
   bind: string;
-  onBlur: string;
+  onBlur: () => void;
```
Mwindo: Todo

web-client/src/views/StartCase/InternationalAddress.tsx
```diff
@@ -7,7 +7,7 @@ import React from 'react';
 
 const props = cerebralProps as unknown as {
   bind: string;
-  onBlur: string;
+  onBlur: any;
```
Mwindo: todo

web-client/src/views/CaseDetail/PetitionPaymentForm.tsx
```diff
@@ -6,172 +6,182 @@ import { sequences, state } from '@web-client/presenter/app.cerebral';
 import React from 'react';
 import classNames from 'classnames';
 
-export const PetitionPaymentForm = connect(
```
Mwindo, line 9: I refactored this file slightly (the diff makes it look like a bigger refactor than it is) as an experiment to reduce typescript errors (`validateSequence` and `updateSequence` were unrecognized).

Mwindo:
This is just clutter currently. I will reopen as needed, although I might just re-create the branch given the number of changes I made to the test branch.

PR #5373, opened 2024-09-16
10001 (and 10002 and 10003) Design Debt Updates

Mwindo:
The validation in the original PR for 10001 et al. correctly caught that [corrupt-pdf](https://github.com/ustaxcourt/ef-cms/blob/staging/cypress/helpers/file/corrupt-pdf.pdf) is corrupt in [the Cypress test](https://github.com/ustaxcourt/ef-cms/blob/2ffbdef46f2ca79333f27f4005978cd30f7c6018/cypress/local-only/tests/integration/fileUpload/upload-court-issued-pdf-validation.cy.ts#L34-L47) but did _not_ catch it outside of Cypress.

This PR thus does two things:

1. Ensures that our PDF file uploads within Cypress mimic the file uploads outside of Cypress by specifying that the upload is in binary format.
2. Implements an unfortunate solution--but as far as I can tell the only viable solution, unless we use something other than pdfjs--to catch corrupt but not fully invalid PDFs.

### DIFF
--- a/web-client/src/views/FileHandlingHelpers/pdfValidation.ts
+++ b/web-client/src/views/FileHandlingHelpers/pdfValidation.ts
@@ -38,10 +38,39 @@ export const validatePdf = ({
 
       const fileAsArrayBuffer = new Uint8Array(result as ArrayBuffer);
 
-      // We try to load the pdf. If we get any errors, we return an errorInformation accordingly.
+      // eslint-disable-next-line spellcheck/spell-checker
+      // As of 2024 September, pdfjs can sometimes render PDFs with invalid characters,
+      // and it only indicates this by logging a warning via console.log.
+      // These PDFs are corrupt and should not be allowed to be uploaded.
+      // Therefore, to catch these warnings, we will temporarily override console.log.
+      const consoleLog = console.log;
+
+      // We will try to load the PDF. If we get any errors, we will return an errorInformation object accordingly.
       try {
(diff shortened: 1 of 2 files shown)


web-client/src/views/FileHandlingHelpers/pdfValidation.ts
```diff
@@ -38,10 +38,39 @@ export const validatePdf = ({
...
+          isEvalSupported: false,
+        }).promise;
+
+        // We raise a custom error even if the load was successful
+        // when invalid hex string warnings were caught
+        if (pdfIsCorrupt) {
+          const corruptPdfError = new Error('PDF has invalid characters');
+          corruptPdfError.name = 'CorruptPDFException';
+          throw corruptPdfError;
+        }
```
Mwindo, line 72: Here is the solution. I hate it.

pdfjs-dist will log a warning when it encounters invalid characters, but it will continue and try to parse the PDF anyway, sometimes successfully. For example, our corrupt-pdf.pdf cannot be opened by MacOS Preview, and our back-end library fails to parse it, but pdfjs can actually display it just fine.

There is no way for us to "capture" these pdfjs warnings in an elegant way. This forces us to temporarily override console.log, see if pdfjs logs the invalid character warning, and then (via `finally`) reset the console.log.

Mwindo:
This approach is unnecessary. I will open a better PR.

PR #5375, opened 2024-09-17
10001 (and 10002 and 10003): File Validation Updates

Mwindo:
TL;DR: This PR adds some updates to make sure that 1) Cypress attachments are mimicking actual browser behavior and 2) we validate pdf headers on the front end just like we do on the back end.

---

This PR addresses two things

1) PDFs with incorrect headers were accepted on the front end but rejected on the back end. While there might be slight differences between front-end and back-end validation, this was a particularly large and unnecessary difference.
2) The validation in [the original PR for 10001 et al.](https://github.com/ustaxcourt/ef-cms/pull/5365) correctly caught that [corrupt-pdf](https://github.com/ustaxcourt/ef-cms/blob/staging/cypress/helpers/file/corrupt-pdf.pdf) is corrupt in [the Cypress test](https://github.com/ustaxcourt/ef-cms/blob/2ffbdef46f2ca79333f27f4005978cd30f7c6018/cypress/local-only/tests/integration/fileUpload/upload-court-issued-pdf-validation.cy.ts#L34-L47) but did _not_ catch it outside of Cypress. This was because we attached files in Cypress with a different encoding than what occurs in the browser. (More details: in Cypress, because the file was loaded [with a non-binary encoding](https://www.npmjs.com/package/cypress-file-upload#working-with-file-encodings), pdfjs happened to fail to parse the PDF and therefore marked it as corrupt. Outside of Cypress, where the file bytes were loaded as a binary array, pdfjs happened to be able to parse the PDF and therefore did not mark it as corrupt. To avoid this divergence between the test environment and actual environment, I defaulted to uploading files in Cypress via binary encoding.)

### DIFF
--- a/web-client/src/views/FileHandlingHelpers/pdfValidation.ts
+++ b/web-client/src/views/FileHandlingHelpers/pdfValidation.ts
@@ -12,6 +12,13 @@ export const PDF_CORRUPTED_ERROR_MESSAGE =
 const GENERIC_FILE_ERROR_MESSAGE =
   'There is a problem uploading the file. Try again later.';
 
+export const validatePdfHeader = (pdfData: Uint8Array): boolean => {
+  const stringDecoder = new TextDecoder('utf8');
+  const pdfHeaderBytes = pdfData.slice(0, 5);
+  const pdfHeaderString = stringDecoder.decode(pdfHeaderBytes);
+  return pdfHeaderString === '%PDF-';
+};
+
 export const validatePdf = ({
   file,
 }: {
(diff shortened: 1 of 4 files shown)


web-client/src/views/FileHandlingHelpers/pdfValidation.ts
```diff
@@ -12,6 +12,13 @@ export const PDF_CORRUPTED_ERROR_MESSAGE =
 const GENERIC_FILE_ERROR_MESSAGE =
   'There is a problem uploading the file. Try again later.';
 
+export const validatePdfHeader = (pdfData: Uint8Array): boolean => {
+  const stringDecoder = new TextDecoder('utf8');
+  const pdfHeaderBytes = pdfData.slice(0, 5);
+  const pdfHeaderString = stringDecoder.decode(pdfHeaderBytes);
+  return pdfHeaderString === '%PDF-';
+};
```
Mwindo, line 20: We do this on the back end [here](https://github.com/ustaxcourt/ef-cms/blob/263df266f4d323a93eca085d9d03a85de1177272/web-api/src/business/useCases/pdf/validatePdfInteractor.ts#L33-L52). I was hoping to have a shared function so that both places used the same logic, but there were enough differences (mainly logging and using a different decoder class) that doing so proved more complicated rather than less.

web-client/src/views/FileHandlingHelpers/pdfValidation.ts
```diff
@@ -38,10 +45,21 @@ export const validatePdf = ({
...
+          corruptPdfError.name = 'CorruptPDFHeaderException';
+          throw corruptPdfError;
+        }
+
+        // Attempt to load the PDF
         const pdfjs = await applicationContext.getPdfJs();
-        await pdfjs.getDocument(fileAsArrayBuffer).promise;
+        await pdfjs.getDocument({
+          data: fileAsArrayBuffer,
+          isEvalSupported: false,
```
Mwindo, line 60: Whoops, see https://github.com/ustaxcourt/ef-cms/pull/5059

jtdevos (approved):

TomElliottFlexion (approved):

PR #5387, opened 2024-09-20
10001 (and 10002 and 10003): Updates to client-side PDF Validation to mark read-only pdfs as invalid

Mwindo:
`pdfjs-dist` throws a `PasswordException` upon open, but this only checks password protection on read. This PR implements checks for any password protections on edit as well. (We need to allow edits so that DAWSON can stamp, add pages, etc.)


web-client/src/views/FileHandlingHelpers/pdfValidationHelpers.ts
```diff
@@ -0,0 +1,15 @@
+import { PDFDocumentProxy } from 'pdfjs-dist/types/src/display/api';
```
Mwindo, line 1: Pulled out these functions into a separate module for easier mocking.

web-client/src/views/FileHandlingHelpers/pdfValidationHelpers.ts
```diff
@@ -0,0 +1,15 @@
...
+  const stringDecoder = new TextDecoder('utf8');
+  const pdfHeaderBytes = pdfData.slice(0, 5);
+  const pdfHeaderString = stringDecoder.decode(pdfHeaderBytes);
+  return pdfHeaderString === '%PDF-';
+};
+
+export const validatePermissions = async (
+  pdfDocument: PDFDocumentProxy,
+): Promise<boolean> => {
+  const permissions = await pdfDocument.getPermissions();
```
Mwindo, line 13: This is null if there is no encryption/password protection for editing. Otherwise, it returns a list of numerical permission codes.

cypress/helpers/file/upload-file.ts
```diff
@@ -23,7 +23,7 @@ export function attachFile({
   selector: string;
   filePath: string;
   selectorToAwaitOnSuccess?: string;
-  encoding?: 'binary'; // Default to binary, which better mimics PDF data handling on the browser
```
Mwindo, line 26: This comment was wrong 🤦

jimlerza (approved):

PR #5397, opened 2024-09-24
10261 Story: Practitioner Case List

Mwindo:
## Description

This PR adds practitioner open and closed case lists for internal DAWSON users. (Ticket [here](https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/gh/flexion/ef-cms/10261))

Screenshot 2024-09-26 at 7 25 57 AM

Screenshot 2024-09-26 at 7 26 05 AM

---

### Overview of changes:

- Adds a `PractitionerCaseList` component and `PractitionerCaseIcons` component. (The latter can probably be extended beyond practitioners and used elsewhere? Icons seem to be rendered differently in different areas of the app.)
- Pulls out some logic from `generatePractitionerCaseListPdfInteractor` into a new interactor, `getPractitionerCasesInteractor`, which is also used in the actions that set up state for `PractitionerCaseList`.
- Adds typing for `state.practitionerDetails`.

### Notes:

- A small number of practitioners have thousands of cases. After tests on the test environment, I determined that back-end pagination approaches were not worth the cost to implement: The worst-case page load time was a few seconds for me (even with poor network speed), and the practitioner with the most cases (4000+) is still be far from our response size limit (~.8 MB out of 6 MB). See conversation [here](https://github.com/ustaxcourt/ef-cms/pull/5430#discussion_r1793782990). Since that comment, I have verified the `responseSize` number for this user in our logs:

Screenshot 2024-10-18 at 10 09 16 AM


shared/src/business/useCases/getPractitionerCasesInteractor.ts
```diff
@@ -0,0 +1,50 @@
...
+import { UnauthorizedError } from '@web-api/errors/errors';
+import { UnknownAuthUser } from '@shared/business/entities/authUser/AuthUser';
+import { partition } from 'lodash';
+
+export const getPractitionerCasesInteractor = async (
+  applicationContext: ServerApplicationContext,
+  { userId }: { userId: string },
+  authorizedUser: UnknownAuthUser,
+) => {
+  console.log('getPractitionerCasesInteractor?!');
```
Mwindo: Remove

shared/src/business/useCases/getPractitionerCasesInteractor.ts
```diff
@@ -0,0 +1,50 @@
...
+    .getDocketNumbersByUser({
+      applicationContext,
+      userId,
+    });
+
+  const cases = await applicationContext
+    .getPersistenceGateway()
+    .getCasesByDocketNumbers({ applicationContext, docketNumbers });
+
+  console.log('CASES?!', cases);
```
Mwindo: Remove

web-client/src/views/Practitioners/PractitionerInformation.tsx
```diff
@@ -11,17 +14,63 @@ import React from 'react';
...
+            totalPages={practitionerInformationHelper.totalClosedCasesPages}
+            onPageChange={selectedPage => {
+              setPractitionerClosedCasesPageSequence({
+                pageNumber: selectedPage,
+              });
+            }}
+          />
+        )
+      );
+    };
```
Mwindo, line 72: Ideally, I would want pagination to live on `PractitionerCaseList` directly and be handled by using `useState` and URL params; however, this goes against our current use of Cerebral.

I chose what I thought was the next best option: Rather than pass in specific global state into `PractitionerCaseList`, which makes the component "messier," I chose to pass in the list of cases into `PractitionerCaseList` and handle pagination here in this component since it is already connected to global state. This keeps us in line with using Cerebral but also keeps `PractitionerCaseList` pretty simple and reusable.

shared/src/business/useCases/generatePractitionerCaseListPdfInteractor.test.ts
```diff
@@ -16,6 +16,11 @@ describe('generatePractitionerCaseListPdfInteractor', () => {
   });
 
   it('returns an unauthorized error on non internal users', async () => {
+    applicationContext
+      .getUseCases()
+      .getPractitionerCasesInteractor.mockImplementation(() => {
+        throw new Error('Unauthorized');
+      });
```
Mwindo, line 23: The unauthorized error is now thrown on `getPractitionerCasesInteractor`. That said, I think it is still important to test that `generatePractitionerCaseListPdfInteractor` independently "bubbles up" that error.

shared/src/business/useCases/generatePractitionerCaseListPdfInteractor.test.ts
```diff
@@ -82,75 +91,19 @@ describe('generatePractitionerCaseListPdfInteractor', () => {
     ).rejects.toThrow('Practitioner not found');
   });
 
-  it('sorts open and closed cases before sending them to document generator', async () => {
```
Mwindo, line 85: A version of this test has been moved to `getPractitionerCasesInteractor`.

jimlerza (approved):

cypress/local-only/tests/integration/advancedSearch/practitioner-information.cy.ts
```diff
@@ -0,0 +1,37 @@
...
+  describe('Open Cases Tab', () => {
+    it('should display number of cases in the tab header and list the correct number of open cases', () => {
+      cy.get('#tabButton-practitionerOpenCases').click();
+      cy.contains('#tabButton-practitionerOpenCases', /\(\d+\)/)
+        .invoke('text')
+        .then(text => {
+          const match = text.match(/\d+/);
+          const numberInParentheses = parseInt(match![0], 10); // Extract and convert the header number
+          // The seeded data has at least 7 cases, but other tests sometimes add them.
+          expect(numberInParentheses).to.be.greaterThan(6);
```
zachrog: Hey I can understand why you would add this because you want to make sure that we are displaying the correct number but this expectation relies on seed data and our cypress tests should ideally never rely on seed data now.

Maybe there is another way to write this such that it does not know about the number of cases in the seed data.

shared/src/business/entities/User.ts
```diff
@@ -13,6 +13,18 @@ import { JoiValidationEntity } from '@shared/business/entities/JoiValidationEnti
 import { formatPhoneNumber } from '../utilities/formatPhoneNumber';
 import joi from 'joi';
 
+export interface UserContact {
```
zachrog: Nit: Typically we use types instead of interfaces

shared/src/proxies/practitioners/getPractitionerCasesProxy.ts
```diff
@@ -0,0 +1,11 @@
+import { get } from '../requests';
+
+export const getPractitionerCasesInteractor = (
+  applicationContext,
+  { userId },
+) => {
```
zachrog: Could you add a type here?

shared/src/business/entities/cases/Case.ts
```diff
@@ -237,17 +237,10 @@ export class Case extends JoiValidationEntity {
...
-      // compare years if they aren't the same;
-      // compare as strings, because they *might* have suffix
-      return aSplit[1].localeCompare(bSplit[1]);
-    } else {
-      // compare index if years are the same, compare as integers
-      return +aSplit[0] - +bSplit[0];
-    }
+    return (
+      (Case.getSortableDocketNumber(docketNumberA) || 0) -
+      (Case.getSortableDocketNumber(docketNumberB) || 0)
```
TomElliottFlexion, line 242: :tada::tada::tada:

shared/src/business/useCases/generatePractitionerCaseListPdfInteractor.test.ts
TomElliottFlexion: Tiniest of nits: could this also `throw new UnauthorizedError('Unauthorized to view practitioners cases');`? Wouldn't even be worthy of another commit, just if you happen to change something else

Mwindo:
@zachrog voiced a concern about the cypress tests using seed data, which is reasonable and a good catch. Thank you! I will put into draft mode and address.

web-client/src/presenter/computeds/practitionerInformationHelper.test.ts
```diff
@@ -1,9 +1,64 @@
...
+} from '@shared/business/entities/EntityConstants';
+import { PractitionerDetail } from '@web-client/presenter/state';
 import { applicationContext } from '../../applicationContext';
 import { practitionerInformationHelper as practitionerInformationHelperComputed } from './practitionerInformationHelper';
 import { runCompute } from '@web-client/presenter/test.cerebral';
+import { v4 } from 'uuid';
 import { withAppContextDecorator } from '../../withAppContext';
 
 describe('practitionerInformationHelper', () => {
+  const pageSize = 100; // This should match the page size in practitionerInformationHelper
```
TomElliottFlexion: Another tiny nit: could this test and `practitionerInformationHelper` use the same exported `PAGE_SIZE` constant?

web-client/src/presenter/computeds/practitionerInformationHelper.test.ts
```diff
@@ -17,9 +72,248 @@ describe('practitionerInformationHelper', () => {
...
+    },
+  );
+
+  it.each([
+    [0, 0],
+    [1, 1],
+    [pageSize, 1],
+    [pageSize + 1, 2],
+    [pageSize * 2 + 1, 3],
+  ])(
```
TomElliottFlexion: Niiice!

web-client/src/presenter/computeds/practitionerInformationHelper.test.ts
Mwindo: Sure! I always waffle between exporting the constant (which keeps the test coupled to the code) and not (which means we are testing that the constant is the value we expect).

TomElliottFlexion (approved):
Looks good (apart from seed data perhaps)! None of my comments are of much import

web-client/src/presenter/computeds/practitionerInformationHelper.test.ts
TomElliottFlexion: Fair point, dealer's choice!

Mwindo: I went with your suggestion, which is usually what I end up doing in these should-I-import-or-redefine cases. Maybe next time I'll just do that first lol rather than waffling. Thank you!

zachrog (approved):

cypress/local-only/tests/integration/advancedSearch/practitioner-information.cy.ts
Mwindo: Done! Thank you!

shared/src/proxies/practitioners/getPractitionerCasesProxy.ts
Mwindo: Sure thing!

shared/src/business/useCases/generatePractitionerCaseListPdfInteractor.test.ts
Mwindo: I like it. I'll keep this comment in mind.

jimlerza (approved):

shared/src/business/useCases/generatePractitionerCaseListPdfInteractor.test.ts
Mwindo: I'm slipping this update into 10461!

PR #5398, opened 2024-09-26
10261 Story: Practitioner Case List

Mwindo:
## Description

This PR adds practitioner open and closed case lists for internal DAWSON users. (Ticket [here](https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/gh/flexion/ef-cms/10261))

Screenshot 2024-09-26 at 7 25 57 AM

Screenshot 2024-09-26 at 7 26 05 AM

---

### Overview of changes:

- Adds a `PractitionerCaseList` component and `PractitionerCaseIcons` component. (The latter can probably be extended beyond practitioners and used elsewhere? Icons seem to be rendered differently in different areas of the app.)
- Pulls out some logic from `generatePractitionerCaseListPdfInteractor` into a new interactor, `getPractitionerCasesInteractor`, which is also used in the actions that set up state for `PractitionerCaseList`.
- Adds typing for `state.practitionerDetails`.

### Notes:

- A small number of practitioners have hundreds/thousands of cases. It might be worth paginating on the back end, loading data asynchronously, etc. as strategies to avoid delays in loading the practitioner information page. However, I want to test the timing of this in our test environment. I suspect that these measures are not worth the cost to implement.


web-client/src/views/Practitioners/PractitionerInformation.tsx
```diff
@@ -11,17 +14,63 @@ import React from 'react';
...
+            totalPages={practitionerInformationHelper.totalClosedCasesPages}
+            onPageChange={selectedPage => {
+              setPractitionerClosedCasesPageSequence({
+                pageNumber: selectedPage,
+              });
+            }}
+          />
+        )
+      );
+    };
```
Mwindo, line 72: Ideally, I would want pagination to live on `PractitionerCaseList` directly and be handled by using `useState` and URL params; however, this goes against our current use of Cerebral.

I chose what I thought was the next best option: Rather than pass in specific global state into `PractitionerCaseList`, which makes the component "messier," I chose to pass in the list of cases into `PractitionerCaseList` and handle pagination here in this component since it is already connected to global state. This keeps us in line with using Cerebral but also keeps `PractitionerCaseList` pretty simple and reusable.

shared/src/business/useCases/generatePractitionerCaseListPdfInteractor.test.ts
```diff
@@ -16,6 +16,11 @@ describe('generatePractitionerCaseListPdfInteractor', () => {
   });
 
   it('returns an unauthorized error on non internal users', async () => {
+    applicationContext
+      .getUseCases()
+      .getPractitionerCasesInteractor.mockImplementation(() => {
+        throw new Error('Unauthorized');
+      });
```
Mwindo, line 23: The unauthorized error is now thrown on `getPractitionerCasesInteractor`. That said, I think it is still important to test that `generatePractitionerCaseListPdfInteractor` independently "bubbles up" that error.

shared/src/business/useCases/generatePractitionerCaseListPdfInteractor.test.ts
```diff
@@ -82,75 +91,19 @@ describe('generatePractitionerCaseListPdfInteractor', () => {
     ).rejects.toThrow('Practitioner not found');
   });
 
-  it('sorts open and closed cases before sending them to document generator', async () => {
```
Mwindo, line 85: A version of this test has been moved to `getPractitionerCasesInteractor`.

akuny (approved):
This looks good! @Mwindo

PR #5403, opened 2024-09-26
#10512: attach mail from domain to identity

mmarcotte:
When reviewing DMARC reports, we are observing that our SPF alignment is not passing. That appears to be due to the fact that the identity is not configured to use a MAIL FROM domain within AWS SES. The domain is attached, but because we are creating an individual identity, we also need to associate that identity to the MAIL FROM domain. This will ensure that the SMTP server sends it on behalf of that subdomain `@from.$EFCMS_DOMAIN` rather than the current and default `@amazonses.com`.

flexion#10512

Looking good after deploying this to exp1:

**AWS Console:**
Screenshot 2024-09-26 at 3 19 51 PM

**EMail Headers:**
Screenshot 2024-09-26 at 3 21 35 PM

**Dmarcian Report:**
Screenshot 2024-09-27 at 9 07 54 AM


web-api/terraform/modules/everything-else-deprecated/ses.tf
```diff
@@ -26,6 +26,11 @@ resource "aws_ses_email_identity" "ses_sender" {
   email = "noreply@${var.dns_domain}"
 }
 
+resource "aws_ses_domain_mail_from" "example" {
```
jimlerza: Is this resource name of "example" a placeholder?

jimlerza (approved):

TomElliottFlexion (approved):

zachrog (approved):

PR #5405, opened 2024-09-26
10313 Bug: Change Email Verification Link Does Not Expire

Mwindo:
The change email verification link is supposed to expire after an hour; it never expires. This PR addresses the issue.

---

Implementation notes:

- When updating a user to have a `pendingEmail` and `pendingEmailVerificationToken`, we now also set a UTC `pendingEmailVerificationTokenTimestamp`. In `verifyUserPendingEmailInteractor`, we verify that the timestamp is within 1 hour of UTC now. When we reset `pendingEmail` and `pendingEmailVerificationToken`, we also reset `pendingEmailVerificationTokenTimestamp`.


web-api/src/business/useCases/practitioner/updatePractitionerUserInteractor.ts
```diff
@@ -149,29 +155,6 @@ export const updatePractitionerUser = async (
   }
 };
 
-const updateUserPendingEmail = async ({
```
Mwindo, line 152: This function was very similar to behavior occurring in `updateUserPendingEmailInteractor`. I refactored the function slightly and put it in `updateUserPendingEmailInteractor`, where it seems a more natural fit.

TomElliottFlexion (approved):

PR #5407, opened 2024-09-27
10395: Improve textarea elements

pixiwyn:
We want textareas to 1) only resize vertically and 2) prevent resizing to a height less than one line of text. This PR adds a class to do so, and removes other, inconsistently applied classes that we had before.


web-client/src/styles/forms.scss
```diff
@@ -134,6 +132,14 @@ legend {
   }
 }
 
+textarea {
+  min-height: calc(
+    1em + 24px
+  ); // Don't allow resizing smaller than the line of text
```
Mwindo: There is nothing sacred about this. It just seemed to look good with a variety of font sizes, and most importantly the font size we use by default.

jimlerza (approved):

PR #5423, opened 2024-10-02
10395: Improve textarea elements

Mwindo:
This PR:

- Disables textarea resize from horizontal scrolling
- Prevents textarea resizing to smaller than one line of text
- Consolidates two existing and inconsistently applied CSS classes into one consistently applied CSS class


nechama-krigsman (approved):

zachrog (approved):
Okay, I gave an approval but wanted to ask something. 

Are we CERTAIN that all textareas should have this? Is this a sensible default to have?

I ask as we have added styles to ALL tables and ALL forms which should not exist. I would only add a style to all elements if 99% of the usages need that style, anything less and it should be passed in.

Mwindo:
> Are we CERTAIN that all textareas should have this? Is this a sensible default to have?
> 
> I ask as we have added styles to ALL tables and ALL forms which should not exist. I would only add a style to all elements if 99% of the usages need that style, anything less and it should be passed in.

This is a good question. Given the ticket description and uses of `textarea`, I think this _is_ a case in which we want the style 99+% of the time. That said, I think there is probably room to clean up our `textarea` classes (e.g., the existing `ustc-message-modal-text-area` is just `resize: vertical`, and there really shouldn't be a need for that now). Let me take care of that.

In the process, I'll consider using a separate CSS class, which I do generally prefer, rather than a default style. The argument for the default here is that it really _is_ a default, and it doesn't touch the page layout like margin (etc.) does. The argument against is that in the future this could be wrong; also, without a separate class, a future developer might be tempted to throw something like margin into the default style.

UPDATE: I saw two similar existing CSS classes -- `ustc-message-modal-text-area` and `resize-vertical-only` -- that did part of what we wanted (`resize: vertical`), but they seem to have been inconsistently applied and maybe forgotten about. Therefore, instead of using a default style, I kept the "existing approach" but with only one CSS class -- `textarea-resize-vertical` -- applied consistently.

Mwindo:
I've made a few small changes and am happy to address feedback on them, but for now I'm going to merge

PR #5424, opened 2024-10-02
File validation bug: Make correct file type check case insensitive

Mwindo:


web-client/src/views/FileHandlingHelpers/fileValidation.ts
```diff
@@ -22,10 +22,12 @@ export interface FileValidationResponse {
   errorInformation?: FileValidationErrorInfo;
 }
 
-function getFileExtension(filename: string) {
+export const getFileExtension = (filename: string) => {
```
Mwindo, line 25: Just keeping function declaration types in this module consistent.

web-client/src/views/FileHandlingHelpers/fileValidation.test.ts
```diff
@@ -244,6 +244,20 @@ describe('fileValidation', () => {
     expect(validationResult).toMatchObject({ isValid: true });
   });
 
+  it('BUG: should return valid for valid file with uppercase extension', async () => {
```
Mwindo, line 247: I've added a few tests and organized them slightly better.

jtdevos (approved):

jimlerza (approved):

PR #5430, opened 2024-10-04
10261 Story: Practitioner Case List (Updates based on UX feedback)

Mwindo:
- Large case numbers can trigger a 502. This PR (theoretically) fixes that.
- The Docket No. column no longer wraps.
- The old top margin is preserved for the non-case-list tab contents; the new top margin is applied only for the case-list tab contents.

### DIFF
--- a/shared/src/business/useCases/getPractitionerCasesInteractor.ts
+++ b/shared/src/business/useCases/getPractitionerCasesInteractor.ts
@@ -7,6 +7,7 @@ import {
 import { ServerApplicationContext } from '@web-api/applicationContext';
 import { UnauthorizedError } from '@web-api/errors/errors';
 import { UnknownAuthUser } from '@shared/business/entities/authUser/AuthUser';
+import { formatCase } from '@shared/business/utilities/getFormattedCaseDetail';
 import { partition } from 'lodash';
 
 export const getPractitionerCasesInteractor = async (
@@ -31,11 +32,8 @@ export const getPractitionerCasesInteractor = async (
     .getPersistenceGateway()
     .getCasesByDocketNumbers({ applicationContext, docketNumbers });
 
-  cases.forEach(
-    aCase => (aCase.caseTitle = Case.getCaseTitle(aCase.caseCaption)),
-  );
-
   const caseDetails: PractitionerCaseDetail[] = cases.map(c => {
+    c = formatCase(applicationContext, c, authorizedUser);
     return {
       caseTitle: c.caseTitle,
       consolidatedIconTooltipText: c.consolidatedIconTooltipText,
@@ -44,9 +42,6 @@ export const getPractitionerCasesInteractor = async (
       inConsolidatedGroup: c.inConsolidatedGroup,
       isLeadCase: c.isLeadCase,
       isSealed: c.isSealed,
-      leadDocketNumber: c.leadDocketNumber,
-      sealedDate: c.sealedDate,
-      sealedToTooltip: c.sealedToToolTip,
       status: c.status,
     };
   });
--- a/shared/src/business/useCases/getPractitionerCasesInteractor.ts
+++ b/shared/src/business/useCases/getPractitionerCasesInteractor.ts
@@ -33,16 +33,28 @@ export const getPractitionerCasesInteractor = async (
     .getCasesByDocketNumbers({ applicationContext, docketNumbers });
 
   const caseDetails: PractitionerCaseDetail[] = cases.map(c => {
-    c = formatCase(applicationContext, c, authorizedUser);
+    const formattedCase = formatCase(applicationContext, c, authorizedUser);
+
+    const {
+      caseTitle,
+      consolidatedIconTooltipText,
+      docketNumber,
+      docketNumberWithSuffix,
+      inConsolidatedGroup,
+      isLeadCase,
+      isSealed,
+      status,
+    } = formattedCase;
+
     return {
-      caseTitle: c.caseTitle,
-      consolidatedIconTooltipText: c.consolidatedIconTooltipText,
-      docketNumber: c.docketNumber,
-      docketNumberWithSuffix: c.docketNumberWithSuffix,
-      inConsolidatedGroup: c.inConsolidatedGroup,
-      isLeadCase: c.isLeadCase,
-      isSealed: c.isSealed,
-      status: c.status,
+      caseTitle,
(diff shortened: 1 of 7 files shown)


shared/src/business/useCases/getPractitionerCasesInteractor.ts
```diff
@@ -30,12 +32,22 @@ export const getPractitionerCasesInteractor = async (
     .getPersistenceGateway()
     .getCasesByDocketNumbers({ applicationContext, docketNumbers });
 
-  cases.forEach(
-    aCase => (aCase.caseTitle = Case.getCaseTitle(aCase.caseCaption)),
```
Mwindo, line 34: `formatCase` below takes care of this. Rather than call it in one place (the PractitionerCaseList) and not in another (the PDF of the case list), I decided to keep it consistent and call `formatCase` by default.

shared/src/business/useCases/getPractitionerCasesInteractor.ts
```diff
@@ -30,12 +32,22 @@ export const getPractitionerCasesInteractor = async (
...
+      caseTitle: c.caseTitle,
+      consolidatedIconTooltipText: c.consolidatedIconTooltipText,
+      docketNumber: c.docketNumber,
+      docketNumberWithSuffix: c.docketNumberWithSuffix,
+      inConsolidatedGroup: c.inConsolidatedGroup,
+      isLeadCase: c.isLeadCase,
+      isSealed: c.isSealed,
+      status: c.status,
+    };
+  });
```
Mwindo, line 47: Before, we were returning a whole "case" object. (I put "case" in quotes because the available properties on a case differ depending on where it is at.) I have restricted this for two reasons:

1. This prevents a 502 error when there are a lot of cases. See [troubleshooting doc](https://github.com/ustaxcourt/ef-cms/blob/staging/docs/additional-resources/troubleshooting.md#502-errors).
2. This makes exactly what we have/need for a practitioner case explicit.

web-client/src/presenter/state.ts
```diff
@@ -911,21 +911,22 @@ export type PractitionerDetail = {
   practitionerType?: string;
   middleName?: string;
   contact?: Partial<UserContact>;
-  openCaseInfo?: PractitionerCaseInfo; // Only for internal users
-  closedCaseInfo?: PractitionerCaseInfo; // Only for internal users
+  openCaseInfo?: PractitionerAllCasesInfo; // Only for internal users
+  closedCaseInfo?: PractitionerAllCasesInfo; // Only for internal users
```
Mwindo, line 915: Renamed for clarity. Otherwise, `PractitionerCaseInfo` and `PractitionerCaseDetail` seem like the same thing.

web-client/src/styles/tabs.scss
```diff
@@ -244,3 +244,7 @@
     margin-bottom: 0 !important;
   }
 }
+
+.tab-content-with-top-margin {
+  margin-top: 2.5rem;
+}
```
Mwindo, line 250: We don't want the margin in the existing tabs, but do want a margin in the case list tabs. I added this class to accomplish this goal. If there is an existing class I can use, please let me know!

shared/src/business/useCases/getPractitionerCasesInteractor.ts
Mwindo: An alternative approach would be to paginate here on the backend, but that would introduce further complexities, especially if we wanted to avoid re-sorting to get the right records on every page change.

shared/src/business/useCases/getPractitionerCasesInteractor.ts
```diff
@@ -30,12 +32,34 @@ export const getPractitionerCasesInteractor = async (
...
+      caseTitle,
+      consolidatedIconTooltipText,
+      docketNumber,
+      docketNumberWithSuffix,
+      inConsolidatedGroup,
+      isLeadCase,
+      isSealed,
+      status,
+    };
+  });
```
Mwindo, line 59: Before, we were returning a whole "case" object. (I put "case" in quotes because the available properties on a case differ depending on where it is used, how it is formatted, etc.) I have restricted the case object here for two reasons:

1. I have strong reason to suspect the restriction will prevent a 502 error we are getting for a particular practitioner with a lot of cases. See [troubleshooting doc](https://github.com/ustaxcourt/ef-cms/blob/staging/docs/additional-resources/troubleshooting.md#502-errors).
2. The restriction makes exactly what we have/need for a practitioner case explicit.

An alternative approach would be to paginate here on the backend, but that would introduce further complexities, especially if we wanted to avoid re-sorting to get the right records on every page change. If it turns out that a particular practitioner will ever have _considerably_ more cases, I suppose we might have to go this route to avoid a 502.

En-8: Do you happen to have an estimate of how much more "considerably" more would be for this? And how performant is this now for the 4000-case petitioner? 

I like the limit on what's returned regardless of choice to paginate on the backend, but I'm curious how long we will have before pagination becomes a thing we have to worry about again, either due to 502 errors or loading times -- 1 month? 12 months? Several years? This will probably be a tough question to get a precise answer on but I am curious if we can come up with a super rough estimate.

Mwindo: Good questions!

The current 502-causing practitioner has ~4300 cases. Based on some local experiments (jsonifying the response and getting blob size), without the filtering we get a ~10.85 MB response size, well over the 6 MB lambda limit. With filtering, we get a ~.8 MB response size, well under. Even with these rough calculations, this seems to indicate enough wiggle room to handle any practical case list size.

Performance-wise, things seem ok. For one, most users have far, far fewer cases, so this is not a common occurrence. For two, other users with 1000+ cases were snappy enough. (That said, I still need to test in the test environment to see the impact of a non-linear slowdown--we do sort, for instance--for this particular user.)

En-8 (approved):

shared/src/business/useCases/getPractitionerCasesInteractor.ts
En-8: Whoa...10MB to 0.8MB is quite the change. Agreed, that sounds like we're in fine shape. Thank you!

Mwindo: Thanks for the good questions!

PR #5435, opened 2024-10-08
10363: AWS Batch

cruzjone-flexion:
Steps for deployment

1. Account Specific
2. ./scripts/dynamo/setup-aws-batch-zipper-minimum-count.sh


cypress-smoketests.config.ts
```diff
@@ -72,6 +77,12 @@ export default defineConfig({
         }) {
           return deleteAllItemsInEmailBucket({ bucketName, retries });
         },
+        doesFileExists(filePath) {
```
Mwindo: Totally minor: maybe `fileExists`? Otherwise `doesFileExist` is fine.

web-api/src/business/useCases/document/batchDownloadDocketEntriesInteractor.ts
```diff
@@ -136,6 +126,66 @@ const batchDownloadDocketEntriesHelper = async (
...
+      .getDispatchers()
+      .sendZipperBatchJob(
+        applicationContext,
+        UUID,
+        zipName,
+        clientConnectionId,
+        authorizedUser.userId,
+      );
+
+    const FAKE_NUMBER = 45;
```
Mwindo: Can we give this a more descriptive name, or else add a comment? From your presentation I know why we are waiting 45 seconds, but if I came at this fresh I would be flummoxed.

web-api/terraform/modules/batch/docker-image/package.json
```diff
@@ -0,0 +1,25 @@
+{
```
Mwindo, line 1: Can we update `dependency-updates.md`? Instead of `note: we have 2 package.json files, be sure to update both` etc., we should probably say, `note: we have 3 package.json files, be sure to update them all` and add this file to the subsequent list.

web-api/terraform/modules/batch/docker-image/src/docketEntryBatchDownloader.ts
```diff
@@ -0,0 +1,263 @@
...
+const WEBSOCKET_ENPOINT = `https://${WEBSOCKET_API_GATEWAY_ID}.execute-api.${AWS_REGION}.amazonaws.com/${STAGE}`;
+const notificationClient = new ApiGatewayManagementApiClient({
+  endpoint: WEBSOCKET_ENPOINT,
+  region: AWS_REGION,
+  requestHandler: new NodeHttpHandler({
+    requestTimeout: 900000,
+  }),
+});
+
+//TYPE THIS INSTEAD OF ANY
```
Mwindo: Is this for future work, or forgotten work for this PR?

cypress/deployed-and-local/integration/aws-batch-docket-entries-download/aws-batch-docket-entries-download.cy.ts
```diff
@@ -0,0 +1,105 @@
...
+          'be.visible',
+        );
+        cy.get('[data-testid="modal-button-confirm"]').click();
+
+        cy.get('[data-testid="progress-bar-description"]').should(
+          'contain.text',
+          'Preparing Files',
+        );
+
+        cy.wait(40 * 1000);
```
Mwindo, line 64: pmu08kmgulu01

Mwindo (commented):
Thanks @cruzjone-flexion! (Et al.!) I have some initial comments and will finish off tomorrow morning. 

One question RE: the loading bar (based on the demo): can we add logic to stop the dummy loading bar as soon as the batch job has actually begun processing to avoid the dueling progress bars?

web-api/terraform/modules/batch/docker-image/package.json
```diff
@@ -0,0 +1,25 @@
...
+  },
+  "dependencies": {
+    "@aws-sdk/client-apigatewaymanagementapi": "3.650.0",
+    "@aws-sdk/client-s3": "3.670.0",
+    "@aws-sdk/client-sqs": "3.670.0",
+    "@aws-sdk/lib-storage": "3.662.0",
+    "@aws-sdk/node-http-handler": "3.374.0",
+    "@aws-sdk/s3-request-presigner": "3.657.0",
+    "fflate": "^0.8.2",
+    "stream": "^0.0.3"
```
Mwindo: Not sure we want the carets

web-api/terraform/modules/batch/docker-image/src/docketEntryBatchDownloader.ts
```diff
@@ -0,0 +1,263 @@
...
+    while (continueReading) {
+      const unzippedChunk = await reader.read();
+      const nextChunk = unzippedChunk.value || new Uint8Array();
+      continueReading = !unzippedChunk.done;
+      compressedPdfStream.push(nextChunk, unzippedChunk.done);
+      while (passThrough.readableLength > 1024 * 1024 * 10) {
+        // Wait for the buffer to be drained, before downloading more files
+        await new Promise(resolve => setTimeout(resolve, 10));
+      }
+    }
```
Mwindo, line 191: Is it possible to use [pipeline](https://nodejs.org/api/stream.html#streampipelinesource-transforms-destination-options) here? E.g., something like `await pipeline(bodyStream, compressedPdfStream, passThrough);`

web-api/terraform/modules/batch/docker-image/src/docketEntryBatchDownloader.ts
```diff
@@ -0,0 +1,263 @@
...
+    wsClient,
+  });
+
+  console.log('Fetching Signed URL');
+  const command = new GetObjectCommand({
+    Bucket: ZIP_TEMP_S3_BUCKET,
+    Key: UNIQUE_ZIP_NAME,
+  });
+
+  const url = await getSignedUrl(zipStorageClient, command, { expiresIn: 120 });
```
Mwindo: Am I misreading, or does this expire after two minutes? If so, this seems too short. (The user might leave their laptop and come back after a few minutes, but the link won't work.)

Mwindo (approved):
I have approved conditional upon a few points being addressed:
- Fixing the dueling progress bars (I know I'm not UX, but I'd almost prefer a spinner that just says, "Preparing files. This process could take a couple of minutes." I tested the flow and was stuck at 97% for ~20 seconds.)
- Some indication in the code of why we are waiting 45 seconds
- Updating `dependency-updates.md`
- Increasing the s3 link expiration time (or else addressing why it is so low)

Something I would like to see soon as a follow-up: not using the number of files as a proxy and relying on file sizes instead

### General thoughts:

There was a lot of great work done on this PR! Setting up new infrastructure is often tricky, and you all did an excellent job. Thank you!

I am happy to approve and support this work going forward given the feedback from the team. That said, I will at least outline my reservations for transparency/documentation purposes because I feel like I don't have enough information to fully evaluate the merits of this solution:

- How many docket records have more than 50 entries? How many of those have more than whatever our GB cutoff is?
- Are there records < 50 entries but > our GB cutoff, in which case we might not have fully fixed the bug?
- What should our GB cutoff be, anyway?
- Are we making the experience worse for certain users (an extra 45-second wait time in cases that aren't needed), and how much does the bugfix outweigh that disadvantage?
- Does 30+ minute download expecting uninterrupted connection address the problem we are trying to solve effectively?
- Is a technically robust solution even necessary? Could we limit file download sizes, or cache big cases (e.g., you can select up to X MB of "random-access" docket entries, or else you can download the full file)?
- How useful will this batch processing infrastructure be for future work in the app?

These aren't all necessarily super difficult questions, but they aren't answered in my mind, and I don't see any trail of documentation outlining the decisions made.

web-api/terraform/modules/batch/docker-image/src/docketEntryBatchDownloader.ts
cruzjone-flexion: The link is generated and immediately sent to front end via web socket message and it starts downloading (all automated).

web-api/src/business/useCases/document/batchDownloadDocketEntriesInteractor.ts
```diff
@@ -175,3 +219,24 @@ const batchDownloadDocketEntriesHelper = async (
     userId: authorizedUser.userId,
   });
 };
+
+async function displayFakeProgressBarUntilBatchBootsUp(
```
Mwindo, line 223: 👍

web-api/terraform/modules/batch/docker-image/src/docketEntryBatchDownloader.ts
Mwindo: What if the pop-up is blocked? In my case, I had to manually reclick the link via the popup icon:

Screenshot 2024-10-29 at 11 00 01 AM

.circleci/config.yml
```diff
@@ -198,6 +198,30 @@ jobs:
           name: 'Deploy - Pending Color Switch'
           command: npm run pending-color-switch
 
+  docker-image-zipper:
+    docker:
+      - image: cimg/node:20.14
```
jimlerza, line 203: Why not use the same image that all of our other steps use? Doing so would eliminate the need to manually install the AWS CLI every time this job runs, and not require circle to also pull down this generic node image.

If there's a technical reason why we need to do this from this generic node image, we need to add something to the dependency updates documentation that instructs folks to keep this image up to date.

cypress/helpers/cypressTasks/dynamo/dynamo-helpers.ts
```diff
@@ -94,6 +94,23 @@ export const getFeatureFlagValue = async ({
   return !!result?.Item?.current;
 };
 
+export const getRawFeatureFlagValue = async ({
+  flag,
+}: {
+  flag: string;
+}): Promise<boolean> => {
```
jimlerza: I think we need to either
```suggestion
}): Promise<boolean | undefined> => {
```

cypress/helpers/cypressTasks/dynamo/dynamo-helpers.ts
```diff
@@ -94,6 +94,23 @@ export const getFeatureFlagValue = async ({
...
+  const dynamoClient = await getDocumentClient();
+  const result = await dynamoClient.get({
+    Key: {
+      pk: flag,
+      sk: flag,
+    },
+    TableName: getCypressEnv().dynamoDbDeployTableName,
+  });
+
+  return result?.Item?.current;
```
jimlerza, line 111: or
```suggestion
  return result?.Item?.current ?? false;
```

scripts/dynamo/setup-aws-batch-zipper-minimum-count.sh
```diff
@@ -0,0 +1,31 @@
...
+        "N": "50"
+    }
+}
+END
+)
+
+aws dynamodb put-item \
+    --region us-east-1 \
+    --table-name "efcms-deploy-${ENV}" \
+    --item "${ITEM}"
```
jimlerza:
```suggestion
    --item "${ITEM}"

```

scripts/setup-all-env-configuration.sh
```diff
@@ -16,4 +16,5 @@
 ./scripts/dynamo/setup-section-outbox-retrieval-days.sh
 ./scripts/dynamo/setup-add-docket-numbers-to-orders-flag.sh
 ./scripts/dynamo/setup-updated-trial-session-types-flag.sh
-./scripts/dynamo/setup-consolidated-cases-group-access-petitioner-flag.sh
\ No newline at end of file
+./scripts/dynamo/setup-consolidated-cases-group-access-petitioner-flag.sh
+./scripts/dynamo/setup-aws-batch-zipper-minimum-count.sh
```
jimlerza:
```suggestion
./scripts/dynamo/setup-aws-batch-zipper-minimum-count.sh

```

web-api/terraform/modules/batch/batch.tf
```diff
@@ -0,0 +1,183 @@
...
+    resourceRequirements = [
+      {
+        type  = "VCPU"
+        value = "4"
+      },
+      {
+        type  = "MEMORY"
+        value = "8192"
+      }
+    ]
```
jimlerza, line 167: We put these sorts of values for the opensearch clusters in secrets manager, though I'm not exactly sure that it makes any difference, since either way we still have to do a deployment to apply the change.

web-api/terraform/modules/batch/batch.tf
```diff
@@ -0,0 +1,183 @@
...
+			assignPublicIp  = "ENABLED"
+		}
+  })
+
+  platform_capabilities = ["FARGATE"]
+
+  retry_strategy {
+    attempts = 1
+  }
+}
```
jimlerza, line 183:
```suggestion
}

```

web-api/terraform/modules/batch/docker-image/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+ENV AWS_REGION=${AWS_REGION}
+ENV WEBSOCKET_API_GATEWAY_ID=${WEBSOCKET_API_GATEWAY_ID}
+
+WORKDIR /usr/src/app
+
+COPY package*.json ./
+RUN npm install
+
+COPY . .
```
jimlerza: This seems curious to me. Can it be simplified?

```suggestion
COPY . .
RUN npm install
```

web-api/terraform/modules/batch/docker-image/check-env-variables.sh
jimlerza, line 1: I don't like when we have multiple copies of the same file in the repo, so I was going to suggest that you just copy the real version of this file in the `Dockerfile` like so:

```
COPY .../.../.../.../../check-env-variables.sh .
```

But I see why we can't do that - the path to the `suppress-output.sh` file is different, so this file is different.

With that in mind, if we're gonna have another copy of this file, can we at least strip out all of the stuff that `suppress-output` is doing so we don't have to _also_ have a second copy of that file too? We aren't calling this version of this file with the `--quiet` flag.

web-api/terraform/modules/batch/docker-image/src/docketEntryBatchDownloader.ts
jimlerza: I would support a longer expiration as well.

web-api/terraform/modules/batch/variables.tf
```diff
@@ -0,0 +1,15 @@
...
+  type = string
+}
+
+variable "current_color" {
+  type = string
+}
+
+variable "region" {
+  type = string
+}
```
jimlerza:
```suggestion
}

```

web-api/terraform/modules/ci-cd/ci-cd-permissions.tf
```diff
@@ -365,11 +378,15 @@ resource "aws_iam_policy" "ci_cd_iam_policy" {
...
+        "arn:aws:iam::${data.aws_caller_identity.current.account_id}:role/lambda_role_*",
+        "arn:aws:iam::${data.aws_caller_identity.current.account_id}:instance-profile/batch_instance_profile_*"
       ]
     }
   ]
 }
 
 EOF
-}
+}
```
jimlerza:
```suggestion
}

```

jimlerza (approved):

PR #5438, opened 2024-10-08
10487 story intermediate branch to test 1728427842

zachrog:


Mwindo (approved):
LGTM! 

That said, one thing I just thought of: it might be worthwhile to add a comment on the test date strings indicating why they were selected the way they were (rather than just being random, pretty good dates to test against).

PR #5448, opened 2024-10-11
10519 Bug: 30 Day NOTT Location Display Issue

Mwindo:
Ticket: https://github.com/flexion/ef-cms/issues/10519

TL;DR: Our 30-Day NOTT PDF generation assumed that the trial location city/state and actual court city/state were always the same and defaulted to the former. This assumption is not correct (see below for why). This PR fixes the issue by using the court city/state.

---

Although a trial session is associated with a particular location (city and state), in some cases the specific courthouse chosen for the trial might actually be in a different city and/or state: some city/state pairs do not have a dedicated courthouse, so scheduling there might require using a nearby courthouse, perhaps one in a neighboring suburb or across a river dividing two states.

Example: a trial session scheduled for Kansas City, MO might actually need to use a courthouse across the river in Kansas City, KS. The trial is still officially associated with Kansas City, MO -- this is the federal location designated for the trial -- but the trial is held in Kansas City, KS, which is the address the court-goers need on their notices.

---

Notes:

- A side effect of this change: the state will now be abbreviated rather than written out. This keeps us consistent with the original Notice of Trial.


web-api/src/business/useCases/trialSessions/serveThirtyDayNoticeInteractor.ts
```diff
@@ -138,6 +138,17 @@ export const serveThirtyDayNoticeInteractor = async (
           caseCaption: caseEntity.caseCaption,
         });
 
+        const formatCityState = ({
```
Mwindo, line 141: Maybe overkill and a templated string is enough, but I did it this way because `trialSession.city` and `trialSession.state` are optional, which could theoretically lead to something silly like `, KS`.

En-8 (approved):

PR #5453, opened 2024-10-14
Devex custom case no appcontext

zachrog:
# What's In this PR
- A full vertical slice(frontend and backend) of customCaseReport with no appContext
- New jest mocking standards
- New singleton of serverAppContext to allow for iterative transition to no appContext


Mwindo (approved):
Overall, this LGTM. I don't see anything here that causes me to worry about not having an applicationContext. That said, I'd definitely want people who support applicationContext to review.

The main pro, as I see it, lies in reducing unnecessary levels of indirection, which makes reasoning about types and implementation details harder (this is a big one), contributes to a massive context object, and leads to extra code. It isn't entirely clear to me what problem abstraction is solving in our case. I think we would benefit from simplicity, adding abstraction as needed. This PR nudges us in that direction and is a template for future work.

The main cons:
- Higher coupling to jest. This doesn't strike me as persuasive. Moving from jest would involve a high cost either way.
- If we as a team decide to switch to this pattern, we need to make sure to get everyone doing it. But this is a con associated with any change.

Thanks @zachrog!

PR #5463, opened 2024-10-16
10261 Story: Practitioner Case List -- UX Updates (Show multiple pages in pagination, and fix sort)

Mwindo:
We want the paginator to show more pages (when applicable) for the practitioner case lists. Also, UX discovered an existing bug pertaining to out-of-order docket numbers. This PR addresses these two issues.

### DIFF
--- a/shared/src/business/entities/cases/Case.ts
+++ b/shared/src/business/entities/cases/Case.ts
@@ -232,15 +232,15 @@ export class Case extends JoiValidationEntity {
    */
   static sortByDocketNumber(cases) {
     return cases.sort((a, b) => {
-      return this.docketNumberSort(a.docketNumber, b.docketNumber);
+      return Case.docketNumberSort(a.docketNumber, b.docketNumber);
     });
   }
 
   // This will not take into account [19]95 vs [20]95
   static docketNumberSort(docketNumberA, docketNumberB) {
     return (
-      (this.getSortableDocketNumber(docketNumberA) || 0) -
-      (this.getSortableDocketNumber(docketNumberB) || 0)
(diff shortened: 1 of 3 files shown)


shared/src/business/entities/cases/Case.ts
```diff
@@ -237,17 +237,10 @@ export class Case extends JoiValidationEntity {
...
-    const bSplit = docketNumberB.split('-');
-
-    if (aSplit[1] !== bSplit[1]) {
-      // compare years if they aren't the same;
-      // compare as strings, because they *might* have suffix
-      return aSplit[1].localeCompare(bSplit[1]);
-    } else {
-      // compare index if years are the same, compare as integers
-      return +aSplit[0] - +bSplit[0];
-    }
```
Mwindo, line 250: This leads to [20]95 and [19]95 being indistinguishable. I don't know if there is any value to the string comparison when we already have `getSortableDocketNumber`.

Mwindo: @TomElliottFlexion: Do you see any issue with this change to docketNumberSort? `docketNumberSort` is used directly in `createCourtIssuedOrderPdfFromHtmlInteractor` and `getColdCases`, and indirectly in `getCasesForUserInteractor` and `getPractitionerCasesInteractor` (the code directly related to this story). I don't see any particular reason why these instances would need to use the `.localeCompare` string comparison rather than the `sortableDocketNumber`.

TomElliottFlexion: Not at all! Looks like you just removed redundant code, thank you

TomElliottFlexion (approved):

PR #5465, opened 2024-10-16
10518-bug: Override problematic default Quill styles

Mwindo:
`react-quill` uses CSS styles that cause issues. In this PR, we override these styles.

**Before (note the d):**
Screenshot 2024-10-16 at 3 50 52 PM

**After:**
Screenshot 2024-10-16 at 3 50 30 PM

----

### Notes:

The error is essentially the same as the one recorded [here](https://github.com/slab/quill/issues/3309).

This is a "minimal" fix for the bug. There are "better" ways to fix this, but they are more involved. (Example: use `Quill` directly, which does not have the same style issue as `react-quill`, which is no longer maintained.) This PR also does not address a related bug: saving an order with a nested list and then editing that order will cause the re-loaded list to be rendered incorrectly. Ultimately, this is because Quill uses its own data structure, the Quill Delta, and that--not the HTML--is what Quill needs to save/reload its state. All of our "save-and-edit-mismatch" bugs boil down to this misuse of Quill. However, since the court is moving towards a Word/Microsoft solution, we decided using Quill properly (or switching to a different editor without a separate data structure) was not worth the investment.

### DIFF
--- a/web-client/src/styles/quill.scss
+++ b/web-client/src/styles/quill.scss
@@ -1,0 +1,155 @@
+.ql-editor {
+  min-height: 640px;
+  max-height: 640px;
+  font-family: 'Times New Roman', Times, serif;
+  font-size: 16px;
+
+  p {
+    max-width: 100%;
+  }
+}
+
+.ql-snow .ql-picker.ql-size .ql-picker-item[data-value='10px']::before,
+.ql-snow
+  .ql-picker.ql-size
+  .ql-picker-label.ql-active[data-value='10px']::before {
+  content: '10px';
+  font-size: 10px !important;
+}
+
+.ql-snow .ql-picker.ql-size .ql-picker-item[data-value='12px']::before,
+.ql-snow
+  .ql-picker.ql-size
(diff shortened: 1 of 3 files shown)


web-client/src/styles/quill.scss
```diff
@@ -0,0 +1,161 @@
...
+  font-size: 18px !important;
+}
+
+.ql-snow .ql-picker.ql-size .ql-picker-item[data-value='20px']::before,
+.ql-snow
+  .ql-picker.ql-size
+  .ql-picker-label.ql-active[data-value='20px']::before {
+  content: '20px';
+  font-size: 20px !important;
+}
```
Mwindo, line 66: This is all just moved from custom.scss.

zachrog: * Cries in CSS *

zachrog (approved):

web-client/src/styles/quill.scss
Mwindo: 😭

PR #5446, opened 2024-10-10
10313 Bug: Change Email Verification Link Does Not Expire - UX Updates

Mwindo:
This PR:

- Updates the verify email notification banner and adds a button to resend the email
- Updates the message displayed to a user if they attempt to verify their email with an expired token
- Fixes a typo in the email in which the anchor tag was accidentally included in the link
- Uses a DateHandler function rather than using luxon directly in `verifyUserPendingEmailInteractor`

PR #5467, opened 2024-10-17
10313 Bug: Change Email Verification Link Does Not Expire 

Mwindo:
The change email verification link is supposed to expire after ~~an hour~~24 hours; it never expires. This PR addresses the issue and allows the user to resend a verification email.

Ticket: https://github.com/orgs/flexion/projects/11/views/1?pane=issue&itemId=76509729&issue=flexion%7Cef-cms%7C10313

Based on to-test PRs here: https://github.com/ustaxcourt/ef-cms/pull/5405, https://github.com/ustaxcourt/ef-cms/pull/5417, https://github.com/ustaxcourt/ef-cms/pull/5446

Note that anyone who has tried to reset their password within the 24-hour window before deploying this code will have their token forcibly expired. Because of our deployment schedule, I do not know of a good way around this. I don't think it should affect many (if any) users; the worst case scenario in this case is that the user will have to log in and click the new "Resend verification link" button.

---

Implementation notes:

- When updating a user to have a `pendingEmail` and `pendingEmailVerificationToken`, we now also set a UTC `pendingEmailVerificationTokenTimestamp`. In `verifyUserPendingEmailInteractor`, we verify that the timestamp is within 24 hours of UTC now. When we reset `pendingEmail` and `pendingEmailVerificationToken`, we also reset `pendingEmailVerificationTokenTimestamp`.
- The notification banner telling the user that a verification email has been sent has a button to allow them to resend a verification email.


web-api/src/business/useCases/practitioner/updatePractitionerUserInteractor.ts
```diff
@@ -149,29 +155,6 @@ export const updatePractitionerUser = async (
   }
 };
 
-const updateUserPendingEmail = async ({
```
Mwindo, line 152: This function was very similar to behavior occurring in `updateUserPendingEmailInteractor`. I refactored the function slightly and put it in `updateUserPendingEmailInteractor`, where it seems a more natural fit.

web-client/src/views/AppComponent.tsx
```diff
@@ -302,6 +303,7 @@ export const AppComponent = connect(
         {showModal === 'AppMaintenanceModal' && <AppMaintenanceModal />}
         {showModal === 'GenericErrorModal' && <GenericErrorModal />}
         {showModal === 'FileUploadErrorModal' && <FileUploadErrorModal />}
+        {showModal === 'VerifyNewEmailModal' && <VerifyNewEmailModal />}
```
Mwindo, line 306: Since the banner with the new "resend" functionality can be shown on any page, the modal needs to be available across the app. I removed it from the place it was appearing "locally" before and added it here for "global" display.

web-client/src/views/MyAccount/VerifyNewEmailModal.tsx
```diff
@@ -9,7 +9,7 @@ export const VerifyNewEmailModal = connect(
     cancelSequence: sequences.clearModalSequence,
     closeVerifyEmailModalAndNavigateToMyAccountSequence:
       sequences.closeVerifyEmailModalAndNavigateToMyAccountSequence,
-    email: state.form.email,
```
Mwindo, line 12: We don't always get the pending email value via a form now since the banner can appear anywhere.

jimlerza (approved):

akuny (approved):
Looks good!

PR #5468, opened 2024-10-17
10483: Display Empty Message to Public Users;

cruzjone-flexion:


zachrog (approved):

web-client/src/presenter/state-public.ts
```diff
@@ -27,7 +27,9 @@ const computeds = {
     typeof practitionerSearchHelper
   >,
   publicCaseDetailHeaderHelper,
-  publicCaseDetailHelper,
+  publicCaseDetailHelper: publicCaseDetailHelper as unknown as ReturnType<
+    typeof publicCaseDetailHelper
+  >,
```
Mwindo, line 32: Is the type not inferred here properly without the `as unknown as ReturnType<...`?

Mwindo (approved):

web-client/src/presenter/state-public.ts
zachrog: This is what we have to do for all helpers for them to appear correct in the .tsx files

Mwindo: Thank you!

cruzjone-flexion: No, helper is a function and we want to type it as the results of the function.

PR #5469, opened 2024-10-17
No App Context Experiment

zachrog:
# What's In this PR
- A full vertical slice(frontend and backend) of customCaseReport with no appContext
- New jest mocking standards
- New singleton of serverAppContext to allow for iterative transition to no appContext


akuny (approved):
This looks good! In my opinion, the tradeoff of having a more involved mock setup is worth it for the reduced indirection we get with ES module imports.

My only concern is that the mock setup could become complicated and inconsistent across multiple test files, but the patterns in this PR provide a good foundation for establishing conventions to prevent that.

pixiwyn (approved):

shared/src/proxies/reports/getCustomCaseReportProxy.ts
```diff
@@ -2,10 +2,10 @@ import {
   GetCustomCaseReportRequest,
   GetCustomCaseReportResponse,
 } from '@web-api/business/useCases/caseInventoryReport/getCustomCaseReportInteractor';
+import { applicationContext } from '@web-client/applicationContext';
```
bdruth, line 5: You added an import for `applicationContext` but removed the reference to `applicationContext` below?

web-api/src/business/useCases/caseInventoryReport/getCustomCaseReportInteractor.test.ts
```diff
@@ -1,14 +1,16 @@
+jest.mock('@web-api/persistence/elasticsearch/getCasesByFilters');
 import {
   GetCustomCaseReportRequest,
   getCustomCaseReportInteractor,
 } from './getCustomCaseReportInteractor';
-import { applicationContext } from '../../../../../shared/src/business/test/createTestApplicationContext';
+import { getCasesByFilters as getCasesByFiltersMock } from '@web-api/persistence/elasticsearch/getCasesByFilters';
```
bdruth, line 6: Just a nit, but I think you have the naming reversed, at least, in how I would think of it ... this import is the _real_ `getCasesByFilters` - as it is importing from the implementation module, right? Below, the return of `jest.mocked` is _not_ `getCasesByFilters`, rather it is, presumably, the mock of it.

shared/src/proxies/reports/getCustomCaseReportProxy.ts
zachrog: Yeah an interesting case. So if we are to adopt this strategy we have to do it iteratively, and nearly function in the entire app requires appContext, so how do you remove appContext from a parent function without having to rewrite every function in the application?

For functions you do not want to update yet, import applicationContext instead of DI it to allow iterative changing.

web-api/src/business/useCases/caseInventoryReport/getCustomCaseReportInteractor.test.ts
zachrog: This works a little differently from how you think, but maybe there is a better way to get that across?

as soon as you do jest.mock('@web-api/persistence/elasticsearch/getCasesByFilters') then wherever you import '@web-api/persistence/elasticsearch/getCasesByFilters' you are getting a mock. So the import you pointed out IS actually a mock, but Typescript does not know that it thinks you have imported the real thing. jest.mocked() Doesn't actually do anything but tell typescript "hey buddy this function is actually mocked".

jimlerza (approved):

PR #5472, opened 2024-10-18
10480 Story: M112 and M116 Motion Issues

Mwindo:
M112 and M116 were not classified as motions, but they are motions. This PR fixes the issue and refactors code slightly.

---

## Notes:

It is unclear to me if a specific test for the fix in this PR is helpful: the code was doing what it was "supposed" to do, we just told it that the documents with codes M112 and M116 were not motions. Nothing else in the code relies on the previous category of M112 and M116 ("Appearance and Representation"). Therefore, adding a test would essentially amount to "make sure that event codes M112 and M116 _in particular_ correspond to motions," but that seems arbitrary (since, in principle, there is nothing stopping us from checking that _every_ motion event code corresponds to a motion.)

We already have [a test](https://github.com/ustaxcourt/ef-cms/blob/9e383752efdf3479345edbbf21429f96558d0957/web-api/src/business/useCases/docketEntry/addPaperFilingInteractor.test.ts#L303) to check that tracked documents are automatically blocked. We don't seem to have a cypress test, but my feeling is that a full e2e for this is maybe too much.

That said, if reviewers feel that a more specific test would still be beneficial, or that a cypress test would be beneficial, I can add one. (The cypress test would probably be something like: Petitioner creates case -> Petitions Clerk Serves it -> Petitions clerk files a motion -> Judge logs in; blocked case marker should be displayed, and the case should be in tracked items ->  Judge grants the motion; blocked case marker should still be displayed, and case should still be in tracked items -> Docket Clerk QCs document and removes it from the tracked items; blocked case marker should be gone, and the item should not longer be in tracked items.)

---

I apologize for the extraneous early commits in the history. It doesn't make any difference to the files changes, but let me know if you want me to clean these up.


shared/src/business/entities/EntityConstants.ts
```diff
@@ -548,13 +564,27 @@ export const OBJECTIONS_OPTIONS_MAP = {
...
+  }),
+  'Application to Take Deposition',
+]);
+
+export const EXTERNAL_OBJECTION_DOCUMENT_TYPES = new Set([
+  ...DOCUMENT_EXTERNAL_CATEGORIES_MAP['Motion'].map(entry => {
+    return entry.documentType;
+  }),
+  'Application to Take Deposition',
+]);
```
Mwindo, line 579: This code was repeated in several places, so I extracted it. Because it was only ever used for lookups, I decided to use a Set.

shared/src/business/entities/EntityConstants.ts
```diff
@@ -600,28 +654,6 @@ export const STATUS_REPORT_ORDER_DOCUMENTS_ALLOWLIST = uniq(
...
-  ).map(x => x.eventCode);
-
-export const TRACKED_DOCUMENT_TYPES_EVENT_CODES = union(
-  EXTERNAL_TRACKED_DOCUMENT_EVENT_CODES,
-  INTERNAL_TRACKED_DOCUMENT_EVENT_CODES,
-  [
-    TRACKED_DOCUMENT_TYPES.proposedStipulatedDecision.eventCode,
-    TRACKED_DOCUMENT_TYPES.orderToShowCause.eventCode,
-  ],
-);
```
Mwindo, line 623: I moved this up to be next to the other logic for `TRACKED_DOCUMENT_TYPES`. I also refactored slightly and removed unused exports.

shared/src/tools/externalFilingEvents.json
```diff
@@ -145,28 +145,6 @@
       "labelFreeText": "Name",
       "ordinalField": "",
       "allowStamp": true
-    },
```
Mwindo, line 148: Moving these and categorizing them as Motion.

shared/src/tools/internalFilingEvents.json
```diff
@@ -156,27 +156,6 @@
       "labelFreeText2": "",
       "ordinalField": ""
     },
-    {
```
Mwindo, line 159: Moving these and categorizing them as Motion.

shared/src/business/entities/EntityConstants.ts
```diff
@@ -548,13 +562,35 @@ export const OBJECTIONS_OPTIONS_MAP = {
...
+];
+
+export const EXTERNAL_DOCUMENT_TYPES_REQUIRING_OBJECTION = new Set(
+  EXTERNAL_DOCUMENTS_REQUIRING_OBJECTION.map(doc => doc?.documentType),
+);
+
+export const EXTERNAL_DOCUMENT_CODES_REQUIRING_OBJECTION = new Set(
+  EXTERNAL_DOCUMENTS_REQUIRING_OBJECTION.map(doc => doc?.eventCode),
+);
+
```
Mwindo, line 586: Similar code was repeated in several places, so I extracted it.

In addition to `Application to Take Deposition`, the original code also had `Motion to Withdraw as Counsel` and `Motion to Withdraw Counsel (filed by petitioner)` (M12 and M116, respectively). Since this PR categorizes them as motions, we do not need to explicitly add them to the set.

shared/src/business/entities/EntityConstants.ts
```diff
@@ -548,13 +562,35 @@ export const OBJECTIONS_OPTIONS_MAP = {
...
+];
+
+export const EXTERNAL_DOCUMENT_TYPES_REQUIRING_OBJECTION = new Set(
+  EXTERNAL_DOCUMENTS_REQUIRING_OBJECTION.map(doc => doc.documentType),
+);
+
+export const EXTERNAL_DOCUMENT_CODES_REQUIRING_OBJECTION = new Set(
+  EXTERNAL_DOCUMENTS_REQUIRING_OBJECTION.map(doc => doc.eventCode),
+);
+
```
Mwindo, line 586: Similar code was repeated in several places, so I extracted it.

In addition to `Application to Take Deposition`, the original code also had `Motion to Withdraw as Counsel` and `Motion to Withdraw Counsel (filed by petitioner)` (M12 and M116, respectively). Since this PR categorizes them as motions, we do not need to explicitly add them to the set.

TomElliottFlexion (approved):

PR #5466, opened 2024-10-17
10074 Bug: Update Add Practitioner and IRS radio button validation

Mwindo:
A few more small updates to the "bugfix" that turned into "make validation more consistent and fix various minor bugs along the way."

PR #5419, opened 2024-10-02
10074 bug: Auto Scroll Issues on Petitions Clerk Create Case Form (and more) -- Updates

Mwindo:
10074 became "fix a bunch of validation quirks." (See [original PR](https://github.com/ustaxcourt/ef-cms/pull/5382).) UX discovered a few other validation quirks that this PR addresses.

PR #5382, opened 2024-09-19
10074 bug: Auto Scroll Issues on Petitions Clerk Create Case Form (and more)

Mwindo:
TL;DR: Fixing ~~some~~ quite a few form validation bugs/quirks.

---

The goals of this PR:
- Fix a couple of bugs
- When applicable, update `ErrorNotification` live to keep it in sync with in-line validations. (The previous behavior had `ErrorNotification` only updating on submissions, not validations.)
- Refactor to make validation slightly more consistent across the app and to eliminate duplicative actions (namely, `setAlertErrorAction` followed by `setValidationAlertErrorsAction`).

In the future, we will probably move away from the `ErrorNotification` altogether--similar to how things work with the new petitioner flow--but this at least keeps behavior consistent until then and hopefully will make a refactor easier. 

---

An outline of the issues addressed:

1. After (almost) every field edit on the petitions clerk create case form, we were scrolling back to the `ErrorNotification` banner. (This was the original error of [10074](https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/gh/flexion/ef-cms/10074)).
~~2. In certain cases on the petitions clerk create case form, the form scrolled weirdly. (This was design debt [10357](https://app.zenhub.com/workspaces/flexionef-cms-5bbe4bed4b5806bc2bec65d3/issues/gh/flexion/ef-cms/10357).)~~ Not fixed!
3. In the `PetitionsPayment` forms, blank Payment Dates and Payment Methods led to 400 errors rather than validation errors. (See video below.)

https://github.com/user-attachments/assets/f6888b74-2aa3-4bea-b6ae-90ed7cdf33e7

4. Certain onBlurs/onChanges weren't providing live validation. (E.g., Status Report Order validation was not providing live validation updates.) There are probably still some that I have missed.

PR #5474, opened 2024-10-21
10074 Bug (And More)

Mwindo:
This ticket ([here](https://github.com/orgs/flexion/projects/11/views/1?pane=issue&itemId=76509709&issue=flexion%7Cef-cms%7C10074)) evolved from "fix a bug on the petitioner-files-a-case screen where the screen scrolls to the top every time the user enters new information" to "fix that bug, plus make our validation more consistent across the app and fix a few related bugs along the way." This PR addresses that ad hoc goal.

---

## Notes:

This PR represents a general improvement, but it _does not_ fix all known related issues, and it does not completely standardize our validation. (E.g., some places use `ErrorNotification`, some don't.) Whatever the future direction, the hope is that this work will at least make a future refactor easier because things will be more consistent, and in the meantime the app is slightly better.

My initial fix of the bug documented in 10074 assumed the `ErrorNofitication` banner used (sometimes) across the app live updated along with the inline validation error messages. That _wasn't_ actually the case, but UX preferred that. This preference led to further changes, so this PR:
- Fixes "the" bug
- Fixes a couple of bugs in which invalid data was leading to a server error after submission rather than a front-end validation error before submission
- Fixes several spots where fields were not being validated on blur, on change, etc., and other quirks. In a few cases, this meant creating a new validation sequence
- Keeps the `ErrorNotification` banner consistent with inline validation errors (except for a few edge cases)
- Adds a "scroll to top" action that is used when users are submitting, but not validating, so they can see all of the errors associated with their submission.

Associated to-test PRs:

- https://github.com/ustaxcourt/ef-cms/pull/5382
- https://github.com/ustaxcourt/ef-cms/pull/5419
- https://github.com/ustaxcourt/ef-cms/pull/5466

A big shout-out to @swongCO, @katiecissell, and @mwestereng1 for all of their testing and feedback!


shared/src/business/entities/cases/Case.ts
```diff
@@ -534,7 +534,7 @@ export class Case extends JoiValidationEntity {
     mailingDate: JoiValidationConstants.STRING.max(25)
       .when('isPaper', {
         is: true,
-        otherwise: joi.allow(null).optional(),
+        otherwise: joi.optional().allow(null),
```
Mwindo, line 530: Merely cosmetic--just to keep the same order throughout this file

shared/src/business/entities/Practitioner.ts
```diff
@@ -122,10 +122,16 @@ export class Practitioner extends User {
       .description('The firm name for the practitioner.'),
     firstName: JoiValidationConstants.STRING.max(100)
       .required()
-      .description('The first name of the practitioner.'),
+      .description('The first name of the practitioner.')
+      .messages({
+        '*': 'Enter first name',
+      }),
```
Mwindo, line 128: Otherwise the user sees something like `"firstName" is a required field`

web-client/src/presenter/actions/validateAddPractitionerAction.ts
```diff
@@ -54,11 +54,11 @@ export const validateAddPractitionerAction = ({
...
-      'contact.country',
-      'contact.address1',
-      'contact.city',
-      'contact.state',
-      'contact.postalCode',
+      'country',
+      'address1',
+      'city',
+      'state',
+      'postalCode',
```
Mwindo, line 61: Otherwise the form did not display validation messages for these fields.

web-client/src/presenter/actions/validateCaseDetailsAction.ts
```diff
@@ -18,6 +18,9 @@ export const validateCaseDetailsAction = async ({
         caseDetail: {
           ...caseDetail,
           ...form,
+          caseType: form.caseType || null,
+          petitionPaymentDate: form.petitionPaymentDate || null,
+          petitionPaymentMethod: form.petitionPaymentMethod || null,
```
Mwindo, line 23: Fixes a "fails after submission rather than validates correctly before" error

web-client/src/presenter/computeds/filingPartiesFormHelper.ts
```diff
@@ -18,11 +18,9 @@ export const filingPartiesFormHelper = (
     PARTY_TYPES,
   } = applicationContext.getConstants();
 
-  const partyValidationError =
+  const partyValidationCheckboxError =
```
Mwindo, line 18: There was a duplicate validation message error for `otherFilingParty`, and the fix meant that this was validation error corresponded not to the entire `partyValidation` area but only the checkbox

web-client/src/presenter/sequences/CaseAssociation/validateEditPetitionerCounselSequence.ts
```diff
@@ -12,7 +13,7 @@ export const validateEditPetitionerCounselSequence = [
       setRepresentingFromRepresentingMapActionFactory('form'),
       validateEditPetitionerCounselAction,
       {
-        error: [setValidationErrorsAction],
+        error: [setValidationErrorsAction, setValidationAlertErrorsAction],
```
Mwindo, line 16: Here and throughout, the `setValidationAlertErrorsAction` is keeping our `ErrorNotification` consistent with our inline messages.

jimlerza (approved):

PR #5475, opened 2024-10-21
10261 Story: Update to sealed icon hover text

Mwindo:
The sealed case icon hover text I was using was only valid for _docket entries_, not full cases. I have changed the code to mimic other places where _cases_ are sealed. I also added a comment to the code just to help other developers avoid the same mistake.

### DIFF
--- a/web-client/src/styles/custom.scss
+++ b/web-client/src/styles/custom.scss
@@ -1794,7 +1794,8 @@ button.change-scanner-button {
   margin-right: 4px;
 }
 
-.sealed-docket-entry {
+.sealed-docket-entry,
+.sealed-case-entry {
   color: color($theme-color-secondary-dark);
 }
 
--- a/web-client/src/views/Practitioners/PractitionerCaseIcons.tsx
+++ b/web-client/src/views/Practitioners/PractitionerCaseIcons.tsx
@@ -11,8 +11,8 @@ export function PractitionerCaseIcons({
     <div className="multi-filing-type-icon">
       {formattedCase.isSealed && (
         <Icon
-          aria-label={formattedCase.sealedToTooltip}
-          className="sealed-docket-entry"
+          aria-label="sealed"
+          className="sealed-case-entry"
           icon="lock"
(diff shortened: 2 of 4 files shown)


web-client/src/styles/custom.scss
```diff
@@ -1794,7 +1794,8 @@ button.change-scanner-button {
   margin-right: 4px;
 }
 
-.sealed-docket-entry {
+.sealed-docket-entry,
+.sealed-case-entry {
```
Mwindo, line 1798: There is an `iconSealed` class, but it doesn't do what I need (it is too opinionated). Rather than using that one or using `sealed-docket-entry` (which is poorly named in my context), I made a new, more appropriately named class.

web-client/src/views/Practitioners/PractitionerCaseIcons.tsx
```diff
@@ -11,8 +11,8 @@ export function PractitionerCaseIcons({
     <div className="multi-filing-type-icon">
       {formattedCase.isSealed && (
         <Icon
-          aria-label={formattedCase.sealedToTooltip}
-          className="sealed-docket-entry"
+          aria-label="sealed"
+          className="sealed-case-entry"
           icon="lock"
           title={formattedCase.sealedToTooltip}
```
TomElliottFlexion: Just double checking that between the `aria-label` and the `title`, one of the two has a full-text explanation of what is sealed? Want to make sure that we're staying accessible

TomElliottFlexion (approved):
TY!

web-client/src/views/Practitioners/PractitionerCaseIcons.tsx
Mwindo: No sir--this is just an oversight. Thank you for catching!

Mwindo: The text we want is "sealed"--I was so focused on the aria-label that I just overlooked title entirely. Thanks again :)

PR #5478, opened 2024-10-21
10313 Bug: Actually use TOKEN_EXPIRATION_TIME_HOURS 🤦 

Mwindo:
Just fixing a stupid mistake I caught after reviewing 10313 code one last time.

### DIFF
--- a/web-api/src/business/useCases/user/verifyUserPendingEmailInteractor.ts
+++ b/web-api/src/business/useCases/user/verifyUserPendingEmailInteractor.ts
@@ -24,7 +24,7 @@ export const userTokenHasExpired = (
     calculateDifferenceInHours(
       createISODateString(),
       tokenExpirationTimestamp,
-    ) > 1
+    ) > TOKEN_EXPIRATION_TIME_HOURS
   );
 };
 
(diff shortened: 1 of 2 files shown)


web-api/src/business/useCases/user/verifyUserPendingEmailInteractor.ts
```diff
@@ -24,7 +24,7 @@ export const userTokenHasExpired = (
     calculateDifferenceInHours(
       createISODateString(),
       tokenExpirationTimestamp,
-    ) > 1
+    ) > TOKEN_EXPIRATION_TIME_HOURS
```
Mwindo, line 27: It is supposed to be 24 hours. I had this wrong originally (because the ticket said 1 hour), then updated it correctly once the ticket was updated, and then refactored to use a DateHelper function and got it wrong again--somehow completely ignoring my constant from right above 🤦 

~~We have good tests for expired and non-expired tokens. Another test to prevent the above stupid mistake would be something like, "Make sure the constant is correct," e.g., "Make sure a token isn't invalid after an hour, or maybe two hours, etc." (I'm not sure what a good value to choose would be, and clearly doing a test for every value up to 24 is silly.) Is that worth it? Since it is testing the specific implementation so rigidly, I think it isn't. But I can add one if anyone disagrees.~~

Nevermind the above crossed-out text, I disagree with myself now. I added a test.

akuny (approved):

Mwindo:
@jimlerza I tested the 24-hour logic on test (I updated the ticket with a little more detail) and it looks good to go!

jimlerza (approved):

PR #5482, opened 2024-10-23
10412 Design Debt: Improving Messages Table

Mwindo:
The diff is not nearly so heavy as it looks. It comes from wrapping multiple tsx components in a div so that they can be horizontally scrollable.

I do not like that there are six very similar message table components (the Cartesian product of individual or section and inbox, outbox, or completed). I think there should one with some config passed in. I am going to try refactoring this but will stop if it isn't straightforward.


Mwindo:
I have been experimenting with a refactor. I will close this PR and reopen based on my refactoring branch.

PR #5487, opened 2024-10-25
10412 Design Debt: Message Table Updates

Mwindo:
The design debt [ticket](https://github.com/orgs/flexion/projects/11/views/1?pane=issue&itemId=76509485&issue=flexion%7Cef-cms%7C10412) is more or less for getting the sort icons in the sortable message table to not wrap independently. We also decided to allow the table to scroll horizontally as needed.

In the process, I decided it was worth embarking upon a refactor. That is why the PR is not tiny.

**Before (icon wraps independently):**
Screenshot 2024-10-28 at 9 42 23 AM

**After (icon wraps with text):**
Screenshot 2024-10-28 at 9 41 14 AM

**If reviewers do not like the refactor, I can just add changes to fix the icon wrapping, scroll the table, and wrap columns with text containing no spaces (as the current behavior is to allow them to expand horizontally without bound).**

---

Refactor notes:

_TL;DR: Instead of 6 separate message tables all independently built but using a lot of the same code, there is 1 message table that each of the 6 message tables instantiate with different configurations. The refactor is not perfect, but I think it is 1) better and 2) a good start to further refactoring._

While fixing "the" message table, I saw there were actually six very similar message tables (inbox, outbox, and completed for both individuals and sections). I do not like this: 1) every update to "the" message table needs to touch multiple files, 2) it is easy for the tables to get out of sync ("whoops, someone forgot to add that class in one of the files"), and 3) every table is independently responsible for making sure the header columns and row columns match (e.g., that we don't have off-by-one errors where someone forgot a `colspan`). Moreover, I saw that what fields one could sort by/display were opaque: to answer that question, one needed to go into each component separately.

My thoughts are that you should be able to see what fields are available in a given context, pass in some subset of them into some `SortableTable` component, pass in the data object (messages, docket entries, whatever), and see everything "just work." Especially with [10525](https://github.com/orgs/flexion/projects/11/views/1?pane=issue&itemId=82760879&issue=flexion%7Cef-cms%7C10525) and likely other stories adding sorting to tables, I want to get closer to this `SortableTable` ideal. (And then extend this to `FilteredSortableTable`.)

This PR is a step in that direction. It creates a `MessageTable` component that each of the six message tables uses, along with a helper class detailing what fields are available to sort by. My hope is that we can continue to move in this direction to make things less message-specific and therefore reusable across the app.

For transparency, here are (I believe) the pros and cons:

Pros: 

1. Changes to one table (e.g., styles) are maintained across the different message tables
2. `MessageTable` automatically creates the header rows and data rows from the passed in columns without the developer needing to coordinate them
3. `MessageColumns` makes it easier for a developer to see how to construct a table for messages
4. Code deduplication

Cons:

1. The obverse of 1 above: changes to one table affect all others.
2. There are a couple of clunky places in `MessageTable`--concessions to "these tables are all basically--but not _quite_--the same. I think there are ways to improve this: this is not a perfect refactor, just a good start.
3. There is slightly more rendering logic.

### DIFF
--- a/cypress/local-only/tests/integration/messages/messages.cy.ts
+++ b/cypress/local-only/tests/integration/messages/messages.cy.ts
@@ -80,7 +80,7 @@ describe('Messages', () => {
       createAndServePaperPetition().then(({ docketNumber }) => {
         cy.wrap(docketNumber).as('DOCKET_NUMBER');
         goToCase(docketNumber);
-        sendMessagesToCompletedTab(DOCKET_CLERK_ID, docketNumber);
+        sendMessagesToCompletedTab(DOCKET_CLERK_ID);
         loginAsPetitionsClerk();
         goToCase(docketNumber);
         sendMessages(DOCKET_CLERK_ID);
@@ -92,79 +92,18 @@ describe('Messages', () => {
     });
 
     describe('Individual Message Boxes', () => {
-      describe('Sorting on the Individual Message Inbox', () => {
-        it('individual inbox subject column', () => {
-          loginAsDocketClerk();
-          cy.get(
-            '[data-testid="message-individual-subject-header-button"]',
-          ).click();
-          verifySubjectTitleOrder({
-            boxType: 'inbox',
-            isAscending: true,
-            prefix: 'Subject Line',
-            queueType: 'individual',
-          });
-          cy.get(
-            '[data-testid="message-individual-subject-header-button"]',
-          ).click();
-          verifySubjectTitleOrder({
-            boxType: 'inbox',
-            isAscending: false,
-            prefix: 'Subject Line',
-            queueType: 'individual',
-          });
-        });
+      before(() => {
+        loginAsDocketClerk();
+      });
 
-        it('individual inbox received at column when defaulted to sort ascending', () => {
-          loginAsDocketClerk();
-          cy.get(
-            '[data-testid="message-individual-received-header-button"]',
-          ).click();
-          verifySubjectTitleOrder({
-            boxType: 'inbox',
-            isAscending: false,
-            prefix: 'Subject Line',
-            queueType: 'individual',
-          });
-          cy.get(
-            '[data-testid="message-individual-received-header-button"]',
-          ).click();
-          verifySubjectTitleOrder({
-            boxType: 'inbox',
-            isAscending: true,
-            prefix: 'Subject Line',
-            queueType: 'individual',
-          });
-        });
-
-        it('individual inbox docket number column', () => {
-          loginAsDocketClerk();
-          cy.get(
-            '[data-testid="message-individual-docket-number-header-button"]',
-          ).click();
-          verifySubjectTitleOrder({
-            boxType: 'inbox',
-            isAscending: false,
-            prefix: 'Subject Line',
-            queueType: 'individual',
-          });
-          cy.get(
-            '[data-testid="message-individual-docket-number-header-button"]',
-          ).click();
-          verifySubjectTitleOrder({
-            boxType: 'inbox',
-            isAscending: false,
-            prefix: 'Subject Line',
-            queueType: 'individual',
-          });
-        });
+      beforeEach(() => {
+        cy.visit('/');
       });
 
       describe('Sorting on the Individual Message Inbox', () => {
         it('individual inbox subject column', () => {
-          loginAsDocketClerk();
           cy.get(
-            '[data-testid="message-individual-subject-header-button"]',
+            '[data-testid="messages-individual-inbox-subject-header-button"]',
           ).click();
           verifySubjectTitleOrder({
             boxType: 'inbox',
@@ -173,7 +112,7 @@ describe('Messages', () => {
             queueType: 'individual',
           });
           cy.get(
-            '[data-testid="message-individual-subject-header-button"]',
+            '[data-testid="messages-individual-inbox-subject-header-button"]',
           ).click();
           verifySubjectTitleOrder({
             boxType: 'inbox',
@@ -184,9 +123,8 @@ describe('Messages', () => {
         });
 
         it('individual inbox received at column when defaulted to sort ascending', () => {
-          loginAsDocketClerk();
           cy.get(
-            '[data-testid="message-individual-received-header-button"]',
+            '[data-testid="messages-individual-inbox-createdAt-header-button"]',
           ).click();
           verifySubjectTitleOrder({
             boxType: 'inbox',
@@ -195,7 +133,7 @@ describe('Messages', () => {
             queueType: 'individual',
           });
           cy.get(
-            '[data-testid="message-individual-received-header-button"]',
+            '[data-testid="messages-individual-inbox-createdAt-header-button"]',
           ).click();
           verifySubjectTitleOrder({
             boxType: 'inbox',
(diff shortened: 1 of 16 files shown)


web-client/src/ustc-ui/Table/SortableColumn.tsx
```diff
@@ -62,27 +62,34 @@ export const SortableColumn = ({
       }}
     >
       <span
-        className={classNames('margin-right-105', {
+        className={classNames('margin-right-1', {
```
Mwindo, line 65: This looked nicer with the icon wrap in place.

web-client/src/ustc-ui/Table/SortableColumn.tsx
```diff
@@ -62,27 +62,34 @@ export const SortableColumn = ({
...
+          )}
+          {!isLoading &&
+            getFontAwesomeIcon({
+              ascText,
+              descText,
+              direction: currentlySortedOrder,
+              isActiveColumn: isActive,
+            })}
+        </span>
+      </span>
```
Mwindo, line 92: This represents the actual fix for the ticket.

web-client/src/views/DocketRecord/DocketRecord.tsx
```diff
@@ -55,7 +55,7 @@ export const DocketRecord = connect(
         <DocketRecordHeader />
 
         <NonPhone>
-          <div style={{ overflowX: 'scroll', width: '100%' }}>
+          <div className="width-full overflow-x-auto">
```
Mwindo, line 58: Just updating in the places i saw it across the app for consistency.

web-client/src/views/Messages/MessageColumns.ts
```diff
@@ -0,0 +1,129 @@
+import {
```
Mwindo, line 1: This is the "here are the available fields for a messages table" class. Eventually, filters should go here too ... but one step at a time.

My hope is that we can pull out and reuse the message-independent stuff in order to standardize our sortable tables.

web-client/src/views/Messages/MessageList.tsx
```diff
@@ -0,0 +1,396 @@
+import {
```
Mwindo, line 1: This is "the" message table. You pass in what columns and filters you want, and it takes care of the rest.

Eventually, I would love to see a message-independent "you pass in what columns, filters, and data you want (maybe with specific components for specific cells), and it takes care of the rest" class.

web-client/src/views/Messages/MessageList.tsx
```diff
@@ -0,0 +1,396 @@
...
+};
+
+const MessageRow = ({
+  columns,
+  message,
+  messageListId,
+  onSelect,
+  selectable,
+}: {
+  message: any;
```
Mwindo: I will try to type this better if there is an existing type that works.

web-client/src/views/Messages/MessagesIndividualCompleted.tsx
```diff
@@ -1,156 +1,29 @@
-import { Button } from '../../ustc-ui/Button/Button';
```
Mwindo, line 1: Here, as in `MessagesIndividualInbox`, `MessagesSectionOutbox`, etc., we just pass in the columns and filters we want into `MessageList`.

cypress/local-only/tests/integration/messages/messages.cy.ts
```diff
@@ -92,79 +92,18 @@ describe('Messages', () => {
...
-          ).click();
-          verifySubjectTitleOrder({
-            boxType: 'inbox',
-            isAscending: false,
-            prefix: 'Subject Line',
-            queueType: 'individual',
-          });
-        });
+      beforeEach(() => {
+        cy.visit('/');
```
Mwindo, line 100: Not logging in before each sort test (we don't need too--we just need to navigate to the home screen) shaved about 15s (~12%) off of this test file.

web-client/src/views/Messages/MessageTable.tsx
```diff
@@ -0,0 +1,394 @@
+import {
```
Mwindo, line 1: Here, as in `MessagesIndividualInbox`, `MessagesSectionOutbox`, etc., we just pass in the columns and filters we want into `MessageList`.

Mwindo, line 1: This is "the" message table. You pass in what columns and filters you want, and it takes care of the rest.

Eventually, I would love to see a message-independent "you pass in what columns, filters, and data you want (maybe with specific components for specific cells), and it takes care of the rest" class.

web-client/src/views/Messages/MessageColumns.ts
```diff
@@ -0,0 +1,133 @@
...
+  DOCKET_NUMBER: {
+    displayField: 'docketNumberWithSuffix',
+    sortField: 'docketNumber',
+    sortType: 'string',
+  },
+  FROM: { sortField: 'from', sortType: 'string' },
+  FROM_SECTION: { sortField: 'fromSectionFormatted', sortType: 'string' },
+  SUBJECT: { sortField: 'subject', sortType: 'string' },
+  TO: { sortField: 'to', sortType: 'string' },
+  TO_SECTION: { sortField: 'toSection', sortType: 'string' },
```
Mwindo, line 47: Curiously, we don't have a `toSectionFormatted` (so `X's Chambers` is rendered `xsChambers`). If this is easy to add, I'll consider it.

web-client/src/views/Messages/MessageTable.tsx
```diff
@@ -0,0 +1,401 @@
...
+            role="alert"
+          >
+            <div className="usa-alert__body">
+              Message(s) completed at{' '}
+              {messagesIndividualInboxHelper.messagesCompletedAt} by{' '}
+              {messagesIndividualInboxHelper.messagesCompletedBy}
+            </div>
+          </div>
+        )}
+        <div className="overflow-x-auto" id={id}>
```
Mwindo, line 145: Making message tables scrollable.

web-client/src/views/Messages/MessageTable.tsx
```diff
@@ -0,0 +1,401 @@
...
+          <div className="grid-row grid-gap">
+            {messageFilters.length > 0 && (
+              <div className="desktop:grid-col-8 tablet:grid-col-12 display-flex flex-align-center">
+                <TableFilters
+                  filters={messageFilters}
+                  onSelect={updateMessageFilterSequence}
+                ></TableFilters>
+              </div>
+            )}
+            {selectable && getCompleteAllButton()}
```
Mwindo, line 155: Coupling selectable and completable ... but for now, I think that's ok. We can separate them if needed.

web-client/src/views/Messages/MessageTable.tsx
```diff
@@ -0,0 +1,401 @@
...
+          </div>
+
+          <table className="usa-table ustc-table subsection">
+            <thead>
+              <tr>
+                {selectable && getSelectAllCheckbox()}
+                {messageColumns.map(columnData => {
```
Mwindo, line 162: High-level: for every passed-in column, construct the headers. Then, for every data point, construct the row based on the same passed-in columns.

akuny (approved):
Looks like a step in the right direction!

PR #5513, opened 2024-11-04
10408: Add Special Trial Session Copy Notes to Trial Sessions Page

JayFlexy:
See [Ticket 10408](https://github.com/orgs/flexion/projects/11/views/1?pane=issue&itemId=76509708&issue=flexion%7Cef-cms%7C10408)


Mwindo (changes requested):
Blocking temporarily because it looks like this is a staging branch marked with to-test! It would be bad to merge this.

PR #5515, opened 2024-11-04
10408 Add Special Trial Session Copy notes To Test

JayFlexy:
See [10408 Ticket](https://github.com/orgs/flexion/projects/11/views/1?pane=issue&itemId=76509708&issue=flexion%7Cef-cms%7C10408)


web-api/src/persistence/dynamo/trialSessions/getTrialSessionWorkingCopy.ts
```diff
@@ -17,11 +39,28 @@ export const getTrialSessionWorkingCopy = ({
...
+}): TrialSessionWorkingCopy =>
   get({
     Key: {
       pk: `trial-session-working-copy|${trialSessionId}`,
       sk: `user|${userId}`,
     },
     applicationContext,
   });
+
+export const getBulkTrialSessionWorkingCopies = async ({
```
Mwindo, line 51: I think the pattern on the app is typically to split these functions out into separate files, e.g., `getBulkTrialSessionWorkingCopy.ts` and `getBulkTrialSessionWorkingCopy.test.ts`.

shared/src/business/entities/trialSessions/SpeciailTrialSessions.ts
```diff
@@ -0,0 +1,14 @@
+export interface SpecialTrialSession {
```
Mwindo: I think it is probably better to put these into the interactor file itself. I also wonder if we can rename `SpecialTrialSession`, which sounds like it is the full SpecialTrialSession type. (BTW: I have been told we use `type` more often than `interface`. I'm not sure how important that is, but I figured I would mention it.)

web-client/src/presenter/actions/TrialSession/getBulkSpecialTrialSessionCopyNotesAction.test.ts
```diff
@@ -0,0 +1,43 @@
...
+          },
+        ],
+      },
+      state: {},
+    });
+
+    expect(
+      applicationContext.getUseCases()
+        .getBulkSpecialTrialSessionCopyNotesInteractor,
+    ).toHaveBeenCalled();
```
Mwindo: In addition, we should check that the action returns the expected value.

web-client/src/presenter/actions/TrialSession/getBulkSpecialTrialSessionCopyNotesAction.ts
```diff
@@ -0,0 +1,26 @@
+import {
+  SpecialTrialSession,
+  TrialSessionWorkingCopyNotes,
+} from '@shared/business/entities/trialSessions/SpeciailTrialSessions';
+const getSpecialTrialSessions = trialSessions =>
+  trialSessions
+    .filter(trialSession => trialSession.sessionType === 'Special')
```
Mwindo: Let's do:
`import { SESSION_TYPES } from '@shared/business/entities/EntityConstants';` and then `SESSION_TYPES.special` instead of the raw string

web-client/src/presenter/state/trialSessionsPageState.ts
```diff
@@ -19,6 +19,7 @@ const filters: TrialSessionsFilters = {
 
 export const initialTrialSessionPageState = {
   filters,
+  specialTrialSessionCopyNotesObject: {},
```
Mwindo: Any reason not to type this?

web-client/src/views/TrialSessions/TrialSessionsTable.tsx
```diff
@@ -87,57 +87,82 @@ export const TrialSessionsTable = connect(
                   );
                 }
                 if (isTrialSessionRow(row)) {
+                  const additionalRow = row.sessionType === 'Special' && (
```
Mwindo: Same comment about using the constant

web-client/src/views/TrialSessions/TrialSessionsTable.tsx
```diff
@@ -87,57 +87,82 @@ export const TrialSessionsTable = connect(
...
+                    <tr className="special-sessions-row">
+                      <td></td>
+                      <td></td>
+                      <td></td>
+                      <td colSpan={6}>
+                        <div className="special-session">
+                          <span>
```
Mwindo: I think we can get rid of this outer span

Mwindo (approved):
Overall, this looks good to me--thanks, @JayFlexy!

I've added a few comments. A more general comment: It is probably worth adding (or better yet, if it exists, modifying an existing) Cypress test, too.

web-client/src/views/TrialSessions/TrialSessionsTable.tsx
JayFlexy: The format for bolding the leading text before the colon was the reason I had to do it this way. Do you have another way of accomplishing `Bolded text: regular text`?

Mwindo: I could be missing something, but here's what it looks like for me without that outer span:

Screenshot 2024-11-05 at 10 51 48 AM

JayFlexy: I stand corrected!

shared/src/business/entities/trialSessions/SpeciailTrialSessions.ts
JayFlexy: I attempted this but because I'm using these types in both client side and server side code, it was easier to put them here without getting errors like 'module not found' etc.

PR #5519, opened 2024-11-06
10491 work item refactor

pixiwyn:
See [10504](https://github.com/ustaxcourt/ef-cms/pull/5520) for more context


web-api/src/business/useCases/docketEntry/completeDocketEntryQCInteractor.test.ts
```diff
@@ -648,54 +620,6 @@ describe('completeDocketEntryQCInteractor', () => {
     );
   });
 
-  it('sets the assigned users section from the selected section when it is defined and the user is a case services user', async () => {
```
Mwindo, line 651: @codyseibert: Do you remember why these tests were removed?

scripts/run-once-scripts/postgres-migration/delete-user-outboxes.ts
```diff
@@ -0,0 +1,69 @@
+/**
```
Mwindo, line 1: This is basically just a copy of delete-section-outboxes with `user` replaced by `section`. I thought a) it would be better to separate the scripts for more granularity and b) it wasn't worth refactoring (e.g., have each call a "delete-outbox-script") since these are one-time scripts, and there are only two.

Mwindo (approved):
Reviewed this as part of my review for 10504.

jimlerza (approved):

PR #5528, opened 2024-11-07
OpEx: Postgres Quality-of-Life Improvements

jimlerza:
### Generate Token Script

1. Added a `--yes` parameter to presume prior confirmation of environment variables
2. Added a `--succinct` parameter to output only the generated token
3. Now supports generating tokens for other database users by overriding the `$DB_USER` environment variable

### Get Endpoint URL Scripts

1. Added a `--yes` parameter to presume prior confirmation of environment variables
2. When run with the `$CI` environment variable set, will output only the endpoint URL

### New "Connect" Script

Running `connect.sh` will generate a database token and connect with it using `psql`:

__Connecting to the read-only endpoint:__
```bash
./scripts/postgres/connect.sh
```

__Connecting to the writeable endpoint:__
```bash
./scripts/postgres/connect.sh --rw
```

__Overriding the default user and connecting to the writeable endpoint:__
```bash
DB_USER="${ENV}_dawson" ./scripts/postgres/connect.sh --rw
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -261,8 +261,7 @@ jobs:
       - run:
           name: Run Postgres Migration
           command: |
-            OUTPUT_ONLY=true npm run deploy:allColors $ENV
-            POSTGRES_HOST=$(cat ./web-api/terraform/applyables/allColors/output.json | jq -r '.rds_host_name.value')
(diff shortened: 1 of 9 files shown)


Mwindo (approved):
Thank you @jimlerza!

jtdevos (approved):

PR #5529, opened 2024-11-07
10461 Story: Public Trial Session Details  - Addressing UX Feedback

Mwindo:
- Removes the term and session status from the header for public users
- Removes unnecessary data from the interactor after the above change
- Standardizes the table **Count** indicator for lists across various trial session pages
- Adds scrolling to the public trial sessions details table when needed
- A few other tweaks, as noted in comments


web-api/src/business/useCases/trialSessions/getPublicTrialSessionDetailsInteractor.ts
```diff
@@ -9,10 +9,6 @@ export type PublicTrialSessionDetails = Pick<
   | 'trialLocation'
   | 'startDate'
   | 'estimatedEndDate'
-  | 'term'
-  | 'sessionStatus'
-  | 'termYear'
-  | 'sessionType'
```
Mwindo, line 15: This data is no longer necessary since we do not display the term or session info for public users.

web-api/src/business/useCases/trialSessions/getPublicTrialSessionDetailsInteractor.ts
```diff
@@ -21,7 +17,6 @@ export type PublicTrialSessionDetails = Pick<
   | 'postalCode'
 > & {
   calendaredCases: RawPublicCase[];
-  isSwingSession: boolean;
```
Mwindo, line 24: This was a helper that can just as well be calculated on the client-side to avoid passing more data than necessary over the network.

web-client/src/styles/custom.scss
```diff
@@ -2355,7 +2355,15 @@ button.change-scanner-button {
   padding: 0;
 }
 
+.trial-session-public-card {
+  line-height: 24px;
+}
```
Mwindo, line 2360: To keep consistent with the internal view.

web-client/src/styles/custom.scss
```diff
@@ -2355,7 +2355,15 @@ button.change-scanner-button {
...
+
+.h1-size {
+  font-size: 2em;
+}
```
Mwindo, line 2364: Our current use of h1 on some trial session views violates axe. This is a silly class so that we can use an h2 that looks the same as before when it was an h1.

web-client/src/ustc-ui/Icon/ConsolidatedCaseIcon.tsx
```diff
@@ -20,7 +20,9 @@ export const ConsolidatedCaseIcon = ({
             icon="copy"
           />
           {showLeadCaseIcon && (
-            <span className="fa-inverse lead-case-icon-text">L</span>
+            <span aria-hidden={true} className="fa-inverse lead-case-icon-text">
+              L
+            </span>
```
Mwindo, line 25: Added aria-hidden to prevent VoiceOver from reading "L," which, while slightly comical, is probably unhelpful.

web-client/src/views/Public/TrialSessions/PublicTrialSessionDetails.tsx
```diff
@@ -77,16 +77,10 @@ const PublicTrialSessionInformation = ({
...
-                <span className="address-line">
-                  {formattedTrialSession.address2}
-                </span>
-                <span className="address-line">
-                  {formattedTrialSession.formattedCityStateZip}
-                </span>
+                <div>{formattedTrialSession.courthouseName}</div>
+                <div>{formattedTrialSession.address1}</div>
+                <div>{formattedTrialSession.address2}</div>
+                <div>{formattedTrialSession.formattedCityStateZip}</div>
```
Mwindo, line 83: Block display lets these fields wrap at word boundaries rather than in the middle of the text.

web-client/src/views/Public/TrialSessions/PublicTrialSessionDetails.tsx
```diff
@@ -131,61 +125,65 @@ function NonMobileOpenCases({
 }) {
   return (
     <React.Fragment>
-      <div className="text-semibold push-right margin-bottom-2">
-        Count: {openCases.length}
+      <div className="text-right margin-bottom-2">
+        <span className="text-semibold">Count: </span>
+        {openCases.length}
       </div>
-      <table
```
Mwindo, line 137: Not much is actually changing here except for wrapping the table in a div so it can be scrolled.

web-client/src/views/Public/TrialSessions/PublicTrialSessionDetails.tsx
```diff
@@ -131,61 +125,65 @@ function NonMobileOpenCases({
...
+      <div className="overflow-x-auto overflow-y-hidden">
+        <table
+          className="usa-table ustc-table trial-sessions subsection"
+          data-testid="public-open-cases"
+          id="public-open-cases"
+        >
+          <thead>
+            <tr>
+              <th
+                aria-label="Icons for consolidated and/or sealed cases"
```
Mwindo, line 141: Recommended text from UX, which I've added in a few other similar places.

web-client/src/views/TrialSessionDetails/AllCases.tsx
```diff
@@ -13,63 +13,66 @@ export const AllCases = connect(
...
-                        title="Manually added"
-                      />
-                    </span>
-                  )}
-                </td>
-                <td>{item.caseTitle}</td>
-                <td>{item.status}</td>
-                <td>{item.disposition}</td>
-                <td>{item.removedFromTrialDateFormatted}</td>
+        <div className="overflow-x-auto overflow-y-hidden">
```
Mwindo, line 16: Again, not much change besides wrapping in div to allow scrolling.

web-client/src/views/TrialSessionDetails/InactiveCases.tsx
```diff
@@ -12,48 +12,51 @@ export const InactiveCases = connect(
   function InactiveCases({ inactiveCases }) {
     return (
       <React.Fragment>
-        <div className="text-semibold push-right margin-bottom-2">
```
Mwindo, line 15: And again

web-client/src/views/TrialSessionDetails/OpenCases.tsx
```diff
@@ -14,72 +14,75 @@ export const OpenCases = connect(
   function OpenCases({ openCases }) {
     return (
       <React.Fragment>
-        <div className="text-semibold push-right margin-bottom-2">
```
Mwindo, line 17: Aaand again

web-client/src/views/TrialSessionDetails/AllCases.tsx
Mwindo: The only other change is standardizing the Count: indicator and using the new "Icons for consolidated and/or sealed cases" text.

nechama-krigsman (approved):

PR #5530, opened 2024-11-08
OpEx: fix terraform build errors

jimlerza:
* Granting the CircleCI user `iam:UpdateAssumeRolePolicy` permission on the `restore_role_*` resources resolves the following error:

   ```
   Error: updating IAM Role (restore_role_stg) assume role policy: operation error IAM: UpdateAssumeRolePolicy, https response error StatusCode: 403, RequestID: 71db920d-0a5f-4dfa-a666-7c94e3ad5811, api error AccessDenied: User: arn:aws:iam::************:user/CircleCI is not authorized to perform: iam:UpdateAssumeRolePolicy on resource: role restore_role_stg because no identity-based policy allows the iam:UpdateAssumeRolePolicy action
   ```

* Granting the `batch_role_*` roles `ecs:DescribeClusters` permission on the `compute_environment_*` resources resolves the following error:
   ```
   CLIENT_ERROR - User: arn:aws:sts::************:assumed-role/batch_role_stg_green_us-east-1/aws-batch is not authorized to perform: ecs:DescribeClusters on resource: arn:aws:ecs:us-east-1:************:cluster/compute_environment_stg_green_us-east-1_Batch_1eef0571-e183-334e-a41e-5c9ffe142236 because no identity-based policy allows the ecs:DescribeClusters action
   ```

* Simplified how the `POSTGRES_HOST` environment variable is set in Circle (via calling a refactored `get-host` script)


scripts/postgres/generate-token.sh
jimlerza, line 1: Two things are happening in this diff:

1. Refactored the parameter parsing to happen all in one place, rather than across 3 files.
2. Explicitly passing the `us-east-1` region into our `aws rds describe-db-clusters` call.

scripts/postgres/get-host.sh
jimlerza, line 1: This file began as a renaming of `get-read-write-endpoint.sh` but I guess it is different enough that github's diff viewer prefers to show it as if it's a new file. 🤷

pixiwyn (approved):

JayFlexy (approved):

PR #5533, opened 2024-11-08
OpEx: Windows development environment setup instructions

jimlerza:
It's gross, but it works.


jimlerza:
Serious question: do we really want these instructions in the repo? I know I don't personally suggest or condone software development in Windows...

Maybe this belongs in the Court's Sharepoint or something instead?

TomElliottFlexion (approved):

TomElliottFlexion:
Just add "We don't suggest or condone software development in Windows" at the top and we should be good

PR #5534, opened 2024-11-11
10461 Story: Addressing UX Feedback Round 2

Mwindo:
- Make sure icons don't wrap on smaller screens, and reduce some spacing
- Fix mobile table: header for petitioner and respondent counsels was in the `map` instead of outside of it
- Move the Count indicators across the trial sessions pages outside of the scroll div so that they are right aligned with the screen (and therefore visible to the user even when the edge of the table is not)

PR #5538, opened 2024-11-12
10461 Story: Add flex-column back in

Mwindo:
Me, in the previous push: Removed a "superfluous" CSS class. (https://github.com/ustaxcourt/ef-cms/pull/5534#discussion_r1836837635)
Narrator: It wasn't.

### DIFF
--- a/web-client/src/presenter/computeds/Public/publicTrialSessionHelper.test.ts
+++ b/web-client/src/presenter/computeds/Public/publicTrialSessionHelper.test.ts
@@ -272,7 +272,7 @@ describe('publicTrialSessionsHelper', () => {
       expect(trialSessionRows).toEqual([
         {
           formattedSessionWeekStartDate: 'November 23, 2020',
-          sessionWeekStartDate: '2020-11-23T05:00:00.000+00:00',
+          sessionWeekStartDate: '2020-11-23T00:00:00.000-05:00',
         },
         {
           alertMessageForNOTT: '',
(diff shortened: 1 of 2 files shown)


web-client/src/views/Public/TrialSessions/PublicTrialSessionDetails.tsx
```diff
@@ -245,18 +245,22 @@ function MobileOpenCases({
                         publicCase.privatePractitioners.length > 0 && (
                           <div>
                             <span className="label">Petitioner Counsel</span>
-                            {publicCase.privatePractitioners?.map(counsel => (
-                              <div key={counsel.name}>{counsel.name}</div>
-                            ))}
+                            <div className="display-flex flex-column gap-1">
```
Mwindo, line 248: While I'm here, adding a tiny bit of spacing between names just in case there is a really long one.

TomElliottFlexion (approved):

web-client/src/presenter/computeds/Public/publicTrialSessionHelper.test.ts
```diff
@@ -272,7 +272,7 @@ describe('publicTrialSessionsHelper', () => {
       expect(trialSessionRows).toEqual([
         {
           formattedSessionWeekStartDate: 'November 23, 2020',
-          sessionWeekStartDate: '2020-11-23T05:00:00.000+00:00',
+          sessionWeekStartDate: '2020-11-23T00:00:00.000-05:00',
```
Mwindo, line 275: Looks like somehow a test was failing due to new timezone handling in test.

TomElliottFlexion: You've handled it correctly, at least as far as we could tell when making these changes (publicTrialSessionHelper is not in staging). `createDateAtStartOfWeekEST` now takes in an ISO date string with _any_ zone offset, and returns an EST offset date in UTC

TomElliottFlexion: (at midnight)

PR #5542, opened 2024-11-13
10460 story

nechama-krigsman:
flexion#10460


web-client/src/presenter/actions/TrialSession/setTimeStampAction.test.ts
```diff
@@ -0,0 +1,14 @@
+import { DateTime } from 'luxon';
+import { runAction } from '@web-client/presenter/test.cerebral';
+import { setTimeStampAction } from '@web-client/presenter/actions/TrialSession/setTimeStampAction';
+
+describe('setTimeStampAction', () => {
+  it('should set time stamp', async () => {
+    const propertyName = 'FetchedTrialSessions';
+    const result = await runAction(setTimeStampAction({ propertyName }), {});
+
+    const expectedDate = DateTime.now().setZone('America/New_York').toISODate();
```
Mwindo, line 10: Can we just use the `formatNow` function (or some other function) in `DateHandler` rather than importing `luxon`? (That way we keep luxon in one spot.)

web-client/src/presenter/actions/TrialSession/setTimeStampAction.ts
```diff
@@ -0,0 +1,8 @@
+import { DateTime } from 'luxon';
+import { state } from '@web-client/presenter/app.cerebral';
+
+export const setTimeStampAction =
+  ({ propertyName }: { propertyName: string }) =>
+  ({ store }: ActionProps) => {
+    store.set(state[propertyName], DateTime.now().setZone('America/New_York'));
```
Mwindo, line 7: Same comment as above.

Mwindo: This comment applies to anywhere we've imported luxon below, too.

web-client/src/presenter/computeds/Public/publicTrialSessionHelper.test.ts
```diff
@@ -0,0 +1,476 @@
+import { DateTime } from 'luxon';
+import { TrialSessionInfoDTO } from '@shared/business/dto/trialSessions/TrialSessionInfoDTO';
+import { publicTrialSessionsHelper } from '@web-client/presenter/computeds/Public/publicTrialSessionsHelper';
+import { runCompute } from '@web-client/presenter/test.cerebral';
+
+describe('publicTrialSessionsHelper', () => {
```
Mwindo, line 6: Minor, but it looks like we alternate between `TrialSession` and `TrialSessions`. Not worth addressing in this PR. (In fact, it's probably better _not_ to address until 10461 goes out too!)

web-client/src/presenter/computeds/Public/publicTrialSessionsHelper.ts
```diff
@@ -0,0 +1,175 @@
...
+  const sessionTypesModified = Object.values(sessionTypes).filter(
+    st => !!st,
+  ).length;
+
+  return (
+    !!proceedingTypeModified ||
+    !!judgesModified ||
+    !!locationsModified ||
+    !!sessionTypesModified
+  );
```
Mwindo, line 57: Totally minor, but maybe:

```
return [proceedingTypeModified, judgesModified, locationsModified, sessionTypesModified].some(Boolean);
```

web-client/src/presenter/test.cerebral.ts
```diff
@@ -10,8 +10,9 @@ type FakeRunComputeType = <T>(
 ) => T;
 export const runCompute = cerebralRunCompute as unknown as FakeRunComputeType;
 
-type FakeRunActionType = <T>(
+type FakeRunActionType = <T, U = ClientState>(
```
Mwindo, line 13: 👍

web-client/src/ustc-ui/Select/Selects.tsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+
+export function Select({ defaultValue, name, onChange, options, value }) {
```
Mwindo, line 3: 👍

web-client/src/ustc-ui/Select/Selects.tsx
```diff
@@ -0,0 +1,44 @@
...
+      <select
+        className="usa-select margin-bottom-2"
+        name={name}
+        value={value || ''}
+        onChange={e => {
+          onChange(e.target.value);
+        }}
+      >
+        {defaultValue && (
+          <option key={defaultValue.label} value={defaultValue.value}>
```
Mwindo, line 15: Minor: Can we ensure label will always be unique?

web-client/src/ustc-ui/Select/Selects.tsx
```diff
@@ -0,0 +1,44 @@
...
+            {defaultValue.label}
+          </option>
+        )}
+
+        {options.map(option => {
```
Mwindo, line 20: Maybe `options?.map` to avoid breaking if no options are passed in

web-client/src/views/Public/TrialsSessions/PublicTrialSessions.tsx
```diff
@@ -0,0 +1,329 @@
...
+          </NonMobile>
+        </div>
+        <div className="grid-row">
+          <Button
+            link
+            data-testid="trial-sessions-reset-filters-button"
+            disabled={!publicTrialSessionsHelper.filtersHaveBeenModified}
+            onClick={() => {
+              resetPublicTrialSessionDataSequence();
+              displayProgressSpinnerSequence({ timeInSeconds: 0.25 });
```
Mwindo, line 127: Is this just a "show the user something happened" spinner?

web-client/src/views/Public/TrialsSessions/PublicTrialSessions.tsx
```diff
@@ -0,0 +1,329 @@
...
+  },
+);
+
+function NonMobilePublicTrialsSessions({
+  displayProgressSpinnerSequence,
+  publicTrialSessionData,
+  publicTrialSessionsHelper,
+  resetPublicTrialSessionDataSequence,
+  updateFormValueSequence,
+}: TrialsSessionsUiParams) {
```
Mwindo, line 98: Nitpick: `TrialSessions`, here and elsewhere

web-client/src/views/Public/TrialsSessions/PublicTrialSessionsFilters.tsx
```diff
@@ -0,0 +1,207 @@
...
+};
+
+const props = cerebralProps as unknown as PublicTrialSessionsFiltersProps;
+
+const PublicTrialSessionsFiltersDeps = {
+  displayProgressSpinnerSequence: sequences.displayProgressSpinnerSequence,
+  publicTrialSessionData: state[props.ROOT],
+  publicTrialSessionsHelper: state.publicTrialSessionsHelper,
+  updateFormValueSequence: sequences.updateFormValueSequence,
+};
```
Mwindo, line 23: This doesn't need to be addressed in this PR, but I wonder how you would feel about passing these in as props rather than connecting to Cerebral? That way Cerebral has one "entry point"--into the parent component--rather than proliferating through the child components.

Mwindo (approved):
Great work, @cruzjone-flexion and @nechama-krigsman! Thank you!

I left several minor comments. I'm not sure any of these need to be addressed in the current PR, especially since my plan is to open up a follow-up "clean-up" PR anyway to make sure my work in 10461 is consistent with yours. (E.g., I handled the "current as of" differently, and it's probably better to be consistent.)

web-client/src/presenter/sequences/displayProgressSpinnerSequence.ts
```diff
@@ -0,0 +1,8 @@
+import { showProgressSequenceDecorator } from '@web-client/presenter/utilities/showProgressSequenceDecorator';
+
+export const displayProgressSpinnerSequence = showProgressSequenceDecorator([
+  async ({ props }: ActionProps<{ timeInSeconds: number }>) =>
+    await new Promise(resolve =>
+      setTimeout(() => resolve(null), props.timeInSeconds * 1000),
+    ),
```
TomElliottFlexion, line 7: Why is this necessary?

Mwindo: I _think_ the idea is that when filtering is fast, it looks like nothing changed, so this is just a signal to the user. I asked a similar question.

web-client/src/views/Public/TrialsSessions/PublicTrialSessions.tsx
```diff
@@ -0,0 +1,329 @@
...
+  ) => {
+    if (displayProgressSpinnerSequence)
+      displayProgressSpinnerSequence({ timeInSeconds: 0.25 });
+    updateFormValueSequence(...args);
+    updateFormValueSequence({
+      key: 'pageNumber',
+      root: ROOT,
+      value: 0,
+    });
+  };
```
TomElliottFlexion, line 173: Can this be done in Cerebral?

web-client/src/presenter/sequences/displayProgressSpinnerSequence.ts
TomElliottFlexion: Thank you! That's a good problem to have

cruzjone-flexion (commented):
All feedback is refactoring related, we will handle it all in a separate PR.

jimlerza (approved):

web-client/src/presenter/sequences/displayProgressSpinnerSequence.ts
cruzjone-flexion: That is correct, mainly in mobile view. UX team stated that the filters would take most of screen and so users will not see changes in the table. So this is to provide a UI indication to the user that something has happened once they change some filter.

PR #5543, opened 2024-11-13
10461 Story: UX updates and coordinating with 10460

Mwindo:
This PR looks huge, but it is not.

What it does:

1. Adds back in the indentation for the `CaseLink` in `PublicTrialSessionDetails.tsx`. (The Figma did not have indentation, but we decided to add it.)
2. Comments in a commented-out link to PublicTrialSessionDetails from 10460.
3. Adds and uses a new sort function to ensure consolidated cases are grouped together. ("Doesn't this already exist?" See inline comment.)
4. Renames `TrialsSessions` folder to `TrialSessions`.
5. Removes the estimated end date for public trial session data, which is unused.


shared/src/business/entities/cases/Case.ts
```diff
@@ -243,6 +243,39 @@ export class Case extends JoiValidationEntity {
     );
   }
 
+  static sortByDocketNumberAndGroupConsolidatedCases<
```
Mwindo, line 246: There is a similar function [here](https://github.com/ustaxcourt/ef-cms/blob/d95d05f9ee81c6a2eb12ae73063c1ca9b693577c/shared/src/business/utilities/trialSession/getFormattedTrialSessionDetails.ts#L203), but 1) it isn't super intuitive and 2) it is wholly separate from our other sorting. I also keep running into this issue of "how, exactly, should I be sorting docket numbers in this context?"; my hope is that we can get better at naming our sorting functions and consolidating them.

web-client/src/views/Public/TrialSessions/PublicMobileTrialSessionsDataRow.tsx
```diff
@@ -0,0 +1,60 @@
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
```
Mwindo, line 1: These files below are just moved from `TrialsSessions` to `TrialSessions`, along with commenting back in the link to `TrialSessionDetails`.

shared/src/business/entities/cases/Case.ts
Mwindo: After this feature is delivered, I will remove the similar function and use this function instead. I would like to get this feature out before adding more work for myself (because we would like this feature delivered soon!).

web-client/src/views/Public/TrialSessions/PublicTrialSessionsTable.tsx
```diff
@@ -0,0 +1,110 @@
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
```
Mwindo, line 1: These files below are just moved from `TrialsSessions` to `TrialSessions`, along with commenting in any links to `PublicTrialSessionDetails`.

package-lock.json
TomElliottFlexion, line 1: Is this intentionally part of the PR?

shared/src/business/entities/cases/Case.ts
TomElliottFlexion: Good points, agreed! I wonder if `Case.ts ` is the best place for this function? Not a big deal, but it's not single case specific, and this entity is already huge

web-client/src/presenter/computeds/Public/publicTrialSessionDetailsHelper.test.ts
```diff
@@ -13,8 +13,12 @@ describe('publicTrialSessionDetailsHelper', () => {
 
   const MOCK_SEALED_CASE = cloneDeep(MOCK_CASE);
   MOCK_SEALED_CASE.isSealed = true;
+  MOCK_SEALED_CASE.docketNumber = '101-23';
+  MOCK_SEALED_CASE.docketNumberWithSuffix = '101-23';
 
   const MOCK_LEAD_CASE = cloneDeep(MOCK_CASE);
+  MOCK_LEAD_CASE.docketNumber = '100-23';
+  MOCK_LEAD_CASE.docketNumberWithSuffix = MOCK_LEAD_CASE.docketNumber;
   MOCK_LEAD_CASE.leadDocketNumber = MOCK_LEAD_CASE.docketNumber;
```
TomElliottFlexion, line 22: Should one of these have a suffix?

shared/src/business/entities/cases/Case.ts
TomElliottFlexion: Eh, after reading further I see why this was added here. Maybe something for long-term consideration, but ignore the above comment for this PR

package-lock.json
Mwindo: I can remove it!

web-client/src/views/AppComponentPublic.tsx
```diff
@@ -12,7 +12,7 @@ import { PublicCaseDetail } from './Public/PublicCaseDetail';
 import { PublicPrintableDocketRecord } from './Public/PublicPrintableDocketRecord';
 import { PublicSearch } from './Public/PublicSearch';
 import { PublicTrialSessionDetails } from '@web-client/views/Public/TrialSessions/PublicTrialSessionDetails';
-import { PublicTrialSessions } from '@web-client/views/Public/TrialsSessions/PublicTrialSessions';
+import { PublicTrialSessions } from '@web-client/views/Public/TrialSessions/PublicTrialSessions';
```
TomElliottFlexion, line 15: THANK YOU

web-client/src/presenter/computeds/Public/publicTrialSessionDetailsHelper.test.ts
Mwindo: `MOCK_CONSOLIDATED_CASE` below does :) These don't need one.

TomElliottFlexion (approved):
Nice!

package-lock.json
TomElliottFlexion: Not a huge deal, just like to avoid updating dependencies (even transitive) out of cycle

shared/src/business/entities/cases/Case.ts
Mwindo: Yes, I totally agree. What I would like to do is have a separate sorting file where we consolidate all of the case/docket sorting functions that seem somewhat scattered.

TomElliottFlexion: I like that idea very much

package-lock.json
Mwindo: >  just like to avoid updating dependencies (even transitive) out of cycle

This is the correct approach. I was just being lazy 😬

PR #5546, opened 2024-11-14
Parallelize smoke tests in circle deploys

En-8:
Does as the title says: splits smoke tests across six parallel executors when running smoke tests in cypress. It also bumps up the resource allocation for the smoke tests job to handle the additional load of parallel cypress runs.

This cuts our smoke tests duration from ~35 minutes to ~10 minutes

Here's a successful deploy to test: https://app.circleci.com/pipelines/github/ustaxcourt/ef-cms/9210/workflows/cbb7c74d-304c-4a74-bc96-9226944d1819

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -348,12 +348,13 @@ jobs:
             ./scripts/reindex/disable-reindex-cron.sh
 
   smoketests:
+    parallelism: 6
     docker:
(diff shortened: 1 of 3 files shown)


Mwindo (approved):
Thank you, @En-8!

jimlerza (approved):

PR #5547, opened 2024-11-14
(To Test) Parallelize smoke tests in circle deploys

En-8:
Does as the title says: splits smoke tests across six parallel executors when running smoke tests in cypress. It also bumps up the resource allocation for the smoke tests job to handle the additional load of parallel cypress runs.

This cuts our smoke tests duration from ~35 minutes to ~8 minutes


jimlerza:
woo-hoo

scripts/run-cypress.sh
```diff
@@ -127,12 +127,15 @@ else
   export CYPRESS_MIGRATE=$CYPRESS_MIGRATE
 fi	
 
+echo "${RUN_SPECIFIC_TEST}"
```
jimlerza: If you wanna keep this let's make it friendly

```suggestion
[ -n "${RUN_SPECIFIC_TEST}" ] && echo "Running ${RUN_SPECIFIC_TEST}" || echo "Running all tests"
```

jimlerza (approved):

scripts/run-cypress.sh
En-8: I see you resolved this since I removed it (I assume), but for documentation's sake I'll respond: this echo was really only a troubleshooting echo. I think the information cypress gives you when the tests execute (or fail to execute) is actually generally enough information, so I don't think it's worth keeping the echo around in a more friendly form.

PR #5548, opened 2024-11-14
10461: Fix router merge conflict mix-up

Mwindo:


web-client/src/routerPublic.ts
```diff
@@ -84,11 +84,6 @@ const router = {
       return app.getSequence('redirectToLoginSequence')();
     });
 
-    route('/trial-sessions', () => {
-      setPageTitle('Trial sessions');
-      return app.getSequence('gotoPublicTrialSessionsSequence')();
-    });
-
```
Mwindo, line 91: Looks like a botched merge conflict -- the route was added twice.

Mwindo: Same route is at line 94

En-8 (approved):

PR #5555, opened 2024-11-14
OpEx: Stale Cases Report

jimlerza:
Outputs a CSV of cases that are not closed or on appeal but haven't had a document filed in over a year.

Unfortunately, I was unable to achieve this without querying for the "most recently filed document" in each case individually.

I tried to do a `terms` query on `docketNumber.S`, providing the not-closed-not-on-appeal docket numbers (roughly 21,000) as the input. According to the OpenSearch documentation, terms queries have a limit of 65,536 terms, so I'm not quite sure why this didn't work.

Undeterred, I was pleased with the performance gains when executing the "get most recently filed document" queries in a priority queue:

Execution time with a `for of` loop: 13 minutes 12 seconds
Execution time with `pqueue`: 34 seconds

Just imagine how much easier these kinds of reports will be when all our data is in postgres!

### DIFF
--- a/scripts/reports/stale-cases.ts
+++ b/scripts/reports/stale-cases.ts
@@ -1,7 +1,10 @@
 // usage:
 // npx ts-node --transpile-only scripts/reports/stale-cases.ts
 
-import { CASE_STATUS_TYPES } from '@shared/business/entities/EntityConstants';
+import {
+  CASE_STATUS_TYPES,
+  CaseStatus,
+} from '@shared/business/entities/EntityConstants';
 import {
   ServerApplicationContext,
   createApplicationContext,
@@ -11,6 +14,7 @@ import {
   calculateDifferenceInDays,
   createISODateString,
 } from '@shared/business/utilities/DateHandler';
+import { compareStrings } from '@shared/business/utilities/sortFunctions';
 import {
   search,
   searchAll,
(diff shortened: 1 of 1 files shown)


scripts/reports/stale-cases.ts
```diff
@@ -0,0 +1,167 @@
...
+  aCase: RawCase;
+  applicationContext: ServerApplicationContext;
+}): Promise<void> => {
+  const mostRecentDocketEntry = await getMostRecentDocketEntry({
+    applicationContext,
+    docketNumber: aCase.docketNumber,
+  });
+  const deRcvdAt = mostRecentDocketEntry?.receivedAt;
+  const deAge = deRcvdAt ? calculateDifferenceInDays(todayISO, deRcvdAt) : 0;
+  if (deAge > 364) {
```
jtdevos: You have awakened the *Constant Curmudgeon*, who will provide little substantive feedback and will instead whine about replacing magic numbers with constants.

jtdevos (changes requested):
Looks good!  Just one tweak requested.

scripts/reports/stale-cases.ts
Mwindo: > You have awakened the Constant Curmudgeon, who will provide little substantive feedback and will instead whine about replacing magic numbers with constants.

This feels like it should be the text in a card game.

Mwindo (approved):

scripts/reports/stale-cases.ts
jimlerza: Ha! Your suggestion is appreciated. :)

jtdevos (approved):

PR #5556, opened 2024-11-15
Dexex organize docket sorting (WIP)

Mwindo:
This is a work in progress. Please ignore for now.

We have several different methods for sorting docket-related things scattered throughout the codebase. This PR attempts to collect them in one place, refactoring and renaming them for clarity as need be.


shared/src/business/utilities/getFormattedCaseDetail.ts
```diff
@@ -448,46 +424,16 @@ const formatCounsel = ({ caseDetail, counsel }) => {
...
-  sortByString = '',
-) => {
-  const sortFunc = getDocketRecordSortFunc(sortByString);
-  const isReversed = sortByString.includes('Desc');
-  docketEntries.sort(sortFunc);
-  if (isReversed) {
-    // reversing AFTER the sort keeps sorting stable
-    return docketEntries.reverse().sort(sortUndefined);
-  }
-  return docketEntries.sort(sortUndefined);
```
Mwindo, line 484: We were sorting the whole list twice rather than improving our a custom comparison function to allow us to sort once.

shared/src/business/utilities/sorting/caseSorting.ts
```diff
@@ -0,0 +1,76 @@
+const getSortableDocketNumber = (docketNumber?: string) => {
```
Mwindo, line 1: This file is just logic moved over from Case.ts.

PR #5557, opened 2024-11-15
Devex case factory (WIP)

Mwindo:
### This will be implemented (I hope) as part of 10502.

This is a WIP. Please ignore.

TODO:
- [ ] Check IRS Superuser permissions are correct
- [ ] Ensure Case, PublicCase, RestrictedCase are constructed correctly. Maybe rename Case to FullCase. Perhaps extend a base Case class. (Essentially: Get the class hierarchy figured out.)
- [ ] Update types
- [ ] Test

The goal of this PR is to streamline the logic behind what subset of case data a user sees:

1. The business rules behind who can see what subset of data are typically not difficult to state (e.g., "an internal user of type X can see everything" or "a petitioner can see everything if they are associated with a case, otherwise they see what the public sees" or "the public sees some data if the case is not sealed but virtually no data if the case is sealed"). However, these business rules are very difficult to tease out in the code.
2. Case entities have a lot of fields that only matter in certain cases and not others (e.g., fields for internal users vs. the public), which makes the code difficult to reason about.
3. Because of 1 and 2, if a developer is not careful, they might end up revealing data they shouldn't. It should be enough for a developer to call the correct constructor and feel confident that a "safe" and relevant subset of data has been defined for them. In other words, if a developer constructs a `PublicCase`, they should feel confident that while some of the `PublicCase` data might need to be modified for their specific purposes, it should nonetheless require no extra modifications in order to be "safe" to show; it should never occur that they find out, "Whoops, I forgot to call the correct formatting function to obfuscate X, which shouldn't be available to the public."


Mwindo:
Closing since the goal is to implement as part of 10502

PR #5562, opened 2024-11-18
10547:Session Copy - Session Notes card: change styling to match other note cards

btejha:


web-client/src/views/TrialSessionWorkingCopy/SessionNotes.tsx
```diff
@@ -19,54 +19,52 @@ export const SessionNotes = connect(
   }) {
```
Mwindo, line 19: Nitpick, but probably something UX will notice: there is still a small window in which the buttons wrap above the header.

Screenshot 2024-11-20 at 9 44 06 AM

Also, I know that the reason for the setup being the way it is here comes about because the CSS in our app is a convoluted mess, but I wonder if there is a way to put the header and buttons inline at the top of the component? Semantically, I expect to see the header row at the top and the note text beneath.

btejha: Thanks for catching that! could you clarify what you mean by "a small window" are you referring to a specific screen size or resolution?

Mwindo: Sure! I just mean that if you resize the window from large to small, there is a brief moment (right before the cards stack) in which the buttons wrap above the header.

TomElliottFlexion (approved):

jimlerza (approved):

PR #5564, opened 2024-11-18
10547:Session Copy - Session Notes card: change styling to match other note cards

btejha:


web-client/src/views/TrialSessionWorkingCopy/SessionAssignments.tsx
```diff
@@ -9,8 +9,8 @@ export const SessionAssignments = connect(
   },
   function SessionAssignments({ formattedTrialSessionDetails }) {
     return (
-      <div className="card trial-session-card">
-        <div className="content-wrapper">
+      <div className="card trial-session-card height-full">
+        <div className="content-wrapper height-full">
```
Mwindo, line 13: Is this class actually necessary? Running locally, it looks the same to me whether I add it or not.

Mwindo (approved):
Added a minor comment, but LGTM overall--thanks, @btejha!

web-client/src/views/TrialSessionWorkingCopy/SessionAssignments.tsx
btejha: Yes, by adding className height-full, styling changes to the grid. Below are the differences. 
Screenshot 2024-11-18 at 4 31 14 PM

Screenshot 2024-11-18 at 4 31 26 PM

Mwindo: Sorry I wasn't clearer: I meant is the height-full class necessary on line 13 here? I think the Session Assignments box looks the same either way.

PR #5566, opened 2024-11-18
[Addendum to 10408] Rm judge edit perms from trial session info page to test

JayFlexy:
This removes the ability of the a Judge to edit Trial Session Info page. This is a request from the court and may not necessarily make it into the release of 10408 so adding here. This should be release closely (preferably in the same release as) [10408](https://github.com/orgs/flexion/projects/11/views/1?pane=issue&itemId=76509708&issue=flexion%7Cef-cms%7C10408)

### DIFF
--- a/web-client/src/presenter/computeds/formattedTrialSessionDetails.test.ts
+++ b/web-client/src/presenter/computeds/formattedTrialSessionDetails.test.ts
@@ -378,7 +378,7 @@ describe('formattedTrialSessionDetails', () => {
       });
     });
 
-    it('should be false when trial session start date is in the future, it is NOT closed, the user is a judge role', () => {
+    it('should  NOT allow judge to edit trial session info page', () => {
       mockTrialSession = {
         ...TRIAL_SESSION,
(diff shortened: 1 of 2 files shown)


Mwindo (approved):

web-client/src/presenter/computeds/formattedTrialSessionDetails.test.ts
```diff
@@ -377,6 +378,25 @@ describe('formattedTrialSessionDetails', () => {
       });
     });
 
+    it('should be false when trial session start date is in the future, it is NOT closed, the user is a judge role', () => {
```
Mwindo: I wonder if it is worth renaming to "should not allow judge to edit a trial session". Otherwise, a dev looking at this might think, "But maybe a judge should be able to edit a trial session if it _is_ closed" etc.

JayFlexy: Good point, implementing!

PR #5567, opened 2024-11-18
Load legacy pdfjs for outdated browsers

Mwindo:
Currently, pdfjs does not play nice with pre-ES2022 browsers. See https://github.com/flexion/ef-cms/issues/10531.

I have access to an outdated version of Safari (14.1.3). I did a quick test on experimental2 with the changes in this PR. I am able to upload valid pdfs on a legacy browser. I also still received error messages for improper (corrupted, secured, etc.) pdfs as expected. Modern browsers continue to work as expected as well, loading the modern version of pdfjs (confirmed via logging).

### Legacy browser with these changes (slightly refactored):

Screen Shot 2024-11-18 at 3 16 39 PM

### Legacy browser without these changes:

Screen Shot 2024-11-18 at 3 16 27 PM

### Modern browser with these changes:
Screenshot 2024-11-18 at 3 39 14 PM


web-client/src/applicationContext.ts
```diff
@@ -365,6 +365,39 @@ const reduce = ImageBlobReduce({
 
 let user;
 let broadcastChannel;
+const clientSupportsES2022 = (() => {
```
Mwindo, line 368: Most sources I saw recommended doing feature detection rather than looking at agent strings, etc. So that is the approach I took.

En-8: Huh interesting...how do you decide what features to check for? Did you "just know", is there a known/recommended list somewhere, something else?

Mwindo: Great question! The most thorough thing to do is to go through the library and see exactly which features the library needs. This is overkill and also assumes you can keep up-to-date.

The next best thing is to find some minimal set of ES2022 features: if a browser client fails to implement any of them, you know that the client does not conform to ES2022.

The features checked here are therefore based on the following:
- The updated comment in the sneakily compatability-breaking release: https://github.com/mozilla/pdfjs-dist/compare/v2.12.313...v2.13.216.
- Our message logs, many of which indicate issues with private fields and with Array prototype `at` (see the ticket for screenshots)
- Basic EMCAScript updates, like `hasOwn` instead of `hasOwnProperty`

Is this 100% foolproof? No: a random browser client could choose to implement some but not all ES2022 features. But this should be rare, and I think it is better to load the modern library when possible rather than always only loading the legacy build.

Mwindo: "See the ticket for screenshots" lol didn't link the ticket. Here we go: https://github.com/flexion/ef-cms/issues/10531. (I did this first as an experiment, so the branch/PR name doesn't have the ticket info.)

En-8 (approved):

web-client/src/applicationContext.ts
En-8: Thanks for the explanation! That makes a lot of sense. I wonder if it's worth somehow documenting the rationale for what we're checking next to where we're checking it? I'm personally unsure, and regardless, don't think that's worth blocking PR approval.

PR #5574, opened 2024-11-20
OpEx: Practitioners' email address, as defined in each of their cases

jimlerza:
Added a report that will show what email address is defined for a given practitioner in each of the cases they are associated with. This is intended to be used by a human to manually determine if any cases failed to update after a practitioner updates their email address.


scripts/reports/practitioners-email-in-cases.ts
```diff
@@ -0,0 +1,114 @@
...
+    role,
+    userId: practitionerId,
+  });
+  const practitionersEmailInCases = {};
+  for (const practitionersCase of practitionersCases) {
+    const practitionerObj = practitionersCase[`${role}s`]?.find(
+      pract => pract.userId === practitionerId,
+    );
+    if (practitionerObj && practitionerObj.email) {
+      practitionersEmailInCases[practitionersCase.docketNumber] =
```
Mwindo, line 109: Is it possible to have a duplicate docketNumber (e.g., maybe missing distinguishing suffixes)?

Mwindo (approved):

scripts/reports/practitioners-email-in-cases.ts
jimlerza: Even for docket numbers that have suffixes, the `docketNumber.S` field will not contain the suffix. The suffix is stored separately in the `docketNumberSuffix.S` field and a composite of the docket number and suffix is stored as `docketNumberWithSuffix.S` for good measure.

Because the docket number (without suffix) is used as the pk/sk for a case, there won't be duplicates.

If, however, it happens that the same practitioner can somehow be in a given case's `[irs|private]Practitioners` array twice, the `.find()` a few lines above will settle for the first match it comes across.

Mwindo: Got it--great explanation, thank you!

cruzjone-flexion (approved):

PR #5575, opened 2024-11-20
10554: Fix disappearing messages bug

En-8:
flexion#10554

- Properly set the parentMessageId when navigating to the "Apply Stamp" page so that the Stamp Order document is either associated with the correct message or no message (if stamped from document view)
- Don't mark message threads as "replied to" until after the new message has been validated, so that if there are any errors during the validation we don't mark the thread (this was what caused messages to disappear if they were impacted by the above stamping issue)
- Wrap "mark thread replied to" and "create message" writes in a single transaction, so that if writing the new message fails for whatever reason, the thread is not marked as "replied to"


Mwindo (approved):
LGTM! Thank you @En-8!

TomElliottFlexion (approved):

jimlerza (approved):

PR #5576, opened 2024-11-21
OpEx: recreate the geocoded-all-cases.csv file for QGIS

jimlerza:
Back in 2022, Mike and I generated a list of all cases starting from 2017, using the fields that were mapped on the `efcms-case` index in elasticsearch at the time.

We then fed that output into the US Census Bureau's geocoder, and generated a resulting CSV that had `lat`s and `lon`s for every case, and Tenille imported that into QGIS and made some slick maps. Hooray!

Now we want to refresh that data, so I tried to recreate the spreadsheet using the [US Census Bureau Geocoder](https://www.npmjs.com/package/us-census-geocoder) package from NPM.


jimlerza:
Closing this PR. While the logic works, I am concerned that there isn't actually enough context around this data to tell the story we think it's telling.

PR #5577, opened 2024-11-21
10553: Process to update contact info fails if DAWSON is also updating other contact info at the same time TO TEST

cruzjone-flexion:
When the change address process starts we fetch user information from persistence. Once we complete updating all the cases associated to this user with new contact info we save user information back in persistence. The issue here is that the user can initiate changing email while we are updating the cases so the user object we fetched in the beginning of the change address process is out of date now and we save this stale information back in persistence losing any changes made.


web-api/src/business/useCaseHelper/generateChangeOfAddressHelper.ts
```diff
@@ -126,12 +126,19 @@ export const generateChangeOfAddressHelper = async ({
...
+        .getUserById({ applicationContext, userId: user.userId });
+
       const userEntity = new Practitioner({
-        ...user,
+        ...liveUser,
+        contactInfo,
         isUpdatingInformation: false,
       });
 
+      resultsUser = userEntity.validate().toRawObject();
```
Mwindo: Is there any reason not to just get the most recent user data regardless of the `websocketMessagePrefix` condition?

Mwindo (approved):
One small comment, but looks good overall! Thanks @cruzjone-flexion!

web-api/src/business/useCaseHelper/generateChangeOfAddressHelper.ts
Mwindo: Eh, nevermind, ignore this.

cruzjone-flexion: Good point. I will look into where the 'admin' option is being called and see if its ok to have latest userInfo as well.

cruzjone-flexion: seems like 'admin' call does not use the user information being sent back so I think it is safe to add it there too. Thanks for the feedback!

PR #5580, opened 2024-11-21
Type Websocket Messages - TO TEST

cruzjone-flexion:


web-api/src/business/useCaseHelper/generateChangeOfAddressHelper.ts
```diff
@@ -145,10 +150,15 @@ export const generateChangeOfAddressHelper = async ({
       });
     }
 
+    const CONTACT_UPDATE_COMPLTET_ACTION:
```
Mwindo: Looks like a small typo (`COMPLTET`)

web-api/src/notifications/sendNotificationToUserTypes.ts
```diff
@@ -0,0 +1,213 @@
+import { RawPractitioner } from '@shared/business/entities/Practitioner';
```
Mwindo, line 1: I wonder if we could have:

```
type BaseNotification = {
  action: string;
};
```

And then extend each type as needed to avoid the long union type at the end.

Mwindo: E.g.

```
type NotificationWithProgress = BaseNotification & {
  filesCompleted: number;
  totalFiles: number;
};
```

web-api/src/business/useCaseHelper/generateChangeOfAddressHelper.ts
```diff
@@ -107,10 +107,15 @@ export const generateChangeOfAddressHelper = async ({
     applicationContext.logger.error(error);
   }
 
+  const NOTIFICATION_ACTION:
+    | 'user_contact_update_progress'
+    | 'admin_contact_update_progress' =
+    `${websocketMessagePrefix}_contact_update_progress`;
```
Mwindo, line 113: Nitpick: Is it worth typing this once in `generateChangeOfAddress` and then importing?

Mwindo (approved):
Small comments, but looks good overall! Thanks @cruzjone-flexion!

web-api/src/notifications/sendNotificationToUserTypes.ts
cruzjone-flexion: If action is not specific string value we can mismatch required fields for the action being used, for example I can use action1 extra props for action2


```typescript
{
action: 'action1',
action2Prop: true
}

{
action: 'action2',
action1Prop: 'string'
}
``` 

and no errors will be thrown because it does not care for specific action as long as it fits one of the extension bodies.

PR #5581, opened 2024-11-21
10554 Fix disappearing messages - to test

En-8:
flexion#10554


cruzjone-flexion (approved):

akuny (approved):

web-api/src/persistence/postgres/messages/createMessageAsReply.ts
```diff
@@ -0,0 +1,41 @@
...
+
+        return await trx
+          .insertInto('dwMessage')
+          .values(toKyselyNewMessage(newMessage))
+          .returningAll()
+          .executeTakeFirstOrThrow(() => {
+            throw new Error('could not create a message');
+          });
+      }),
+  );
```
Mwindo, line 38: I think this introduces a new pattern into our codebase: storing data before validating. In other words, we update directly in the db here rather than setting the field on the entity, validating, and calling `upsertMessages`. Is there a reason not to use `upsertMessages([themessages])`?

Mwindo: To be clear: This isn't "wrong"--updating the field in the database is a reasonable thing to do--it just isn't our normal pattern, and I'm not sure how everyone will feel about that.

web-api/src/persistence/postgres/messages/createMessageAsReply.ts
```diff
@@ -0,0 +1,41 @@
...
+  parentMessageId: string;
+  newMessage: RawMessage;
+}): Promise<RawMessage> => {
+  const messages = await getMessageThreadByParentId({
+    parentMessageId,
+  });
+
+  const createdMessage = await getDbWriter(
+    async writer =>
+      await writer.transaction().execute(async trx => {
```
Mwindo, line 19: ❤️

web-api/src/persistence/postgres/messages/createMessageAsReply.ts
Mwindo: Also, regardless of the update, I suspect we should be using `upsertCaseMessages` for the insert at the end.

En-8: I think in this case we explicitly _don't_ won't to use upsert, but I honestly hadn't thought of that until now (nor did I remember it existed, so thank you for pointing it out!). Beyond simply "it was like this when I got here" (which is obviously not the greatest), here's my reasoning:

From the postgresql docs: "ON CONFLICT DO UPDATE updates the existing row that conflicts with the row proposed for insertion as its alternative action.". In the case of messages, we only ever want to create a new message as a reply to a previous message. Everything about this message should be brand new -- including the messageId. In the unlikely event that two writes to this table conflict on message ID, `ON CONFLICT DO UPDATE` (which is what our upsert appears to do) is likely going to result in a loss of message data (whichever message came first will be overwritten). For two independently initiated message replies, this doesn't seem like what we want. If we really wanted to be resilient to clashing message IDs here, we could catch the exception that is raised by default on conflict, then reattempt the entire transaction; however, I think that's unecessary since we're using UUID messageIds as our primary key.

It's entirely possible I'm misunderstanding something though, so I'd be happy to chat through this one further if you'd like.

As for your "new pattern" comment -- let's chat if you have a moment. I'm not sure I'm 100% following.

En-8: I think in this case we explicitly don't won't to use upsert, but I honestly hadn't thought of that until now (nor did I remember it existed, so thank you for pointing it out!). Beyond simply "it was like this when I got here" (which is obviously not the greatest), here's my reasoning:

From the postgresql docs: "ON CONFLICT DO UPDATE updates the existing row that conflicts with the row proposed for insertion as its alternative action.". In the case of messages, we only ever want to create a new message as a reply to a previous message. Everything about this message should be brand new -- including the messageId. In the unlikely event that two writes to this table conflict on message ID, ON CONFLICT DO UPDATE (which is what our upsert appears to do) is likely going to result in a loss of message data (whichever message came first will be overwritten). For two independently initiated message replies, this doesn't seem like what we want. If we really wanted to be resilient to clashing message IDs here, we could catch the exception that is raised by default on conflict, then reattempt the entire transaction; however, I think that's unecessary since we're using UUID messageIds as our primary key.

It's entirely possible I'm misunderstanding something though, so I'd be happy to chat through this one further if you'd like.

As for your "new pattern" comment -- let's chat if you have a moment. I'm not sure I'm 100% following.

En-8: For documentation's sake: @Mwindo and I discussed this and agreed:
- "insert" is probably better than "upsert" in this scenario
- We're going to leave the "mark as replied to" portion of this as is. We both agree there are likely better ways to handle this write, but it's similar to how it was before the bugfix, and the bugfix is important enough to warrant shipping over continuing to flesh out the "better way", for now.

Thanks for the discussion, @Mwindo, and please feel free to comment with any additional info or corrections if I misrepresented our convo in any way!

PR #5585, opened 2024-11-22
10525: Full Spectrum Docket Sorting

cruzjone-flexion:
flexion#10525


cypress/helpers/caseDetail/docketRecord/paperFiling/fill-paper-filing-form.ts
```diff
@@ -0,0 +1,26 @@
+import { attachFile } from '../../../file/upload-file';
+import { selectTypeaheadInput } from '../../../components/typeAhead/select-typeahead-input';
+
+export function fillPaperFilingForm({
```
Mwindo, line 4: 👍

cypress/local-only/tests/integration/caseDetail/docketRecord/docketRecordSort.cy.ts
```diff
@@ -0,0 +1,156 @@
+import { createAndSaveForLaterPaperFiling } from '../../../../../helpers/caseDetail/docketRecord/paperFiling/create-and-save-for-later-paper-filing';
```
Mwindo: I think naming convention would have this as `docket-record-sort.cy.ts`

shared/src/business/entities/EntityConstants.ts
```diff
@@ -6,6 +6,10 @@ import courtIssuedEventCodesJson from '../../tools/courtIssuedEventCodes.json';
 import externalFilingEventsJson from '../../tools/externalFilingEvents.json';
 import internalFilingEventsJson from '../../tools/internalFilingEvents.json';
 
+export const KEYS = {
+  DOCKET_RECORD_TABLE_SORT: 'DOCKET_RECORD_TABLE_SORT',
+};
```
Mwindo: Can you explain why we use `KEYS` and `ROOT`? It seems like we are using this as a way to dynamically set state based on keys, but this worries me: working with Cerebral can already be opaque enough (always interacting with a global state object), and this approach seems to "hide" state (it cannot be seen in `state.ts`, and it isn't immediately obvious where this state is defined/typed).

To be clear: I'm not saying this approach is wrong or bad, I just want to understand it better!

In any case, if we do use ROOT/KEY, I think we should make these names more descriptive. A root into what? What sort of key? Etc. These terms need context.

web-client/src/views/DocketRecord/DocketRecord.tsx
```diff
@@ -287,3 +377,54 @@ export const DocketRecord = connect(
 );
 
 DocketRecord.displayName = 'DocketRecord';
+
+export function SortableHeader({
```
Mwindo: I feel like exporting `SortableHeader` as a thin wrapper around `SortableColumn` might be a little confusing: e.g.,"Header" vs. "Column" even though they are the "same" under the hood, `SortableHeader` in principle has little to do with DocketRecord, etc. I wonder if it is worth passing in `hideOnMobile` straight into `SortableColumn` (with default = false) and maybe even renaming it (reflecting your new, better name) to `SortableColumnHeader` (and then using only this `SortableColumnHeader`)?

Mwindo (commented):
Thanks @cruzjone-flexion and @nechama-krigsman! This looks good overall.

I did see a discrepancy with the sorted order in the UI vs. the printable record when sorting by pages locally (I let the former of you know), and I think I need to understand the purpose of `KEYS` and `ROOT` a little better before giving my full thumbs up.

Mwindo (approved):
Thanks for addressing the feedback :)

nechama-krigsman (approved):

jimlerza (approved):

PR #5582, opened 2024-11-21
10293: use client-side Paginator component in three reports

akuny:
See [10293](https://github.com/flexion/ef-cms/issues/10293).

**Edit:** Goal for this PR is to get these changes into test in order to test at scale, given the fact that client-side pagination might not be the right fit for all three of these reports.

This PR integrates the client-side paginator into the following reports:

- **Case Inventory**
- **Deadlines**
- **Pending Report**

#### **Summary of Changes**
1. **Client-Side Paginator Refactoring**:
   - Extracted the `useClientSidePaginator` hook from `ColdCaseReportList.tsx` to allow for consistent client-side pagination across different reports.
   - Since pagination state is exclusively relevant to the React components displaying these reports, I could not think of a compelling reason to manage this state within individual slices of the Cerebral state tree.

2. **Removal of Server-Side Pagination Logic**:
   - The following persistence functions no longer include server-side pagination logic:
     - `getCaseInventoryReport`
     - `fetchPendingItems`
     - `getCaseDeadlinesByDateRange`
   - Corresponding interactors and proxies have been updated to reflect this change.

3. **Removal of Obsolete Cerebral Functionality**:
   - The following sequences and actions have been removed, and associated Cerebral integration tests have been updated accordingly:
     - `caseInventoryReportLoadMoreSequence`
     - `loadMoreCaseDeadlinesSequence`
     - `loadMorePendingItemsSequence`
     - `incrementPendingItemsPageAction`
     - `incrementScreenMetadataPageAction`

PR #5593, opened 2024-11-26
10293 To Test - Repaginate Case Inventory Report

akuny:
This PR build on [recent changes](https://github.com/ustaxcourt/ef-cms/pull/5582) in pursuit of [10293](https://github.com/flexion/ef-cms/issues/10293).

It reincorporates server-side pagination for the Case Inventory Report and uses Cerebral to manage frontend state.

TODO: Cypress test for this change.


shared/src/business/utilities/getFormattedCaseDetail.ts
```diff
@@ -18,6 +18,17 @@ import { UnknownAuthUser } from '@shared/business/entities/authUser/AuthUser';
...
+  docketNumber: string;
+  caseTitle: string;
+  consolidatedIconTooltipText: string;
+  inConsolidatedGroup: boolean;
+  isLeadCase: boolean;
+  associatedJudge?: string;
+  status: string;
+  [key: string]: unknown;
+};
+
```
TomElliottFlexion, line 31: TY

web-client/src/presenter/actions/CaseInventoryReport/getCaseInventoryReportAction.ts
```diff
@@ -10,21 +10,26 @@ import { state } from '@web-client/presenter/app.cerebral';
...
       });
 
-    store.set(state.caseInventoryReportData, {
-      foundCases: reportData.foundCases,
-    });
+    store.set(state.caseInventoryReportData.foundCases, reportData.foundCases);
+    store.set(
+      state.caseInventoryReportData.foundCasesCount,
+      reportData.totalCount,
+    );
```
Mwindo: Not a big deal either way, but is there any particular reason we want to split apart `foundCases` and `foundCasesCount`? The latter seems redundant given that the time complexity of `Array.length` is O(1). (I first noticed this because some tests look strange when we say `foundCases: []` but `foundCasesCount: 3`, etc.)

Mwindo (approved):
Left a small comment, but LGTM overall--thanks @akuny!

web-client/src/presenter/actions/CaseInventoryReport/getCaseInventoryReportAction.ts
akuny: Great point! I think unclear naming conventions are the culprit:

`foundCases` and `foundCasesCount` are tracked separately in state because the former actually contains the _current page's_ found cases, while the latter represents the total number of found cases _across all pages_.

`foundCases` should really be something like `foundCasesForCurrentPage`, while `foundCasesCount` should be something like `foundCasesTotalCount`. I'll clean this up.

Mwindo: Ah, that makes sense. Thanks!

PR #5545, opened 2024-11-14
10472 story

JayFlexy:
See Ticket: [10472 - Open PDF in new tab](https://github.com/orgs/flexion/projects/11/views/1?pane=issue&itemId=76486952&issue=flexion%7Cef-cms%7C10472)

PR #5607, opened 2024-11-29
10472: PDF Preview in new Tab to test

JayFlexy:
10472: open pdf previews in new tab. See #5545


web-client/src/views/PDFPreviewButton.tsx
```diff
@@ -9,7 +9,9 @@ import { state } from '@web-client/presenter/app.cerebral';
 import React from 'react';
 
 const pdfPreviewButtonDeps = {
-  openPdfPreviewModalSequence: sequences.openPdfPreviewModalSequence,
+  loadPdfForTabSequence: sequences.loadPdfForTabSequence,
+  openCaseDocumentDownloadUrlSequence:
+    sequences.openCaseDocumentDownloadUrlSequence,
```
Mwindo: Is this used? I don't see it below.

web-client/src/presenter/actions/PDFPreviewTab/loadPdfForTabAction.ts
```diff
@@ -0,0 +1,44 @@
+export const loadPdfForTabAction = ({
```
Mwindo, line 1: Do we need `loadPdfAction` and `loadPdfSequence` anymore after adding this action and its related sequence? A cursory glance suggests to me that we might not, since these look like they are used for loading PDFs in modals, which this ticket is changing. If this is true, let's remove those files and associated tests and remove the sequence from `presenter.ts`; otherwise, if we can't remove the old action/sequence, I wonder if there is some way to avoid duplicating logic since this action is almost a copy of `loadPdfAction`.

web-client/src/presenter/actions/PDFPreviewTab/loadPdfForTabAction.ts
```diff
@@ -0,0 +1,44 @@
...
+    reader.onerror = function (err) {
+      reject(err);
+    };
+
+    if (isBase64Encoded) {
+      // @ts-ignore
+      reader.readAsDataURL(file);
+    } else {
+      reader.readAsArrayBuffer(file);
+    }
```
Mwindo: I would spend some time investigating this since the type mismatch looks like a mistake. `readAsDataURL` expects a `Blob` or `file` ... but file is a `string` here. It might be as simple as something like `reader.readAsText(new Blob([file]));`, but I'm not sure.

TomElliottFlexion (approved):

web-client/src/views/PDFPreviewButton.tsx
JayFlexy: Good catch!

web-client/src/presenter/actions/PDFPreviewTab/loadPdfForTabAction.ts
JayFlexy: Removed loadPdfAction & Sequence along with some related code.

web-client/src/presenter/actions/PDFPreviewTab/loadPdfForTabAction.ts
JayFlexy: Checked on this. Awaiting for test env to merge

PR #5608, opened 2024-11-29
DevEx: add fix for docker compose invocation - Test

JayFlexy:
creates compatibility for docker compose or docker-compose


jimlerza:
The container that the github actions run in has `docker-compose` installed in it so this doesn't need to go to `test` before it goes to `staging`.

PR #5609, opened 2024-11-29
DevEx: add fix for docker compose invocation

JayFlexy:
creates compatibility for docker compose or docker-compose

### DIFF
--- a/run-local.sh
+++ b/run-local.sh
@@ -1,15 +1,21 @@
 #!/bin/bash
 # Used for running the API and necessary services (dynamo, s3, elasticsearch) locally
 
+# Determine the docker compose invocation.
+if command -v docker-compose &> /dev/null; then
+  DOCKER_COMPOSE="docker-compose"
+else
+  DOCKER_COMPOSE="docker compose"
+fi
 # shellcheck disable=SC1091
 . ./setup-local-env.sh
 
 if [[ -z "$CI" ]]; then
   echo "Stopping postgres in case it's already running"
-  docker compose -f web-api/src/persistence/postgres/docker-compose.yml down || true
+  $DOCKER_COMPOSE -f web-api/src/persistence/postgres/docker-compose.yml down || true
 
   echo "Starting postgres"
-  docker compose -f web-api/src/persistence/postgres/docker-compose.yml up -d || { echo "Failed to start Postgres containers"; exit 1; }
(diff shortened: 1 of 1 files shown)


Mwindo (approved):
As @pixiwyn mentioned in standup, we do have a change to use `podman` on the way ... but I think I prefer your version here which is more agnostic. Curious to hear other thoughts though. Thanks, @JayFlexy!

En-8 (approved):

TomElliottFlexion (approved):

run-local.sh
```diff
@@ -1,15 +1,21 @@
 #!/bin/bash
 # Used for running the API and necessary services (dynamo, s3, elasticsearch) locally
 
+# Determine the docker compose invocation.
+if command -v docker-compose &> /dev/null; then
+  DOCKER_COMPOSE="docker-compose"
+else
+  DOCKER_COMPOSE="docker compose"
+fi
```
En-8, line 9: There are some possibly-breaking changes between `docker-compose` and `docker compose` regarding container names:

> Compose generates container names based on the project name, service name, and scale/replica count. In Compose V1, an underscore (_) was used as the word separator. In Compose V2, a hyphen (-) is used as the word separator.

I am not sure if this makes a huge difference for our local development environment, but something still to be mindful of.

Link for reference: https://docs.docker.com/compose/releases/migrate/

pixiwyn (approved):

jimlerza:
> As @pixiwyn mentioned in standup, we do have a change to use `podman` on the way ... but I think I prefer your version here which is more agnostic. Curious to hear other thoughts though. Thanks, @JayFlexy!

While I am all for enabling and embracing choice for manual workflows, I think the app's build toolchain should be standardized so we can eliminate discrepancies at compilation and runtime. In theory this can reduce instances of "why does this work for me and not for you?" frustration.

That said, I dunno...how a developer runs the app locally feels like a pretty personal thing. If this were anywhere other than `run-local.sh` this would be an easier call.

JayFlexy:
> > As @pixiwyn mentioned in standup, we do have a change to use `podman` on the way ... but I think I prefer your version here which is more agnostic. Curious to hear other thoughts though. Thanks, @JayFlexy!
> 
> While I am all for enabling and embracing choice for manual workflows, I think the app's build toolchain should be standardized so we can eliminate discrepancies at compilation and runtime. In theory this can reduce instances of "why does this work for me and not for you?" frustration.
> 
> That said, I dunno...how a developer runs the app locally feels like a pretty personal thing. If this were anywhere other than `run-local.sh` this would be an easier call.

Should we discuss in parking lot? I would like to hear more on this

PR #5615, opened 2024-12-02
10549 design debt intermediate branch to test 1733176098

btejha:


web-client/src/views/CaseDetail/CaseDetailHeader.tsx
```diff
@@ -168,7 +168,10 @@ export const CaseDetailHeader = connect<
                         size="1x"
                       />
                     )}
-                    <CaseLink formattedCase={formattedCaseDetail}>
+                    <CaseLink
+                      className="no-wrap-text"
+                      formattedCase={formattedCaseDetail}
+                    >
```
Mwindo, line 174: Testing locally, it seems like the following works:

```
<CaseLink
  className="mobile-text-wrap"
  formattedCase={formattedCaseDetail}
>
```

I could be missing something, however. If the above _does_ work, I think we can avoid making a new CSS class and just use the existing one. (In fact, if the above doesn't work, I think we can still avoid making a new CSS class. But let's worry about that only if it doesn't work!)

Mwindo: And if we don't want the docket-number to wrap, we can do:

```
<CaseLink
  className="mobile-text-wrap"
  formattedCase={formattedCaseDetail}
>
  Docket Number:{' '}
  <span className="no-wrap">
    {formattedCaseDetail.docketNumberWithSuffix}
  </span>
</CaseLink>
```

Mwindo (approved):
As we discussed, I _think_ we can use existing CSS classes to get the behavior we want; if so, let's do that. If it turns out we can't, then so be it! Thanks @btejha!

PR #5638, opened 2024-12-07
Remove chromium from dockerfile experiment

Mwindo:
Just trying to see if I can get the pdf test to pass


Mwindo:
Handled in https://github.com/ustaxcourt/ef-cms/pull/5639

PR #5641, opened 2024-12-09
BugFix: Fix hanging Node.js scripts by adding explicit exit signals

JayFlexy:
Currently, on some local machines, running `npm run start:api` along with `npm run postgres:seed` and other scripts fails to successfully hand off to subsequent scripts. This causes a state where Node.js scripts hang indefinitely without clear indications of the issue.

To address this, we have added explicit exits using `process.exit(0)` in the promise resolution block of the affected scripts. This provides a clear 'done' signal, ensuring that scripts terminate correctly. While this issue is currently observed on one machine, it is best practice to communicate explicit signals from scripts to accommodate variations in local environments.

### DIFF
--- a/web-api/src/persistence/postgres/utils/migrate/migrate.ts
+++ b/web-api/src/persistence/postgres/utils/migrate/migrate.ts
@@ -36,4 +36,12 @@ async function migrateToLatest() {
   });
 }
 
-migrateToLatest().catch;
+migrateToLatest()
+  .then(() => {
+    console.log('Postgres migration completed Successfully!');
+    process.exit(0);
+  })
+  .catch(err => {
+    console.log('Migration failed.');
+    console.log(err);
+  });
--- a/web-api/src/persistence/postgres/utils/seed/seed.ts
+++ b/web-api/src/persistence/postgres/utils/seed/seed.ts
@@ -11,4 +11,12 @@ export const seed = async () => {
   );
 };
 
-seed().catch;
+seed()
+  .then(() => {
+    console.log('Database data seeded successfully!');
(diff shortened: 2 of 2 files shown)


Mwindo (approved):

pixiwyn (approved):

web-api/src/persistence/postgres/utils/migrate/migrate.ts
```diff
@@ -36,4 +36,12 @@ async function migrateToLatest() {
...
-migrateToLatest().catch;
+migrateToLatest()
+  .then(() => {
+    console.log('Postgres migration completed Successfully!');
+    process.exit(0);
+  })
+  .catch(err => {
+    console.log('Migration failed.');
+    console.log(err);
+  });
```
jimlerza, line 47: Other than the helpful log message, why `catch` here? Doesn't this process still hang if we don't re-`throw` or do a `process.exit` inside this `catch`?

jimlerza (approved):

web-api/src/persistence/postgres/utils/migrate/migrate.ts
JayFlexy: In our testing, even when the script was hanging it wasn't because of a local error to be caught so we never actually saw that bit of code running. Was just adding that there just in case but can remove if you want @jimlerza

jimlerza: Up to you! I'm fine with it as-is, but if we think this script should always exit explicitly, then we could add `process.exit(1)` inside the `catch` (in both files).

PR #5642, opened 2024-12-10
Modify restoreDbFromSource to drop all target tables before restore

Mwindo:
Our current implementation of `restoreDbFromSource` only drops and re-creates tables in our target db that exist in our `pg_dump`. Instead, we want a clean slate in our target db.

We could drop and re-create the target database or drop and re-create the target schema, but both of these involve dealing with stricter permissions. Instead, the approach in this PR is to query for all of the tables from the target schema and drop them.


scripts/postgres/restoreDbFromSource.ts
```diff
@@ -215,8 +215,61 @@ async function restoreFromBackup({
   });
   const targetPassword = await targetSigner.getAuthToken();
 
+  // pg_restore only drops tables in the dump, so we drop all tables before pg_restore.
```
Mwindo: This script file is getting long/clumsy. I will see if there are easy ways to simplify in the near future.

jimlerza (approved):

PR #5644, opened 2024-12-10
Modify restoreDbFromSource to drop all target tables before restore

Mwindo:
Our current implementation of `restoreDbFromSource` only drops and re-creates tables in our target db that exist in our `pg_dump`. Instead, we want a clean slate in our target db. Without a clean slate, migrations can fail :(

We could drop and re-create the target database or drop and re-create the target schema, but both of these involve dealing with stricter permissions. Instead, the approach in this PR is to query for all of the tables from the target schema and drop them.


scripts/postgres/restoreDbFromSource.ts
```diff
@@ -215,8 +215,61 @@ async function restoreFromBackup({
   });
   const targetPassword = await targetSigner.getAuthToken();
 
+  // pg_restore only drops tables that exist in the source dump, so we drop all target tables before calling pg_restore.
+  // We could drop the whole target db or the schema, but then we would have to deal with stricter permissions.
   await new Promise(resolve => {
-    const result = spawn(
+    // For each table in the target db public schema, we will create a SQL DROP command and then execute it.
```
akuny, line 221: Small piece of feedback: this new step in the script is lumped into the `restoreFromBackup` function. The `main` function may be more explicit about what is happening step by step if it called a new function named something to the effect of `dropAllTables`, instead of the dropping happening as a result of calling `restoreFromBackup`.

akuny (approved):
@Mwindo Looks good, one small suggestion!

scripts/postgres/restoreDbFromSource.ts
Mwindo: Good call. I will do this as a follow-up :)

Mwindo: (Want to make sure this gets delivered before the next glue job, and I've tested as is.)

PR #5647, opened 2024-12-11
Address slower page loads in test

Mwindo:
This PR moves us from 3 db requests (1 to Dynamo, 2 to Postgres) to 2 db requests (1 to Dynamo, 1 to Postgres) in `getCaseByDocketNumber`. The hope is that eliminating these db requests will speed up page loads on pages that are getting case information for possibly hundreds of cases. (E.g., the trial session details page.)


web-api/src/persistence/postgres/utils/migrate/migrations/0007-init-case-correspondence-indexes.ts
```diff
@@ -0,0 +1,16 @@
...
+    .column('docketNumber')
+    .execute();
+}
+
+export async function down(db: Kysely<any>): Promise<void> {
+  await db.schema
+    .dropIndex('idx_caseCorrespondence_docketNumber')
+    .on('dwCaseCorrespondence')
+    .execute();
+}
```
Mwindo: Lack of an index was causing a slowdown. (The index is currently on test, added through psql.) On our next migration, it would be nice to consolidate some of these migration files to avoid one-offs like 0008-forgot-this-or-that-index.

pixiwyn (approved):

PR #5648, opened 2024-12-12
10553: Process to update contact info fails if DAWSON is also updating other contact info at the same time

cruzjone-flexion:
flexion#10553


web-api/src/business/useCases/user/queueEmailUpdateAssociatedCasesWorker.test.ts
```diff
@@ -0,0 +1,186 @@
...
+import { mockPetitionerUser } from '@shared/test/mockAuthUsers';
+import { petitionerUser } from '@shared/test/mockUsers';
+
+function sleep(timeInMilliseconds: number) {
+  return new Promise(resolve => {
+    setTimeout(() => {
+      resolve(null);
+    }, timeInMilliseconds);
+  });
+}
```
Mwindo: It looks like we have a function already: https://github.com/ustaxcourt/ef-cms/blob/af8e9ca30d8392ccd630cd25b57139120509f3ba/shared/src/tools/helpers.ts#L61

web-api/src/business/useCases/user/queueEmailUpdateAssociatedCasesWorker.ts
```diff
@@ -0,0 +1,111 @@
...
+}: {
+  applicationContext: ServerApplicationContext;
+  user: RawUser | RawPractitioner;
+}): Promise<void> {
+  user.isUpdatingInformation = false;
+  let userEntity;
+  if (
+    user.role === ROLES.privatePractitioner ||
+    user.role === ROLES.irsPractitioner ||
+    user.role === ROLES.inactivePractitioner
```
Mwindo: This comment isn't really related to your bugfix, but I've seen this check in a few places, and I feel nervous about continuing this pattern. (It is very easy for a dev to think, "Ah, let me create a User" without realizing they need to create a _Practitioner_ instead, or it is easy for a dev to forget one of the disjuncts like `ROLES.inactivePractitioner`.)

My first thought is that it would be nicer if we could construct the appropriate user with a factory, but that's more complicated and outside your ticket's scope. However, I wonder about your thoughts pulling this check (here and wherever else it is used, which includes admin.ts, changePasswordInteractor, and updateAssociatedCaseWorker) into a function like "isPractitionerUser" or something, which is at least a little cleaner.

web-api/src/business/useCases/user/queueEmailUpdateAssociatedCasesWorker.ts
```diff
@@ -0,0 +1,111 @@
...
+    .getPersistenceGateway()
+    .getDocketNumbersByUser({
+      applicationContext,
+      userId: user.userId,
+    });
+
+  if (!docketNumbersByUser.length) {
+    await disableIsUserUpdatingFlag({ applicationContext, user });
+    return;
+  }
```
Mwindo, line 41: Scattered thoughts:

Is there any path that could lead to us failing to disable and then locking a user out from updating their info? I suppose a failed request could result in this. Seems like this is something we can improve once we get user data into postgres. (We can update one field at a time, for instance, and/or lock the table.)

Also wondering if our existing [withLocking](https://github.com/ustaxcourt/ef-cms/blob/af8e9ca30d8392ccd630cd25b57139120509f3ba/web-api/src/business/useCaseHelper/acquireLock.ts#L136) would be able to handle this in the meantime ... although, frankly, I don't know much about it.

Feel free to respond to or ignore this.

web-api/src/business/useCases/user/queueEmailUpdateAssociatedCasesWorker.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  await waitUntilAllExpectedCasesAreUpdatedWithEmail({
+    applicationContext,
+    userEmail: user.email!,
+  })
+    .catch(error =>
+      console.error(`ERROR CHECKING COUNT OF UPDATED CASES -> ${error}`),
+    )
+    .finally(async () => {
+      await disableIsUserUpdatingFlag({ applicationContext, user });
```
Mwindo, line 59: 👍

web-client/src/presenter/sequences/updateUserInformationSequence.ts
```diff
@@ -0,0 +1,4 @@
+import { getUserAction } from '../actions/getUserAction';
+import { setUserAction } from '@web-client/presenter/actions/setUserAction';
+
+export const updateUserInformationSequence = [getUserAction, setUserAction];
```
Mwindo: Where is this being used?

Mwindo (approved):
It is difficult for me to give a _great_ review of this--I am definitely relying on the tests that you and Tenille have done--but it looks reasonable? I don't want to hold up code.

My main concern is that I think a lot of this work will change substantially as we move off of Dynamo and mostly off of Open Search and onto Postgres. In other words, this is a good deal of somewhat complex code to solve a problem that will likely be easier to handle in a couple of months (when we can avoid doing updates via upserts that overwrite the entire entity).

shared/src/business/entities/factories/UserFactory.ts
```diff
@@ -0,0 +1,27 @@
+import { Practitioner } from '@shared/business/entities/Practitioner';
+import { ROLES, Role } from '@shared/business/entities/EntityConstants';
+import { User } from '@shared/business/entities/User';
+
+type MinimalFactoryInfo = {
+  role: Role;
+};
+
+export class UserFactory {
```
Mwindo, line 9: I love that you are tackling this! That said, I am a little wary given that we have `Practitioner`, `PrivatePractitioner`,  and `IrsPractitioner`. I am not sure how these all play together, and if we only get the UserFactory halfway working, someone might call it expecting it to get the right thing in all cases when it doesn't.

Mwindo: (And then depending on complexity this maybe involves more testing, etc.)

cruzjone-flexion (commented):
It is true @Mwindo that once this is migrated to Postgres it will simplify this logic. That is why we decided to make sure these process do not overlap and they wait until one finishes before starting the next (instead of modifying process to work in parallel). In the worker we check that the expected count of cases are updated and then toggle the flag again. We also error handle and timeout handle to make sure user is not stuck in a update state.

jimlerza (approved):

PR #5654, opened 2024-12-17
DevEx: Implement the argument parser in our typescript scripts

jimlerza:
See #5606 for the fancy marketing blahblah

[Successful migration deployment](https://app.circleci.com/pipelines/github/ustaxcourt/ef-cms/9506/workflows/c2506794-f45f-470c-b89b-5a896430223f)


zachrog (approved):
Overall this looks good and helps us standardize how we run our scripts. Thank you for cleaning all of this up. Big fan.

Personally, I think that handling the argument parsing should be done by a package rather than rolling our own, or custom functionality like "requireActiveAwsSession" be built on top of a package. Maintaining our own arg parser does not seem valuable to me, however I recognize the court is the one who uses scripts far more often than Flexion so trust your judgement and needs more than my own. Definitely not a show stopper but wanted to be direct.

jimlerza:
> Overall this looks good and helps us standardize how we run our scripts. Thank you for cleaning all of this up. Big fan.
> 
> Personally, I think that handling the argument parsing should be done by a package rather than rolling our own, or custom functionality like "requireActiveAwsSession" be built on top of a package. Maintaining our own arg parser does not seem valuable to me, however I recognize the court is the one who uses scripts far more often than Flexion so trust your judgement and needs more than my own. Definitely not a show stopper but wanted to be direct.

Argument parsing is done by `node:utils`, which is not a package, but built into node itself. We started implementing that in a few scripts, and it worked well, but it required a _lot_ of code to set up the object and then get the values out, so I abstracted most of that into a wrapper.

This is a follow-up PR which implements that wrapper in the scripts. The [original PR](https://github.com/ustaxcourt/ef-cms/pull/5606) was just the wrapper and its tests.

TomElliottFlexion (approved):

PR #5679, opened 2025-01-07
10502 dxox

### LINKED ISSUE #8836: Move Case to Postgres
branch off: origin/10391-dynamo-migrations-kysely-aurora-rds-tokens (till merged in)

DynamoDB Entities to Replace:
- case

OpenSearch Indexes to Keep:
- efcms-case (used to search cases)

convert the functions in web-api/src/persistence/dynamo/cases, and web-api/src/persistence/dynamo/caseWorksheet to postgres kysely functions.

DynamoDB Functions to Move (unit tests can be removed):

pixiwyn:
flexion#10502

Pre-deployment:
- Update Migration Files and Order
- **Ensure AWS terraform provider for new opensearch sync stuff is up-to-date with staging**
- After maintenance mode, but before deploying: take snapshot of Postgres RDS

Post-deployment:
- Run `scripts/run-once-scripts/postgres-migration/delete-cases.ts`
- Monitor DB load/performance

This PR is giant: it is changing how cases are stored/fetched in a case management system.

Some highlights:
- Move case data out of Dynamo and into Postgres.
- Set up a way to index from Postgres to OpenSearch since we can no longer rely on DynamoDB Streams.
- Switch traffic to point to a single region. This eliminates "we've written in the east but haven't replicated that data to the west" issues and should speed the app up in general for west users. (Fewer average round trips to east.)
- Change our persistence functions related to cases to make the transition from Dynamo to Postgres more efficient. A one-to-one mapping of data storage/access patterns was not possible because it caused efficiency issues (i.e. timeouts).
- Improve our Postgres connection code to reduce number of unnecessary connections and health checks.
- Prefer Promise.allSettled to Promise.all when persisting data as one imperfect means of reducing inconsistencies in interrupted workflows.
- Translate several OpenSearch queries to Postgres queries.
- Better typing for case-related data. In the past, we often lied.
- Fix, and adding better error handling, for a devilish bug that was basically unrelated to 10502 but might have come up at some point later (and been even _more difficult_ to debug then!): https://github.com/flexion/ef-cms/issues/10658
- Fix an issue with LambdaClient having too many concurrent connections: https://github.com/flexion/ef-cms/issues/10664.
- Prevent lambda functions from retrying when called asynchronously.


shared/src/authorization/authorizationClientService.test.ts
```diff
@@ -8,27 +8,14 @@ import {
...
-  });
-
-  it('should return true for any user whose userId matches the 3rd owner argument, in this case "someUser" === "someUser"', () => {
-    expect(
-      isAuthorized(
-        mockPetitionerUser,
-        'unknown action' as any,
-        mockPetitionerUser.userId,
-      ),
-    ).toBeTruthy();
```
Mwindo, line 31: Before, `isAuthorized` allowed an optional `userId` to be passed (e.g., a user "isAuthorized" to see some case). However, this made authorization in general more complicated: we were doing similar checks in other function calls. Therefore, the `userId` parameter has been eliminated. See `CaseFactory.ts` for more on related refactoring.

shared/src/authorization/authorizationClientService.test.ts
```diff
@@ -108,14 +95,6 @@ describe('Authorization client service', () => {
     });
   });
 
-  describe('irsPractitioner', () => {
-    it('should be authorized to get a case', () => {
-      expect(
-        isAuthorized(mockIrsPractitionerUser, ROLE_PERMISSIONS.GET_CASE),
```
Mwindo, line 126: Users either can get all case data (e.g., internal users) or else only data for a case they are associated with. This permission itself did not do anything. (Any user can get a case. Finer-grained permissions determine which ones.)

shared/src/business/entities/Message.test.ts
```diff
@@ -1,7 +1,7 @@
 import { CASE_STATUS_TYPES, PETITIONS_SECTION } from './EntityConstants';
 import { Message, RawMessage } from './Message';
 import { createISODateString } from '../utilities/DateHandler';
-import { getTextByCount } from '../utilities/getTextByCount';
+import { getTextByCount } from '@shared/test/getTextByCount';
```
Mwindo, line 4: Moved this because it is only used for testing.

shared/src/business/entities/Message.ts
```diff
@@ -50,9 +51,10 @@ export class Message extends JoiValidationEntity {
     this.createdAt = rawMessage.createdAt || createISODateString();
     this.leadDocketNumber = rawMessage.leadDocketNumber;
     this.docketNumber = rawMessage.docketNumber;
-    this.docketNumberWithSuffix = `${rawMessage.docketNumber}${
-      rawMessage.docketNumberSuffix ? rawMessage.docketNumberSuffix : ''
-    }`;
+    this.docketNumberWithSuffix = Case.getDocketNumberWithSuffix({
+      docketNumber: rawMessage.docketNumber,
+      docketNumberSuffix: rawMessage.docketNumberSuffix,
+    });
```
Mwindo, line 57: We store `docketNumber` and `docketNumberSuffix` in the DB. We could store `docketNumberWithSuffix` as a column in our DB rather than stitching the two pieces of data together using this function. However, we opted for the latter approach instead because 1) unless it is computationally expensive, it is good practice to persist only the data you need and 2) it avoids a developer seeing `docketNumber`, `docketNumberSuffix`, and `docketNumberWithSuffix` and wondering, "Wait, is there a difference between `docketNumber + docketNumberSuffix` and `docketNumberWithSuffix`? Will `docketNumberWithSuffix` always exist? Should I do something like `docketNumberWithSuffix || docketNumber + docketNumberSuffix` just to be safe?" and so on.

web-api/src/persistence/postgres/cases/getCaseMetadataByDocketNumber.ts
```diff
@@ -30,6 +30,9 @@ export const getCaseMetadataByDocketNumber = async ({
         ...transformNullToUndefined(rawCaseEntity(dbCase)),
         petitioners:
           (dbCase.petitioners as RawPetitioner[]).map(p => {
+            if (!p) {
+              return;
+            }
```
Mwindo: Probably best to let validation catch this rather than erroring out here.

web-api/src/business/useCases/processStreamRecords/processPractitionerMappingEntries.ts
```diff
@@ -26,49 +29,31 @@ export const processPractitionerMappingEntries = async ({
...
-      });
+      const docketNumber = practitionerMappingData.pk.S.substring(
+        'case|'.length,
+      );
 
-      if (!caseMetadataWithCounsel) {
-        throw Error(
-          `Unable to index ${practitionerMappingData.pk.S.substring('case|'.length)} case data not found`,
+      // After case records have been moved into postgres, we need to get the case data associated with the practitioner from postgres.
+      // However, when we try to fetch case data here during the initial blue-green migration re-indexing step (to get case records
```
Mwindo, line 37: Meant to say in-lining. Will update in a more substantive commit.

web-api/src/business/useCases/processStreamRecords/processPractitionerMappingEntries.ts
```diff
@@ -26,49 +29,31 @@ export const processPractitionerMappingEntries = async ({
...
+      // into postgres in the first place), the case data might not yet have been moved over to postgres. Therefore, we fallback to the case data from Dynamo.
+      // TODO after 10502: Only rely on postgres by lining getCaseDataFromPostgres here.
```
Mwindo: Meant to say in-lining. Will update in a more substantive commit.

shared/src/business/utilities/setupPdfDocument.ts
```diff
@@ -1,14 +1,8 @@
-export const setupPdfDocument = async ({
-  alternateFont = false,
-  applicationContext,
-  pdfData,
-}) => {
+export const setupPdfDocument = async ({ applicationContext, pdfData }) => {
```
Mwindo, line 1: The `alternateFont` parameter was never used. (In fact, looking at the type--`Promise<PdfFont>` rather than the expected `PdfFont`--I suspect it wouldn't have worked even if used.) I removed it.

shared/src/business/entities/cases/CaseFactory.ts
```diff
@@ -0,0 +1,201 @@
+import {
```
Mwindo, line 1: The original intent behind this file was to clean up a lot of the convoluted code in `getCaseInteractor.ts`, throughout which we were sometimes making redundant calls, calls that did not do anything, and calls that were distributed in ways that made following the code tricky. (@zachrog deserves credit for working through a lot of this with me!) The factory was meant to be a "one-stop shop" constructor helper for getting clean Case data: a dev could call this and be confident that they had a "safe" subset of data.

Working through more of the PR and seeing the way we get case data, I do not think the "one-stop shop" idea will be so simple. (Long story short: we access different subsets of case data, so a single "Factory" really only works in limited contexts.) It's therefore best to think of this as easier-to-follow code for `getCaseInteractor.ts` and `getPublicCaseInteractor.ts` that might have limited usefulness elsewhere.

shared/src/business/entities/cases/RestrictedCase.ts
```diff
@@ -0,0 +1,26 @@
...
+import { JoiValidationEntity } from '@shared/business/entities/JoiValidationEntity';
+
+// An entity for case details for a case a user does not have access to
+export class RestrictedCase extends JoiValidationEntity {
+  public docketNumber?: string;
+  public docketNumberSuffix?: string;
+  public isPaper?: boolean;
+  public isSealed?: string;
+  public leadDocketNumber?: boolean;
+  public docketEntries: {}[];
```
Mwindo, line 11: This is a compromise: we sometimes _expect_ "Case" data to have docketEntries. Rather than doing a falsy check in these situations, I thought it was cleaner to just make this an empty list. Now I'm having second thoughts, but I don't have strong feelings either way.

shared/src/business/entities/contacts/Contact.test.ts
```diff
@@ -2,13 +2,13 @@ import { Contact } from './Contact';
 import { MOCK_CONTACT_PRIMARY } from '../../../test/mockContact';
 
 describe('Contact', () => {
-  describe('hasEAccess validation', () => {
+  describe('hasElectronicAccess validation', () => {
```
Mwindo, line 5: We renamed columns with `E` abbreviation for two reasons:
1. It was not consistent with our other, non-abbreviated nomenclature
2. Our library had problems converting to/from Postgres snake_case

shared/src/business/entities/customCaseReportSearch/CustomCaseReportSearch.ts
```diff
@@ -35,10 +35,7 @@ export class CustomCaseReportSearch extends JoiValidationEntity {
   public pageSize: number;
```
Mwindo, line 35: `CustomCaseSearch` was moved from OpenSearch to Postgres. The changes below reflect that.

shared/src/business/useCases/getCaseDeadlinesInteractor.test.ts
```diff
@@ -170,108 +172,4 @@ describe('getCaseDeadlinesInteractor', () => {
       startDate: START_DATE,
     });
   });
-
-  it('logs an error for any case that fails validation and includes all cases that pass validation', async () => {
```
Mwindo, line 199: We are returning a smaller subset of data now, not a case. Validating it the same way felt incorrect.

shared/src/business/useCases/getCaseExistsInteractor.ts
```diff
@@ -1,34 +1,21 @@
...
-    .getCaseByDocketNumber({
-      applicationContext,
-      docketNumber: Case.formatDocketNumber(docketNumber),
-    });
+export const getCaseExistsInteractor = async ({
+  docketNumber,
+}: {
+  docketNumber: string;
+}) => {
+  const caseRecord = await getCaseMetadataByDocketNumber({
```
Mwindo: We don't need all the case data returned by `getCaseByDocketNumber` to tell if it exists. Getting the metadata is sufficient.

web-api/terraform/modules/opensearch-sync/opensearch-sync.tf
```diff
@@ -0,0 +1,59 @@
...
+}
+
+resource "aws_sqs_queue" "opensearch_sync_queue" {
+  name                       = "opensearch_sync_queue_${var.environment}_${var.color}"
+  visibility_timeout_seconds = 901
+  redrive_policy = jsonencode({
+    deadLetterTargetArn = aws_sqs_queue.opensearch_sync_dl_queue.arn
+    maxReceiveCount     = 1
+  })
+}
```
Mwindo, line 21: Hmm. I think we assumed that this would preserve message order, but SQS only preserves message order if it is set up at creation as a FIFO Queue. See [here](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_CreateQueue.html). We probably need to do this?

shared/src/business/useCases/getCaseInteractor.ts
```diff
@@ -1,122 +1,15 @@
+import { Case } from '@shared/business/entities/cases/Case';
+import { CaseFactory } from '@shared/business/entities/cases/CaseFactory';
+import { NotFoundError, UnauthorizedError } from '@web-api/errors/errors';
+import { ServerApplicationContext } from '@web-api/applicationContext';
 import {
-  AuthUser,
```
Mwindo, line 2: Most of the logic in this file has been extracted into the CaseFactory.

shared/src/business/utilities/caseFilter.ts
```diff
@@ -30,24 +19,10 @@ const CASE_CONTACT_ATTRIBUTE_ALLOWLIST = [
   'title',
 ];
 
-export type SealedCase = Record<CaseAttributeAllowlistKeys, any>;
```
Mwindo, line 33: There is now a SealedCase entity.

shared/src/business/utilities/setServiceIndicatorsForCase.ts
```diff
@@ -1,26 +0,0 @@
-import { Case } from '../entities/cases/Case';
```
Mwindo, line 1: Renamed to `setServiceIndicatorsForPetitionersOnCase.ts`

shared/src/proxies/getPublicCaseExistsProxy.ts
```diff
@@ -1,19 +1,12 @@
...
- * @returns {Promise<*>} the promise of the api call
- */
 export const getPublicCaseExistsInteractor = (
   applicationContext,
   { docketNumber },
 ) => {
   return head({
     applicationContext,
     endpoint: `/public-api/cases/${docketNumber}`,
+    params: undefined,
```
Mwindo, line 10: Ew. Done to get types happy.

web-api/src/business/useCaseHelper/consolidatedCases/formatPublicCase.test.ts
```diff
@@ -1,122 +0,0 @@
-import { DOCKET_ENTRY_SEALED_TO_TYPES } from '@shared/business/entities/EntityConstants';
```
Mwindo, line 1: Handled by CaseFactory now.

web-api/src/business/useCaseHelper/service/createChangeItems.ts
```diff
@@ -173,9 +173,7 @@ const createWorkItemForChange = async ({
 export const generateAndServeDocketEntry = async ({
   applicationContext,
   authorizedUser,
-  barNumber,
```
Mwindo, line 176: These arguments weren't used.

web-api/src/business/useCaseHelper/trialSessions/generateNoticeOfChangeToInPersonProceeding.ts
```diff
@@ -68,7 +67,7 @@ export const generateNoticeOfChangeToInPersonProceeding = async (
       data: {
         caseCaptionExtension,
         caseTitle,
-        docketNumberWithSuffix,
+        docketNumberWithSuffix: docketNumberWithSuffix || docketNumber,
```
Mwindo, line 70: TODO: Let's make `docketNumberWithSuffix` a computed property. `docketNumberWithSuffix` should always be defined and just return `Case.getDocketNumberWithSuffix`. The `||` here is just a hack to get the types right.

Mwindo: In fact, I am just now realizing that we set the `docketNumberWithSuffix` in different ways that we should standardize. In `Case.ts`, despite having `docketNumberSuffix` in the DB, we always recompute it [here](https://github.com/ustaxcourt/ef-cms/blob/b9cbce59f8841c5ddb02f9ab133995883d422556/shared/src/business/entities/cases/Case.ts#L761). Elsewhere, we sometimes rely on that field being in the DB and then stitch it on.

Maybe `docketNumberWithSuffix` should just be a computed field that always calls `getDocketNumberSuffix`; if so, we just store `docketNumber` and `caseType` on the DB.

Mwindo: I've looked into this a little. It isn't quite as straightforward as I had hoped, but I'll work on this in a separate branch.

web-api/src/persistence/postgres/cases/getCaseByDocketNumber.ts
```diff
@@ -1,19 +1,104 @@
...
+  applicationContext,
+  docketNumber,
+  includeConsolidatedCases = true,
+  user = undefined, // Only needed to check permissions on sealed addresses for consolidated cases
+}: {
+  docketNumber: string;
+  applicationContext: ServerApplicationContext;
+  includeConsolidatedCases?: boolean;
+  user?: UnknownAuthUser;
+}): Promise<RawCase> => {
```
Mwindo, line 24: In other Postgres work, we were able to return the Entity itself (e.g., Message) to avoid using the "Raw" entities. For Case, this is much more difficult because we choose different subsets of the Case table data in different situations. (For instance, sometimes we get practitioners, sometimes we don't. Sometimes we get the consolidated cases, sometimes we don't.) In other words, while a Message is always a full Message entity, a "Case" is not always a full Case. There are a few options to handle these discrepancies:

1. We create new entities for all the different subsets of data we need. This would involve a lot of work, and it is outside the scope of this already giant ticket.
2. We always get all Case data. This is inefficient.
3. We say, "Well, darn it," and use RawCase.

We opted for 3.

shared/src/business/utilities/caseFilter.ts
Mwindo: Called `RestrictedCase`

shared/src/business/entities/Statistic.test.ts
```diff
@@ -271,15 +263,6 @@ describe('Statistic', () => {
...
     });
 
-    it('should update the penalty in the penalties array', () => {
-      expect(statistic.penalties.length).toEqual(penaltyArrayLength);
-
-      statistic.updatePenalty(MOCK_UPDATED_PENALTY);
-
-      expect(statistic.penalties.length).toEqual(penaltyArrayLength);
-      expect(statistic.penalties[0]).toEqual(MOCK_UPDATED_PENALTY);
-    });
```
Mwindo, line 281: `updatePenalty` was never used, so I got rid of it and its test.

web-api/src/business/useCases/createCaseInteractor.ts
```diff
@@ -1,26 +1,36 @@
-import { Case } from '../../../../shared/src/business/entities/cases/Case';
```
Mwindo, line 1: A few notes on this file and `createCaseFromPaperInteractor` (the two places we need to generate a docket number and create a case)

1. Ideally, we should have _one_ place we are creating a case. The logic in these two files is very similar. We did not address that.
2. The changes in these files: pulling out the minimum data needed to create the case metadata and putting a lock around it. Ideally, we wouldn't lock, but there are reasons why it is completed with the current architecture to do that. (E.g., it would be great for the database to create the docket number by autoincrementing ... but, in the current setup, we need the docket number _before_ saving to the DB.)
3. ~~Looking at this now, I am realizing that we might need to wrap the _entire_ logic in a lock (not just the metadata) to avoid a case being created without some data it needs, like petitioners. However, this involves multiple writes that might slow things down, in which case locking seems like it could become a bigger issue. That said, I really don't know how to get around this at the moment. We either need to say, "Well, the app wasn't handling concurrency properly before, and it is out of scope to fix that in 10502" or else "Well, let's rearchitect the way the backend creates cases," in which case 10502 will probably never go out.~~ Nevermind, I was thinking about transactions.
4. In an ideal world, we would do all the stuff necessary for a case with a random unique id as the pk and _then_ assign a docket number only upon saving all the necessary data. No locking (docket number is automatically generated, and only on successful writes), no worries about inconsistent state (everything is tied to a random id that can be safely discarded on failure), etc.

web-api/elasticsearch/index-cases.ts
```diff
@@ -0,0 +1,106 @@
+import { marshall } from '@aws-sdk/util-dynamodb';
```
Mwindo, line 1: This file defines how to take a change to `dwCase` and get it into OpenSearch. Here is what happens:

1. In `database.ts`'s `getDbWriter` (called by `pgInsertInto`, `pgUpdateTable`, `pgDeleteFrom`), we get the result of an update to `dwCase`. We then call `transformOpenSearchCase` (indirectly: obtained dynamically from the definition of `DatabaseSchema` in `database-types.ts`) before sending the transformed result to a queue. In this case, the transformed result is just docket numbers for cases that need updating.
2. On the other side of the queue, a lambda listening to SQS gets us to `openSearchSyncHandler`, which calls `indexOpenSearchCase` (indirectly: obtained dynamically as before), which then takes those docket numbers from above, gets the case data from the database as a `RawCase`, transforms it into a Dynamo record (since our existing `bulkIndexRecords` expects dynamo records, and we aren't in the business of refactoring that), and (finally) indexes the data. 

Note: Notice that _every time_ we write to `dwCase`, we will need to re-read from the DB the data that was just written in order to index it. This is what we were doing before, and I hate it. While it _does_ have the advantage of not requiring a FIFO queue, it puts more load on our DB for no compelling reason. We should be able to just pass in the data we need to update through the queue and update the index directly based on that. Unfortunately, this is not straightforward:
- By default, I believe OpenSearch overwrites, so we need all of the `RawCase` data when we index, not just changes to `dwCase` (which doesn't include, for instance, petitioners and practitioners) or changes to `dwPetitionerOnCase`, etc.
- Practitioners are not in Postgres.
- We then _would_ need to guarantee the order of messages in the queue.

web-api/src/persistence/postgres/cases/mapper.ts
```diff
@@ -0,0 +1,185 @@
...
+//       ...pick(rawCase, [...DatabaseSchema['dwCase'].columns, 'caseCaption']),
+//     },
+//     (value, key) =>
+//       valueShouldBeDate(key, value)
+//         ? calculateDate({ dateString: value as string })
+//         : value || null,
+//   );
+// };
+
+export function fromKyselyCase<T extends object>(record: T) {
```
Mwindo, line 101: Wow, what? Holy cow, the typing in that `DatabaseToAppCodeMapper` tranform call at the end is hideous! what the heck is _this_? Why??? Strap in:

The general problem is the following: we have a "case" entity, but we play fast and loose with case data and often lie about it. Examples: Our entity says, "You need to have `consolidatedCases`." But sometimes we are not actually fetching consolidated cases from persistence. Our entity says, "It's ok, you don't necessarily _need_ practitioners." But in some cases, something will fail because you have not set them. It's sort of a mess.

So, what to do when we retrieve data from persistence? Here are a few options:

1. Lie. We'll say we are returning a `RawCase` from persistence even when we are not. That is more or less what we do now.
2. Don't lie. Make a new DTO/entity for every endpoint we need.
3. Don't lie. Punt on figuring out how to organize our entities better. In the meantime, make a convoluted mechanism that maps DB types to app-code types. It makes whatever better solution we _do_ come up with easier because at least we can see what types we are working with.

There are benefits to 2. The reason we went with 3 is because it avoids an explosion of different DTOs and because we as a team need to think through how our entities should work. 3 is a stopgap that let's us move forward without figuring out how to organize our entities.

Mwindo: "Ok, fine. Then why didn't you do something similar for `toNewKyselyCase`?" Because we don't need to know the types--they come from RawCase. `fromKyselyCase` needs to tell the developer "here is what you are working with based on what you selected from the DB." `toNewKyselyCase` just needs to map things to the database, not tell the developer which types they are working with. (Go to the schema!) This could change in the future, but that's a future issue.

web-api/src/persistence/postgres/cases/getCaseByDocketNumber.ts
```diff
@@ -1,19 +1,110 @@
...
+    'consolidatedCases' | 'correspondence' | 'docketEntries'
+  >[] = [];
+  if (includeConsolidatedCases) {
+    consolidatedCases = await getCasesMetadataWithCounselByLeadDocketNumber({
+      applicationContext,
+      leadDocketNumber: dbCaseMetadata.leadDocketNumber!,
+    });
+    if (user) {
+      consolidatedCases = consolidatedCases.map(c =>
+        formatSealedAddresses(c, user),
```
Mwindo, line 69: Should this `formatSealedAddresses` call be elsewhere? Probably. Was it already here before to solve a high-priority bug, but it's hard to tell because of the git diff? Also yes.

web-api/terraform/modules/opensearch-sync/opensearch-sync.tf
Mwindo: Nevermind. We no longer need to worry about message order since we will re-read data from the database. See https://github.com/ustaxcourt/ef-cms/pull/5679#discussion_r2019142335.

web-api/src/persistence/postgres/utils/getColumnsForTable.ts
```diff
@@ -0,0 +1,5 @@
+import { Database, DatabaseSchema } from '@web-api/database-types';
+
+export const getColumnsForTable = (table: keyof Database): any[] => {
```
Mwindo, line 3: I am experimenting with ways to improve the typing. This function is currently only used in `pgInsertInto` and typed this way because kysely expects a certain (fairly complex) type for `updateSet`. The idea here is that `pgInsertInto` can then automatically infer columns rather than the user always specifying which columns to update. (The latter is better practice, but the former more faithfully translates our current "upsert everything" approach.)

These columns are essentially just strings, but Kysely wraps them in a custom Column type. In a quick experiment, I managed to make the typing here "better" (e.g., not "any"), but it leads to more convoluted, kysely-specific typing that still had a clunky bit I couldn't fully resolve. Rather than go down this rabbit hole further, I aborted.

shared/src/authorization/authorizationClientService.test.ts
zachrog: Thank you for making this change! Love it!

Mwindo (approved):

zachrog (approved):

jimlerza (approved):

PR #5683, opened 2025-01-08
DEP updates 2025 01 06(Lint Overhaul)

zachrog:
## In This PR
- Updated ESLint to v9 which uses new "flat config"
- Removed most lint packages in favor of well supported ones
- Lint fixes everywhere
- Update major version of Opensearch v2.x.x -> v3.0.0
- Remove unused custom opensearch analyzers and filters
- Fix RDS terraform destroy/recreate of reader instance
- Update Terraform providers to be consistent and use latest patch version.

### Important Files
- package.json
- eslint.config.mjs
- web-api/elasticsearch/elasticsearch-settings.ts
- web-api/terraform/modules/rds/rds.tf
- web-api/src/middleware/apiGatewayHelper.ts
- web-api/src/app.ts

Successful Migration
https://app.circleci.com/pipelines/github/flexion/ef-cms/59507/workflows/66f90c23-2982-4b02-a85c-42cfbe846c60


eslint.config.mjs
```diff
@@ -0,0 +1,196 @@
...
+    ...eslint.configs.recommended,
+    ignores: [
+      '**/node_modules/',
+      '**/dist/',
+      '**/dist-lambdas/',
+      '**/*.js',
+      '**/build/',
+      '*.config.js',
+      '**/*_.js',
+      'scripts/run-once-scripts/**/*',
```
jimlerza:
```suggestion
```

eslint.config.mjs
```diff
@@ -0,0 +1,196 @@
...
+    },
+    ignores: [
+      '**/node_modules/',
+      '**/dist/',
+      '**/dist-lambdas/',
+      '**/*.js',
+      '**/build/',
+      '*.config.js',
+      '**/*_.js',
+      'scripts/run-once-scripts/**/*',
```
jimlerza:
```suggestion
```

eslint.config.mjs
```diff
@@ -0,0 +1,196 @@
...
```
jimlerza:
```suggestion
```

zachrog: Marking resolved as there will be a new archive directory

scripts/npm/upgrade-npm-packages.ts
TomElliottFlexion, line 1: Well this looks very cool

scripts/upload-practitioner-application-packages.ts
```diff
@@ -243,7 +243,7 @@ const asyncTifToPdf = ({
         if (result.code === 0) {
           resolve(fileName);
         } else {
-          reject();
+          reject(new Error('Unable convert tif to pdf'));
```
TomElliottFlexion, line 246: Love the details

shared/src/business/entities/DocketEntry.validate.test.ts
```diff
@@ -464,6 +480,7 @@ describe('validate', () => {
 
       expect(docketEntry.isValid()).toBeFalsy();
       if (item.expectValidationErrors) {
+        // eslint-disable-next-line jest/no-conditional-expect
```
TomElliottFlexion, line 465: Would it make sense to disable this in the config? Don't know why this rule is a default

zachrog: I did have to add a couple of these lint exceptions, but I generally agree with this rule from jest. Unfortunately I see us break it a lot.

For instance I saw some people do this in a test

try {
function()
} catch(e) {
expect(e).toBe(blah)
}

This is bad because if the function does not explode and works the catch block never runs and the test passes. So I think it helps point things like this out.

TomElliottFlexion (approved):

scripts/npm/upgrade-npm-packages.ts
zachrog: Thanks Tom!

shared/src/business/entities/DocketEntry.unsealEntry.test.ts
```diff
@@ -1,8 +1,22 @@
-import { A_VALID_DOCKET_ENTRY } from './DocketEntry.test';
-import { DOCKET_ENTRY_SEALED_TO_TYPES } from './EntityConstants';
+import { DOCKET_ENTRY_SEALED_TO_TYPES, ROLES } from './EntityConstants';
 import { DocketEntry } from './DocketEntry';
 
 describe('unsealEntry', () => {
+  const mockPrimaryId = '7111b30b-ad38-42c8-9db0-d938cb2cb16b';
+
+  const A_VALID_DOCKET_ENTRY = {
```
Mwindo: Is the idea to re-create the test docket entry object as needed in each test rather than to use a centralized test object? I see benefits to both approaches, although I am curious about the decision here given that the object is unchanged in every test. If I just want any old valid docket entry for testing, it is helpful to have one defined rather than re-creating it every time and keeping it up-to-date with any schema changes (even if rare).

Mwindo (approved):
Awesome! Thank you @zachrog! This is great. My only comment is regarding `A_VALID_DOCKET_ENTRY` being redefined in all those tests. Without more context that's not my cup of tea, but it's no big deal, and I can see why one might prefer it.

scripts/run-once-scripts/postgres-migration/delete-messages.ts
```diff
@@ -30,7 +30,7 @@ const { tableNameInput } = parseArgsAndEnvVars(scriptConfig) as {
 const dynamoDbClient = new DynamoDBClient({ region: 'us-east-1' });
 const dynamoDbDocClient = DynamoDBDocumentClient.from(dynamoDbClient);
 
-let totalItemsDeleted = 0;
+const totalItemsDeleted = 0;
```
jimlerza: This change gave me pause. I legitimately thought the linter wasn't smart enough to identify actual, legitimate uses of `let`. Turns out, this variable is never incremented.

Not a huge fan of this particular change - we should've either fixed this to make it do what it purports or removed it altogether - but this script is on the chopping block anyway.

shared/src/business/entities/DocketEntry.test.ts
jimlerza, line 1: This change resulted in a lot of copypasta in the other `DocketEntry.*.test.ts` files. Is there a technical or lint-related reason for this change?

shared/src/business/entities/Statistic.test.ts
```diff
@@ -81,31 +81,6 @@ describe('Statistic', () => {
...
-      expect(statistic.getFormattedValidationErrors()!).toEqual({
-        penalties: [
-          {
-            index: 0,
-            penaltyAmount: 'Enter penalty amount.',
-          },
-        ],
-      });
-    });
-
```
jimlerza, line 108: Why was this removed?

shared/src/business/test/getFakeFile.ts
```diff
@@ -2,11 +2,9 @@ import fs from 'fs';
 import path from 'path';
 const testAssetsPath = path.join(__dirname, '../../../test-assets/');
 
-// eslint-disable-next-line spellcheck/spell-checker
 // contains "Congrations, you found the Easter Egg."
```
jimlerza, line 5: Congrations? :grimacing:

web-api/src/business/useCases/trialSessions/generateTrialCalendarPdfInteractor.ts
```diff
@@ -40,8 +40,9 @@ export const generateTrialCalendarPdfInteractor = async (
 
   let startTimeFormatted;
   if (trialSession.startTime) {
+    // eslint-disable-next-line prefer-const
     let [hour, min]: any = trialSession.startTime.split(':');
```
jimlerza, line 44: Just so I understand, does this mean that the linter can't tell when variable(s) assigned by destructuring are later modified? How confident are we that we haven't accidentally converted a `let` to `const` somewhere we shouldn't have?

shared/src/business/entities/DocketEntry.unsealEntry.test.ts
zachrog: TBH I was in caveman brain in trying to make lint errors go away everywhere, so don't read too much into it. The big problem I was trying to solve was a test file was exporting a constant. NEVER DO THIS. When a test file exports a constant and you import it in some other test file then you will automatically run all test files in the original. So we had like 10 files running duplicate tests because of this.

I think my preference is to use centrally located, well typed test objects so that if a type changes we don't need to update 100 test files. However, I did not do that here as I had 300 files to update for lint errors. Oops.

web-api/src/lambdas/v2/getCaseLambda.test.ts
jimlerza, line 1: I appreciate that this change gets us away from mocking app context and I think it is a good change, however I think larger diffs like this (and others) should have been made in a different PR.

web-api/src/lambdas/v2/getDocumentDownloadUrlLambda.test.ts
jimlerza, line 1: This is test has changed substantially. I would like to see this in a separate PR.

web-api/terraform/modules/kms/providers.tf
```diff
@@ -0,0 +1,9 @@
+terraform {
+  required_providers {
+    aws = {
+      source                = "hashicorp/aws"
+      version               = ">= 5.78.0"
+      configuration_aliases = [aws.us-west-1]
```
jimlerza, line 6: I notice this `configuration_aliases` line is defined in a few providers, but not all of them. I read through the [documentation](https://developer.hashicorp.com/terraform/language/providers/configuration#alias-multiple-provider-configurations) to better understand, but now I'm unsure why this is defined here in the module and not in a region-specific declaration.

web-client/src/presenter/actions/AdvancedSearch/setDocketNumberFromAdvancedSearchAction.test.ts
```diff
@@ -40,22 +40,6 @@ describe('setDocketNumberFromAdvancedSearchAction', () => {
...
-        state: {
-          advancedSearchForm: {
-            caseSearchByDocketNumber: { docketNumber: '101-18S' },
-          },
-        },
-      },
-    );
-    expect(output.docketNumber).toEqual('101-18');
-  });
   it('matches a docket number with R suffix', async () => {
```
jimlerza, line 43: Why was this test case removed?

web-client/src/presenter/actions/CaseConsolidation/setDocketNumberFromModalSearchAction.test.ts
```diff
@@ -34,19 +34,6 @@ describe('setDocketNumberFromModalSearchAction', () => {
...
-        presenter,
-      },
-      state: {
-        modal: {
-          searchTerm: '101-18S',
-        },
-      },
-    });
-    expect(output.docketNumber).toEqual('101-18');
-  });
```
jimlerza, line 49: Again, why remove this? Can we no longer use suffixes in search because of a lint change?

web-client/src/presenter/actions/Login/updateAuthenticationFormValueAction.test.ts
```diff
@@ -7,7 +7,7 @@ describe('updateAuthenticationFormValueAction', () => {
   const email = 'isbell@example.com';
   const code = 'TEST_CODE';
 
-  it('should set state.authentication.form.email to the passed in email prop', async () => {
+  it('should set state.authentication.form.password to the passed in email prop', async () => {
```
jimlerza: It should do what now? Sets password to the email prop?

web-client/src/presenter/actions/TrialSessionWorkingCopy/generatePrintableTrialSessionCopyReportAction.test.ts
```diff
@@ -155,48 +155,6 @@ describe('generatePrintableTrialSessionCopyReportAction', () => {
...
-            },
-          },
-          trialSessionWorkingCopyHelper: {
-            trialStatusFilters,
-          },
-        },
-      },
-    );
-    expect(result.output.pdfUrl).toEqual(url);
-  });
```
jimlerza, line 198: Why was this removed?

web-client/src/presenter/actions/setDocketNumberFromSearchAction.test.ts
```diff
@@ -34,19 +34,6 @@ describe('setDocketNumberFromSearchAction', () => {
...
-        presenter,
-      },
-      state: {
-        header: {
-          searchTerm: '101-18S',
-        },
-      },
-    });
-    expect(output.docketNumber).toEqual('101-18');
-  });
```
jimlerza, line 49: Same as above

web-client/src/presenter/actions/shouldRefreshCaseAction.test.ts
```diff
@@ -39,17 +39,4 @@ describe('shouldRefreshCaseAction', () => {
...
-      modules: {
-        presenter,
-      },
-      props: {
-        docketNumber: '101-19',
-      },
-      state: {},
-    });
-    expect(pathYesStub).toHaveBeenCalled();
-  });
```
jimlerza, line 54: Why was this removed?

web-client/src/presenter/computeds/TrialSession/addTrialSessionInformationHelper.test.ts
```diff
@@ -84,30 +84,6 @@ describe('addTrialSessionInformationHelper', () => {
...
-        state: {
-          form: { sessionScope: TRIAL_SESSION_SCOPE_TYPES.locationBased },
-          user: { role: 'docketclerk' },
-        },
-      });
-
-      expect(result.displayRemoteProceedingForm).toEqual(false);
-    });
-  });
-
```
jimlerza, line 110: Why was this removed?

jimlerza (changes requested):
Overall I am glad we were able to upgrade eslint and appreciate all of the work that went into making our code compliant with the latest lint rules. On the other hand, I noted some changes that don't appear to have anything to do with lint rules or dependencies.

In particular, I'd like to understand why several test cases have been outright removed. I know we needed to get the entire suite passing in this PR, but I need to understand why the tests were removed before I can approve this.

shared/src/business/test/getFakeFile.ts
TomElliottFlexion: Congrations.

scripts/run-once-scripts/postgres-migration/delete-messages.ts
zachrog: I updated it to count correctly.. I ran a lint fix on the whole project and 90% of the const problems it auto fixed, I believe this is one.

shared/src/business/entities/DocketEntry.test.ts
zachrog: Copying response from Chris B.:

TBH I was in caveman brain in trying to make lint errors go away everywhere, so don't read too much into it. The big problem I was trying to solve was a test file was exporting a constant. NEVER DO THIS. When a test file exports a constant and you import it in some other test file then you will automatically run all test files in the original. So we had like 10 files running duplicate tests because of this.

I think my preference is to use centrally located, well typed test objects so that if a type changes we don't need to update 100 test files. However, I did not do that here as I had 300 files to update for lint errors. Oops.

I will update to use an exported constant.

shared/src/business/test/getFakeFile.ts
zachrog: Congrations.

shared/src/business/entities/Statistic.test.ts
zachrog: So you probably have questions about Zach why did you go delete so many random tests? Fair.

When I added the new jest lint rules one of the rules was no duplicate it() test names and no duplicate describe() test names. That makes sense, two tests should not have the same name. Unfortunately we did this in dozens of places. So I went through each duplicate and saw if I could give it a better name or just delete it entirely if it was a duplicate test. In many cases one of the devs copy/pasted a test and did not change anything, so I removed those tests. If you look at Statistic.test.ts test file there is an EXACT copy of this test beneath the deleted one.

shared/src/business/test/getFakeFile.ts
zachrog: Haha I think this misspelling was on purpose for fun. It made me chuckle when I first found this PDF.

web-api/src/business/useCases/trialSessions/generateTrialCalendarPdfInteractor.ts
zachrog: The linter only auto fixes let -> const in scenarios it is 100% confident about, Otherwise it just complains.

In this scenario we destructure hour, min using let but only reassign hour. So it was complaining that min COULD be a const while hour could be a let. This felt needlessly nitpicky so I ignored the rule here.

web-api/src/lambdas/v2/getCaseLambda.test.ts
zachrog: I will pull out application Context into its own PR. It was not my intention to try and make this one so large. I started with appContext but this dep updates ballooned because of all the eslint changes.

web-api/src/lambdas/v2/getDocumentDownloadUrlLambda.test.ts
zachrog: yes

web-api/terraform/modules/kms/providers.tf
zachrog: As I understand it you can pass in providers to modules in two ways, explicitly and implicitly. In most scenarios we just do it implicitly (Whatever the parent uses as primary provider the child module will use.). Sometimes we have modules which deploy to both east and west(kms.tf), in those instances we need the module to indicate to the outside that you have to pass it aws(the default) and (aws.us-west-1) for places that want to deploy stuff to west. 

For my understanding and reading of the docs this is the correct way to declare a module has a default implicit aws provider and a second required aws provider in another aws region.

web-client/src/presenter/actions/CaseConsolidation/setDocketNumberFromModalSearchAction.test.ts
zachrog: As stated before and for all other modified test it was because it was a duplicate.

web-client/src/presenter/actions/Login/updateAuthenticationFormValueAction.test.ts
zachrog: Fixed. Only modified this test because it had a duplicate test name

web-client/src/presenter/actions/AdvancedSearch/setDocketNumberFromAdvancedSearchAction.test.ts
zachrog: As stated before and for all other modified test it was because it was a duplicate.

web-client/src/presenter/actions/TrialSessionWorkingCopy/generatePrintableTrialSessionCopyReportAction.test.ts
zachrog: As stated before and for all other modified test it was because it was a duplicate.

web-client/src/presenter/actions/setDocketNumberFromSearchAction.test.ts
zachrog: As stated before and for all other modified test it was because it was a duplicate.

web-client/src/presenter/actions/shouldRefreshCaseAction.test.ts
zachrog: As stated before and for all other modified test it was because it was a duplicate.

web-client/src/presenter/computeds/TrialSession/addTrialSessionInformationHelper.test.ts
zachrog: As stated before and for all other modified test it was because it was a duplicate.

jimlerza (approved):

web-api/terraform/modules/kms/providers.tf
jimlerza: Thanks for the clarification! That makes sense.

PR #5685, opened 2025-01-10
Devex: clear judgeUser state on logout

Mwindo:
Currently, `judgeUser` is not cleared upon logging out. This can cause issues if a user (e.g., a test user) logs in as a judge, logs out, then logs in as a non-judge.


TomElliottFlexion (approved):
Might need a test before going to staging, nbd

Mwindo:
> Might need a test before going to staging, nbd

Yep, it will be! This was caught as part of 10491 and will be checked accordingly. We might tack it onto the 10491 staging PR, but since it is only tangentially related, it could also go out separately. In either case, I will document it.

jimlerza (approved):

PR #5687, opened 2025-01-10
OpEx: more reports for 2025 Congressional Budget Justification

jimlerza:
Added 3 more CBJ reports:

* All petitions filed by non-attorney practitioners (not pro-se petitioners) in a given (fiscal or calendar) year.
* All attorney practitioners admitted in a given (fiscal or calendar) year.
* Statistics for Limited Entry of Appearance documents and their companion Notice of Completion documents.


jimlerza:
Rather than hard-code the `P` event code in the `documents-filed-by-non-attorneys` script, I made an `eventCode` parameter that can be passed in so we can re-use the report for other document types.

Mwindo (approved):
I didn't dive too deep into this, but it looks reasonable. Thanks @jimlerza!

PR #5693, opened 2025-01-13
10491-dxox: Fix processWorkItemEntries

Mwindo:
Currently, in staging/prod environments, we store "a" WorkItem in different ways:
- The "main" WorkItem record with pk `case|<docketNumber>`. This is conditionally updated as part of `updateCaseAndAssociations`. (For instance, if a case's `associatedJudge` is switched, these WorkItem records are updated.)
- User-outbox "copies" with pk `user-outbox...`.
- Section-outbox "copies" with pk `section-outbox...`.

The "main" records are always up-to-date: anytime we make "copies," we also save the "main" record:
- In `saveWorkItemForDocketClerkFilingExternalDocument`, we save the original directly via a call to `put`
- In `putWorkItemInUsersOutbox` and `putWorkItemInOutbox`, we save the original in all spots via `saveWorkItem`

In Postgres (test environment), we have gotten rid of the copies since we can query for relevant "main" WorkItems more easily. However, thanks to Tenille, we noticed that `processWorkItemEntries` does not distinguish between the "main" WorkItem DynamoDB records and the "outbox copy" WorkItem DynamoDB records. This means that the WorkItem that we write to Postgres could be stale: we might write the (up-to-date) "main" record first and then overwrite it with a stale "copy" afterwards. This PR rectifies the issue by filtering out these "copies."

### DIFF
--- a/web-api/src/business/useCases/processStreamRecords/processWorkItemEntries.test.ts
+++ b/web-api/src/business/useCases/processStreamRecords/processWorkItemEntries.test.ts
@@ -34,6 +34,44 @@ describe('processWorkItemEntries', () => {
     },
   };
 
+  const mockSectionOutboxWorkItemRecord = {
+    dynamodb: {
+      NewImage: {
+        docketNumber: {
+          S: '123-45',
+        },
+        entityName: {
+          S: 'WorkItem',
+        },
+        pk: {
+          S: 'section-outbox|docket|2021-04-14',
+        },
+        sk: {
+          S: '2021-04-14T13:20:24.0732',
+        },
+      },
+    },
+  };
+
+  const mockUserOutboxWorkItemRecord = {
+    dynamodb: {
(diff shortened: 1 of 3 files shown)


web-api/src/business/useCases/processStreamRecords/processWorkItemEntries.test.ts
```diff
@@ -34,6 +34,44 @@ describe('processWorkItemEntries', () => {
...
+          S: '123-45',
+        },
+        entityName: {
+          S: 'WorkItem',
+        },
+        pk: {
+          S: 'section-outbox|docket|2021-04-14',
+        },
+        sk: {
+          S: '2021-04-14T13:20:24.0732',
```
Mwindo, line 50: These pks/sks (and those below) are based on real data.

pixiwyn (approved):

PR #5697, opened 2025-01-14
10504: Fix bug in updating case deadlines

Mwindo:
When updating case deadlines, we are writing to the DB, then immediately trying to read the update from the DB. If the user is reading from the read replica (west coast), the update might not get to the main db (east coast) in time. To fix this, we update the `caseDeadlines` state directly in `updateCaseDeadlineAction` based on the API response to the "writing to the DB" request.

### DIFF
--- a/web-client/src/presenter/actions/CaseDeadline/updateCaseDeadlineAction.ts
+++ b/web-client/src/presenter/actions/CaseDeadline/updateCaseDeadlineAction.ts
@@ -4,6 +4,7 @@ export const updateCaseDeadlineAction = async ({
   applicationContext,
   get,
   path,
+  store,
 }: ActionProps) => {
   const { associatedJudge, associatedJudgeId, docketNumber, leadDocketNumber } =
     get(state.caseDetail);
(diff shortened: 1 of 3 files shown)


web-client/src/presenter/sequences/updateCaseDeadlineSequence.ts
```diff
@@ -26,7 +25,6 @@ export const updateCaseDeadlineSequence = [
       clearScreenMetadataAction,
       clearModalAction,
       clearModalStateAction,
-      getCaseDeadlinesForCaseAction,
```
Mwindo, line 29: This is where we were reading directly after writing.

web-client/src/presenter/actions/CaseDeadline/updateCaseDeadlineAction.ts
```diff
@@ -22,6 +23,16 @@ export const updateCaseDeadlineAction = async ({
...
 
+  const existingCaseDeadlines = get(state.caseDeadlines);
+
+  const updatedCaseDeadlines = existingCaseDeadlines.map(deadline =>
+    deadline.caseDeadlineId === updateCaseDeadlineResult.caseDeadlineId
+      ? updateCaseDeadlineResult
+      : deadline,
+  );
+
+  store.set(state.caseDeadlines, updatedCaseDeadlines);
```
Mwindo, line 34: This assumes that `caseDeadlines` is in state. Since the only place we use this action is on the Tracked Items tab of the Case Detail page, and since we have loaded the `caseDeadlines` when we come to this page, this assumption should be ok.

pixiwyn (approved):

PR #5700, opened 2025-01-15
Devex modify postgres restore to test

Mwindo:
- Remove `--clean` flag in `pg_restore` since we drop all target tables beforehand
- Slight refactor of `scripts/postgres/restoreDbFromSource.ts`
- Run `scripts/postgres/create-rds-users.sh` in config.yml to avoid having to manually reset user table permissions

The changes to `restoreDbFromSource` were tested on ustc-test to ustc-dev.


jimlerza (changes requested):
Nothing wrong with the code in this diff, but the `sed` call inside the `create-postgres-users.sh` is using a syntax that will fail in the linux container that runs in circleci.

The correct syntax for achieving the same behavior in both OSX and linux is:

```
sed -i '' "s/ENVREPLACEME/${ENV}/g;" create-users.sql > "${GENERATED_SQL_FILE}"
```

Mwindo:
I could be missing something, but I think the command in `scripts/postgres/create-rds-users.sh` is ok. In fact, the `-i` flag (combined with `>`) will behave differently. As far as I know, sed with `g` and `>` is supported by the major versions of sed.

jimlerza (approved):

.circleci/config.yml
```diff
@@ -869,9 +874,14 @@ jobs:
...
       - run:
-          name: Restore Postgres
+          name: Restore Postgres Data
           command: |
             TARGET_ENV=$LOWER_ENV TARGET_ACCOUNT_ID="${LOWER_ENV_ACCOUNT_ID}" ./scripts/postgres/restoreDbFromSource.ts
+      - run:
+          name: Restore Postgres User Permissions
+          command: |
+            cd ./scripts/postgres
+            ./scripts/postgres/create-postgres-users.sh
```
jimlerza:
```suggestion
            ./create-postgres-users.sh
```

PR #5701, opened 2025-01-15
Devex modify postgres restore to staging

Mwindo:
- Remove `--clean` flag in `pg_restore` since we drop all target tables beforehand
- Slight refactor of `scripts/postgres/restoreDbFromSource.ts`
- Run `scripts/postgres/create-rds-users.sh` in config.yml to avoid having to manually reset user table permissions


.circleci/config.yml
```diff
@@ -869,9 +874,14 @@ jobs:
...
       - run:
-          name: Restore Postgres
+          name: Restore Postgres Data
           command: |
             TARGET_ENV=$LOWER_ENV TARGET_ACCOUNT_ID="${LOWER_ENV_ACCOUNT_ID}" ./scripts/postgres/restoreDbFromSource.ts
+      - run:
+          name: Restore Postgres User Permissions
+          command: |
+            cd ./scripts/postgres
+            ./scripts/postgres/create-postgres-users.sh
```
jimlerza:
```suggestion
            ./create-postgres-users.sh
```

jimlerza (approved):

.circleci/config.yml
Mwindo: This suggestion got lost among my other Github notifications--thank you!!

PR #5705, opened 2025-01-16
Dep updates 2025 01 13

TomElliottFlexion:
- **!!Requires account specific deploy!!** (on prod account)
- `dynamodb:DeleteTableReplica` permission added to CircleCI (temporarily?) to allow for reconfiguration of Dynamo table replica terraform
- Canvas forced to 3.0.1
- @aws-sdk/client-s3 (and peer dependency) are locked to 3.726.0. Upgrading client-s3 results in s3 files not being downloaded properly in AWS Batch jobs. Still not sure why- will keep investigating
- [Migration](https://app.circleci.com/pipelines/github/flexion/ef-cms/59765/workflows/42eb7951-7706-4c71-bb5a-8cde36f64264)


docs/dependency-updates.md
```diff
@@ -73,10 +76,6 @@ regex search the entire project for `version = ">= \d+.\d+.\d+"` and make sure i
 
 - fortawesome packages are locked down to pre-6.x.x to maintain consistency of icon styling until there is usability feedback and research that determines we should change them. This includes `@fortawesome/free-solid-svg-icons`, `@fortawesome/free-regular-svg-icons`, and `@fortawesome/fontawesome-svg-core`.
 
-# canvas
```
zachrog, line 76: Yaahhhhhhh

scripts/npm/upgrade-npm-packages.ts
```diff
@@ -5,14 +5,12 @@ import { execSync, spawnSync } from 'child_process';
 // Caveats list: packages to exclude from upgrade
 const caveats: string[] = [
   'react',
-  '@types/react',
```
zachrog: If we are not Upgrading react then we should not be upgrading types I think. 

If our React is on v18 then the types should match it.

web-api/terraform/modules/dynamo-table/dynamo.tf
```diff
@@ -1,80 +1,62 @@
...
 
-  replica {
-    region_name = "us-west-1"
+  timeouts {
+    create = "2h"
+    update = "2h"
   }
+}
+
+resource "aws_dynamodb_table_replica" "efcms_table_west" {
```
zachrog, line 56: WOOOOOOO. Nice find.

web-api/terraform/modules/dynamo-table/dynamo.tf
```diff
@@ -1,80 +1,62 @@
...
+  global_table_arn = aws_dynamodb_table.efcms-table-east.arn
 
+  provider = aws.us-west-1
```
zachrog, line 59: Because the module "dynamo-table" has multiple providers(aws and aws.us-west-1) it should have a providers.tf file to make the providers explicit. An example of this is the rds module: 

web-api/terraform/modules/rds/providers.tf

Terraform gives warnings in the console if you do not have this. I ran a terraform plan on this branch and got this warning.
image

zachrog (approved):
Approving but wanted to ask that you add the providers.tf file. Nice job Tom

scripts/npm/upgrade-npm-packages.ts
TomElliottFlexion: Fair and reasonable, will revert

web-api/terraform/modules/dynamo-table/dynamo.tf
TomElliottFlexion: Hmm, I thought since they were being passed in explicitly at the parent level and then defined in dynamo.tf, that it wasn't needed. Looks like I was wrong! Or at least that it's better to provide a mapping

TomElliottFlexion: Warning be gone, good catch

web-api/terraform/modules/dynamo-table/dynamo.tf
```diff
@@ -1,80 +1,62 @@
...
     attribute_name = "ttl"
     enabled        = true
   }
 
-  replica {
-    region_name = "us-west-1"
+  timeouts {
+    create = "2h"
+    update = "2h"
   }
```
jimlerza, line 53: Is this bandaging a problem in lieu of solving it?

Are we satisfied with assuming that creating dynamo databases "just takes longer than it used to" without figuring out why?

TomElliottFlexion: I don't believe it's taking longer than it used to; there is an issue reading the state of the table after it's (quickly) created. This PR doesn't change the timeouts, only changes the way providers are passed to dynamo.tf and explicitly defines the replica (with the same timeout)

zachrog (approved):

jimlerza (approved):

PR #5714, opened 2025-01-20
10522: Notice of Change of Trial Location

cruzjone-flexion:
[10522 - Task](https://github.com/flexion/ef-cms/issues/10522)


web-client/src/views/TrialSessions/EditTrialSession.tsx
```diff
@@ -88,7 +98,7 @@ export const EditTrialSession = connect(
               data-testid="submit-edit-trial-session"
               type="submit"
               onClick={() => {
-                updateTrialSessionSequence();
+                displayTrialSessionLocationChangeModalSequence();
```
Mwindo: Can we rename this since the user isn't necessarily editing a trial session location? (Maybe something like `handleEditedTrialSessionSequence`?)

web-client/src/presenter/state.ts
```diff
@@ -640,6 +641,9 @@ export const baseState = {
   createOrderSelectedCases: [] as any[],
   currentJudges: [],
   currentPage: 'Loading',
+  currentTrialSessionLocation: undefined as
+    | TrialSessionLocationInfo
+    | undefined,
```
Mwindo: Would you be open to wrapping this and `updatedTrialSessionLocation` in an object to avoid having two miscellaneous but closely coupled pieces of state in the root of the state tree? Also, do we ever unset this state, and is it worth doing? (Unlikely to be a big deal. Mostly just worried about continuing the pattern of not clearing state since that has caused problems in other cases.)

web-client/src/presenter/actions/hasTrialSessionLocationChangedAction.test.ts
```diff
@@ -0,0 +1,253 @@
...
+          proceedingType: TRIAL_SESSION_PROCEEDING_TYPES.inPerson,
+        },
+      },
+    });
+
+    expect(unchangedMock.mock.calls.length).toEqual(1);
+    expect(updatedMock.mock.calls.length).toEqual(0);
+  });
+
+  it('should call "unchanged" path when location information has been updated but there are no cases to generate notices', async () => {
```
Mwindo: I'm a little nervous about the naming of `hasTrialSessionLocationChangedAction` if it is checking more than whether or not the trial session location has changed. I wonder if something like `needsToGenerateNoticeOfChangedTrialLocationAction` would be better.

shared/src/business/utilities/pdfGenerator/documentTemplates/NoticeOfChangeOfTrialLocation.tsx
```diff
@@ -0,0 +1,108 @@
...
+              <th>Judge</th>
+            </thead>
+            <tr>
+              <td>{updatedTrialSession.judge?.name}</td>
+            </tr>
+          </table>
+        </div>
+      </div>
+      <div>
+        &emsp;The parties are hereby notified that the Court’s Notice of Trial
```
Mwindo: Do we want `&apos;` here too?

shared/src/business/utilities/trialSession/shouldGenerateNoticeOfChangeTrialLocation.test.ts
```diff
@@ -0,0 +1,114 @@
...
+    courthouseName: 'TEST_COURT_HOUSE_NAME',
+    isCalendared: true,
+    postalCode: 'TEST_POSTAL_CODE',
+    proceedingType: TRIAL_SESSION_PROCEEDING_TYPES.inPerson,
+    state: 'TEST_STATE',
+    trialLocation: 'TEST_TRIAL_LOCATION',
+    sessionType: SESSION_TYPES.regular,
+  };
+
+  it('return "false" when location has not been updated', () => {
```
Mwindo: `should ...`, here and below

web-api/src/business/useCases/trialSessions/getTrialSessionOpenCasesCountInteractor.ts
```diff
@@ -0,0 +1,42 @@
...
+import {
+  ROLE_PERMISSIONS,
+  isAuthorized,
+} from '@shared/authorization/authorizationClientService';
+import { applicationContext } from '@web-api/applicationContext';
+import { UnknownAuthUser } from '@shared/business/entities/authUser/AuthUser';
+import { getCasesInTrialSession } from '@web-api/business/useCases/trialSessions/updateTrialSessionInteractorHelper';
+import { getTrialSessionById } from '@web-api/persistence/dynamo/trialSessions/getTrialSessionById';
+
+export const getTrialSessionOpenCasesCountInteractor = async (
```
Mwindo: It feels like the name of this interactor isn't quite in line with what it does.

web-api/src/business/useCases/trialSessions/updateTrialSessionInteractorHelper.ts
```diff
@@ -0,0 +1,287 @@
...
+import { TrialSessionWorkingCopy } from '@shared/business/entities/trialSessions/TrialSessionWorkingCopy';
+import { get } from 'lodash';
+
+type GetCasesInTrialSessionParams = {
+  applicationContext: ServerApplicationContext;
+  trialSession: RawTrialSession;
+  authorizedUser: AuthUser;
+};
+
+export async function getCasesInTrialSession({
```
Mwindo, line 19: A note you can ignore: Part of me wonders if it would be better to get the trial session from the db by trial session id here to avoid needing a one-off interactor to get the _counts_ of cases that need notices. (E.g., we could expose this as an interactor and check the count in the frontend.) But I can see why it was done this way if we already have the trial session info, among other reasons.

Mwindo (approved):
Thanks @cruzjone-flexion! LGTM overall! I left a few smallish comments, mostly about naming.

jimlerza (approved):

pavanyenduri (approved):

PR #5715, opened 2025-01-21
10522: Refactor + Improve typing;

cruzjone-flexion:
1. Refactor
2. Improve Typings
3. Update Import Statements to use aliases


web-api/src/business/useCases/trialSessions/updateTrialSessionInteractorHelper.ts
```diff
@@ -0,0 +1,237 @@
+import { AuthUser } from '@shared/business/entities/authUser/AuthUser';
```
Mwindo, line 1: I wonder if any of the associated tests in `updateTrialSessionInteractor.test.ts` should also be moved to `updateTrialSessionInteractorHelper.test.ts` for better modularity/fit between test and file.

Mwindo (approved):

web-api/src/business/useCases/trialSessions/updateTrialSessionInteractorHelper.ts
cruzjone-flexion: Good Point; That will lower the complexity of the unit test;
Will start working on that;

PR #5717, opened 2025-01-21
10566: Idle Inactivity Monitor

zachrog:
# In This PR
- Remove old idle inactivity monitor and replace with IdleActivityMonitor.tsx.
- Now all idle related functionality lies with IdleActivityMonitor.tsx
- Replace some cerebral integration tests with cypress tests


pavanyenduri (approved):

web-client/src/presenter/actions/IdleLogout/navigateToIdleLogoutAction.test.ts
```diff
@@ -0,0 +1,21 @@
...
+
+const mockRoute = jest.fn();
+describe('navigateToIdleLogoutAction', () => {
+  beforeAll(() => {
+    presenter.providers.router = {
+      route: mockRoute,
+    };
+  });
+
+  it('calls history.back()', async () => {
```
Mwindo, line 13: Copy-paste artifact?

cypress/local-only/tests/integration/logoutBehavior/idle-logout.cy.ts
```diff
@@ -49,10 +55,11 @@ describe('Idle Logout Behavior', () => {
     cy.puppeteer('closeTab', newTabUrl);
   });
 
-  it('should sign out of all tabs after idle', () => {
+  it.only('should sign out of all tabs after idle', () => {
```
Mwindo: Looks like an `only` slipped in.

Mwindo (approved):
Overall, so much better. Thank you @zachrog, @pavanyenduri! I left a couple of comments to address.

cypress.config.ts
```diff
@@ -104,50 +104,49 @@ export default defineConfig({
...
+            browser: any,
+            close: boolean = true,
+          ) {
+            // Note that browser.pages is *not* sorted in any particular order.
+            // Therefore we pass in the URL we want to find rather than an index.
+            const pages: Page[] = await browser.pages();
+            for (const page of pages) {
+              if (page.url().includes('.cy.ts')) {
+                continue;
+              }
```
Mwindo: Anecdotal: I have noticed this test being flakier. I think the test is flakier because it no longer uses retry. See import [here](https://github.com/ustaxcourt/ef-cms/blob/79010436763d7b933b104d91e15f6f49a6aa4d45/cypress.config.ts#L24) and use [here](https://github.com/ustaxcourt/ef-cms/blob/79010436763d7b933b104d91e15f6f49a6aa4d45/cypress.config.ts#L96)

zachrog: Chris thank you so much! We have been having flaky problems with this test. We will implement this.

zachrog: Done!

jimlerza (approved):

PR #5721, opened 2025-01-22
10585: Fix json

Mwindo:
Tests as fast-follow.


btejha (approved):

TomElliottFlexion (approved):

jimlerza (approved):

jimlerza (approved):
Nice use of the `UnknownAuthUser` type, which still allows this to be used without an authorized user, and for defaulting the value so other uses of `getCaseByDocketNumber` are unaffected.

There is one other use (excluding scripts) - `getCaseByDocketNumber` is called by `getCalendaredCasesForTrialSession`, which I assume is only ever used internally?

jimlerza:
> Nice use of the `UnknownAuthUser` type, which still allows this to be used without an authorized user, and for defaulting the value so other uses of `getCaseByDocketNumber` are unaffected.
> 
> There is one other use (excluding scripts) - `getCaseByDocketNumber` is called by `getCalendaredCasesForTrialSession`, which I assume is only ever used internally?

Oop, nevermind. I see that the addresses aren't returned in the `trial-sessions/<id>` response. Nice!

PR #5729, opened 2025-01-28
no-new-dates

TomElliottFlexion:
Adds custom local eslint plugin that disallows the use of `new Date()`. There were no eslint v9 compatible plugins out there, so I wrote one. This may not be important enough to start creating custom rules; please feel free to object here


eslint-no-new-dates-plugin.mjs
```diff
@@ -0,0 +1,7 @@
+import noNewDatesRule from './eslint-no-new-dates-rule.mjs';
+
+const noNewDatesPlugin = {
+  rules: { 'no-new-dates': noNewDatesRule },
+};
+
+export default noNewDatesPlugin;
```
Mwindo: What do you think about the following?

1. A directory for any custom eslint rules to avoid further cluttering the root directory
2. Calling the plugin something more general like `eslint-custom-rules-plugin.mjs`
3. Adding all custom rules (currently, 'no-new-dates') to `eslint-custom-rules-plugin.mjs`

Mwindo (approved):
Thanks for setting this up @TomElliottFlexion! Small comment to take or leave, but no objections from me--custom rules are easy enough to undo if we want.

eslint-no-new-dates-plugin.mjs
TomElliottFlexion: Excellent idea, done!

jimlerza (approved):

PR #5753, opened 2025-02-05
OpEx: Once a month, generate a fresh "stale cases" CSV and email it to stakeholders

jimlerza:
Here's a lambda that runs a report which generates a CSV and then emails the report to stakeholders.

Things to note:

   * added a new dispatcher for sending email with an attachment
     * the dispatcher does not use the existing mail queue because BulkTemplatedEmail messages can't contain attachments
     * there is no retry logic
     * ~we might need to add some verification headers or perhaps a replyTo or something~ update: nope
   * added a new lambda with a cron trigger
     * since this does so little I opted for a new role with just-enough permissions rather than the existing lambda role
     * because emails with attachments don't use templates, currently the email content is in the lambda, but it could probably live somewhere else


.circleci/config.yml
```diff
@@ -186,6 +186,12 @@ jobs:
           name: 'Deploy - Public Web Client - S3'
           command: |
             ./web-client/deploy-public.sh
+      - run:
+          name: 'Deploy - Inactive Cases Email Lambda and Monthly Cron Trigger'
+          command: |
+            if [ -n "$INACTIVITY_REPORT_RECIPIENTS" ]; then
```
jimlerza, line 192: We'll only set this secret in prod, so this will only deploy there.

web-api/src/dispatchers/ses/sendEmailWithAttachment.ts
```diff
@@ -0,0 +1,89 @@
...
+  filename: string;
+  recipient: string;
+  sender: string;
+  subject: string;
+}): Uint8Array => {
+  const boundary = '----=_Part_0_123456789.123456789';
+  const rawEmail = [
+    `From: ${sender}`,
+    `To: ${recipient}`,
+    `Subject: ${subject}`,
```
jimlerza, line 73: We might need to add some additional headers here - replyTo/bounce handler, sender verification headers, etc.

web-api/src/lambdas/staleCasesEmail/staleCasesEmailLambda.ts
```diff
@@ -0,0 +1,75 @@
...
+  'Attached is a DAWSON-generated list of cases that have had no ' +
+  'activity in DAWSON for at least 365 days. There should be no issue with ' +
+  'most or all of the cases listed; but it is also possible that an error ' +
+  'occurred. If a case is on this list and you believe it was included in ' +
+  'error, please let Case Services know.\n\n' +
+  'If your mail program does not open this file automatically (usually by ' +
+  'double clicking), you can open the file with Excel. The file is sorted ' +
+  'first by judge and then by number of days since the last activity. ' +
+  '"Chief Judge" indicates cases in the General Docket.\n\n' +
+  'Thank You,\nThe DAWSON Team';
```
jimlerza, line 21: Perhaps this email content belongs elsewhere?

web-api/src/dispatchers/ses/sendEmailWithAttachment.ts
```diff
@@ -0,0 +1,89 @@
...
+    `Content-Type: multipart/mixed; boundary="${boundary}"`,
+    '',
+    `--${boundary}`,
+    'Content-Type: text/plain; charset=UTF-8',
+    'Content-Transfer-Encoding: 7bit',
+    '',
+    body,
+    '',
+    `--${boundary}`,
+    `Content-Type: text/csv; name="${filename}"`,
```
jimlerza: Hard-coded `Content-Type` header for the attachment. Tsk tsk.

web-api/src/dispatchers/ses/sendEmailWithAttachment.ts
jimlerza: Update: checked into this with @mmarcotte, and those are set at the SES identity level, not on individual messages. Woohoo!

TomElliottFlexion (approved):

web-api/src/lambdas/staleCasesEmail/staleCasesEmailLambda.ts
zachrog: If we are okay with needing to do a code deploy in order to change the content of the email. I am okay if it lives directly here. Although not "pretty" I think it makes it very simple and easy to find.

zachrog (approved):

PR #5767, opened 2025-02-11
Devex: add nvmrc file

akuny:
This PR adds an `.nvmrc` file to the repository's root directory. This lets developers who have nvm installed quickly switch to whichever version of Node is specified by this file.


TomElliottFlexion (approved):

.nvmrc
```diff
@@ -0,0 +1 @@
+22.13.1
```
jimlerza:
```suggestion
22.13.1

```

jimlerza (approved):
Should we also mention somewhere in the dependency updates documentation that the value in this file needs to be incremented manually when the node version changes?

akuny:
> Should we also mention somewhere in the dependency updates documentation that the value in this file needs to be incremented manually when the node version changes?

Good call, added a note to the dependency updates doc.

akuny:
Closing in favor of https://github.com/ustaxcourt/ef-cms/pull/5889

PR #5768, opened 2025-02-11
Remove flaky multi-tab tests

zachrog:
This tests has been flaky on PRs with three engineers trying to fix it. We still have tests around idle-logout, just not multi-tab idle logout.


TomElliottFlexion (approved):

Mwindo:
Given that 1) this test has served its functionality for the foreseeable future (supporting a refactor of the idle logout behavior) and 2) the test currently does little more than check that our library is working (and that we broadcast), I am ok with removing this. That said, I do think it is worthwhile trying to add it back if we can at some point.

Mwindo (approved):

jimlerza (approved):

PR #5772, opened 2025-02-12
10598 Bug: Case Deadlines Blocked Issues in West

Mwindo:
Currently, when creating or deleting a case deadline, we immediately thereafter re-fetch the case deadlines as part of updateCaseAutomaticBlock. If we created/deleted the case deadline in the East and then are immediately reading from the West, the West might not yet have the update, which leads to the case not being properly blocked/unblocked.

This PR rectifies the issue by passing in data that results from create/delete rather than re-fetching.

### DIFF
--- a/shared/src/business/entities/cases/Case.ts
+++ b/shared/src/business/entities/cases/Case.ts
@@ -72,12 +72,11 @@ import { Statistic } from '../Statistic';
 import { TrialSession } from '../trialSessions/TrialSession';
 import { UnprocessableEntityError } from '../../../../../web-api/src/errors/errors';
 import { User } from '../User';
-import { clone, compact, includes, isEmpty, startCase } from 'lodash';
+import { clone, compact, includes, startCase } from 'lodash';
 import { compareStrings } from '../../utilities/sortFunctions';
 import { getDocketNumberSuffix } from '../../utilities/getDocketNumberSuffix';
 import { shouldGenerateDocketRecordIndex } from '../../utilities/shouldGenerateDocketRecordIndex';
 import joi from 'joi';
-import { RawCaseDeadline } from '@shared/business/entities/CaseDeadline';
 
 export class Case extends JoiValidationEntity {
   public associatedJudge?: string;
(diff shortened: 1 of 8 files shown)


web-api/src/business/useCases/caseDeadline/createCaseDeadlineInteractor.ts
```diff
@@ -33,13 +33,14 @@ export const createCaseDeadline = async (
...
 
-  await upsertCaseDeadlines([newCaseDeadline.validate().toRawObject()]);
+  const caseDeadlines = await upsertCaseDeadlines([newCaseDeadline.validate().toRawObject()]);
 
   caseEntity = await applicationContext
     .getUseCaseHelpers()
     .updateCaseAutomaticBlock({
       applicationContext,
       caseEntity,
+      deadlines: caseDeadlines
```
Mwindo: NOTE: This _isn't_ all of the case deadlines, but it still works because of how updateAutomaticBlocked works (it just checks for empty). This is a bit more opaque than I would like. Can we modify updateAutomaticBlocked to take in a more sensible parameter since we don't actually _use_ the case deadlines except for the isEmpty check?

Mwindo: > This is a bit more opaque than I would like

Because now a future developer might make changes under the reasonable assumption that they have all case deadlines in updateAutomaticBlocked ... but they don't!

shared/src/business/entities/cases/Case.ts
```diff
@@ -1400,18 +1399,14 @@ export class Case extends JoiValidationEntity {
    * @param {object} caseDeadlines - the case deadlines
    * @returns {Case} the updated case entity
    */
-  updateAutomaticBlocked({
-    caseDeadlines,
-  }: {
-    caseDeadlines?: RawCaseDeadline[];
-  }) {
+  updateAutomaticBlocked({ hasCaseDeadline }: { hasCaseDeadline: boolean }) {
```
Mwindo, line 1402: I changed `caseDeadlines` to `hasCaseDeadline` because 1) we don't actually need all of the case deadlines, 2) getting all of the case deadlines upon insert would have meant an extra db query following the upsert in `createCaseDeadline.ts`, and 3) passing in only a subset of deadlines (e.g., the case deadline upserted in `createCaseDeadline.ts`) seemed misleading to future devs.

web-api/src/business/useCaseHelper/automaticBlock/updateCaseAutomaticBlock.ts
```diff
@@ -3,30 +3,30 @@ import { ServerApplicationContext } from '@web-api/applicationContext';
...
 }) => {
   if (caseEntity.trialDate || caseEntity.highPriority) {
     return caseEntity;
   }
-  const caseDeadlines = await getCaseDeadlinesByDocketNumber({
-    docketNumber: caseEntity.docketNumber,
-  });
 
-  caseEntity.updateAutomaticBlocked({ caseDeadlines });
+  if (hasCaseDeadline === undefined) {
```
TomElliottFlexion, line 21: Is it safe to assume that we don't have access to this information in the other places where `updateCaseAutomaticBlock` is called?

TomElliottFlexion (approved):

web-api/src/business/useCaseHelper/automaticBlock/updateCaseAutomaticBlock.ts
Mwindo: Good call out. That is correct.

zachrog (approved):

pixiwyn (approved):

PR #5775, opened 2025-02-13
10598 Bug: Case Deadlines Blocked Issues in West

Mwindo:
Currently, when creating (manually or autogenerated from an order) or deleting a case deadline, we immediately thereafter re-fetch the case deadlines as part of `updateCaseAutomaticBlock`. If we created/deleted the case deadline in the East and then are immediately reading from the West, the West might not yet have the update, which leads to the case not being properly blocked/unblocked.

This PR rectifies the issue by passing in data that results from create/delete rather than re-fetching.


shared/src/business/entities/cases/Case.ts
```diff
@@ -1411,14 +1411,14 @@ export class Case extends JoiValidationEntity {
    * @param {object} caseDeadlines - the case deadlines
    * @returns {Case} the updated case entity
    */
-  updateAutomaticBlocked({ caseDeadlines }) {
+  updateAutomaticBlocked({ hasCaseDeadline }: { hasCaseDeadline: boolean }) {
```
Mwindo, line 1402: I changed `caseDeadlines` to `hasCaseDeadline` because 1) we don't actually need all of the case deadlines, 2) getting all of the case deadlines upon insert would have meant an extra db query following the upsert in `createCaseDeadline.ts`, and 3) passing in only a subset of deadlines (e.g., the case deadline upserted in `createCaseDeadline.ts`) seemed misleading to future devs.

pixiwyn (approved):

jimlerza (approved):

web-client/src/styles/custom.scss
```diff
@@ -1306,7 +1306,9 @@ hr.lighter {
     }
 
     font-weight: $font-bold;
+    /* stylelint-disable declaration-property-value-keyword-no-deprecated */
     overflow-y: overlay;
+    /* stylelint-enable declaration-property-value-keyword-no-deprecated */
```
Mwindo, line 1311: The pre-commit linter was complaining about deprecated keywords. This is completely orthogonal to the work in this PR. I can address in a follow-up, but I did not want to fiddle with styling in a staging PR that we expect to go live.

PR #5776, opened 2025-02-14
10598 Bug: Case Deadlines Blocked Issues in West (Part 2)

Mwindo:
- This PR addresses an East-West issue with autogenerated deadlines
- It also corrects a known existing issue in which the Blocked indicator does show immediately after adding a deadline but instead requires a refresh


web-api/src/business/useCases/caseDeadline/createCaseDeadlineInteractor.ts
```diff
@@ -43,13 +43,15 @@ export const createCaseDeadline = async (
...
+  const result = await applicationContext
+    .getUseCaseHelpers()
+    .updateCaseAndAssociations({
+      applicationContext,
+      authorizedUser,
+      caseToUpdate: caseEntity,
+    });
 
-  return newCaseDeadline;
+  return new Case(result, { authorizedUser }).validate().toRawObject();
```
Mwindo, line 54: This mimics what `deleteCaseDeadlineInteractor.ts` does. The response from `createCaseDeadlineInteractor` was never really used, so changing it here does not break any existing behavior.

web-client/src/presenter/actions/CaseDeadline/createCaseDeadlineAction.ts
```diff
@@ -25,16 +27,17 @@ export const createCaseDeadlineAction = async ({
...
   };
 
-  const createCaseDeadlineResult = await applicationContext
+  const updatedCase = await applicationContext
     .getUseCases()
     .createCaseDeadlineInteractor(applicationContext, {
       caseDeadline,
     });
 
+  setAutomaticBlockedFieldsOnState(updatedCase, store);
```
Mwindo, line 36: Just mimicking what we do with deleteCaseDeadlineAction to update the Blocked indicator without needing a refresh. Now we actually use the response to update state.

web-client/src/presenter/actions/CaseDeadline/createCaseDeadlineAction.ts
```diff
@@ -25,16 +27,17 @@ export const createCaseDeadlineAction = async ({
...
+
   return path.success({
     alertSuccess: {
       message: 'Deadline saved.',
     },
-    caseDeadline: createCaseDeadlineResult,
```
Mwindo, line 38: This wasn't used anywhere.

TomElliottFlexion (approved):

PR #5781, opened 2025-02-17
10536: Fix messages jumping when clicking message complete checkbox;

nechama-krigsman:
When sorting in descending order, the function first sorts the array and then reverses it using array.reverse(). When a checkbox is clicked, it triggers a state update and re-renders, which unexpectedly reverses the array again. Since array.reverse() mutates the original array, the order was already changed before the re-render. Cloning the array before reversing fixes this bug.

### DIFF
--- a/web-client/src/presenter/utilities/processFormattedMessages.ts
+++ b/web-client/src/presenter/utilities/processFormattedMessages.ts
@@ -58,26 +58,8 @@ export const sortFormattedMessages = (
   );
 
   if (tableSort && tableSort.sortOrder === DESCENDING) {
-    return sortedFormattedMessages.reverse();
+    return [...sortedFormattedMessages].reverse();
   }
-  return sortedFormattedMessages;
-};
-
-export const sortCompletedMessages = (
-  sortedMessages,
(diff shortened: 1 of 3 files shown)


web-client/src/presenter/utilities/processFormattedMessages.ts
```diff
@@ -58,7 +58,7 @@ export const sortFormattedMessages = (
   );
 
   if (tableSort && tableSort.sortOrder === DESCENDING) {
-    return sortedFormattedMessages.reverse();
+    return [...sortedFormattedMessages].reverse();
```
Mwindo, line 61: Since we always already clone `formattedCaseMessages` as part of setting `sortedFormattedMessages`, is this destructuring necessary?

Mwindo (approved):
Thanks @nechama-krigsman!

web-client/src/presenter/utilities/processFormattedMessages.ts
nechama-krigsman: After cloning `formattedCaseMessages` there was still some weirdness happening when toggling the sort between ascending and descending.

PR #5792, opened 2025-02-18
Add  GitHub actions kysely migrations check

Mwindo:
This adds a rough-and-ready GitHub Actions check to ensure that the migration files in a PR do not conflict with the migration files in the target branch.

The test passes below when the migrations are compatible. It does not pass [here](https://github.com/ustaxcourt/ef-cms/actions/runs/13401494223/job/37433148565?pr=5790) where I purposefully renamed a migration.

The new check will make renaming merged migrations, if we ever need to, more tedious, but it will prevent us from merging out-of-order/conflicting migration files. This seems like a good trade-off.


TomElliottFlexion (approved):
Fantastic idea

akuny (approved):

zachrog (approved):

PR #5797, opened 2025-02-20
opex: waf expensive requests monitor

mmarcotte:
This introduces a new waf rule to handle expensive requests in prod. Its eventual aim is to set a rate limit for order and opinion search on the public app as these requests can tax the opensearch cluster due to the parent:child relationship. Recent stress tests have shown that the DynamoDB rate limiting becomes ineffective when put under significant load due to eventual consistency and throttling. 

We are putting this rule in place with CloudWatch monitoring enabled so that we can evaluate the rate of these requests in production in order to ensure that our rule is configured at a reasonable value: 300 requests / 60sec. This value is subject to change after performing additional stress testing in the `test` environment as we rightsize the opensearch cluster instance class. 

This work has been tested and verified in the `dev` environment. 

https://github.com/flexion/ef-cms/issues/10600


zachrog (approved):
Very Very cool. I am happy to see our rate limiting move into managed resources like WAF.

Eventually will this get rid of the IPLimiter we have in the lambda code?

mmarcotte:
> Very Very cool. I am happy to see our rate limiting move into managed resources like WAF.
> 
> Eventually will this get rid of the IPLimiter we have in the lambda code?

Yes, `IPLimiter`, `advancedQueryLimiter`, and `userIdLimiter` are [on the chopping block](https://github.com/ustaxcourt/ef-cms/commit/b9f7f9cfce644ce30aa834c6b5220bd0598c6c73).

TomElliottFlexion (approved):

Mwindo (approved):

jimlerza (approved):

PR #5801, opened 2025-02-20
10569 story

pavanyenduri:
- Added casetype field onto the case search form.


pavanyenduri:
This PR has to point to Staging.

PR #5805, opened 2025-02-21
10600 opex opensearch rightsizing

mmarcotte:
flexion#10600


web-api/elasticsearch/elasticsearch-settings.ts
```diff
@@ -30,7 +30,7 @@ export const settings = ({
       'mapping.total_fields.limit': '1000',
       max_result_window: 20000,
       number_of_replicas: actualNumberOfReplicas,
-      number_of_shards: 1,
+      number_of_shards: 1, // TODO: would like to set to 2 or 3 for prod-like environments; will require a re-index
```
TomElliottFlexion: Does this need to be done here?

web-api/terraform/modules/waf/main.tf
```diff
@@ -64,18 +64,18 @@ resource "aws_wafv2_web_acl" "apis" {
...
+        custom_response {
+          response_code = 429
+        }
+      } // change to `block {}` when confident in ruleset
     }
 
     statement {
       rate_based_statement {
-        limit                 = 300
+        limit                 = 6000 # very high for now during load testing; current target is 350
```
TomElliottFlexion: Does this need to be lowered before going to staging?

TomElliottFlexion (approved):

web-api/terraform/modules/waf/main.tf
```diff
@@ -64,18 +64,18 @@ resource "aws_wafv2_web_acl" "apis" {
...
-      # block {
-      #   custom_response {
-      #     response_code = 429
-      #   }
-      # } // change to `block {}` when confident in ruleset
+      block {
+        custom_response {
+          response_code = 429
+        }
+      } // change to `block {}` when confident in ruleset
```
Mwindo: Do we need this comment still?

Mwindo (commented):
Can we update code related to [this](https://github.com/ustaxcourt/ef-cms/blob/cad304238a4830e4c7660967efb6ba5120b2f455/shared/src/sharedAppContext.ts#L16-L29)? It looks like it is no longer needed.

jimlerza:
> Can we update code related to [this](https://github.com/ustaxcourt/ef-cms/blob/cad304238a4830e4c7660967efb6ba5120b2f455/shared/src/sharedAppContext.ts#L16-L29)? It looks like it is no longer needed.

Nice find!

web-api/terraform/modules/waf/main.tf
jimlerza: > Do we need this comment still?

Sure don't!

jimlerza:
> > Can we update code related to [this](https://github.com/ustaxcourt/ef-cms/blob/cad304238a4830e4c7660967efb6ba5120b2f455/shared/src/sharedAppContext.ts#L16-L29)? It looks like it is no longer needed.
> 
> Nice find!

Upon further inspection, I actually think we still need this. The search interactors are performing GET/POST requests (which could possibly receive a 429 response) and then setting that custom error message in `state` if necessary.

In other words, this map is being used for handling these errors (which will still exist), not just for sending them (which we're not doing in lambda code anymore).

Mwindo (approved):
Looks like a few easy-to-fix unit test failures, but happy to approve. Thanks @jimlerza @mmarcotte :)

PR #5810, opened 2025-02-21
Fix Flaky Cypress test

zachrog:
This fixes judge-activity-report.cy.ts. The problem was the assertExists() will IMMEDIATELY look for something in the DOM and does not use the implicit cypress retries. This caused the Submitted/CAV test to be flaky as the selector showed on the DOM milliseconds after cypress searched for it.


cypress/helpers/retry.ts
```diff
@@ -15,6 +15,13 @@ export function retry(cb: () => Cypress.Chainable<boolean>, maxAttempts = 5) {
 }
 
 export function assertExists(selector: string) {
+  /*
+   This is a necessary wait as the below find command does not have built in cypress retries. 
+   This makes the below command VERY flaky because if the browser does not have the selector the moment you look for it this fails.
+   This function should only ever be used with the retry() mechanism.
+  */
+  // eslint-disable-next-line cypress/no-unnecessary-waiting
+  cy.wait(2000);
```
Mwindo, line 24: Good find! However, I wonder if there a way to use a timeout instead? From what I can tell, this will add ~30 seconds to our testing. In other words, why aren't we using built-in cypress retries?

Mwindo: Why aren't we just using `.should('exist');` for instance?

zachrog: Let me know if you want a flurry of information. Come chat with me.

Mwindo:
I think the retry mechanism we have needs improving, but this is a good stopgap measure in the meantime! Thanks, @zachrog!

Mwindo (approved):

PR #5816, opened 2025-02-24
10570: Add missing properties for sorting;

cruzjone-flexion:


shared/src/business/utilities/pendingItem/sortPendingReportItems.ts
```diff
@@ -10,11 +11,37 @@ export function sortPendingReportItems(
     return sortBy(pendingItems, ['receivedAt']);
   }
 
-  const sortedPendingItems = sortBy(pendingItems, [
-    pendingItemSortField,
+  const KEY = `sortable__${pendingItemSortField}`;
```
Mwindo: Is there a reason we are adding a temporary key, sorting, and then removing that key vs. not using a key at all and using a custom comparator instead?

Mwindo (approved):
Added a question about the way we are sorting (I'd prefer not adding a temporary key), but generally LGTM.

shared/src/business/utilities/pendingItem/sortPendingReportItems.ts
cruzjone-flexion: That was the next refactor, use JS sort instead of lodash method.

PR #5839, opened 2025-02-27
10561: Term Generator - Rulesheet UI

cruzjone-flexion:


web-api/src/business/useCaseHelper/trialSessions/trialSessionCalendaring/createProspectiveTrialSessions.ts
```diff
@@ -90,24 +95,18 @@ export const createProspectiveTrialSessions = ({
       trialLocation: city,
     });
 
-    // Since the min of reg cases is 40, and the min of small cases is 40,
-    // and the sum of these two values is below the hybrid case max of 100,
-    // we can safely assume that if the combination of remaining regular
-    // cases and remaining small cases is above the minimum of 50, we can
-    // assign all of those remaining cases to a hybrid session.
```
Mwindo, line 97: Is it worth updating rather than removing this comment?

web-client/src/views/TermBuilderView.tsx
```diff
@@ -0,0 +1,201 @@
...
+                  currentValue={hybridCaseMaxQuantity}
+                  updateFormValueSequence={updateFormValueSequence}
+                />{' '}
+                per trial session per location.
+              </p>
+
+              <p>
+                If there has been no trial (other than specials) in the last two
+                terms for a location, a session will be added if there are any
+                cases. (This will ignore any minimum cases rules). Motions and
```
Mwindo: Nitpick, but Figma shows the period in the parens

web-client/src/views/TermBuilderView.tsx
```diff
@@ -0,0 +1,201 @@
...
+                Hearings will not count as a trial visit to a location.
+              </p>
+
+              <p>
+                Washington, D.C. - Only one of the two court rooms can be taken
+                by a non-special session per week. Only South can be scheduled
```
Mwindo: Probably worth asking @katiecissell about the casing here since it looks like we usually uppercase Special but don't here.

web-client/src/views/TermBuilderView.tsx
```diff
@@ -0,0 +1,201 @@
...
+        </section>
+      </>
+    );
+  },
+);
+
+type TermBuilderInputParams = {
+  propertyName: string;
+  currentValue: number;
+  updateFormValueSequence: typeof sequences.updateFormValueSequence;
```
Mwindo: Can we make this type more agnostic? Otherwise, we are not really decoupling from Cerebral here.

web-client/src/views/TermBuilderView.tsx
```diff
@@ -0,0 +1,201 @@
...
+  propertyName,
+  updateFormValueSequence,
+}: TermBuilderInputParams) {
+  return (
+    <input
+      autoCapitalize="none"
+      name={propertyName}
+      placeholder="Number"
+      type="number"
+      min="0"
```
Mwindo: `min` is a good idea in principle, but it is leading to maybe unexpected behavior for users? Example: I want to delete 6 and say 5. I backspace to remove the 6. Now zero appears. If I type 5, I have 05, which (as a user) looks wrong. So then I have to highlight the whole thing and enter 5.

https://github.com/user-attachments/assets/2d85c70e-c80a-4329-b2c3-ee8adf0aba2e

Mwindo (approved):
Added a few small notes, but looks good overall. Thanks, @cruzjone-flexion!

web-api/src/business/useCaseHelper/trialSessions/trialSessionCalendaring/createProspectiveTrialSessions.ts
cruzjone-flexion: Don't think so because now we are using the same process as Small/Regular. 

This comment was regarding hard coding logic for hybrid (before we would only have 0 or 1 hybrid session now it checks if the minimum is met then is smart enough to calculate how many, similar to S/R)

web-client/src/views/TermBuilderView.tsx
Mwindo: Not sure what I was looking at--the Figma has the period as you have it here 🤦 Sorry!

PR #5846, opened 2025-03-03
devex: update test case with hardcoded date

akuny:
This test case was failing because over the weekend March 1, 2025 became a date in the pass, not the future.

This test has been updated to use a date that will remain in the future for the duration of DAWSON's lifecycle. It also now explicitly expects the particular exception that the function under test should throw under these conditions--the previous arrangement of the test case only implied this outcome.


Mwindo (approved):
Thanks, @akuny !

nechama-krigsman (approved):

jimlerza (approved):

PR #5866, opened 2025-03-06
10594: Disable checkbox for petitioner corresponding to currently logged in user when e-filing

En-8:
flexion#10594


Mwindo (approved):
Nitpick, but is it worth enforcing this on the back end too? (I can imagine myself on an off day thinking, "Man, this checkbox is broken, let me modify the DOM." Definitely an edge case though.)

akuny (approved):
This looks good, although I am curious what @Mwindo is thinking in terms of corresponding changes in the backend.

En-8:
> Nitpick, but is it worth enforcing this on the back end too? (I can imagine myself on an off day thinking, "Man, this checkbox is broken, let me modify the DOM." Definitely an edge case though.)

This is a good point, and something I think is worth addressing. Standby...

Mwindo:
> This looks good, although I am curious what @Mwindo is thinking in terms of corresponding changes in the backend.

My thought was just enforcing the "petitioner x cannot _not_ file on behalf of petitioner x, even if they file on behalf of others too" in the file-a-document back end. So even if a user de-selected the radio button, the back end would roll its eyes and make sure the document was filed on their behalf anyway.

En-8:
> > This looks good, although I am curious what @Mwindo is thinking in terms of corresponding changes in the backend.
> 
> My thought was just enforcing the "petitioner x cannot _not_ file on behalf of petitioner x, even if they file on behalf of others too" in the file-a-document back end. So even if a user de-selected the radio button, the back end would roll its eyes and make sure the document was filed on their behalf anyway.

Unfortunately, doing this validation _actually_ server-side introduces an egregious amount of complexity due to the wizard-like workflow only being submitted to the backend after the user has already made it through all other client-side validation and is sitting on the "review" page (which currently implies everything is valid according to the system).

However, I was able to hook into the current client-side validation that occurs when the user hits the "Review Filing" button on the same screen they'd be selecting parties. It's not perfect, but it's at least one additional layer of stuff.

jimlerza (approved):

PR #5869, opened 2025-03-07
Fix broken import in test

Mwindo:
We removed `navigateToCaseDetail` from [here](https://github.com/ustaxcourt/ef-cms/blob/test/cypress/local-only/support/pages/case-detail.ts) in staging but were still trying to import it in test [here](https://github.com/ustaxcourt/ef-cms/blob/test/cypress/local-only/tests/integration/trialSessionMinutes/create-and-print-minute-sheet.cy.ts).


cypress/local-only/tests/integration/trialSessionMinutes/create-and-print-minute-sheet.cy.ts
```diff
@@ -20,7 +17,7 @@ describe('Create a minute sheet, fill out sections of the form, navigate away an
     createAndServePaperPetition({ trialLocation: 'Birmingham, Alabama' }).then(
       ({ docketNumber }) => {
         loginAsIrsPractitioner();
-        navigateToCaseDetail('irspractitioner', docketNumber);
```
Mwindo, line 23: This was doing a duplicate login (when it worked).

cruzjone-flexion (approved):

zachrog (approved):

PR #5873, opened 2025-03-07
10569 story intermediate branch to test

zachrog:


web-client/src/presenter/computeds/AdvancedSearch/CaseSearchByNameHelper.test.ts
```diff
@@ -4,28 +4,32 @@ import { formatNow, FORMATS } from '@shared/business/utilities/DateHandler';
...
-        value: CASE_TYPES_MAP.whistleblower,
-      },
-    ];
+  const caseTypeOptions = [
+    {
+      label: CASE_TYPES_MAP.deficiency,
+      value: CASE_TYPES_MAP.deficiency,
+    },
+    {
+      label: 'L - Collection (Lien/Levy)',
```
Mwindo: I think we should probably be using our constants here, at the very least for `Collection (Lien/Levy)`. Maybe L/P/R/W/X are constants somewhere too?

web-client/integration-tests-public/journey/unauthedUserSearchesByMeta.ts
```diff
@@ -32,7 +32,7 @@ export const unauthedUserSearchesByMeta = (cerebralTest, overrides = {}) => {
       cerebralTest.getState(`searchResults.${ADVANCED_SEARCH_TABS.CASE}`),
     ).toBeUndefined();
     expect(cerebralTest.getState('advancedSearchForm')).toEqual({
-      caseSearchByName: { countryType: ALL_COUNTRY_TYPE },
+      caseSearchByName: { countryType: ALL_COUNTRY_TYPE, procedureType: 'All' },
```
Mwindo: I wonder if it is worth having a more generic "all options" constant. This feels a little clunky given that we use both a constant and a non-constant. (Same comment below.)

web-client/src/views/AdvancedSearch/CaseSearchByName.tsx
```diff
@@ -107,22 +108,20 @@ export const CaseSearchByName = connect(
...
                                 .countryType === 'all'
                             }
                             className="usa-radio__input"
                             data-testid="all-country-selection"
                             id="all-country-selection"
-                            name="all"
+                            name="country"
                             type="radio"
-                            value="All"
+                            value="all"
```
Mwindo: This should probably be the constant.

Mwindo (approved):
Overall this looks good. I'd prefer a few more constants. Ideally, in my head, we should have an `ALL_OPTIONS` constant (= 'all' or 'All' or whatever) and use that consistently to avoid an inevitable case mismatch in the future.

Thanks @zachrog!

web-client/src/presenter/computeds/AdvancedSearch/CaseSearchByNameHelper.test.ts
zachrog: Updated to use const!

web-client/integration-tests-public/journey/unauthedUserSearchesByMeta.ts
zachrog: Updated to use const! good idea.

PR #5876, opened 2025-03-11
10502: Fix generateDocketNumber

Mwindo:
We were ordering by `docketNumber` (lexicographical), not `sortableDocketNumber` (numerical), which meant new docket numbers were not being generated properly. Plus a few miscellaneous changes from the past few days, including unit tests for CaseFactory and pulling out one-off database reads into functions.

We also fixed an issue with the case inventory report pagination.


pixiwyn (approved):

shared/src/business/entities/Statistic.ts
```diff
@@ -192,3 +192,7 @@ const itemizeTotalPenalties = function (
     });
   }
 };
+
+declare global {
```
zachrog: When you have a moment please remove global declarations, we try to avoid these now as when types come out of nowhere it can be confusing.

akuny (approved):

zachrog (approved):

PR #5886, opened 2025-03-13
10502 Cases: Some updates

Mwindo:
- Fix statistic penalties, which were not being attached to their parent statistics correctly, and make sure their order is maintained


shared/src/business/entities/Statistic.test.ts
```diff
@@ -271,15 +263,6 @@ describe('Statistic', () => {
...
     });
 
-    it('should update the penalty in the penalties array', () => {
-      expect(statistic.penalties.length).toEqual(penaltyArrayLength);
-
-      statistic.updatePenalty(MOCK_UPDATED_PENALTY);
-
-      expect(statistic.penalties.length).toEqual(penaltyArrayLength);
-      expect(statistic.penalties[0]).toEqual(MOCK_UPDATED_PENALTY);
-    });
```
Mwindo, line 281: `updatePenalty` was never called anywhere in the codebase, so I got rid of it and its test here.

Mwindo:
Will just open a new branch with changes from 10502-dxox

PR #5888, opened 2025-03-13
10502 Updates

Mwindo:
- Update entity diagram
- Fix itemized penalties not always pointing towards their parent statistic. (This wasn't a problem in Dynamo because, despite storing a `statisticId` "foreign key" on the penalty, we really relied on the parent-child relationship in the Dynamo record rather than the "foreign key.")
- Enforce sort orders on statistics and penalties. This involves adding a new column to both tables. (This wasn't a problem in Dynamo because we were relying on the array-based index in the Dynamo records.)
- Try to speed up getCaseByDocketNumber slightly by 1) not getting practitioners twice and 2) using Promise.all
- Fix some type errors

**Note:** after merging this PR, we will need to run a migration to 1) fix orphaned penalties (whose `statisticId` does not point to a statistic) and 2) ensure statistics and penalties have proper timestamps (based on their array-based index in their DynamoDB records) so that we can display them in the appropriate order using Postgres.


web-api/src/persistence/postgres/cases/getCaseByDocketNumber.ts
```diff
@@ -27,25 +27,28 @@ export const getCaseByDocketNumber = async ({
   // These case items are no longer in dynamoDB
   const SK_FILTER_OUT = ['work-item', 'correspondence', 'case'];
 
-  const dbCaseMetadata = await getCaseMetadataWithCounsel({
-    applicationContext,
+  const dbCaseMetadata = await getCaseMetadataByDocketNumber({
```
Mwindo, line 30: We don't need to get the counsel--that will be gotten below as part of our catch-all dynamo query

shared/src/business/entities/Statistic.test.ts
```diff
@@ -271,15 +263,6 @@ describe('Statistic', () => {
...
     });
 
-    it('should update the penalty in the penalties array', () => {
-      expect(statistic.penalties.length).toEqual(penaltyArrayLength);
-
-      statistic.updatePenalty(MOCK_UPDATED_PENALTY);
-
-      expect(statistic.penalties.length).toEqual(penaltyArrayLength);
-      expect(statistic.penalties[0]).toEqual(MOCK_UPDATED_PENALTY);
-    });
```
Mwindo, line 281: `updatePenalty` was never called, so I removed it and its tests.

shared/src/business/entities/trialSessions/TrialSession.ts
```diff
@@ -106,7 +106,7 @@ export class TrialSession extends JoiValidationEntity {
   public termYear: string;
   public trialClerk?: TTrialClerk;
   public trialLocation?: string;
-  public trialSessionId?: string;
+  public trialSessionId: string;
```
Mwindo, line 109: We always set this. See constructor.

web-api/src/business/useCases/trialSessions/updateTrialSessionInteractor.test.ts
```diff
@@ -353,23 +352,5 @@ describe('updateTrialSessionInteractor', () => {
         `Trial session ${TEST_TRIAL_SESSION_ID} was not found.`,
       );
     });
-
-    it('should call with empty string if no trial session id is provided', async () => {
```
Mwindo, line 357: A TrialSession entity always has an id.

web-api/src/persistence/postgres/cases/statistics/getCaseStatistics.ts
```diff
@@ -1,58 +1,39 @@
...
+        sql`jsonb_agg(to_jsonb(sp) ORDER BY sp.updated_at)`.as('penalties'),
+      )
+      .orderBy('cs.year')
+      .orderBy('cs.updatedAt', 'desc')
+      .groupBy(['cs.docketNumber', 'cs.statisticId'])
       .execute(),
   );
 
-  // Group penalties by statisticId
-  const statisticsWithPenalties: Record<string, RawStatistic> =
```
Mwindo, line 20: Doing this grouping in the query itself now.

Mwindo:
Opened new branch since there were more changes from 10502-dxox

PR #5890, opened 2025-03-14
10436 to test

TomElliottFlexion:


web-api/src/persistence/postgres/cases/reports/getBlockedCasesForTrialLocation.ts
```diff
@@ -21,9 +22,10 @@ export type BlockedCaseData = Pick<
 
 export const getBlockedCasesForTrialLocation = async (
   trialLocation: string,
+  filterStatusForTrialLocation: boolean = false,
```
Mwindo, line 25: I wonder how you would feel about making this more general, like, `statuses: string[] | undefined`? Looking at this without context, I wouldn't know what `filterStatusForTrialLocation` would mean (e.g., "What statuses? Why?").

web-api/src/persistence/postgres/cases/getEligibleCasesForTrialCity.ts
```diff
@@ -0,0 +1,105 @@
...
+        'isSealed',
+      ])
+      .where('preferredTrialCity', '=', trialCity)
+      .where('status', '=', CASE_STATUS_TYPES.generalDocketReadyForTrial)
+      .where(eb =>
+        eb.and([
+          eb.or([
+            eb('automaticBlocked', '=', false),
+            eb('automaticBlocked', 'is', null),
+          ]),
```
Mwindo: I think you could do `'automaticBlocked', 'is not', true`

Mwindo: Similar comment in the next line

web-api/src/persistence/postgres/cases/mapper.ts
```diff
@@ -316,3 +318,13 @@ export const indexCaseEntity = ({
     }),
   );
 };
+
+export const convertDbRowToRawEligibleCase = (dbCase: any): RawEligibleCase => {
```
Mwindo: We should probably just call this `rawEligibleCase`. Apologies--this `convert..` nomenclature was mine, but it was not consistent with others!

web-client/src/ustc-ui/Icon/CaseIcons.tsx
```diff
@@ -3,7 +3,13 @@ import { Icon } from '@web-client/ustc-ui/Icon/Icon';
 import React from 'react';
 import classNames from 'classnames';
 
-export function CaseIcons({ formattedCase }: { formattedCase: any }) {
+export function CaseIcons({
+  formattedCase,
+  shouldIndent = true,
```
Mwindo: Is `shouldIndent` necessary? I did a quick test locally (with Anchorage), and it looked like the indentation was the same. If it _is_ necessary, I wonder if there is some way we can avoid it. (My general thought is that "shouldIndent" feels like a strange attribute to be adding to a "case" object, and a tend to dislike boolean flags when avoidable. But maybe this is difficult 🤷 )

web-client/src/presenter/computeds/trialLocationHelper.ts
```diff
@@ -0,0 +1,135 @@
...
+    isLeadCase: boolean;
+  },
+): BlockedFormattedCase => {
+  const blockedFormattedCase: BlockedFormattedCase = {
+    ...blockedCase,
+    blockedDateEarliest: '',
+    caseTitle: '',
+  };
+
+  if (blockedCase.blockedDate && blockedCase.automaticBlocked) {
```
Mwindo: It looks like the top-level conditions are all based on `blocked` or `automaticBlocked` except this first `blockedCase.blockedDate`. Is that correct? (Just worried about any possible discrepancy between blocked and blockedDate in this logic.)

Mwindo (approved):
Thanks @TomElliottFlexion! I left a few comments, but nothing pressing. I'll probably do a more thorough review once the whole thing is ready.

akuny (approved):

web-api/src/persistence/postgres/cases/getEligibleCasesForTrialCity.ts
TomElliottFlexion: Thank you! For some reason `'!=', true` does not work; this does

web-client/src/ustc-ui/Icon/CaseIcons.tsx
TomElliottFlexion: Certainly not! This is a vestige of a misconception that no longer exists

web-client/src/presenter/computeds/trialLocationHelper.ts
TomElliottFlexion: This _is_ confusing. Refactored to use the same method between `blockedCasesReportHelper` and `trialLocationHelper`

web-api/src/persistence/postgres/cases/reports/getBlockedCasesForTrialLocation.ts
TomElliottFlexion: You're right, I'll refactor this. I'm not 100% sure I want to pass in the values to filter, since there is only one instance in which we _don't_ filter blocked case statuses, but this is confusing

PR #5901, opened 2025-03-18
10502 Experiment: Better Entity Typing

Mwindo:
An experiment to improve the typing for our entities. The basic idea: don't throw away type information with transformNullToUndefined, and try to make a single mapper function for the case table that is robust enough to return different subsets of RawCase data.

This is an experimental branch meant to be merged into 10502-dxox.


Mwindo:
Closing this branch since the work has been merged into 10502-dxox

PR #5903, opened 2025-03-18
10606 Add AWS Real User Monitoring Infrastructure and UI instrumentation

En-8:
flexion#10606

This PR sets up the infrastructure and UI instrumentation to enable collection of Real User Monitoring metrics using the AWS CloudWatch Real User Monitoring service.

There's a few key pieces:
- Required Infrastructure: RUM App Monitor, corresponding Cognito Identity pool allowing unauthenticated access so that the web client (running in the user's browser) is able to retrieve tokens for signing requests to Put RUM events. We opt for a single app monitor per environment in us-east-1 so that RUM events are available in a centralized location per environment, regardless of color (blue/green).
- Client instrumentation. This is done for both the public site and internal, authenticated site.
- **Account Specific**: Additional permissions for CircleCI to be able to provision and manage RUM-related resources via Terraform.

## Manual Deploy Step (Prior to deployment)
For lower environments, `"RUM_SAMPLE_RATE": "1"` needs to be added to env config in secrets manager
For production, `"RUM_SAMPLE_RATE": "0.1"`

### Secret Set in USTC Environments

- [x] test
- [x] staging
- [x] exp1
- [x] dev
- [x] irs
- [x] prod

### Account-Specific Deployed in USTC Accounts

- [x] lower
- [ ] prod

### Secret Set in Flexion Environments

- [x] exp1
- [x] exp2
- [x] exp3
- [x] exp4
- [x] exp5

### Account-Specific Deployed in Flexion Account

- [x] account


web-api/terraform/modules/ci-cd/ci-cd-permissions.tf
```diff
@@ -85,6 +85,22 @@ resource "aws_iam_policy" "ci_cd_policy" {
...
+      "Action": [
+        "cognito-identity:*"
+      ],
+      "Resource": "arn:aws:cognito-identity:us-east-1:${data.aws_caller_identity.current.account_id}:identitypool/*"
+    },
+    {
+      "Sid": "CloudWatchRUM",
+      "Effect": "Allow",
+      "Action": [
+        "rum:*"
```
En-8, line 100: We had to keep these wildcard Actions because we were receiving errors stating that the overall policy size is reaching AWS size limits.

web-api/terraform/modules/rum/rum.tf
```diff
@@ -0,0 +1,68 @@
+
+resource "aws_cognito_identity_pool" "rum_identity_pool" {
+  identity_pool_name               = "${var.environment}-rum-identity-pool"
+  allow_unauthenticated_identities = true
```
En-8, line 3: Unauthenticated access is required for RUM to function for both the public and authenticated sites.

https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-RUM-get-started-authorization.html

web-client/src/providers/realUserMonitoring.ts
```diff
@@ -0,0 +1,30 @@
...
+  if (process.env.ENV === 'local') return;
+  let awsRum: AwsRum;
+  try {
+    const config: AwsRumConfig = {
+      allowCookies: true,
+      enableXRay: false,
+      endpoint: 'https://dataplane.rum.us-east-1.amazonaws.com',
+      identityPoolId: process.env.RUM_IDENTITY_POOL_ID,
+      sessionSampleRate: process.env.ENV === 'prod' ? 0.1 : 1, // sample 10% of sessions in production until we know how expensive this will be, 100% in lower environments
+      telemetries: ['performance'],
```
En-8, line 13: There are other potentially useful metrics to collect, but starting with "performance" since this story is solely focused on ensuring we can measure performance. We also want to gather more data about how expensive this will be to run in production.

web-client/src/providers/realUserMonitoring.ts
```diff
@@ -0,0 +1,30 @@
...
+    awsRum = new AwsRum(
+      APPLICATION_ID,
+      APPLICATION_VERSION,
+      'us-east-1',
+      config,
+    );
+
+    awsRum.enable();
+  } catch (error) {
+    // Ignore errors thrown during CloudWatch RUM web client initialization
```
En-8, line 28: We ignore errors on RUM Client init so that we don't create disruptions for users if the client fails to initialize for some reason.

jimlerza (approved):

web-client/src/providers/realUserMonitoring.ts
```diff
@@ -0,0 +1,30 @@
...
+export const initializeRealUserMonitoring = () => {
+  if (process.env.ENV === 'local') return;
+  let awsRum: AwsRum;
+  try {
+    const config: AwsRumConfig = {
+      allowCookies: true,
+      enableXRay: false,
+      endpoint: 'https://dataplane.rum.us-east-1.amazonaws.com',
+      identityPoolId: process.env.RUM_IDENTITY_POOL_ID,
+      sessionSampleRate: parseInt(process.env.RUM_SAMPLE_RATE!),
```
jimlerza: I think `parseInt()` only likes whole integers!

```suggestion
      sessionSampleRate: Number(process.env.RUM_SAMPLE_RATE!),
```

web-client/src/providers/realUserMonitoring.ts
Mwindo: Is it worth logging this failure?

Mwindo (approved):
Awesome job @En-8!

web-client/src/providers/realUserMonitoring.ts
En-8: I went back-and-forth on this. It might have some use, at least for us in lower environments, but in production it's going to be logging errors to the user's browser console, which we're not currently collecting anywhere (I don't think).

My concern with turning it on in production is that I don't know what it might log. I would hope that it wouldn't end up logging anything sensitive, but I can't with 100% confidence say it wouldn't.

I'm on the fence, and could pretty easily be convinced to go ahead and add a log here.

En-8: I went ahead and added the log. After some thought, the potential for it to be useful for troubleshooting does seem worth it enough to me.

Mwindo: Oh, I was thinking just calling the [logError endpoint](https://github.com/ustaxcourt/ef-cms/blob/57d0fce3ff870699f7c0a6a64dbcced362973f8d/web-api/src/app.ts#L1090) with a generic message like `RUM failed to initialize` so we could monitor the monitoring lol (good fodder for a potential, and maybe useful, alarm). Whatever you decide is fine, I don't have strong feelings.

PR #5907, opened 2025-03-19
10606 story intermediate branch to test 1742402482

En-8:
flexion#10606
https://github.com/ustaxcourt/ef-cms/pull/5903

This has been deployed successfully to an experimental environment https://app.circleci.com/pipelines/github/flexion/ef-cms/61069/workflows/f597809b-439c-4026-88bf-29af9865ab1e

https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#rum:dashboard?tab=overview

## Manual Deploy Step
- `"RUM_SAMPLE_RATE": "1"` needs to be added to environment config in secrets manager, e.g. `test_deploy`. We'll be using 100% session sampling in lower environments and 10% in production, to start with.


jimlerza (approved):

.circleci/config.yml
```diff
@@ -138,6 +138,13 @@ jobs:
           name: 'Deploy - All Color Terraform'
           command: |
             npm run deploy:allColors $ENV
+            cd ./web-api/terraform/applyables/allColors
```
zachrog: Is this necessary if you are looking the values up with the AWS CLI? in 00-common?

web-api/terraform/bin/deploy-app-all-colors.sh
```diff
@@ -101,6 +101,7 @@ export TF_VAR_dynamsoft_product_keys=$DYNAMSOFT_PRODUCT_KEYS
 export TF_VAR_postgres_master_username="${POSTGRES_MASTER_USERNAME}"
 export TF_VAR_postgres_master_password="${POSTGRES_MASTER_PASSWORD}"
 export TF_VAR_restoring_aws_account_id=$PROD_ENV_ACCOUNT_ID
+export TF_VAR_rum_sample_rate=$RUM_SAMPLE_RATE
```
zachrog, line 104: Who is responsible for setting RUM_SAMPLE_RATE? Is it in secrets manager?

web-api/terraform/modules/rum/rum.tf
```diff
@@ -0,0 +1,67 @@
+resource "aws_cognito_identity_pool" "rum_identity_pool" {
+  identity_pool_name               = "${var.environment}-rum-identity-pool"
+  allow_unauthenticated_identities = true
+}
+
+resource "aws_iam_role" "rum_unauthenticated_role" {
```
zachrog, line 6: I think this has to do with my lack of knowledge of RUM but why do we need a separate identity pool for RUM? I assume this is so the webapp can post to RUM directly but when does our WebApp assume this role?

.circleci/config.yml
zachrog: I think I have a preference for accessing these infrastructure ids in a more durable way than environment variables.

Either by storing them in something like AWS-SSM to be looked up later, or by looking up the values when they are needed using the awsCli or AWS SDK

zachrog (changes requested):
Just wanted to see infrastrcture ids being stored or accessed in another way so that we do not have to rely on EXACTLY this pipeline for our deploys to work.

Overall this looks awesome and aligns perfectly with how we organized terraform. Nice job Nate!

web-api/terraform/bin/deploy-app-all-colors.sh
zachrog: NVM I missed your comment!

TomElliottFlexion (approved):

web-api/terraform/modules/rum/rum.tf
En-8: Your assumption is correct -- the web app needs credentials so that it can sign requests sent directly to AWS RUM. ([more info here](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-RUM-get-started-authorization.html)).

The details of when this is happening are a little opaque from what I could find, but I believe the role is assumed upon initialization of the RUM client in the user's browser , then the RUM client manages refreshing credentials as needed. We allow unauthenticated access, so this is not tied to the user authentication flow.

There are more options for setting up a RUM monitor, at least one of which doesn't require an identity pool (using a resource-based policy). You could also restrict the identity pool to require authentication for access, which would involve configuring the identity pool to work with our user pool. I opted against this because then we would be unable to collect metrics for the public site and internal site with a single app monitor.

.circleci/config.yml
En-8: 1. If by "is this necessary" you're asking why we need the values here and also in 00-common, my understanding was that 00-common was exclusively for the environment switcher, whereas here we're getting them so that they're available to the `build-dist.sh` script in CircleCI. If you mean "why do we need to use this pattern instead of the pattern we have in 00-common (to your second question), read on...
2. This is a good point. Initially I wasn't a huge fan of the "look it up with the CLI" pattern because it adds latency to something we already have via TF outputs and requires specifying the name of the resource in a second place that's not tied to the terraform configuration. However, I wasn't considering how this environment variable approach couples us directly to this exact pipeline. Since I already have the cli commands I'd need to look up the appropriate values, I'll use those instead.

zachrog (approved):

PR #5908, opened 2025-03-19
10502 Updates

Mwindo:
- Update entity diagram
- Fix itemized penalties not always pointing towards their parent statistic. (This wasn't a problem in Dynamo because, despite storing a `statisticId` "foreign key" on the penalty, we really relied on the parent-child relationship in the Dynamo record rather than the "foreign key.")
- Enforce sort orders on statistics and penalties. This involves adding a new column (we chose `updatedAt`) to both tables. (This wasn't a problem in Dynamo because we were relying on the array-based index in the Dynamo records.)
- Add cypress tests around statistics and statistics penalties
- Try to speed up `getCaseByDocketNumber` slightly by 1) not getting practitioners twice and 2) using `Promise.all`
- Also use `Promise.all` in `getCaseMetadataWithCounsel`
- Add an index for `leadDocketNumber` on the Case table
- Fix the Cold Case Report timing out--query was taking 200+ seconds, now takes << 1
- Revamp the database schema definitions so that we can get metadata information more easily; it looks uglier, but I think it is better overall
- Update `transformNullToUndefined` so that it does not wash away types and so that it works recursively (thanks to @zachrog for pointing this out and helping!)
- Revamp the Case mapper function (now called `fromKyselyCase`). We were pretending to return a RawCase entity when we weren't. The new mapper function keeps us honest about our types and should be flexible. (Thanks to @zachrog and @akuny for their helpful discussions on this!)
- Fix some type errors

---

**Notes** 
After merging this PR, we will need to do the following:

1. Roll back migrations 0015, 0014, 0013. We added an index in 0013.
2. Run a migration`*`

`*` Why? To 1) fix orphaned penalties (whose `statisticId` does not point to a statistic) and 2) ensure statistics and penalties have proper timestamps (based on their array-based index in their DynamoDB records) so that we can display them in the appropriate order using Postgres.


web-api/src/persistence/postgres/cases/getCaseByDocketNumber.ts
```diff
@@ -27,25 +27,28 @@ export const getCaseByDocketNumber = async ({
   // These case items are no longer in dynamoDB
   const SK_FILTER_OUT = ['work-item', 'correspondence', 'case'];
 
-  const dbCaseMetadata = await getCaseMetadataWithCounsel({
-    applicationContext,
+  const dbCaseMetadata = await getCaseMetadataByDocketNumber({
```
Mwindo, line 30: We don't need to get the counsel--that will be gotten further down in the file as part of our catch-all dynamo query

shared/src/business/entities/Statistic.test.ts
```diff
@@ -271,15 +263,6 @@ describe('Statistic', () => {
...
 
-    it('should update the penalty in the penalties array', () => {
-      expect(statistic.penalties.length).toEqual(penaltyArrayLength);
-
-      statistic.updatePenalty(MOCK_UPDATED_PENALTY);
-
-      expect(statistic.penalties.length).toEqual(penaltyArrayLength);
-      expect(statistic.penalties[0]).toEqual(MOCK_UPDATED_PENALTY);
-    });
-
```
Mwindo, line 282: `updatePenalty` was never called, so I removed it and its tests.

cypress/local-only/tests/integration/caseDetail/caseInformation/edit-case-caption-via-case-detail-header.cy.ts
```diff
@@ -1,23 +0,0 @@
-import { loginAsDocketClerk } from 'cypress/helpers/authentication/login-as-helpers';
```
Mwindo, line 1: This test was wrapped into the new edit-case-detail.cy.ts test.

cypress/local-only/tests/integration/caseDetail/caseInformation/edit-case-detail.cy.ts
```diff
@@ -0,0 +1,327 @@
+import { loginAsDocketClerk } from 'cypress/helpers/authentication/login-as-helpers';
```
Mwindo, line 1: We can add more to this file since it currently just tests renaming a case caption, changing the case type, and statistics-related things. I thought it more efficient to create the case once and do multiple case-detail edits rather than having separate files that each create a case.

shared/src/business/entities/trialSessions/TrialSession.ts
```diff
@@ -106,7 +106,7 @@ export class TrialSession extends JoiValidationEntity {
   public termYear: string;
   public trialClerk?: TTrialClerk;
   public trialLocation?: string;
-  public trialSessionId?: string;
+  public trialSessionId: string;
```
Mwindo, line 109: We always set this. See constructor.

shared/src/business/entities/cases/Case.ts
```diff
@@ -134,7 +133,7 @@ export class Case extends JoiValidationEntity {
   public useSameAsPrimary?: boolean;
   public initialDocketNumberSuffix?: string;
   public noticeOfTrialDate?: string;
-  public docketNumberWithSuffix?: string;
+  public docketNumberWithSuffix!: string;
```
Mwindo, line 136: We always set this. See constructor. The `!` is because we set it via constructor -> `assignFieldsForAllUsers`, not directly in the constructor itself, and TypeScript complains that we don't set it directly in the constructor.

shared/src/business/useCases/getCaseDeadlinesInteractor.ts
```diff
@@ -72,5 +72,5 @@ export const getCaseDeadlinesInteractor = async (
     });
   }
 
-  return { deadlines: deadlinesWithFullInfo, totalCount: Number(totalCount) };
+  return { deadlines: deadlinesWithFullInfo, totalCount };
```
Mwindo, line 75: The persistence method casts this now so we don't have to do it here.

web-api/src/business/useCases/trialSessions/updateTrialSessionInteractor.test.ts
```diff
@@ -353,23 +352,5 @@ describe('updateTrialSessionInteractor', () => {
         `Trial session ${TEST_TRIAL_SESSION_ID} was not found.`,
       );
     });
-
-    it('should call with empty string if no trial session id is provided', async () => {
```
Mwindo, line 357: A TrialSession entity always has an id.

web-api/src/persistence/postgres/cases/statistics/getCaseStatistics.ts
```diff
@@ -1,58 +1,40 @@
...
+      .select(
+        sql`jsonb_agg(to_jsonb(sp) ORDER BY sp.updated_at)`.as('penalties'),
+      )
+      .orderBy('cs.year')
+      .orderBy('cs.updatedAt', 'desc')
+      .groupBy(['cs.docketNumber', 'cs.statisticId'])
       .execute(),
   );
 
-  // Group penalties by statisticId
```
Mwindo, line 19: All this grouping is done in the query now rather than in app code.

web-api/src/persistence/dynamo/helpers/purgeDynamoKeys.ts
```diff
@@ -1,11 +1,11 @@
...
+export function purgeDynamoKeys<T, S>(obj: T): S {
   for (const property in obj) {
     if (dynamoKeys.includes(property)) {
       delete obj[property];
     } else if (typeof obj[property] === 'object') {
       obj[property] = purgeDynamoKeys(obj[property]);
     }
   }
-  return obj;
+  return obj as unknown as S;
```
Mwindo, line 10: We aren't returning type T--we are returning a new, unknown type. Since this function is hardly used anymore, I allowed specifying the return type S for ease of use.

web-api/src/database-types.ts
```diff
@@ -1,310 +1,447 @@
...
-  isRepliedTo: boolean;
-  message: string;
-  messageId: string;
-  parentMessageId: string;
-  subject: string;
-  to: string;
-  toSection: string;
-  toUserId: string;
-}
+const messageTableDefinition = {
```
Mwindo, line 5: The new general idea in `database-types`: define dummy objects with type information and dynamically generate stuff from it to avoid having to re-specify data like, e.g., column names.

web-api/src/database-types.ts
```diff
@@ -1,310 +1,447 @@
...
 export type MinuteSheetKysely = Selectable<MinuteSheetTable>;
 export type NewMinuteSheetKysely = Insertable<MinuteSheetTable>;
 export type UpdateMinuteSheetKysely = Updateable<MinuteSheetTable>;
+
+type DatabaseTableMetadata<TTable> = {
+  table: TTable;
+  columns: string[];
+};
+
+export const DatabaseSchema = {
```
Mwindo, line 391: Now we can associate metadata with each table.

web-api/src/persistence/postgres/cases/getCaseMetadataByDocketNumber.ts
```diff
@@ -26,18 +25,13 @@ export const getCaseMetadataByDocketNumber = async ({
...
-            }).toRawObject();
-          }) || [],
-      }
+    ? fromKyselyCase({
+        ...dbCase,
+        petitioners: dbCase.petitioners as TPetitioner[], // This is a hack because our typing for Petitioners is yucky
+        docketNumberWithSuffix: Case.getDocketNumberWithSuffix({
+          docketNumber: dbCase.docketNumber,
+          docketNumberSuffix: dbCase.docketNumberSuffix,
+        }),
```
Mwindo, line 34: We could put this in the mapper. Why didn't I? I don't have super strong feelings, but the way the mapper currently works is "here is some database data and random stuff please map it or let is pass through unchanged." `docketNumberWithSuffix` is not database data.

web-api/src/persistence/postgres/cases/getEligibleCasesForTrialCity.ts
```diff
@@ -57,44 +60,46 @@ export const getEligibleCasesForTrialCity = async ({
...
+        ...fromKyselyCase(c),
+        isSealed: !!c.isSealed,
+        irsPractitioners: c.irsPractitioners,
+        privatePractitioners: c.privatePractitioners,
+        docketNumberWithSuffix: Case.getDocketNumberWithSuffix({
+          docketNumber: c.docketNumber,
+          docketNumberSuffix: c.docketNumberSuffix,
+        }),
+      };
+    });
```
Mwindo, line 102: @TomElliottFlexion I'm not sure how you feel about the changes in this file, but it is more type-safe now with the new mapper.

web-api/src/persistence/postgres/cases/getEligibleCasesForTrialCity.ts
```diff
@@ -57,44 +60,46 @@ export const getEligibleCasesForTrialCity = async ({
...
       }),
     ]);
 
-    return purgeDynamoKeys({
+    const dynamoData = purgeDynamoKeys<
+      any,
+      {
+        irsPractitioners: IrsPractitioner[];
+        privatePractitioners: PrivatePractitioner[];
+      }
```
Mwindo, line 78: Basically saying, "Look, TypeScript: I know you _think_ these are each `TDynamoRecord[]` ... but I promise, they aren't!"

web-api/src/persistence/postgres/cases/mapper.ts
```diff
@@ -221,28 +107,72 @@ export const toKyselyNewCase = (rawCase: RawCase) => {
...
-    noticeOfTrialDate: caseRecord.noticeOfTrialDate?.toISOString(),
-    petitionPaymentDate: caseRecord.petitionPaymentDate?.toISOString(),
-    petitionPaymentWaivedDate:
-      caseRecord.petitionPaymentWaivedDate?.toISOString(),
-    receivedAt: caseRecord.receivedAt?.toISOString(),
-    sealedDate: caseRecord.sealedDate?.toISOString(),
-    trialDate: caseRecord.trialDate?.toISOString(),
-  };
-};
+export function fromKyselyCase<T extends object>(record: T) {
```
Mwindo, line 110: We need mapper functions from tables basically to solve some of what an ORM solves: the data is stored in format X in the database, but we need it in format Y in app code. 

Before, the main case mapper function was lying: we said we were returning `RawCase`, but often we weren't. We also planned to have a different mapper function for all use cases, but as more people were working with Postgres (and as I was fixing types so we were no longer lying), this seemed problematic (an explosion of mappers, often with redundant logic). The idea behind this new mapper is to 1) not lie and 2) transform case data and let specific use cases deal with massaging the data further for the type of object they need to construct.

I am also trying to think about ways we might be able to do the mapping independent of specific tables at all since a lot of the mapping is "convert this date type to a string."

web-api/src/persistence/postgres/cases/statistics/createCaseStatistic.ts
```diff
@@ -23,8 +27,22 @@ export const createCaseStatistic = async ({
...
         statisticId: statistic.statisticId,
         year: statistic.year ? parseInt(statistic.year) : null,
-        yearOrPeriod: statistic.yearOrPeriod,
+        yearOrPeriod: statistic.yearOrPeriod || null,
+        updatedAt: calculateDate({ dateString: formatNow() }),
       },
     ],
   });
+
+  await pgInsertInto({
```
Mwindo, line 36: Whoops, forgot to create the statistic's penalties when creating a statistic.

web-client/src/views/CaseDetail/Statistics.tsx
```diff
@@ -52,7 +52,7 @@ export const Statistics = connect(
           <div className="grid-row grid-gap flex-justify">
```
Mwindo, line 52: The diff below is confusing. I just added `data-testid`s.

web-client/src/views/AdvancedSearch/SearchDateRangePickerComponent.tsx
```diff
@@ -46,6 +46,16 @@ export const SearchDateRangePickerComponent = connect(
         'input[name="startDate-date-start"]',
       );
 
+      if (
```
Mwindo, line 49: A little bit of https://github.com/flexion/ef-cms/issues/10449

shared/src/business/entities/EntityConstants.ts
```diff
@@ -1821,6 +1821,9 @@ export const TROUBLESHOOTING_INFO = {
     'https://ustaxcourt.gov/dawson_faqs_case_management.html#FileUpload',
 };
 
+export const MAX_NUMBER_DEFICIENCY_STATISTICS = 12;
+export const MAX_NUMBER_DEFICIENCY_STATISTIC_PENALTIES = 10;
```
Mwindo, line 1825: Surely there is a better spot for these.

web-api/src/persistence/postgres/cases/mapper.ts
Mwindo: We are still currently lying in other places, but I think it is best to go back and fix that after this ticket.

shared/src/business/entities/cases/EligibleCase.ts
```diff
@@ -15,7 +15,7 @@ export class EligibleCase extends JoiValidationEntity {
   public caseType: string;
   public docketNumber: string;
   public docketNumberSuffix?: string;
-  public docketNumberWithSuffix?: string;
+  public docketNumberWithSuffix: string;
```
Mwindo, line 18: We always set this. See constructor.

web-api/src/persistence/postgres/cases/getEligibleCasesForTrialCity.ts
Mwindo: I should have said, "Please let me know how you feel about the changes in this file!" So pretend I said that :)

web-api/src/persistence/postgres/utils/migrate/migrations/0013-add-case-related-indices.ts
```diff
@@ -91,6 +91,12 @@ export async function up(db: Kysely<any>): Promise<void> {
     .column('closedDate')
     .execute();
 
+  await db.schema
+    .createIndex('idxCaseLeadDocketNumber')
+    .on('dwCase')
```
TomElliottFlexion, line 96: Thank you!

web-api/src/persistence/postgres/cases/getEligibleCasesForTrialCity.ts
TomElliottFlexion: I'll take your lead! I am fairly agnostic when it comes to "entities" here; I can see the value of having a `RawEligibleCase` that uses `fromKyselyCase`, but a proliferation of single use entities has downsides too

cypress/local-only/tests/integration/caseDetail/caseInformation/edit-case-detail.cy.ts
```diff
@@ -0,0 +1,327 @@
...
+      getButton('Statistics').click();
+      for (const statistic of getDeficiencyStatistics(docketNumber)) {
+        cy.contains('a', 'Add New Year/Period').click();
+        cy.get(`[data-testid="year-or-period-${statistic.yearOrPeriod}"]`);
+
+        if (statistic.yearOrPeriod === 'Period') {
+          cy.get('[data-testid="year-or-period-Period"]').click({
+            force: true,
+          });
+          // Get data-testid to work ... using id because we multiply the data-testid apparently
```
TomElliottFlexion, line 169: Well now I'm curious about this id multiplication

shared/src/business/entities/cases/Case.ts
```diff
@@ -109,7 +108,7 @@ export class Case extends JoiValidationEntity {
   public closedDate?: string;
   public createdAt: string;
   public docketNumber: string;
-  public docketNumberSuffix?: string;
+  public docketNumberSuffix?: string | null;
```
TomElliottFlexion, line 111: Forgive me if I'm misunderstanding: would this be handled by the mapper? I can see why it would be undefined, but not null

cypress/local-only/tests/integration/caseDetail/caseInformation/edit-case-detail.cy.ts
Mwindo: Ack, thank you for the reminder on this comment. I'm not sure why this was happening, but trying to select by data-testid gave me a "multiple things have this" error.

shared/src/business/entities/cases/Case.ts
Mwindo: The issue is here:

https://github.com/ustaxcourt/ef-cms/blob/fb8372c8911b56c7e086dc308dda5fb04950c8cf/shared/src/business/utilities/getDocketNumberSuffix.ts#L14

Probably should just return undefined instead of null, but I think that broke something and I said, "Ugh, let's deal with that later and just be honest here for now."

cypress/local-only/tests/integration/caseDetail/caseInformation/edit-case-detail.cy.ts
Mwindo: I think it is USWDS funkiness with DateSelector wrapping our input (the highlighted portion and the element above both have the data-testid):

Screenshot 2025-03-19 at 5 46 02 PM

web-api/src/persistence/postgres/cases/getCaseMetadataByDocketNumber.ts
TomElliottFlexion: Won't die on this hill, but counterpoint: `docketNumberWithSuffix` is required in 90% of usecases, shouldn't we make it difficult to forget to add in new functions?

TomElliottFlexion: I don't know, I keep going back and forth on this

TomElliottFlexion (approved):
This is a lot of good work, thank you!

cypress/local-only/tests/integration/caseDetail/caseInformation/edit-case-detail.cy.ts
TomElliottFlexion: I should have guessed...

shared/src/business/entities/cases/Case.ts
TomElliottFlexion: Fair enough!

cypress/local-only/tests/integration/caseDetail/caseInformation/edit-case-detail.cy.ts
```diff
@@ -0,0 +1,327 @@
...
+            penaltyType: PENALTY_TYPES.DETERMINATION_PENALTY_AMOUNT,
+            statisticId: 'cb557361-50ee-4440-aaff-0a9f1bfa30ed',
+            updatedAt: '',
+          },
+        ],
+      },
+    ];
+  };
+
+  describe('should change the things', () => {
```
zachrog, line 130: It SHOLD absolutely change the things.

Mwindo: Thank you ... definitely meant to update this. Will do that along with addressing other feedback.

cypress/local-only/tests/integration/caseDetail/caseInformation/edit-case-detail.cy.ts
```diff
@@ -0,0 +1,327 @@
...
+        getButton('Calculate and Save').click();
+
+        getButton('Calculate penalties as determined by Court').click();
+        for (let i = 0; i < courtDeterminationPenalties.length; i++) {
+          cy.get(`[data-testid="penalty_${i}"]`).type(
+            courtDeterminationPenalties[i].penaltyAmount,
+          );
+          getButton('Add another penalty').click();
+        }
+        getButton('Calculate and Save').click();
```
zachrog, line 209: No need to change it but just an FYI. We generally avoid the Page View Model style of getting elements like getButton() or getCaseCaptionText(). If it is a single selector then just do cy.get().

shared/src/business/useCases/getCasesForUserInteractor.ts
zachrog, line 1: I think line 59 is a dangerous type conversion. I think right now the best we can do is be honest about our types and fix the problems it causes.

web-api/src/persistence/postgres/cases/getCaseByDocketNumber.ts
```diff
@@ -62,7 +65,10 @@ export const getCaseByDocketNumber = async ({
     ),
   );
 
-  let consolidatedCases: RawCase[] = [];
+  let consolidatedCases: Omit<
```
zachrog, line 68: Okay fair warning I am going to be very picky about this function because it is both at the core of this PR and the core of our application. Overall know that you are doing a much better job than what existed before.

- The queryFull() can be added to the Promise.all() for speed.
- Because this function is meant to just be the data layer, it seems odd to me that we are throwing an error if we do not find the case we are looking for. Would you expect a SQL client to throw an error when running a query that returns no results? I recognize that this means ALL places that call getCaseByDocketNumber() need to check that a result was returned, but that seems common in a data layer. Maybe I am being too strict because in 98% of conditions where a case is not found we would throw an error anyway but I wanted to at least talk about it.
- Similarly as the last comment, this function is meant to be part of the data layer. Scrubbing the addresses of sealed consolidated cases seems like something our application should be in charge of. Should an entity be doing that? I am glad that you found too much data being returned from this function but I feel like we have pushed business rules into the data layer.

web-api/src/persistence/postgres/cases/getCaseMetadataByDocketNumber.ts
```diff
@@ -1,17 +1,16 @@
...
-} from '@shared/business/entities/contacts/Petitioner';
+import { Case } from '@shared/business/entities/cases/Case';
 
 export const getCaseMetadataByDocketNumber = async ({
   docketNumber,
 }: {
   docketNumber: string;
-}): Promise<RawCase | undefined> => {
+}): Promise<
+  | Omit<RawCase, 'consolidatedCases' | 'correspondence' | 'docketEntries'>
```
zachrog, line 11: This is what I expect most of our data layer functions to return. It returns either something or undefined.

web-api/src/persistence/postgres/cases/getCaseByDocketNumber.ts
Mwindo: 1. Good catch!
2. I agree. _Sigh._
3. Yes. That said, this feels like it is maybe outside the scope of the 10502 ticket. I think there are a few ways to handle this, and they have more to do with different entities/different data fetch patterns (e.g., maybe we don't even get the sensitive data in the first place) than with Dynamo->Postgres.

web-api/src/persistence/postgres/cases/getCaseMetadataWithCounsel.ts
```diff
@@ -10,26 +9,30 @@ export const getCaseMetadataWithCounsel = async ({
 }: {
   applicationContext: ServerApplicationContext;
   docketNumber: string;
-}): Promise<RawCase | undefined> => {
+}): Promise<
+  | Omit<RawCase, 'consolidatedCases' | 'correspondence' | 'docketEntries'>
+  | undefined
+> => {
   const caseMetaData = await getCaseMetadataByDocketNumber({ docketNumber });
```
zachrog, line 16: All of these calls could be made within a Promise.all()

You could still have the same signature of the function and return undefined if no caseMetaData

shared/src/business/useCases/getCasesForUserInteractor.ts
Mwindo: Agreed. The question is how deep to go in fixing dishonest types related to Case but not so much related to Dynamo->Postgres.

web-api/src/persistence/postgres/cases/getCaseByDocketNumber.ts
akuny: Broader architectural comment:

Re scrubbing the addresses, this is a good example of the sort of work that is suited to an intermediate layer such as a repository between the application-level interactors and the data access-level mapper. That way persistence functions only care about hitting the database and returning something or nothing, the repository handles data transformation and filtering and instantiation of entities, and the interactors can then just do their important work without having to worry about the validity of an entity because they will always-already know that the entities they receive from the repository are good to go.

akuny (approved):

web-api/src/persistence/postgres/utils/databaseToAppCodeMapper.ts
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * @example
+ * const mapper = new DatabaseToEntityMapper({ keyRenameMap, transformMap });
+ * const rawData = mapper.transform(data);
+ *
+ * @param keyRenameMap - An object that maps database keys (e.g., "caption") to application keys (e.g., "caseCaption").
+ * @param transformMap - An object that maps keys to transformation functions. Each function receives the original value,
+ *                       along with the full record if needed, and returns the transformed value.
+ */
+export class DatabaseToAppCodeMapper<
```
zachrog, line 22: No major points right now, it may just take time but right now my feeling is these mappers are overly complex both in implementation and types for what they are doing. Convert Date -> string, rename a variable, transform null to undefined. I applaud you for making a generic option to be used by others and I do not have a replacement in my mind right now.

I will continue to think about this as I work through WorkItems to see other options.

zachrog (changes requested):
Awesome work. I am requesting changes to talk through getCaseByDocketNumber.

web-api/src/persistence/postgres/cases/reports/fetchPendingItems.ts
```diff
@@ -60,6 +60,7 @@ export const fetchPendingItems = async ({
         'd.documentType',
         'd.documentTitle',
         'd.receivedAt',
+        'd.createdAt',
```
Mwindo, line 63: I think this is unnecessary 🤦

Mwindo:
Thank you all for the reviews! I am going to open a new PR based on all of the updates to 10502-dxox (the staging branch).

PR #5916, opened 2025-03-20
10607: Add custom instrumentation for measuring execution duration of…

En-8:
… sequences and actions; send to info cluster


En-8:
We decided as a group today that, given the discussions around reducing our reliance on Cerebral, using a Cerebral-specific approach to performance measurement probably doesn't make sense anymore. We feel that, between AWS RUM and API latency metrics, we currently have a good-enough view into the performance of the system.

PR #5920, opened 2025-03-21
10436 eligible cases refactored to test

TomElliottFlexion:


web-api/src/persistence/postgres/cases/getEligibleCasesForTrialSession.ts
```diff
@@ -0,0 +1,105 @@
...
+          ),
+        ]),
+      )
+      .limit(limit);
+
+    if (sessionType !== SESSION_TYPES.hybrid) {
+      query = query.where('procedureType', '=', sessionType);
+    }
+    return await query.execute();
+  });
```
En-8, line 70: This isn't a request for changes, merely an observation. It also is a decent example of something I've been pondering as we've been migrating things over to PostgreSQL: where should one draw the line on how much business logic resides in a database query?

Something I'm noticing in this refactor is that it seems as if the rules around what constitutes a case's eligibility to be included in any given trial session is now wholly baked into this single query. I can see how this could be helpful: now, instead of having the trial session catalog (which was ultimately, I think, a form of "derived state" that was persisted) without a centralized way to manage updates to case eligibility, we now have a single place to ask "hey, get me all the cases that are eligible to be a part of this trial session" AND it's computed as needed based on the most up-to-date state of the case.

However, the trade-off with our current approach is that now, since the rules defining eligibility are baked into the query, and we don't have a pattern for putting persistence functions under test, our rules for determining eligibility don't seem to be directly under test or a part of any "executable specification". Additionally, if the persistence model every changes again, these rules could very well change, too (or at least the way they're defined could change).

While it's precisely the thing we're refactoring away from, I can understand how this particular use case is one in which it could've been useful to store the derived state of "is this eligible for this trial session", but perhaps with a better abstraction for managing updates to that persisted state.

Again, I don't have any suggested changes or anything, only sharing my musings.

web-api/src/persistence/postgres/cases/getFullEligibleCasesForTrialSession.ts
```diff
@@ -0,0 +1,95 @@
...
+            ),
+          ),
+        ]),
+      )
+      .limit(limit);
+
+    if (sessionType !== SESSION_TYPES.hybrid) {
+      query = query.where('procedureType', '=', sessionType);
+    }
+    return await query.execute();
```
En-8, line 59: Are the columns included in the query the only thing that differs between this and the other persistence function for getting eligible cases? To my previous comment, our rules for case eligibility are now baked into these queries -- is there some way we can share those rules between the two queries so that if those rules change, we don't have to remember to change them in both places?

E.g. maybe we there's some way to dynamically choose between the two different column sets (all vs. subset)?

Curious your thoughts

En-8 (approved):
Nothing I think that warrants blocking going to test: one thought and one related comment I'm curious for thoughts on.

web-api/src/persistence/postgres/cases/getEligibleCasesForTrialSession.ts
TomElliottFlexion: This is interesting, and I don't think I disagree! I'd be more than happy to discuss

web-api/src/persistence/postgres/cases/getFullEligibleCasesForTrialSession.ts
TomElliottFlexion: There is, it was just breaking type safety. Likely worth spending more time finding a workaround

akuny (approved):

PR #5925, opened 2025-03-21
PdfJS-Dist Upgrade, React V19, Cerebral Fork. Update Extravaganza

zachrog:
# In This PR
- Upgrade react to 19 (In test for 1 month)
- Fork @cerebral/react and use modified version which is compatible with react v19 (In test for 1 month)
- remove 'react-quill' and bring in-house to make minor adjustments and allow it to work with react v19 (In test for 1 month)
- Upgrade pdfjs-dist from v2 -> v5
- Use direct import for pdfjs-dist instead of appContext

[successful deploy](https://app.circleci.com/pipelines/github/flexion/ef-cms/61140/workflows/aee4e94c-d8ff-477b-a577-d7f1fb8180df)


docs/dependency-updates.md
```diff
@@ -29,9 +29,6 @@ This command informs us of known security vulnerabilities. If transitive depende
 > **Why am I seeing a medium severity for `quill`?**
```
akuny, line 29: With the removal of `react-quill`, I think we can remove this chunk of text too.

akuny (approved):

Mwindo (changes requested):
@zachrog, this is great overall! Thank you!

Unfortunately, I do see a difference in behavior in validation. I am confused as to why the test in upload-court-issued-pdf-validation.cy.ts did not catch this.

Here is current staging with a password-protected pdf:

Screenshot 2025-03-22 at 8 40 13 AM

Here is the result in this branch (generic error message):

Screenshot 2025-03-22 at 8 45 10 AM

I would like the pdfjs-dist part of this PR to live on test so we can figure out the issue and make sure there are not others.

akuny:
> @zachrog, this is great overall! Thank you!
> 
> Unfortunately, I do see a difference in behavior in validation. I am confused as to why the test in upload-court-issued-pdf-validation.cy.ts did not catch this.
> 
> Here is current staging with a password-protected pdf:
> 
> Screenshot 2025-03-22 at 8 40 13 AM
> Here is the result in this branch (generic error message):
> 
> Screenshot 2025-03-22 at 8 45 10 AM
> I would like the pdfjs-dist part of this PR to live on test so we can figure out the issue and make sure there are not others.

Where in the UI were you able to make the incorrect modal appear? This morning I tested out "Upload PDF" as docket clerk locally with a password-protected PDF to recreate the issue and the correct message was displayed. I could well be missing something.

Mwindo:
@akuny @zachrog I am trying to re-create the error this morning, and now I am seeing the correct modal. I am confused as to why I saw what I saw (it was the PDF Upload option on a case, but the validation logic is the same across all upload components). I will investigate a little further and then un-request changes if I cannot re-create. Apologies for the confusion!

Mwindo (approved):
Hmm. I cannot re-create the issue I found earlier. Chalk it up to end-of-week brain I suppose? I'm sorry :/

I still think my preference would be for pdfjs-dist to live in test for a bit before going into staging, but I will leave that up to you. Thanks for all of the great work here!

docs/dependency-updates.md
zachrog: Unfortunately we cannot. react-quill used quill as a dependency and it was quill that had a medium security vulnerability. 

In this PR we started to use quill directly, but we kept the exact same version that react-quill was using meaning that we still have the same medium security vulnerability. In the future we can try to upgrade quill and see if we can get the security vulnerability to go away, but that is for another update.

jimlerza (approved):

PR #5928, opened 2025-03-25
10502: Some query optimization

Mwindo:
1. Avoid `selectAll` for `dwCase`, which is typically frowned upon anyway. Because the case table is so big, this seems to have a non-trivial impact on performance. (Example: I could not get `16017-21` to load before locally when pointed to test. Now I can.)
2. Avoid race conditions when generating a new docket number by using a mutex lock. (We can explore other solutions in the meantime, like offloading the sequence generation to the database, although there seem to be complexities with this approach.)

### DIFF
--- a/web-api/src/persistence/postgres/caseCorrespondences/getCaseCorrespondenceByDocketNumber.ts
+++ b/web-api/src/persistence/postgres/caseCorrespondences/getCaseCorrespondenceByDocketNumber.ts
@@ -10,7 +10,6 @@ export const getCaseCorrespondenceByDocketNumber = async ({
   const correspondence = await getDbReader(reader =>
     reader
       .selectFrom('dwCaseCorrespondence as cc')
-      .leftJoin('dwCase as c', 'c.docketNumber', 'cc.docketNumber')
       .where('cc.docketNumber', '=', docketNumber)
       .selectAll()
       .select('cc.docketNumber')
--- a/web-api/src/persistence/postgres/cases/generateDocketNumber.ts
+++ b/web-api/src/persistence/postgres/cases/generateDocketNumber.ts
@@ -4,9 +4,7 @@ import {
   formatNow,
   getMonthDayYearInETObj,
 } from '@shared/business/utilities/DateHandler';
-import { getCaseMetadataByDocketNumber } from '@web-api/persistence/postgres/cases/getCaseMetadataByDocketNumber';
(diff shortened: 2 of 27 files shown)


web-api/src/persistence/postgres/caseCorrespondences/getCaseCorrespondenceByDocketNumber.ts
```diff
@@ -10,7 +10,6 @@ export const getCaseCorrespondenceByDocketNumber = async ({
   const correspondence = await getDbReader(reader =>
     reader
       .selectFrom('dwCaseCorrespondence as cc')
-      .leftJoin('dwCase as c', 'c.docketNumber', 'cc.docketNumber')
```
Mwindo, line 13: No need to join. All the info we need is in `dwCaseCorrespondence`.

web-api/src/persistence/postgres/cases/getCasesMetadataWithCounselByLeadDocketNumber.ts
```diff
@@ -9,16 +9,16 @@ export const getCasesMetadataWithCounselByLeadDocketNumber = async ({
   applicationContext: ServerApplicationContext;
   leadDocketNumber: string;
 }): Promise<RawCase[]> => {
-  const dbCases = await getDbReader(reader =>
+  const dbCaseData = await getDbReader(reader =>
```
Mwindo, line 12: Only _slightly_ more rigorous than "anecdotal" (as it is based on a handful of elapsed time stamps being run against test locally): this seems to be speeding up `getCaseByDocketNumber` for consolidated cases by 1+ seconds for some of the cases I have been experimenting with (15631-22, 16017-21, 16017-22, 12235-20).

web-api/src/persistence/postgres/cases/generateDocketNumber.ts
```diff
@@ -50,32 +47,5 @@ export const generateDocketNumber = async ({
     ? formatDateString(receivedAt, FORMATS.YEAR)
     : formatNow(FORMATS.YEAR);
 
-  let attempt = 0;
```
Mwindo, line 53: If we go the mutex route, we guarantee only one process is generating a docket number at a time, so the previous attempt to handle concurrency is no longer necessary.

pixiwyn (approved):

PR #5930, opened 2025-03-25
10474: Practitioner Search Filters

nechama-krigsman:


web-client/src/presenter/actions/AdvancedSearch/getPractitionersByNameAction.ts
```diff
@@ -1,3 +1,4 @@
+import { ALL_SELECTION } from '@shared/business/entities/cases/CaseSearch';
```
Mwindo: This constant feels like it is in a funny place now (we are searching for practitioners, not cases), but I'm not sure exactly where else to put it at the moment.

web-client/src/presenter/actions/AdvancedSearch/clearAdvancedSearchFormAction.ts
```diff
@@ -39,6 +40,7 @@ export const clearAdvancedSearchFormAction = ({
   if (formType === 'practitionerSearchByName') {
     defaultForm.lastKeysOfPages = [];
     defaultForm.total = 0;
+    defaultForm.practitionerType = 'All';
```
Mwindo: Any reason not to use the constant here?

web-client/src/presenter/actions/AdvancedSearch/updateAdvancedSearchFormAction.ts
```diff
@@ -5,12 +5,29 @@ export const updateAdvancedSearchFormAction =
...
+    isArray?: boolean;
   }>) => {
     const formType = formName || props.formType;
+
+    if (props.isArray) {
+      const currentValue =
+        get(state.advancedSearchForm[formType][props.key]) || [];
+      const updatedValue = currentValue.includes(props.value)
+        ? currentValue.filter((value: string) => value !== props.value)
+        : [...currentValue, props.value];
```
Mwindo: I know that the trial sessions filters call a separate sequence entirely (`setTrialSessionsFiltersSequence`). I wonder if that is worth doing in this case, too? 

It might make things more readable: we would know what state is in this specific form, we would avoid a boolean flag, and we could encapsulate certain assumptions we make in the form. Here, for instance, the assumption is that we want to toggle, i.e., if we re-add the same value, we should actually remove it. One might just as easily expect the value to be included just once, e.g., `[...new Set(blah blah)]`. This is a nitpick; I am just thinking out loud.

web-client/src/views/AdvancedSearch/PractitionerSearchByName.tsx
```diff
@@ -66,7 +75,197 @@ export const PractitionerSearchByName = connect(
...
+                  <SelectSearch
+                    aria-labelledby="admission-status-label"
+                    data-testid="admission-status-filter"
+                    inputId="admission-status"
+                    name="admissionStatus"
+                    options={ADMISSIONS_STATUS_OPTIONS.map(admissionStatus => ({
+                      label: admissionStatus,
+                      value: admissionStatus,
+                    }))}
+                    placeholder="- Select one or more -"
```
Mwindo: We use this in a few places now. I wonder if it is worth pulling out as a constant.

web-client/src/styles/custom.scss
```diff
@@ -2458,6 +2458,27 @@ button.change-scanner-button {
   background-color: #ffe396;
 }
 
+.display-gap {
```
Mwindo: Can we remove this and use a utility class instead? `gap-4` seems about right.

Mwindo (approved):
Thanks @nechama-krigsman! I've left a few small comments. The largest change: I think I would prefer the frontend state to be managed more like it is in the trial session filters so that the state is well-defined (which has the added bonus that it would be easier to port away from Cerebral) and we avoid a boolean flag. That said, I will leave this up to you.

PR #5931, opened 2025-03-25
10616 Bug: Navigating Back to Case Detail from Edit Petitioner

akuny:
Update the `editPetitionerInformationHelper` computed so that it handles undefined properties on the state tree without erroring out.


Mwindo (approved):
Thanks @akuny!

TomElliottFlexion (approved):

nechama-krigsman (approved):

PR #5933, opened 2025-03-26
10502: Use the existing locking mechanism

Mwindo:
Replace the mutex lock created in #5928 with our existing lock.


web-api/src/business/useCases/createCaseFromPaperInteractor.ts
```diff
@@ -322,24 +321,32 @@ export const createCaseFromPaperInteractor = async (
...
-          corporateDisclosureFileId,
-          petitionFileId,
-          petitionMetadata,
-          requestForPlaceOfTrialFileId,
-          stinFileId,
-          user,
-        },
-        authorizedUser,
-      );
+  await acquireLock({
```
zachrog, line 324: I think you probably want a try catch lock around createCaseMetadata as you want to release the lock if it throws an error

web-api/src/business/useCases/createCaseInteractor.ts
```diff
@@ -316,22 +315,30 @@ export const createCaseInteractor = async (
...
-          attachmentToPetitionFileIds,
-          corporateDisclosureFileId,
-          petitionFileId,
-          petitionMetadata,
-          stinFileId,
-          user,
-        },
-        authorizedUser,
-      );
+  await acquireLock({
```
zachrog, line 318: Here too

pixiwyn (approved):

PR #5934, opened 2025-03-26
Fix flaky cypress test

zachrog:
- Test used a random first name which would occasionally be a very common name and cause the test to break. Fixed


nechama-krigsman (approved):

jimlerza (approved):

TomElliottFlexion (approved):
Thank you!

PR #5940, opened 2025-03-27
10540: Fix Calculate Penalties on IRS Notice link for cases that have pre-existing penalties/statistics

En-8:
When checking whether or not to render "confirmation text", we weren't first checking that the "statistics" array existed before attempting to access a value by index.

flexion#10540


Mwindo (approved):
Thanks @En-8!

jimlerza (approved):

PR #5942, opened 2025-03-27
10619 Bug: Filtering by Judge on Case Deadlines Report

akuny:
See [10619](https://github.com/flexion/ef-cms/issues/10619). The root cause of this issue was the fact that some judges had varying, inconsistent, values in the case deadlines table in persistence: because the case deadline report had been filtering by judge name, this led to inconsistencies in the report.

This PR uses `associatedJudgeId` instead of `associatedJudge` to filter the case deadline report.

### DIFF
--- a/web-api/src/persistence/postgres/caseDeadlines/getCaseDeadlinesByDateRange.ts
+++ b/web-api/src/persistence/postgres/caseDeadlines/getCaseDeadlinesByDateRange.ts
@@ -71,7 +71,11 @@ export const getCaseDeadlinesByDateRange = async ({
 };
 
 const applyJudgeFilter = (query, judgeId) => {
-  return !!judgeId || judgeId === null
-    ? query.where('associatedJudgeId', '=', judgeId)
-    : query;
+  if (judgeId === null) {
+    return query.where('associatedJudgeId', 'is', null);
+  } else if (judgeId !== undefined) {
+    return query.where('associatedJudgeId', '=', judgeId);
+  }
+
+  return query;
 };
--- a/web-api/src/persistence/postgres/caseDeadlines/getCaseDeadlinesByDateRange.ts
+++ b/web-api/src/persistence/postgres/caseDeadlines/getCaseDeadlinesByDateRange.ts
@@ -46,10 +46,10 @@ export const getCaseDeadlinesByDateRange = async ({
       const results = await deadlineQuery.execute();
 
       let countQuery = reader
-        .selectFrom('dwCaseDeadline')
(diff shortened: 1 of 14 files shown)


web-client/src/views/CaseDeadlines/CaseDeadlines.tsx
```diff
@@ -131,7 +131,6 @@ export const CaseDeadlines = connect(
                     totalPages={caseDeadlineReportHelper.pageCount}
                     onPageChange={pageChange => {
                       setActivePage(pageChange);
-                      console.log('pageChange', pageChange);
```
Mwindo, line 134: TY!

web-api/src/persistence/postgres/caseDeadlines/getCaseDeadlinesByDateRange.ts
```diff
@@ -67,3 +69,9 @@ export const getCaseDeadlinesByDateRange = async ({
     totalCount,
   };
 };
+
+const applyJudgeFilter = (query, judgeId) => {
+  return !!judgeId || judgeId === null
+    ? query.where('associatedJudgeId', '=', judgeId)
```
Mwindo: I am worried about this. SQL wants `IS NULL` and not `= NULL`. I don't know if Kysely accounts for this?

For example, here are the results (truncated in screen 2) I get if I run = NULL:

Screenshot 2025-03-28 at 9 25 48 AM

And is NULL:

Screenshot 2025-03-28 at 9 25 58 AM


Even if it works with Kysely, it is probably safer to do `=` if `judgeId !== null` and `is` if `judgeId === null`.

Mwindo (approved):
Thank you for handling this @akuny!

web-client/src/presenter/computeds/caseDeadlineReportHelper.ts
```diff
@@ -31,14 +31,17 @@ export const caseDeadlineReportHelper = (
...
-    .map(i => applicationContext.getUtilities().formatJudgeName(i.name))
-    .concat(CHIEF_JUDGE)
-    .sort();
+  const judgeOptions = (get(state.judges) || [])
+    .map(judge => ({
+      id: judge.userId,
+      name: applicationContext.getUtilities().formatJudgeName(judge.name),
+    }))
+    .concat({ id: CHIEF_JUDGE, name: CHIEF_JUDGE })
+    .sort((a, b) => a.name.localeCompare(b.name));
```
Mwindo, line 44: 👍

TomElliottFlexion (approved):

zachrog (approved):

PR #5954, opened 2025-04-01
10502/10436: Case updates

TomElliottFlexion:
- Test environment should be on 0008-add-minute-sheet before deployment. 

We expect the kysely migration check to fail because we are renaming some migrations.


akuny (approved):
This all looks good to me. My only comment is that `DatabaseToAppCodeMapper` and `transformNullToUndefined` are both important enough and testable enough to merit having unit test coverage in place.

PR #5956, opened 2025-04-02
10576: New Tests

cruzjone-flexion:


cypress/local-only/tests/integration/caseDeadline/case-deadline-on-a-consolidated-group.cy.ts
```diff
@@ -0,0 +1,164 @@
...
+      loginAsDocketClerk1();
+      goToCase(leadDocketNumber);
+      updateCaseStatus(CASE_STATUS, JUDGE_NAME);
+      cy.get<GroupInfoType>('@GROUP_INFO').then(GROUP_INFO => {
+        GROUP_INFO.leadDocketNumber = leadDocketNumber;
+        cy.wrap(GROUP_INFO).as('GROUP_INFO');
+      });
+    });
+
+    for (let index = 0; index < 4; index++) {
```
zachrog: You probably already saw it but I believe we have a create consolidatedGroup helper function. Wanted to see if it was necessary to create all of these cases for your test. This is creating 5 cases for the setup of the tests which seems like a lot, but may be necessary for your consolidation test.

cypress/local-only/tests/integration/caseDeadline/case-deadline-on-a-consolidated-group.cy.ts
```diff
@@ -0,0 +1,164 @@
...
+        const docketNumber = memberDocketNumbers[index];
+        goToCase(docketNumber);
+        cy.get('[data-testid="tab-tracked-items"]').click();
+        cy.get('.case-deadline-row').should('have.length', 1);
+
+        cy.get('[data-testid="case-deadline-description"]').should(
+          'have.text',
+          'John Is Testing',
+        );
+      }
```
Mwindo, line 61: Minor, but I think we can do this all in the same loop.

Mwindo: Meant to select the preceding `cy.get('[data-testid="tab-tracked-items"]').click();` too, but you get the idea.

web-client/src/presenter/actions/CaseDeadline/getConsolidatedCaseDeadlineAction.test.ts
```diff
@@ -0,0 +1,42 @@
...
+
+describe('getConsolidatedCaseDeadlineAction', () => {
+  const TEST_CASE_DEADLINE_ID = getUniqueId();
+  const TEST_INTERACTOR_RESULTS = 'TEST_INTERACTOR_RESULTS';
+
+  presenter.providers.applicationContext = applicationContext;
+
+  applicationContext
+    .getUseCases()
+    .getConsolidatedCaseDeadlinesInteractor.mockReturnValue(
```
Mwindo: Minor: can we make this mockResolvedValue?

zachrog (approved):

PR #5983, opened 2025-04-15
10586: fix inline validation to test

JayFlexy:


shared/src/business/entities/MotionOrderResponseForm.ts
```diff
@@ -49,7 +48,7 @@ export class MotionOrderResponseForm extends JoiValidationEntity {
         'string.max': 'Limit is 240 characters.',
       }),
     issueOrderFor: joi
-      .when('isOnLeadCase', {
+      .when('isLeadCase', {
         is: joi.equal(true),
```
TomElliottFlexion: Good catch

TomElliottFlexion (approved):

shared/src/business/entities/MotionOrderResponseForm.ts
TomElliottFlexion: My bad, I misread. This still needs to be `isOnLeadCase`: see line 31

PR #5987, opened 2025-04-17
10623 Anonymize email addresses during postgres dump & restore prod->test

En-8:
flexion#10623


scripts/postgres/restoreDbFromSource.ts
```diff
@@ -75,8 +76,11 @@ async function main() {
     username: sourceUsername,
   });
 
+  const sanitizedFileName = `sanitized-${backUpFileName}`;
+  replaceEmailAddresses(backUpFileName, sanitizedFileName);
```
Mwindo: I'm not sure this really matters, but I suspect there is a way to do all the redaction in memory rather than storing the unsanitized data at all, something like removing the `-f` flag in `pg_dump` (so that `pg_dump` uses STDOUT) and then piping through the redactor function. (I realize this is a draft, I was just curious 😄 .)

nechama-krigsman (approved):

jimlerza (approved):

PR #5994, opened 2025-04-18
Return fake host string when running Jest tests;

cruzjone-flexion:
When running tests locally and having your Postgres DB running, test will connect to DB;
This caused an issue for me where I could not replicate an error locally that I was getting in PR;

When Jest is running "NODE_ENV" is set to "test", when that is the case we provide a random message as host and it will display when a unit test has not mocked out a Postgres method; 
image


jimlerza (approved):

Mwindo (approved):
Thanks @cruzjone-flexion!

PR #5996, opened 2025-04-21
10612: fix infinite spinner on edit petitioner after removing email, v2

En-8:
flexion#10612

The previous attempt at this fix was a naive attempt at relying on the front end reloading "case details" to ensure the petitioner information in state was up-to-date. This worked consistently locally and when hitting us-east, but was an inconsistent fix when hitting us-west. Current thought is that this was due to replication lag being sometimes longer than it took for the front end to reload case details state.

This fix changes the approach such that the API returns the updated state of the petitioner for the front end to use to set that petitioner's state, rather than relying on read-after-write.


TomElliottFlexion (approved):
This is the way

PR #6005, opened 2025-04-25
10502 Cases: Updates

Mwindo:
- Use settlePromises to avoid interruption of in-flight promises
- validate _before_ setting the trial session
- Fix sort order of consolidated cases when adding/removing a member case
- Prefer ?? to || in mapper files (minus one or two exceptions)


zachrog (approved):

web-api/src/business/useCases/trialSessions/setTrialSessionCalendarInteractor.ts
```diff
@@ -164,10 +173,10 @@ export const setTrialSessionCalendarInteractor = async (
     // If firing all at once, we exhaust the available connections and will run into connection timeouts.
     const chunkedFunctions = chunk(funcs, CHUNK_SIZE);
     for (const singleChunk of chunkedFunctions) {
-      await Promise.all(singleChunk.map(func => func()));
+      await settlePromises(singleChunk.map(func => func()));
```
Mwindo, line 176: @jimlerza We discovered that in a Lambda, a failure with Promise.all freezes the event loop, and any promises that had been started are paused. The event loop is unfrozen if the Lambda is re-invoked while warm; otherwise, "in-flight" promises that were on the event loop are never resolved. In a node script, the in-flight promises are never resolved at all.

This can lead to data inconsistency: an error on _any_ promise execution within `Promise.all` can cause any other promises to fail in arbitrary spots--for instance, in the middle of `updateCaseAndAssociations`. (Example: Maybe we just added a docket entry but haven't added the work item for it when the event loop is interrupted.) This was already an issue before 10502, but it was caught (and exacerbated) because in 10502 we have separated petitioners on the case and case statistics into their own tables. In our specific situation, a failure occurred when we ran a script, and some in-flight promises had just cleared statistics before reuploading them, leading to bad data in test. (Q: Why do we _clear_ and reupsert at all?? A: It is stupid, but we currently need to do this in order to properly mimic existing behavior.)

In the future, this line of code (and similar places) should be wrapped in a transaction. Since some data is in Postgres and some in Dynamo, that is not practical currently.

Waiting for all the promises to settle does not mean we avoid data inconsistency--we need the transaction for that. However, waiting for all promises to settle _does_ at least markedly improve things: the failure is at least not entirely _arbitrary_. In-flight promises will never be interrupted midway (unless the execution environment just crashes). In `setTrialSessionCalendarInteractor`, for instance, any failure in the promises means that the relationship between trial session and updated cases is not correct because we never get to the code that updates the trial session (some cases will say, "Ok, I am associated with a trial session," but the trial session will say, "No you ain't!"). Nevertheless, a failure now at least does _not_ lead to a partially updated case, which was possible before, and the subset of failures is now more easy to reason about. (E.g., maybe only 1 case update among 100 failed, in which situation you know the other 99 have been updated. Without that knowledge, you would have a harder time determining which subset of cases succeeded and which failed.)

Waiting for all the promises to settle _does_ mean we lose any advantages of a fast-failure (which Promise.all affords), but we thought data consistency was more important.

PR #6013, opened 2025-04-29
Add some error catching to avoid potential re-indexing failure in the glue job

Mwindo:
We have had failures in an entire glue job because, e.g., one data point is off while re-indexing. Rather than fail the entire glue job, we want to log the error.

This PR adds a try/catch with a consistent error message to look for in:

- processDocketEntries
- processCaseEntries
- processPractitionerMappingEntries


web-api/src/business/useCases/processStreamRecords/processCaseEntries.ts
```diff
@@ -5,6 +5,7 @@ import { upsertCaseStatusUpdates } from '@web-api/persistence/postgres/cases/ups
 import { Statistic } from '@shared/business/entities/Statistic';
 import { unmarshall } from '@aws-sdk/util-dynamodb';
 import { settlePromises } from '@web-api/utilities/settlePromises';
+import { getLogger } from '@web-api/utilities/logger/getLogger';
```
Mwindo, line 8: The diffs look more confusing than they are. I have just wrapped everything in each non-test file in a try/catch, which throws off indentation.

jimlerza (approved):

PR #6020, opened 2025-05-05
Fix Query Param Arrays; 

cruzjone-flexion:
There was a dep update that upgraded express which changed some default settings;
We set those default settings for the internal app but forgot to do so on the public app;

Breaking how we receive arrays from the query parameters;


nechama-krigsman (approved):

Mwindo (approved):
Thank you!

jimlerza (approved):

PR #6022, opened 2025-05-06
10502 dxox intermediate branch to test 1746566138

zachrog:
This PR moves caseStatusHistory, petitioners, and statistics all back onto the case table to increase performance, and reduce data loss


Mwindo (approved):
This looks reasonable!

PR #6025, opened 2025-05-08
10676: Follow-up to 10502

Mwindo:
This ticket is for cleaning up stuff after 10502 is deployed, and changes related to that:

**TL;DR**: Remove unnecessary code, refactor updateCaseAndAssociations to be more type safe and easy-to-follow, and always get case data via `getCasesByDocketNumbers`. No more random subsets of case data gotten in different ways--hooray!

---

**Details**: 
- Removing the `processCaseEntries` indexing code
- Removing `aggregateCaseItems` code
- Adding an `index-cases.ts` script
- Refactoring persistence functions so that there are fewer (see notes below)
- Refactoring `updateCaseAndAssociations`. We identified a fairly serious, long-standing bug--promises not awaited quite right--that we were just lucky not to be running into in practice. The refactor exposed it, maybe because we were querying more efficiently. See [this PR comment](https://github.com/ustaxcourt/ef-cms/pull/6025/files#r2132531317). In the meantime, we patched this bug in staging [here](https://github.com/ustaxcourt/ef-cms/pull/6108).
- Replacing `documentTitleJourney.test.ts` (cerebral) with `document-title.cy.ts` (cypress) (which had been failing due to the above)
- Getting rid of `applicationContext` argument in `getCaseByDocketNumber` and `updateCaseAndAssociations`.

**About the `getCasesByDocketNumbers` refactor**: 
In `getCaseByDocketNumber`, we have a call to `getCasesMetadataWithCounselByLeadDocketNumber`. This latter function is inefficient: it fetches several records in a loop rather that bulk fetching from Postgres. In the process of making it better, it seemed best to get rid of persistence functions that are all sort of doing the same thing to avoid repeating logic. To that end, this PR:

- Removes `getCaseMetadataWithCounsel`, `getCasesMetadataByDocketNumbers`, `getCasesInConsolidatedGroup`, `getCasesByLeadDocketNumber`, `getCasesMetadataWithCounselByLeadDocketNumber`
- Instead, relies on a more dynamically typed `getCasesByDocketNumbers` and `getConsolidatedCases` (which uses getCasesByDocketNumbers under the hood in most situations)
- The more we can simplify as the project switches hands, the better!


cypress/deployed-and-local/integration/myAccount/change-address.cy.ts
```diff
@@ -13,7 +13,6 @@ describe('change of address', () => {
   it('changing the address of a private practitioner should generate NCA and update their cases', () => {
     const newAddress = faker.location.streetAddress();
     loginAsPrivatePractitioner('privatePractitioner2@example.com');
-    loginAsPrivatePractitioner('privatePractitioner2@example.com');
```
Mwindo, line 16: Duplicate login was probably a merge issue.

cypress/local-only/tests/integration/documentQC/document-title.cy.ts
```diff
@@ -0,0 +1,100 @@
+import { attachFile } from '../../../../helpers/file/upload-file';
```
Mwindo, line 1: This file replaces `documentTitleJourney.test.ts`

web-api/src/persistence/postgres/cases/getCaseByDocketNumber.ts
```diff
@@ -22,76 +19,58 @@ export const getCaseByDocketNumber = async ({
...
+  return {
+    ...theCase,
     consolidatedCases: consolidatedCases.map(
       c => new ConsolidatedCaseSummary(c),
     ),
-  });
+  };
 };
+
+// This function is super stupid. We need it because certain spots
```
Mwindo, line 56: Before, we called `aggregateCaseItems`. We have removed that because it is no longer necessary. (The aggregation of case data is being taken care of in `getCasesByDocketNumbers`.) The only thing it did that we still need is associating work items with docket entries. We could do this more efficiently than an `O(n^2)` loop, but I don't think the performance impact is very high, and hopefully this is temporary anyway.

web-api/src/persistence/postgres/cases/getCasesByDocketNumbers.ts
```diff
@@ -14,37 +14,82 @@ import { fromKyselyCase } from '@web-api/persistence/postgres/cases/mapper';
 import { CaseKysely } from '@web-api/persistence/postgres/cases/schema';
 import { difference, isEmpty, partition, sortBy } from 'lodash';
 
-export async function getCasesByDocketNumbers({
+export const ALL_OMITTABLE_CASE_FIELDS = [
```
Mwindo, line 18: We started down _including_ fields, which might be more intuitive, rather than _excluding_ fields. We didn't finish, and we think this is still a good improvement.

web-api/src/persistence/postgres/cases/getConsolidatedCases.ts
```diff
@@ -0,0 +1,59 @@
...
+export const getConsolidatedCases = async <
+  T extends OmittableCaseFields[] = [],
+>({
+  leadDocketNumber,
+  excludeFields,
+}: {
+  leadDocketNumber: string;
+  excludeFields?: T;
+}): Promise<Omit<RawCase, 'consolidatedCases' | T[number]>[]> => {
+  // If we only need case metadata, make one query for all of it
```
Mwindo, line 19: The tradeoff:

1. Slightly simpler code (always make two db requests, one to get docket numbers and one to `getCasesByDocketNumbers`) that is slightly less efficient sometimes
2. Slightly more complex code (only make one db query in cases we can do so) that is slightly more efficient sometimes

Implemented here is option 2.

web-api/src/business/useCaseHelper/caseAssociation/updateCaseAndAssociations.ts
```diff
@@ -371,89 +407,11 @@ const updateCaseDeadlines = async ({
...
-  // wait for all validation tasks to complete and for callbacks to be generated
-  const persistenceCallbacks = (await Promise.all(validationRequests)).flat();
-
-  
-  // Persist primary case data first to ensure no errors
-  await upsertCases([validNewRawCaseEntity]);
-  
-  // Then persist related data
-  // all validation has passed, so now execute all persistence callbacks from results
-  const persistenceRequests = persistenceCallbacks.map(persistFn => {
```
Mwindo, line 401: This is a bug. We should be returning the `persistFn()`. `persistenceRequests` is just an array of `undefined`, so we weren't actually awaiting anything below.

Mwindo: This has been the case for 4 years, based on git diff.

shared/src/business/useCases/getCaseDeadlinesInteractor.ts
```diff
@@ -54,8 +54,9 @@ export const getCaseDeadlinesInteractor = async (
   const validatedCaseDeadlines =
     CaseDeadline.validateRawCollection(foundDeadlines);
 
-  const associatedCases = await getCasesMetadataByDocketNumbers({
+  const associatedCases = await getCasesByDocketNumbers({
     docketNumbers: validatedCaseDeadlines.map(item => item.docketNumber),
+    excludeFields: ['docketEntries', 'hearings', 'correspondence'],
```
jimlerza, line 59: 😍

web-api/elasticsearch/index-cases.ts
jimlerza, line 1: This is great!

web-api/src/business/useCaseHelper/caseAssociation/updateCaseAndAssociations.ts
jimlerza, line 1: Excellent refactor. Thanks for moving `updateCaseAndAssociations` to the top. This will be a lot easier to maintain, and it should get a lot slimmer eventually.

web-api/src/persistence/dynamo/documents/updateDocketEntry.ts
jimlerza, line 1: Huh. Good catch

jimlerza (approved):

PR #6065, opened 2025-05-20
chief judge script: remove cognito updates

TomElliottFlexion:
Since we no longer rely on Cognito to store judge titles, this script no longer needs to update Cognito for the Chief Judge change on June 1st. This has been tested in `test`


Mwindo (approved):
Zach + Chris Review: We don't think that this will ever come up but wanted to give a heads up that the Auth token that we use in requests comes from cognito and holds the users name in it. So the AuthUserType which has userId, email, and name will still have the old name. This probably will not matter as we do not use AuthUser.name much but wanted to give a heads up.

jimlerza:
If we're not going to remove the name field from Cognito, then why are we removing the logic that updates it?

PR #6069, opened 2025-05-22
10506: Move Notifications and Jobs to Postgres

pavanyenduri:


web-api/src/business/useCases/user/generateChangeOfAddressForPractitioner.irsPractitioners.test.ts
```diff
@@ -75,7 +75,7 @@ describe('generateChangeOfAddress', () => {
 
     applicationContext
       .getPersistenceGateway()
-      .setChangeOfAddressCaseAsDone.mockReturnValue({ remaining: 0 });
+      .setChangeOfAddressCaseAsDone.mockReturnValue([{ remaining: 0 }]);
```
Mwindo: Can we do `mockResolvedValue`?

Mwindo: Here and where relevant below?

web-api/src/persistence/postgres/connections/schema.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export type ConnectionTable = typeof connectionTableDefinition;
+
+export const DW_CONNECTION_COLUMNS = Object.keys(
+  connectionTableDefinition,
+) as Array<keyof ConnectionTable>;
+
+export type ConnectionKysely = Selectable<ConnectionTable>;
+export type NewConnectionKysely = Insertable<ConnectionTable>;
+export type UpdateConnectionKysely = Updateable<ConnectionTable>;
```
Mwindo: We don't use these types anywhere, which seems suspicious. If we really do not need them, I suppose they should just be deleted.

web-api/src/persistence/postgres/jobs/changeOfAddress/schema.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export type ChangeOfAddressTable = typeof changeOfAddressTableDefinition;
+
+export const DW_CHANGE_OF_ADDRESS_COLUMNS = Object.keys(
+  changeOfAddressTableDefinition,
+) as Array<keyof ChangeOfAddressTable>;
+
+export type ChangeOfAddressKysely = Selectable<ChangeOfAddressTable>;
+export type NewChangeOfAddressKysely = Insertable<ChangeOfAddressTable>;
+export type UpdateChangeOfAddressKysely = Updateable<ChangeOfAddressTable>;
```
Mwindo: Similar comment

web-api/src/persistence/postgres/notifications/schema.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export type NotificationTable = typeof notificationTableDefinition;
+
+export const DW_NOTIFICATION_COLUMNS = Object.keys(
+  notificationTableDefinition,
+) as Array<keyof NotificationTable>;
+
+export type NotificationKysely = Selectable<NotificationTable>;
+export type NewNotificationKysely = Insertable<NotificationTable>;
+export type UpdateNotificationKysely = Updateable<NotificationTable>;
```
Mwindo: Similar comment

web-api/src/persistence/postgres/notifications/getDispatchNotification.ts
```diff
@@ -0,0 +1,12 @@
+import { getDbReader } from '@web-api/database';
+
+export const getDispatchNotification = async (topic: string) => {
+  const TTL = Math.floor(Date.now() / 1000);
+  return await getDbReader(reader =>
+    reader
+      .selectFrom('dwNotification')
+      .where('topic', '=', topic)
```
Mwindo, line 8: Is it worth having an index for topic? (I don't know--I'm just calling out places we filter on a column without an index.)

web-api/src/persistence/postgres/jobs/changeOfAddress/setChangeOfAddressCaseAsDone.ts
```diff
@@ -0,0 +1,12 @@
+import { pgUpdateTable } from '@web-api/persistence/postgres/utils/operation/pgUpdateTable';
+import { sql } from 'kysely';
+
+export async function setChangeOfAddressCaseAsDone(jobId: string) {
+  return await pgUpdateTable({
+    table: 'dwChangeOfAddress',
+    values: {
+      remaining: sql<number>`remaining - 1`,
+    },
+    where: qb => qb.where('jobId', '=', jobId),
```
Mwindo, line 10: Is it worth having an index for jobId? (I don't know--I'm just calling out places we filter on a column without an index.)

Mwindo: Nevermind, this is the probably the primary key

web-api/src/persistence/postgres/connections/getAllWebSocketConnections.ts
```diff
@@ -0,0 +1,7 @@
+import { getDbReader } from '@web-api/database';
+import { Connection } from '@web-api/notifications/sendNotificationToConnection';
+
+export const getAllWebSocketConnections = async (): Promise<Connection[]> =>
```
Mwindo: This is where we could use a `ConnectionKysely`, for instance. The reason this works here without that is because the TS type for `Connection` and the type for the data in the database basically intersect entirely. Not saying that one is necessarily better than the other. On the one hand, it is nice to have fewer types; on the other, it is less consistent with other Postgres work. Just an FYI. I think we either be consistent and use the Kysely types, transforming to app-code types as needed, or else just delete them.

web-api/src/lambdas/rdsExpiredRecordsCleanup/rdsExpiredRecordsCleanupLambda.ts
```diff
@@ -0,0 +1,30 @@
...
+import type { Handler } from 'aws-lambda';
+
+const TIME_TO_LIVE_COLUMN = 'ttl';
+
+export const handler: Handler = async (_event, context) => {
+  await getDbReader(async reader => {
+    const ALL_TABLES = await reader.introspection.getTables();
+    const TABLES_WITH_TTLS = ALL_TABLES.filter(tableMetaData => {
+      return tableMetaData.columns.some(
+        columnMetaData => columnMetaData.name === TIME_TO_LIVE_COLUMN,
```
Mwindo, line 13: Something I imagine happening:

- Dev sets TTL on some new table to 30 mins
- "Wait the record is still there in the DB and it's been two hours!"
- cron job runs that night
- "Oh, it's gone now. What is going on?"

I am trying to think about the best way to avoid that. On the one hand, a timestamp is flexible and allows us to run cron jobs at other intervals. On the other, it signals to the developer that they can expire things whenever when they really can't--it depends on the cron interval. I don't know the best solution. I think at least a comment is warranted, though, something like, "The ttl is only as granular as the cron interval in `<file>`"

Mwindo: It would be great to have the cron run per-table at the TTL interval, but that's getting too far afield.

Mwindo (approved):
Thanks @cruzjone-flexion, @pavanyenduri!

I left a few comments. My gut says the nightly TTL mechanism will trip someone up at some point, but I think this is a good first pass.

jimlerza (approved):

PR #6071, opened 2025-05-22
10436 to test once again

TomElliottFlexion:
- Mostly rehash of https://github.com/ustaxcourt/ef-cms/pull/5890
- Vast majority of changes are either deletions related to `eligible-for-trial-case-catalog` or filling code coverage requirements left by those deletions with random tests
- No schema change!


web-api/src/persistence/postgres/cases/getEligibleCasesCount.ts
```diff
@@ -0,0 +1,47 @@
+import {
```
Mwindo: One recommendation: we really want just one definition of what counts as an eligible-for-trial case. We currently have three (because we repeat the logic across three files). Can we do this instead:

1) Combine `getEligibleCasesForTrialCity` and `getElibibleCasesForTrialSession` into one file, `X.ts`, since they do almost the same thing.
2) In that file, define the following or similar (the query without any `select`/`execute` added onto it):

```
export const eligibleCasesQuery = ({
  db,
  trialCity,
  procedureType,
}: {
  db: Kysely<Database>;
  trialCity: string;
  procedureType: ProcedureType;
}) => {
  return db
    .selectFrom('dwCase')
    .where('preferredTrialCity', '=', trialCity)
    .where('procedureType', '=', procedureType)
    .where('status', '=', CASE_STATUS_TYPES.generalDocketReadyForTrial)
    .where(eb =>
      eb.and([
        eb('automaticBlocked', 'is not', true),
        eb('blocked', 'is not', true),
        eb.not(
          eb.exists(sq =>
            sq
              .selectFrom('dwCase as c2')
              .select('c2.leadDocketNumber')
              .where('c2.preferredTrialCity', '=', trialCity)
              .whereRef('c2.leadDocketNumber', '=', 'dwCase.leadDocketNumber')
              .where(qb =>
                qb.or([
                  qb('c2.automaticBlocked', '=', true),
                  qb('c2.blocked', '=', true),
                ]),
              ),
          ),
        ),
      ]),
    );
};
```

3) Use that query as the basis for both `X.ts` and `getEligibleCasesCount.ts`, e.g., in `getEligibleCasesCount.ts`:

```
export const getEligibleCasesCount = async ({
  trialCity,
  procedureType,
}: {
  trialCity: string;
  procedureType: ProcedureType;
}): Promise<number> => {
  const caseCount = await getDbReader(async reader => {
    const baseQuery = eligibleCasesQuery({
      trialCity,
      procedureType,
      db: reader,
    });
    return await baseQuery
      .select(reader.fn.countAll().as('count'))
      .executeTakeFirst();
  });

  return Number(caseCount?.count) || 0;
```

TomElliottFlexion: You are right, and I will do this!

web-api/src/persistence/postgres/cases/getEligibleCasesCount.ts
```diff
@@ -0,0 +1,47 @@
...
+      .selectFrom('dwCase')
+      .select(reader.fn.countAll().as('count'))
+      .where('preferredTrialCity', '=', trialCity)
+      .where('procedureType', '=', procedureType)
+      .where('status', '=', CASE_STATUS_TYPES.generalDocketReadyForTrial)
+      .where('manuallyAddedToTrial', 'is not', true)
+      .where(eb =>
+        eb.and([
+          eb('automaticBlocked', 'is not', true),
+          eb('blocked', 'is not', true),
```
Mwindo: I think these two lines can be moved outside of the callback, i.e., 

```
 .where('manuallyAddedToTrial', 'is not', true)
 .where('automaticBlocked', 'is not', true)
 .where('blocked', 'is not', true)
 .where(eb => eb.not(...
```

I suspect they are in this `eb.and` because this was based off of the `eb.or` in the blocked cases queries.

web-api/src/persistence/postgres/cases/reports/getBlockedCasesCount.ts
```diff
@@ -0,0 +1,41 @@
+import { getDbReader } from '@web-api/database';
+import { CASE_STATUS_TYPES } from '@shared/business/entities/EntityConstants';
+
+export const getBlockedCasesCount = async (trialLocation: string) => {
```
Mwindo, line 5: A similar comment here about reusing the query here.

In fact, this hasn't come up much yet, but at some point we maybe want to think about how to export and reuse queries. Really, the eligible cases query is saying something like "filter by yada yada and also make sure the case isn't blocked." Ideally, the query for "case is blocked" should be defined once and used in a modular fashion elsewhere. Just food for though, nothing to tackle in this PR or maybe even by the current team.

web-api/src/persistence/postgres/cases/getEligibleCasesCount.ts
TomElliottFlexion: Well now that you point it out...

TomElliottFlexion: Thank you!

web-api/src/persistence/postgres/cases/schema.ts
```diff
@@ -39,6 +39,7 @@ export const caseTableDefinition = {
   leadDocketNumber: DEFAULT as string | null | undefined,
   litigationCosts: DEFAULT as number | undefined,
   mailingDate: DEFAULT as string | undefined,
+  manuallyAddedToTrial: DEFAULT as boolean | undefined,
```
Mwindo: I think we can remove this `| undefined`. Related to this, in the mapper `toKyselyNewCase`, I think we can do 

`
manuallyAddedToTrial: rawCase.manuallyAddedToTrial ?? false,
`

web-client/src/views/TrialSessions/TrialSessionPlanningReportView.tsx
```diff
@@ -155,7 +155,11 @@ function TrialSessionPlanningReportTable({
                       />
                     )}
                   </td>
-                  <td>{trialLocation.trialCityState}</td>
+                  <td aria-label="hyperlink to trial location details">
```
Mwindo, line 158: 👍

web-api/src/persistence/postgres/cases/reports/getBlockedCasesCount.ts
TomElliottFlexion: It's a good point, maybe worth bringing up to the team. I'll draw the line for extracting queries here, but I do think what you're proposing is better long-term

shared/src/business/entities/cases/Case.ts
```diff
@@ -1785,20 +1787,14 @@ export class Case extends JoiValidationEntity {
...
-  }
-
   /**
    * set as calendared
    * @param {object} trialSessionEntity - the trial session that is associated with the case
    * @returns {Case} the updated case entity
    */
   setAsCalendared(trialSessionEntity) {
     this.updateTrialSessionInformation(trialSessionEntity);
+    this.manuallyAddedToTrial = true;
```
Mwindo: Won't this also be set from `setTrialSessionCalendarInteractor` for eligible, non-manually added cases? (Here: https://github.com/ustaxcourt/ef-cms/blob/610462e91d383ff3d5e021c33e5536d0fa37dfad/web-api/src/business/useCases/trialSessions/setTrialSessionCalendarInteractor.ts#L129)

web-api/src/persistence/postgres/cases/reports/getBlockedCasesCount.ts
TomElliottFlexion: ✅

web-api/src/persistence/postgres/cases/getEligibleCasesCount.ts
TomElliottFlexion: ✅

web-api/src/persistence/postgres/cases/getEligibleCasesCount.ts
TomElliottFlexion: ✅

shared/src/business/entities/cases/Case.ts
TomElliottFlexion: Dang it

shared/src/business/entities/cases/Case.ts
```diff
@@ -2099,69 +2095,6 @@ export class Case extends JoiValidationEntity {
   }
 }
 
-/**
- * generates sort tags used for sorting trials for calendaring
- * @returns {object} the sort tags
- */
-export const generateTrialSortTags = function ({
```
Mwindo, line 2106: 🎉

shared/src/business/entities/cases/Case.ts
TomElliottFlexion: Not a _problem_ per se, but misleading

Mwindo (commented):
Looks good so far overall! I left a few comments, and I will finish up the review tomorrow. Thank you @TomElliottFlexion! And souviens-toi que tu mourras!

shared/src/business/entities/cases/Case.ts
TomElliottFlexion: `addedToTrial` it is

shared/src/business/entities/cases/Case.ts
```diff
@@ -1521,6 +1523,7 @@ export class Case extends JoiValidationEntity {
     this.trialLocation = undefined;
```
TomElliottFlexion: I suspect `removeFromTrialWithAssociatedJudge` is redundant relative to `removeFromTrial`, unless there is a good reason for not updating the case status. This isn't relevant to this PR, but I'll investigate further

shared/src/business/entities/cases/Case.ts
```diff
@@ -1785,20 +1788,14 @@ export class Case extends JoiValidationEntity {
...
-    return generateTrialSortTags(this);
-  }
-
   /**
    * set as calendared
    * @param {object} trialSessionEntity - the trial session that is associated with the case
    * @returns {Case} the updated case entity
    */
   setAsCalendared(trialSessionEntity) {
     this.updateTrialSessionInformation(trialSessionEntity);
```
Mwindo: I could be totally missing something here, but aren't we setting `trialSessionId`, `trialDate`, etc. right above the line setting `addedToTrialSession = true`? And if so, couldn't we just filter by `trialSessionId` in our eligible cases query and avoid `addedToTrialSession`? From what I see in the code, we set this field to true/false precisely when we update `trialSessionId`, `trialDate`, etc. I apologize if I am overlooking something!

web-client/src/views/TrialSessions/TrialLocationEligibleCasesTable.tsx
```diff
@@ -0,0 +1,124 @@
...
+        </div>
+        <div className="text-right">
+          <span className="text-semibold">Count: </span>
+          {trialLocationHelper.formattedEligibleCases.length}
+        </div>
+        <div className="padding-1"></div>
+        <div className="overflow-x-auto">
+          <div className="minw-tablet-lg">
+            <table
+              aria-label={`${trialLocationPage.currentTab}`}
```
Mwindo, line 49: Do we need this aria-label? If so, should we add it to the TrialLocationBlockedCasesTable?

web-api/src/persistence/postgres/cases/getEligibleCasesForTrialCity.ts
```diff
@@ -0,0 +1,57 @@
+import { CASE_STATUS_TYPES } from '@shared/business/entities/EntityConstants';
+import { getDbReader } from '@web-api/database';
+import { getCasesByDocketNumbers } from '@web-api/persistence/postgres/cases/getCasesByDocketNumbers';
+import { Kysely } from 'kysely';
+import { Database } from '@web-api/database-schema';
+
+export const eligibleCasesQuery = ({
```
Mwindo, line 7: I'm not sure exactly where to put this comment, but I think we may not be properly hiding private data. See below, where I sealed the address: I can't see the address as a petitions clerk on the main case page, but can see it in the network response on the trial session planning location screen.
Screenshot 2025-05-23 at 9 54 49 AM
Screenshot 2025-05-23 at 9 54 21 AM

shared/src/business/entities/cases/Case.ts
TomElliottFlexion: Good heavens. You are not overlooking anything, I do not know how this was not apparent. Fantastic!

shared/src/business/entities/cases/Case.ts
TomElliottFlexion: Nevermind!

web-api/src/persistence/postgres/cases/getEligibleCasesForTrialCity.ts
TomElliottFlexion: Good catch, this will be a problem anywhere we use `getCasesByDocketNumbers` and don't use `CaseFactory`

TomElliottFlexion: Or another filtering method

Mwindo: 100%. CaseFactory had its heart in the right place, but as I've worked on 10502 more, I've realized it doesn't really solve the problem. Its main benefit at this point is just clarifying the logic for how to clean the data.

web-api/src/business/useCases/trialSessions/addCaseToTrialSessionInteractor.ts
```diff
@@ -57,6 +57,7 @@ export const addCaseToTrialSession = async (
   });
 
   const caseEntity = new Case(caseDetails, { authorizedUser });
+  console.log('caseEntity 1', caseEntity);
```
Mwindo: Looks like a stray log :)

web-api/src/business/useCases/trialSessions/addCaseToTrialSessionInteractor.ts
```diff
@@ -95,7 +89,7 @@ export const addCaseToTrialSession = async (
       authorizedUser,
       caseToUpdate: caseEntity,
     });
-
+  console.log('caseEntity 2', updatedCase);
```
Mwindo: Here too!

Mwindo (approved):
Thank you @TomElliottFlexion !

TomElliottFlexion:
Thank _you_ so much!

PR #6075, opened 2025-05-27
10633 Fix condition PR to Test

pavanyenduri:


web-client/src/presenter/computeds/internalPetitionPartiesHelper.ts
```diff
@@ -238,8 +239,9 @@ export const internalPetitionPartiesHelper = (
   const showSecondaryContactEmailFieldAndConsentBox =
     E_CONSENT_FIELDS_ENABLED_FEATURE_FLAG &&
     !isPaper &&
-    (filingType === 'Myself and my spouse' ||
-      filingType === 'Petitioner and spouse') &&
+    (filingType === FILING_TYPES_DICT.MYSELF_AND_SPOUSE ||
+      filingType === FILING_TYPES_DICT.PETITIONER_SPOUSE ||
+      partyType === PARTY_TYPES.petitionerSpouse) &&
```
TomElliottFlexion: The dictionary is definitely good! Do you know why we aren't updating filingType here?

TomElliottFlexion: Or rather just checking partyType, since filingType can't be updated here

web-client/src/presenter/computeds/internalPetitionPartiesHelper.test.ts
```diff
@@ -497,41 +497,6 @@ describe('internalPetitionPartiesHelper', () => {
...
-            filingType: FILING_TYPES_DICT.MYSELF_AND_SPOUSE,
-            isPaper: false,
-          },
-          user: petitionsClerkUser,
-        },
-      });
-
-      expect(result.showSecondaryContactEmailFieldAndConsentBox).toEqual(true);
-    });
-
```
TomElliottFlexion: Doesn't have to happen before going to test, but we should probably replace this test with one that checks for the `showSecondaryContactEmailFieldAndConsentBox` based on `partyType`

TomElliottFlexion (approved):
Looks good!

PR #6088, opened 2025-06-02
10436: trial location screen and eligible cases refactored

TomElliottFlexion:
- Vast majority of changes are either deletions related to `eligible-for-trial-case-catalog` or filling code coverage requirements left by those deletions with random tests
- No schema change!
- After deploy: run `npx ts-node --transpile-only scripts/run-once-scripts/postgres-migration/delete-case-catalog.ts`
  - These records are no longer needed as all eligible case queries are based on case data


web-api/src/business/useCases/trialSessions/getTrialSessionPlanningReportDataInteractor.ts
```diff
@@ -150,32 +152,21 @@ const getTrialLocation = async (
...
   const stateAbbreviation = invert(US_STATES)[trialLocation.state];
 
-  const eligibleCasesSmall = await applicationContext
-    .getPersistenceGateway()
-    .getEligibleCasesForTrialCity({
-      applicationContext,
-      procedureType: 'Small',
-      trialCity: trialCityStateStripped,
-    });
+  const regularCaseCount = await getEligibleCasesCount({
```
Mwindo, line 157: Just a throwaway note that we could probably do this as one query if we wanted to optimize.

scripts/run-once-scripts/postgres-migration/delete-case-catalog.ts
```diff
@@ -0,0 +1,71 @@
...
+    const itemsToDelete = items.map(item => ({
+      DeleteRequest: {
+        Key: {
+          pk: item.pk,
+          sk: item.sk,
+        },
+      },
+    }));
+
+    const itemsDeletedCount = await batchDeleteDynamoItems(
```
zachrog, line 56: You may need to break up the deleting from the querying. If you delete records while you scan then there is the possibility that you delete the lastEvaluatedKey so when you make your next query it does not know where to start the next query from. I have had this happen to me :(

Have you run this in test yet to confirm it does not error and deletes all records on the first try?

Mwindo (approved):
Thank you @TomElliottFlexion!

scripts/run-once-scripts/postgres-migration/delete-case-catalog.ts
TomElliottFlexion: It did run ~~fine~~ swimmingly, beautifully, wonderfully on an exp and on test, but I'll make sure it's fool-proof! Good call out

web-api/src/business/useCases/trialSessions/getTrialSessionPlanningReportDataInteractor.ts
TomElliottFlexion: I was torn- would that still allow me to use `.select(reader.fn.countAll().as('count'))`?

scripts/run-once-scripts/postgres-migration/delete-case-catalog.ts
TomElliottFlexion: Talked it over, I think this is structured correctly. Since this is using a query rather than a scan, and only deleting after the query completes, it is safe

TomElliottFlexion: And can be rerun ad nauseam

jimlerza (approved):

PR #6091, opened 2025-06-03
10505 locks only to test

TomElliottFlexion:
- Replaces custom dynamo lock records with postgres advisory locks
- Advisory locks are
  - Session-specific
  - Automatically released when a DB session ends or manually released
  - Not record or row-specific, they use arbitrary integer keys. In this case we use a consistent hashed version of the old pk.
- To be 100% sure that we are using the same db connection for the duration of these locks (even if we up the max number of connections in the pool), this uses a scoped connection that is created and destroyed by `acquireLock`


web-api/src/business/useCases/caseStatistics/deleteDeficiencyStatisticInteractor.ts
```diff
@@ -33,6 +33,24 @@ export const deleteDeficiencyStatistic = async (
   return validRawCase;
 };
 
+// export const deleteDeficiencyStatisticInteractor = async (
```
Mwindo: Is this supposed to be commented out?

web-api/src/business/useCases/checkForReadyForTrialCasesInteractor.ts
```diff
@@ -31,41 +30,19 @@ export const checkForReadyForTrialCasesInteractor = async (
...
-    }
-  };
-
   const checkReadyForTrial = async (
     caseRecord: Omit<RawCase, 'consolidatedCases'>,
   ) => {
     const { docketNumber } = caseRecord;
-    await acquireLockForCase({ docketNumber });
+
+    // Make sure this is still good
```
Mwindo: Do you still need this comment? If so, can we make it clearer?

web-api/src/business/useCases/checkForReadyForTrialCasesInteractor.ts
```diff
@@ -31,41 +30,19 @@ export const checkForReadyForTrialCasesInteractor = async (
...
+    const removeLockFunction = await acquireLock({
```
Mwindo, line 38: This `const` seems oddly named.

Mwindo: And are we locking appropriately in this file? It looks like we call remove lock, which actually acquires the lock, but at the end, after the thing we are trying to lock.

web-api/src/business/useCases/caseConsolidation/addConsolidatedCaseInteractor.ts
```diff
@@ -107,7 +107,7 @@ export const determineEntitiesToLock = (
     item => `case|${item}`,
   ),
 });
-
+// 10505: this replicates the existing functionality, but may need to account for `allCasesToConsolidate`?
 export const addConsolidatedCaseInteractor = withLocking(
   addConsolidatedCase,
```
TomElliottFlexion, line 112: This _looks_ like an existing gap in locking; if `caseToUpdate.leadDocketNumber !== caseToConsolidateWith.leadDocketNumber` we may query for more cases that wouldn't be locked. Trying to decide whether to fix that as part of this work

web-api/src/business/useCases/caseStatistics/deleteDeficiencyStatisticInteractor.ts
TomElliottFlexion: It's supposed to be removed!

web-api/src/business/useCases/caseConsolidation/addConsolidatedCaseInteractor.ts
Mwindo: I'd say if it's quick, go for it. If not, document and leave it be.

web-api/src/business/useCases/checkForReadyForTrialCasesInteractor.ts
TomElliottFlexion: As it happens, this one was to remind me to check with _you_ (just as the last one to touch it). The old implementation used some retry logic in the interactor... but there was already retry logic in `acquireLock`. Wanted to make sure there wasn't a reason for that I was missing

web-api/src/business/useCases/trialSessions/setTrialSessionCalendarInteractor.ts
```diff
@@ -32,6 +27,9 @@ export const setTrialSessionCalendarInteractor = async (
   authorizedUser: UnknownAuthUser,
 ): Promise<void> => {
   let docketNumbersToLock: string[] = [];
+  let removeLockFunction: Function = () => {
+    console.log('No locks to release');
+  };
```
Mwindo: Probably old testing code

web-api/src/business/useCases/checkForReadyForTrialCasesInteractor.ts
TomElliottFlexion: Open to suggestions! Here we call `acquireLock` which locks the case, and _returns_ the `removeLock` function (to preserve the locking db connection). Then we call the `removeLockFunction` after updating the case.

web-api/src/business/useCases/trialSessions/setTrialSessionCalendarInteractor.ts
TomElliottFlexion: More of a fallback, but it doesn't look good; will remove

web-api/src/business/useCases/checkForReadyForTrialCasesInteractor.ts
Mwindo: Ah, I see now. I didn't see the update to the relevant function yet, which returns a function to remove the lock. Got it.

web-api/src/business/useCases/checkForReadyForTrialCasesInteractor.ts
TomElliottFlexion: The `acquireLock` interface takes retries and wait time, so I used that

web-api/src/persistence/postgres/utils/mutex.ts
```diff
@@ -1,146 +1,100 @@
...
-  return true;
+/**
+ * Converts a string into a consistent 32-bit integer to use as an advisory lock ID.
+ */
+export const hashLockId = (input: string): number => {
+  const hash = crypto.createHash('sha256').update(input).digest();
+  return hash.readInt32BE(0);
 };
 
+// nope!!!
```
Mwindo: I assume this `nope!!!` means something needs updating!

TomElliottFlexion: Alt Text

TomElliottFlexion: ...just the jsdocs

web-api/src/persistence/postgres/utils/mutex.ts
```diff
@@ -1,146 +1,100 @@
...
-
-  if (!isCaseLockingEnabled) {
-    return false;
-  }
-
-  return true;
+/**
+ * Converts a string into a consistent 32-bit integer to use as an advisory lock ID.
+ */
+export const hashLockId = (input: string): number => {
```
Mwindo, line 15: Debugging this is going to be very, very hard. I wonder if it is worth avoiding a hash and making an invertible map so that we could go from some inscrutable integer to a meaningful string like `case|101-20`. E.g. (probably a terrible idea off the top of my head) represent every letter (case-insensitive) as a two digit code (a = 11, b=12...) with 0 as a separator (actually, with fixed-length, no separator would be needed--but you get the idea). Something like that would preserve uniqueness but also allow us, if needed, to know what was actually locked.

Mwindo: In fact, this would let us log just like usual--we'd map the ids back to strings

TomElliottFlexion: That is a good idea. The only thing that stopped me from putting effort into something like that in the first place, is that the locks are session-scoped, meaning that they only live for the duration of the time that `acquireLock` is live.  Still, you're right, hashing them makes them unnecessarily inscrutable

Mwindo: The issue here is integer size, but the basic idea still applies.

Mwindo: Or if we hash, we could save the original identifiers in a lock object.

TomElliottFlexion: Curious about that last idea. Save them in the DB? We do keep them in the `identifierObjects` for logging

Mwindo (commented):
Overall, I think this looks good. Scattered thoughts:

- `getScopedConnection`, which sidesteps all the connection stuff, makes sense but points out there is still work to do making the connection stuff better and more versatile. Nothing for you to do here, just wanted to highlight it.

~~I am leaning towards maybe creating a lock object with a destroy() method and a list of the original (unhashed) identifiers. This makes the code a little more readable (it is weird to see const removeLockFn = await acquireLock(); cons lockObj = await acquireLock() followed by lockObj.destroy() is less weird), and it also allows us to log the actual things being locked.~~

~~Regardless, however, I think before I can approve, we need to have some way to see what is actually locked, not just a hash. Otherwise debugging lock issues will be a nightmare.~~

Nevermind, I totally missed the logging. It might still be worth making a lock object just for code readability, but I'll leave that to you.

web-api/src/persistence/postgres/utils/mutex.ts
Mwindo: 🤦

Mwindo: Oh man, I totally missed this. Doh! Ignore all of my comments about logging.

Mwindo (approved):
Thanks @TomElliottFlexion! Nice work!

web-api/src/persistence/postgres/utils/mutex.ts
TomElliottFlexion: I'm torn, you make a good point!

TomElliottFlexion: Talked about it; decided that since the locks are so short lived, this will only ever be debugged using the `identifierObjects` which preserve the original human-readable lock id

web-api/src/persistence/postgres/utils/mutex.ts
```diff
@@ -1,145 +1,112 @@
...
-  authorizedUser: UnknownAuthUser,
-) => {
-  if (!authorizedUser?.userId || !clientConnectionId) return;
-  await applicationContext.getNotificationGateway().sendNotificationToUser({
-    applicationContext,
-    clientConnectionId,
-    message: { action: 'async_service_unavailable_error' },
-    userId: authorizedUser?.userId,
-  });
+  return async () => {
```
akuny: Re the conversation about the usage of the returned function that removes the lock earlier in this PR:

To make it clearer that this function itself returns a function that removes the lock, we could assign this function to a removeLockFunction constant and then return the const. That could help maintainers a smidge.

akuny (approved):

web-api/src/persistence/postgres/utils/mutex.ts
TomElliottFlexion: Good idea!

pavanyenduri (approved):

PR #6100, opened 2025-06-03
Fix regex in emailReplacer

Mwindo:
We had a case with a caseNote with a linebreak character (`\n`) immediately followed by an email address.

The previous regex assumed the `n` character to be part of the email address and happened to generate a hash that started with `0`. This became `\0` which is a "null" character and that prevented the case table from being restored.

This diff updates the regex to avoid this.


jimlerza (approved):

TomElliottFlexion (approved):
❤️

jimlerza (approved):

PR #6104, opened 2025-06-04
10673 - Upgrade & Refactor Dynamsoft

codyseibert:
flexion#10673

removing the ec2 instances that we original needed to host dynamsoft due to how their license was worded.  Recently, they changed they way they host the resources and licenses which means we can instead easily depend on a unpkg distribution and remove all of the ec2 dynamsoft files we host behind a load balancer.

You'll need to clear out the dynamsoft bucket before deploying:

`./scripts/run-once-scripts/clear-dynamsoft-bucket.sh`


jimlerza (approved):

web-client/src/persistence/dynamsoft/loader.ts
```diff
@@ -0,0 +1,4 @@
+export async function loadDWTLibrary() {
+  // loading dwt will add a DWT object to the window
+  await import('dwt');
```
zachrog, line 3: Good job, code splitting the dwt import. The package is huge

zachrog (approved):
Nice. Removed two EC2 instances. Just a heads up, because dynamsoft is in allColors when we deploy this will break production scanning temporarily.

Mwindo (approved):

PR #6107, opened 2025-06-05
10502 cleanup: fix the delete-route53-latency-records script

jimlerza:
The `delete-route53-latency-records.ts` script didn't work in prod because the prod URLs don't follow the same format that the lower environment URLs do. We can fix this by using the `$EFCMS_DOMAIN` variable instead.


TomElliottFlexion (approved):
Thank you!

PR #6120, opened 2025-06-11
10505 locks only to test draft

TomElliottFlexion:


zachrog (approved):

web-api/src/persistence/postgres/utils/operation/tryGetLock.ts
```diff
@@ -1,15 +1,26 @@
...
+  db,
+  identifiers,
+}: {
+  db: Kysely<Database>;
+  identifiers: string[];
+}): Promise<{ identifier: string; successfullyLocked: boolean }[]> => {
+  const hashedIdentifiers = identifiers.map(id => hashLockId(id));
+  const idMap = new Map(hashedIdentifiers.map((h, i) => [h, identifiers[i]]));
+
+  const sql = hashedIdentifiers
```
Mwindo: Ah, clever. I wonder if we can use `unnest` rather than `union all`. That would give us one `SELECT` rather than multiple.

Mwindo: E.g., something like (untested!):

```
const sql = `
  SELECT x.id,
         pg_try_advisory_lock(x.id) AS locked
    FROM UNNEST($1::bigint[]) AS x(id)
`;
```

web-api/src/persistence/postgres/utils/mutex.ts
```diff
@@ -68,33 +65,17 @@ export const acquireLock = async ({
...
-            getDawsonLogger().info(
-              `Lock not released explicitly for ${idObj.lockId}, falling back to release by connection destroy`,
-            );
-          }
-        }),
-      );
-    } finally {
-      await destroy();
-    }
+    await destroy(); // Destroying connnection releases all locks
```
Mwindo: 🚀

web-api/src/persistence/postgres/utils/operation/tryGetLock.ts
Mwindo: But might not be material, seems like timing is already fast.

TomElliottFlexion: Just tested it, it _is_ marginally faster (~42 ms vs ~36ms for 200 ids). Why not? I'll update it

Mwindo: Lol okey doke :)

PR #6125, opened 2025-06-12
10494  Docket Entries to test

zachrog:


web-api/elasticsearch/docketEntries/indexOpenSearchDocketEntries.ts
```diff
@@ -0,0 +1,145 @@
...
+
+const deleteDocketEntriesFromOpenSearch = async ({
+  message,
+}: {
+  message: OpenSearchSyncMessage;
+}): Promise<void> => {
+  // eslint-disable-next-line prefer-destructuring
+  const payload: { docketNumber: string; docketEntryId: string }[] =
+    message.payload;
+  const body: { delete: { _index: string; _id: string } }[] = payload.map(
```
zachrog, line 45: Come back here and look to see if we are deleting/adding mapping records correctly.

web-api/src/database.ts
```diff
@@ -19,11 +20,15 @@ export function getDbReader<T>(cb: (r: Kysely<Database>) => T): Promise<T> {
 export async function getDbWriter<T>({
   cb,
   table,
+  action,
 }: {
   cb: (db: Kysely<Database>) => Promise<T>;
   table: keyof Database | null;
+  action: OpenSearchSyncAction | null;
```
Mwindo, line 27: We went back and forth a little on the best way to do this. We could pass the `docketNumber` and `docketEntryId` even for deleted documents without this `action` and let the `indexOpenSearchDocketEntries` code handle things (e.g., `indexOpenSearchDocketEntries` sees that the document does not exist and therefore deletes it). We decided to be explicit here and pass in an action.

web-api/elasticsearch/docketEntries/indexOpenSearchDocketEntries.ts
Mwindo: Seems like it works locally after this commit: https://github.com/ustaxcourt/ef-cms/pull/6125/commits/9514309a3f4ec91bda8249962876c7ddbad013ea

Mwindo: In that I can search for X and see it in search, delete it, and then no longer see it in search.

Mwindo (approved):

PR #6131, opened 2025-06-13
Review  -  Removed Amazon pay language

AnthonyLoeraTaxCourt:
https://github.com/flexion/ef-cms/issues/10677

On page 7 of our petitioner flow we indicate that pay.gov will accept Amazon pay. They do not. We should remove that language so we don't mislead anyone.

AC: Remove Amazon Pay from the list of accepted ways to pay.


Mwindo (approved):
Thanks @AnthonyLoeraTaxCourt and @alexdziarn!

PR #6144, opened 2025-06-17
10562 delete mismatch error messages to test

alexdziarn:
https://github.com/flexion/ef-cms/issues/10562
flexion#10562

### DIFF
--- a/web-client/src/presenter/actions/TrialSession/computeTrialSessionFormDataAction.ts
+++ b/web-client/src/presenter/actions/TrialSession/computeTrialSessionFormDataAction.ts
@@ -28,6 +28,9 @@ export const computeTermAndUpdateState = (
 
     store.set(state.form.term, term);
     store.set(state.form.termYear, date.year);
+  } else {
+    store.set(state.form.term, undefined);
+    store.set(state.form.termYear, undefined);
   }
 };
 
--- a/web-client/src/presenter/actions/TrialSession/validateTrialSessionAction.ts
+++ b/web-client/src/presenter/actions/TrialSession/validateTrialSessionAction.ts
@@ -17,6 +17,8 @@ export const validateTrialSessionAction = ({
     return path.success();
   } else {
     errors.startDate = errors.startDate || errors.term;
+    errors.term = undefined;
+    errors.termYear = undefined;
     const errorDisplayOrder = [
       'startDate',
       'startTime',
--- a/web-client/src/presenter/actions/TrialSession/validateTrialSessionAction.ts
+++ b/web-client/src/presenter/actions/TrialSession/validateTrialSessionAction.ts
@@ -16,7 +16,8 @@ export const validateTrialSessionAction = ({
   if (!errors) {
     return path.success();
   } else {
-    errors.startDate = errors.startDate || errors.term;
+    errors.term = undefined;
+    errors.termYear = undefined;
     const errorDisplayOrder = [
       'startDate',
       'startTime',
--- a/web-client/src/presenter/actions/TrialSession/validateTrialSessionAction.ts
+++ b/web-client/src/presenter/actions/TrialSession/validateTrialSessionAction.ts
@@ -17,8 +17,10 @@ export const validateTrialSessionAction = ({
     return path.success();
   } else {
     errors.startDate = errors.startDate || errors.term;
-    errors.term = undefined;
-    errors.termYear = undefined;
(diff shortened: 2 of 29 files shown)


web-client/src/presenter/actions/TrialSession/computeTrialSessionFormDataAction.ts
```diff
@@ -28,6 +28,9 @@ export const computeTermAndUpdateState = (
 
     store.set(state.form.term, term);
     store.set(state.form.termYear, date.year);
+  } else {
+    store.set(state.form.term, undefined);
+    store.set(state.form.termYear, undefined);
```
Mwindo, line 33: Is this necessary? (I don't know--just double-checking!) If so, I'd recommend using `store.unset`, e.g., `state.unset(state.form.term)`. (It does the same thing, just a little clearer and easier to search for.) But it's not a biggie.

web-client/src/presenter/actions/TrialSession/validateTrialSessionAction.ts
```diff
@@ -17,6 +17,10 @@ export const validateTrialSessionAction = ({
     return path.success();
   } else {
     errors.startDate = errors.startDate || errors.term;
+    if(errors.startDate) {
+      errors.term = undefined;
+      errors.termYear = undefined;
+    }
```
Mwindo, line 23: This is a much safer fix than the earlier one that was done in entity-land. Nice work!

My only concern--and I'll leave it up to you whether to address or not--is that this makes things more fragile. What do I mean?

We currently use the same TypeScript entities and validate them the same way on both the front end and back end. This is a double-edged sword: the great thing is convenience and consistency. The not-so-great thing is that sometimes--as you are discovering here--the front end and back end have slightly different validation needs.

The solution here says: "Hey, look: I know that you need a `term` and a `termYear` on a trial session. But the _user_ can't select any `term` or `termYear`--we set those behind the scenes based on `startDate`! So if there is no `startDate`, don't show those fields, silly!" This is totally reasonable, and it should work. (From what I can tell, this action is only used in the case where we set `termYear` and `term` based on `startDate`.) However, a few things to consider:
- Actions are globally available functions. (I do not necessarily like this, but it's the way things are!) If this action happens to be called in a different context ("Oh, it looks like we already have an action to validate a trial session, so I'll use that")--one in which a form is set up differently (i.e., we _aren't_ setting `term` and `termYear` automatically)--it could lead to spurious indications that the trial session is valid.
- More generally, this couples `term` and `termYear` to `startDate`. If the way we set up this form handling changes in the future, this logic here needs to change too. But it would be extremely easy for a developer to overlook this.
- A developer looking at this might not understand what is going on. ("Wait, why are we randomly ignoring these errors?") This is a small thing, and it might only take a good developer a minute or two to figure out. However, my experience is that the app is riddled with these small "a-minute-or-two-to-figure-out" things, and they add up and sometimes even lead to bugs.

I point this all out not because it is super important here or because it is likely to cause issues or because you've made a mistake. I think your solution works! (Although it might be worth putting in a comment about why we are ignoring these errors here.) I am saying all this simply to share context. Two of the underlying tensions of the app involve 1) the convenience but thorniness of Cerebral actions and global state (since you need to consider all of the places an action is being used or might be used, and all the side effects modifying one piece of state might have) and 2) the way entities are carved up, validated, and shared. Hopefully as you become stewards of the application you will find better ways to address these tensions!

web-client/src/presenter/actions/validateSelectDocumentTypeAction.ts
```diff
@@ -25,6 +25,7 @@ export const validateSelectDocumentTypeAction = ({
   if (!errors) {
     return path.success();
   } else {
+    if(errors.documentType) delete errors.category;
```
Mwindo, line 28: More or less the same comment as above.

Mwindo (approved):
Thank you @alexdziarn! I left a couple of comments--one of them a long-winded one mostly to context-dump--but this is a much safer solution than before for a deceptively tricky design debt :) Nice job.

Mwindo:
Also, just a reminder since this branch has work the to-staging branch does not--make sure to sync these up somehow! (I recommend always updating the to-staging branch and then creating/updating the to-test branch from it. It can be a pain, but--if you are anything like me, at least!--it will probably prevent you from making a mistake in the future!)

PR #6145, opened 2025-06-17
10495 DxOx - Users to Postgres (branched off of docket entries)

codyseibert:
this work is branched off the docket entries work as the overlap between the docket entries and userId seemed good to just have this depend on that refactoring.


web-api/src/business/useCases/auth/changePasswordInteractor.test.ts
```diff
@@ -363,7 +371,7 @@ describe('updateUserPendingEmailRecord', () => {
     const updateUserCalls = updateUser.mock.calls;
     expect(updateUserCalls.length).toEqual(1);
     expect(updateUserCalls[0][0].userToUpdate).toMatchObject({
-      isUpdatingInformation: false,
+      isUpdatingInformation: undefined,
```
TomElliottFlexion, line 367: We _might_ want to keep guaranteeing true/false

PR #6159, opened 2025-06-18
10583: fix bug in focus component

jimlerza:
flexion#10583


jimlerza:
Closing this PR because I've realized that I shouldn't be the one to make PRs to staging

PR #6178, opened 2025-06-23
10614: add counters to trial session copies headers

alexdziarn:
https://github.com/orgs/flexion/projects/11/views/1?pane=issue&itemId=103073796&issue=flexion%7Cef-cms%7C10614


shared/src/business/utilities/pdfGenerator/components/SelectedFiltersSection.tsx
```diff
@@ -1,26 +1,34 @@
 import React from 'react';
 
-export const SelectedFiltersSection = ({ count, selectedFilters }) => (
```
Mwindo, line 3: Heads-up: I think this only affects half of the code. This screen shot below is for `WorkingCopySessionList`, and I suspect that needs to be updated as well. Easy to miss. Unfortunately, our frontend code has a lot of similar and potentially redundant components.

Screenshot 2025-06-24 at 10 09 08 AM

Mwindo: Another heads-up that I don't think the following requirement of the ticket is met: "Whether a particular status is selected or deselected has no bearing on the count shown for that particular status." Selecting/deselecting a status does seem to change the parenthetical counts.

shared/src/business/utilities/pdfGenerator/components/SelectedFiltersSection.tsx
```diff
@@ -1,26 +1,34 @@
...
-        <div className="filter">{selectedFilters[0]?.label || ''}</div>
-        <div className="filter">{selectedFilters[2]?.label || ''}</div>
-        <div className="filter">{selectedFilters[4]?.label || ''}</div>
-        <div className="filter">{selectedFilters[6]?.label || ''}</div>
-        <div className="filter">{selectedFilters[8]?.label || ''}</div>
+        {[0, 2, 4, 6, 8].map(index => (
+          <div key={index} className="filter">
+            {selectedFilters[index]?.label || ''}
+            {selectedFilters[index]?.key in trialStatusCounts && (
+              <strong>({trialStatusCounts[selectedFilters[index].key]})</strong>
```
Mwindo: Two nitpicks:

- I would recommend HTML for semantics and CSS classes for pure styling. `<strong>` _does_ change the styling, but it also signals "This is important!" In this case, because the count is not "important," I would recommend using a CSS class.
- It looks like there is no gap between the text and the parenthetical. I'm sure @katiecissell or @Evelyn-Durbin would catch this, but putting it here for fast feedback.

Screenshot 2025-06-24 at 1 04 34 PM

shared/src/business/utilities/pdfGenerator/documentTemplates/PrintableWorkingCopySessionList.tsx
```diff
@@ -25,6 +25,14 @@ export const PrintableWorkingCopySessionList = ({
       ? `${formattedTrialSession.formattedStartDateFull} - ${formattedTrialSession.formattedEstimatedEndDateFull}`
       : `${formattedTrialSession.formattedStartDateFull}`;
 
+  const trialStatusCounts = formattedCases.reduce((counters, c) => {
```
Mwindo: This will be up to Gunnison to decide, but in general we have tried to keep logic outside of the components and inside, e.g., Cerebral computed or state. (The idea, although imperfect, was to separate logic out from rendering. You'll notice we don't really have react-specific testing in our unit tests; we rely on the separation between state and rendering, testing the state.) I don't think I have the authority to say, "Don't do this!" but I wanted to let you know.

web-client/src/views/TrialSessionWorkingCopy/WorkingCopyFilterHeader.tsx
```diff
@@ -10,13 +10,23 @@ export const WorkingCopyFilterHeader = connect(
...
+    trialSessionWorkingCopy: state.trialSessionWorkingCopy,
   },
   function WorkingCopyFilterHeader({
     autoSaveTrialSessionWorkingCopySequence,
     filters = {},
     trialSessionWorkingCopyHelper,
     trialStatusFilters,
+    trialSessionWorkingCopy,
   }) {
+    const trialStatusCounts = Object.values(trialSessionWorkingCopy.caseMetadata).reduce((counters, c) => {
```
Mwindo: Similar comment to above, with the added comment that this feels duplicative. Can we consolidate these two aggregations somehow? (I realize that the whole components are duplicative, but no reason to avoid adding salt to the wound!)

web-client/src/views/TrialSessionWorkingCopy/WorkingCopySessionList.tsx
```diff
@@ -4,21 +4,32 @@ import { WorkingCopyFilterHeader } from './WorkingCopyFilterHeader';
...
+    autoSaveTrialSessionWorkingCopySequence: sequences.autoSaveTrialSessionWorkingCopySequence
   },
   function WorkingCopySessionList({
     constants,
     toggleWorkingCopySortSequence,
     trialSessionWorkingCopy,
     trialSessionWorkingCopyHelper,
+    autoSaveTrialSessionWorkingCopySequence
   }) {
+    useEffect(() => {
```
Mwindo: I vaguely understand the issue here. My vague understanding is something like this: "Man, on first loading the page, `undefined` won't show up in the count properly. So let's add a `useEffect` to get that in the right format so we see the `undefined`--which corresponds to the "status unassigned"--counts." However, I think there are issues with this solution. I will list them in rough order of least-to-most importance:
- In general, given how the current codebase is set up at least, I think it is better to get the state correct _before_ going to the page. In other words, here we are relying on Cerebral sequences and computeds to do work setting up state in the way the page needs and then _also_ relying on a `useEffect` to set up further state stuff.
- This could be very inefficient. `autoSaveTrialSessionWorkingCopySequence` makes a call to update data in the database (eventually, via `updateTrialSessionWorkingCopyInteractor`). If `formattedCases` includes 100 cases each with `undefined` status, we are now going to make 100 requests to the database to update persistence. It would be much better to do one call with 100 docket numbers rather than 100 calls with one docket number.
- But we probably should be making _zero_ requests to persist data in the database. From what I can tell, we have solved what is essentially a frontend style issue--making sure a count label displays correctly on first rendering--by changing state in our database. Are there side-effects to this decision to change data in the database from one type to another? (Looks like `undefined` vs. `""`, but I haven't dug deeply.) Do we _need_ the data to be in this state in the database? If it turns out we do--i.e., that this really is an issue with our data--how do we handle that in a less ad-hoc way?

web-client/src/presenter/computeds/trialSessionWorkingCopyHelper.ts
```diff
@@ -101,12 +102,21 @@ export const trialSessionWorkingCopyHelper = (
...
-
+  const trialStatusCounts = trialSession.calendaredCases.reduce((counters, c) => {
+    if(caseMetadata[c.docketNumber] === undefined || caseMetadata[c.docketNumber].trialStatus === "" 
+      || caseMetadata[c.docketNumber].trialStatus === undefined) {
+      counters["statusUnassigned"] = (counters["statusUnassigned"] || 0) + 1;
+    } else {
+      counters[caseMetadata[c.docketNumber].trialStatus] = (counters[caseMetadata[c.docketNumber].trialStatus] || 0) + 1;
+    }
+    return counters
+  }, {});
```
Mwindo: Nice work @alexdziarn--this is a _much_ better approach. The disjunctive statement (x || y || z) can probably be improved, maybe with a self-documenting helper function like `trialStatusIsUnassigned`, but otherwise I think this is getting close. I'll take a look again tomorrow :)

Mwindo: Or maybe, rather than a helper function:

```
const trialStatusCounts = trialSession.calendaredCases.reduce((counters, { docketNumber }) => {
  const status = caseMetadata[docketNumber]?.trialStatus || '';
  const key = status || 'statusUnassigned';
  counters[key] = (counters[key] ?? 0) + 1;
  return counters;
}, {});
```

shared/src/business/utilities/pdfGenerator/documentTemplates/PrintableWorkingCopySessionList.tsx
jimlerza: I agree with Chris! In keeping with the established patterns, we should compute this elsewhere and pass it in here.

jimlerza (changes requested):

jimlerza (approved):

web-client/src/presenter/computeds/trialSessionWorkingCopyHelper.ts
```diff
@@ -101,12 +102,23 @@ export const trialSessionWorkingCopyHelper = (
       key: 'statusUnassigned',
       label: 'Unassigned',
     });
-
+  const trialStatusCounts = trialSession.calendaredCases
+    .filter(isOpenCaseInATrial)
+    .reduce((counters, c) => {
+    if(caseMetadata[c.docketNumber] === undefined || caseMetadata[c.docketNumber].trialStatus === "" 
+      || caseMetadata[c.docketNumber].trialStatus === undefined) {
```
Mwindo, line 109: I'd probably pull this out into a function to be self-documenting, but that's super minor and need not be done right now.

Mwindo (approved):
Thanks @alexdziarn!

PR #6181, opened 2025-06-24
10614 quickfix

alexdziarn:
[10614-sesssion-copy-trial-status-counters](https://github.com/orgs/flexion/projects/11/views/1?pane=issue&itemId=103073796&issue=flexion%7Cef-cms%7C10614)


Mwindo (changes requested):
Thanks @alexdziarn! I have commented on the draft to-staging PR and requested changes here. The `useEffect` code gives me pause because I believe it is solving a frontend problem by persisting data in the backend. Let me know if that doesn't make sense or if I am missing something!

Mwindo (approved):
Thanks @alexdziarn! Good work.

I don't love that we do a similar aggregation both in `trialSessionWorkingCopyHelper` and in the `printableWorkingCopy` file, but I won't hold the PR back because of that.

Note that you will need to add some tests for this before it can go to staging, probably in both `trialSessionWorkingCopyHelper.test.ts` and in an existing or new Cypress test.

PR #6182, opened 2025-06-24
10532: update login & service email language

AnthonyLoeraTaxCourt:
https://github.com/orgs/flexion/projects/11/views/1?pane=issue&itemId=83470075&issue=flexion%7Cef-cms%7C10532


web-client/src/views/Practitioners/PractitionerLoginServiceEmailForm.tsx
```diff
@@ -68,7 +71,7 @@ export const PractitionerLoginServiceEmailForm = connect<
                 id={emailFormName}
                 name={emailFormName}
                 type="text"
-                value={form[emailFormName] || ''}
+                value={form[emailFormName] ?? ''}
```
akuny: It's unlikely, but be cautious that form.confirmEmail cannot be set to something falsey apart from null or undefined, such that we don't end up in a situation where one of those is rendered instead of the desired fallback.

akuny (approved):

web-client/src/views/Practitioners/PractitionerLoginServiceEmailForm.tsx
```diff
@@ -34,31 +33,35 @@ export const PractitionerLoginServiceEmailForm = connect<
...
-          </div>
+              <h4>Change Login & Service Email</h4>
+            </>
+          )}
           <div>
-            <h4>Change Login & Service Email</h4>
             <FormGroup
               errorText={
-                validationErrors.updatedEmail || validationErrors.email
+                validationErrors.updatedEmail ?? validationErrors.email
```
jimlerza:
```suggestion
                validationErrors.updatedEmail || validationErrors.email
```

web-client/src/views/Practitioners/PractitionerLoginServiceEmailForm.tsx
jimlerza:
```suggestion
                value={form[emailFormName] || ''}
```

web-client/src/views/Practitioners/PractitionerLoginServiceEmailForm.tsx
```diff
@@ -90,7 +93,7 @@ export const PractitionerLoginServiceEmailForm = connect<
                 id="confirm-email"
                 name="confirmEmail"
                 type="text"
-                value={form.confirmEmail || ''}
+                value={form.confirmEmail ?? ''}
```
jimlerza:
```suggestion
                value={form.confirmEmail || ''}
```

web-client/src/views/Practitioners/PractitionerLoginServiceEmailForm.tsx
```diff
@@ -13,7 +12,7 @@ const practitionerLoginServiceEmailFormDependencies = {
   createPractitionerUserHelper: state.createPractitionerUserHelper,
   form: state.form,
   updateFormValueSequence: sequences.updateFormValueSequence,
-  validateSequence: sequences[props.validateSequenceName],
+  validateSequence: sequences[props`validateSequenceName`],
```
jimlerza, line 15: This doesn't look right to me, but I'm not a cerebral pro. Do you mind to explain what this is doing?

jimlerza (changes requested):
I agree with Andy - we should keep the `||` fallbacks.

web-client/src/views/Practitioners/PractitionerLoginServiceEmailForm.tsx
AnthonyLoeraTaxCourt: I committed your changes. Thank you!

web-client/src/views/Practitioners/PractitionerLoginServiceEmailForm.tsx
AnthonyLoeraTaxCourt: Committed your changes. Thank you!

jimlerza (approved):

PR #6183, opened 2025-06-24
10678: remove old files to share documents

alexdziarn:
https://github.com/orgs/flexion/projects/11/views/1?pane=issue&itemId=114759547&issue=flexion%7Cef-cms%7C10678


codyseibert:
heads up, I think this change might cause the smoketests to run using the wrong api.  The color switch is what updates that subdomain to point to the correct api gateway deployment, but that doesn't run until after the smoke tests are successful

alexdziarn:
> heads up, I think this change might cause the smoketests to run using the wrong api. The color switch is what updates that subdomain to point to the correct api gateway deployment, but that doesn't run until after the smoke tests are successful

I did see that the in route53 the "app/*" points to the current cloudfront distribution wether it be blue or green. Not sure if files are transferred over when a blue/green deployment is made, if they are, then this solution should work. I will just need to edit tests if needed. If they are not, then we'll have to discuss a more complex solution.

web-api/src/environment.ts
```diff
@@ -20,7 +20,7 @@ function getJestDBConnectionError(): string {
 
 export const environment = {
   appEndpoint: process.env.EFCMS_DOMAIN
-    ? `app-${currentColor}.${process.env.EFCMS_DOMAIN}`
+    ? `app.${process.env.EFCMS_DOMAIN}`
```
Mwindo: Just keying off of what you and @codyseibert were saying--I think it will be more complex than this, unfortunately.

PR #6108, opened 2025-06-06
Fix promise returns in updateCaseAndAssociations

zachrog:

PR #6185, opened 2025-06-25
10676 followup to 10502 to test

Mwindo:
This ticket is for cleaning up stuff after 10502 is deployed, and changes related to that:

**TL;DR**: Remove unnecessary code, refactor `updateCaseAndAssociations` to be more type safe and easy-to-follow, and always get case data via `getCasesByDocketNumbers`. No more random subsets of case data gotten in different ways--hooray!

---

**Details**: 
- Removing the `processCaseEntries` indexing code
- Removing `aggregateCaseItems` code
- Adding an `index-cases.ts` script
- Refactoring persistence functions so that there are fewer (see notes below)
- Refactoring `updateCaseAndAssociations`. We identified a fairly serious, long-standing bug--promises not awaited quite right--that we were just lucky not to be running into in practice. The refactor exposed it, maybe because we were querying more efficiently. See [this PR comment](https://github.com/ustaxcourt/ef-cms/pull/6025/files#r2132531317). In the meantime, we patched this bug in staging [here](https://github.com/ustaxcourt/ef-cms/pull/6108).
- Replacing `documentTitleJourney.test.ts` (cerebral) with `document-title.cy.ts` (cypress) (which had been failing due to the above)
- Getting rid of `applicationContext` argument in `getCaseByDocketNumber` and `updateCaseAndAssociations`.

**About the `getCasesByDocketNumbers` refactor**: 
In `getCaseByDocketNumber`, we have a call to `getCasesMetadataWithCounselByLeadDocketNumber`. This latter function is inefficient: it fetches several records in a loop rather that bulk fetching from Postgres. In the process of making it better, it seemed best to get rid of persistence functions that are all sort of doing the same thing to avoid repeating logic. To that end, this PR:

- Removes `getCaseMetadataWithCounsel`, `getCasesMetadataByDocketNumbers`, `getCasesInConsolidatedGroup`, `getCasesByLeadDocketNumber`, `getCasesMetadataWithCounselByLeadDocketNumber`
- Instead, relies on a more dynamically typed `getCasesByDocketNumbers` and `getConsolidatedCases` (which uses getCasesByDocketNumbers under the hood in most situations)
- The more we can simplify as the project switches hands, the better!


cypress/local-only/tests/integration/documentQC/document-title.cy.ts
```diff
@@ -0,0 +1,108 @@
+import { attachFile } from '../../../../helpers/file/upload-file';
```
Mwindo, line 1: This file replaces `documentTitleJourney.test.ts`

scripts/run-once-scripts/postgres-migration/index-cases.ts
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env -S npx ts-node --transpile-only
```
Mwindo, line 1: We can no longer rely on Dynamo streams to re-index cases after a blue-green migration. Eventually we may not need a separate blue/green OpenSearch instance at all. In the meantime, this is a script we can use to re-index cases. It is not perfect--we will need to talk through how to do this well--but it's a start.

scripts/run-once-scripts/postgres-migration/index-docket-entries.ts
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env -S npx ts-node --transpile-only
```
Mwindo, line 1: Similar comment to the above. I believe this is in the to-staging docket entries branch, just not merged into test.

web-api/src/business/useCaseHelper/caseAssociation/updateCaseAndAssociations.ts
```diff
@@ -5,7 +5,7 @@ import { DocketEntry } from '@shared/business/entities/DocketEntry';
 import { IrsPractitioner } from '@shared/business/entities/IrsPractitioner';
```
Mwindo, line 5: This is probably my least favorite function in the whole system. We at least tried to make it a little easier to follow. The previous iteration had a bug in it (now patched in staging) uncaught for years due to the lack of types and a difficult-to-spot misuse of promises.

web-api/src/persistence/postgres/cases/getCasesByDocketNumbers.ts
```diff
@@ -15,37 +15,82 @@ import { fromKyselyDocketEntry } from '@web-api/persistence/postgres/docketEntri
 import { DocketEntryKysely } from '@web-api/persistence/postgres/docketEntries/schema';
 import { difference, isEmpty, sortBy } from 'lodash';
 
-export async function getCasesByDocketNumbers({
+export const ALL_OMITTABLE_CASE_FIELDS = [
```
Mwindo, line 18: "Omittable" case fields made things easier type-wise than "includable" case fields. Also, we thought it might be safer for the default to be "include everything" (less efficient, but nothing will break due to missing data) vs. "include only the minimum" (more efficient, but maybe missing data leads to the app breaking somewhere).

web-api/src/business/useCases/caseConsolidation/removeConsolidatedCasesInteractor.ts
```diff
@@ -44,9 +44,10 @@ const removeConsolidatedCases = async (
 
   const { leadDocketNumber } = caseToUpdate;
 
-  const allConsolidatedCases = await getCasesByLeadDocketNumber({
+  const allConsolidatedCases = await getConsolidatedCases({
     leadDocketNumber,
   });
+  console.log('allConsolidatedCases', allConsolidatedCases);
```
TomElliottFlexion: Just a few stray console logs here

web-api/src/business/useCases/courtIssuedDocument/fileAndServeCourtIssuedDocumentInteractor.consolidatedCases.test.ts
```diff
@@ -23,15 +23,24 @@ import { getCaseByDocketNumber as getCaseByDocketNumberMock } from '@web-api/per
...
 describe('consolidated cases', () => {
   const getCaseByDocketNumber = getCaseByDocketNumberMock as jest.Mock;
-  const getCasesByDocketNumbers = jest.mocked(getCasesByDocketNumbersMock);
+  const getCasesByDocketNumbers =
+    getCasesByDocketNumbersMock as jest.MockedFunction<
+      (args: {
+        docketNumbers: string[];
+        excludeFields?: OmittableCaseFields[];
+      }) => Promise<Omit<RawCase, 'consolidatedCases'>[]>
+    >;
```
TomElliottFlexion, line 43: Sheerly out of curiosity, do you know why we have to do this?

web-api/src/business/useCases/docketEntry/updateDocketEntryMetaInteractor.test.ts
```diff
@@ -600,6 +598,7 @@ describe('updateDocketEntryMetaInteractor', () => {
   });
 
   it('should update the document pending status and the automatic blocked status of the case when setting pending to true', async () => {
+    getCasesByDocketNumbers.mockResolvedValue([MOCK_CASE]);
```
TomElliottFlexion: Not a big deal at all, but ideally we'd either mock this in a beforeEach/All, or use a mockResolvedValueOnce. This mocks it for all subsequent tests

web-api/src/business/useCases/processStreamRecords/processPractitionerMappingEntries.ts
```diff
@@ -32,26 +31,9 @@ export const processPractitionerMappingEntries = async ({
           'case|'.length,
         );
 
-        // After case records have been moved into postgres, we need to get the case data associated with the practitioner from postgres.
-        // However, when we try to fetch case data here during the initial blue-green migration re-indexing step (to get case records
-        // into postgres in the first place), the case data might not yet have been moved over to postgres. Therefore, we fallback to the case data from Dynamo.
-        // TODO after 10502: Only rely on postgres by in-lining getCaseDataFromPostgres here.
```
TomElliottFlexion, line 38: Nice!

web-api/src/business/useCases/docketEntry/updateDocketEntryMetaInteractor.test.ts
Mwindo: Good call!

TomElliottFlexion (approved):
Thank you!

web-api/src/business/useCases/courtIssuedDocument/fileAndServeCourtIssuedDocumentInteractor.consolidatedCases.test.ts
Mwindo: The basic issue is that because of the dynamic typing in `getCasesByDocketNumbers`, TypeScript cannot easily determine exactly which subset of things you are returning in the test. When we specify `docketEntries: Blah` in the resolved values, it therefore complains: you might not *have* docketEntries! The above is one way to fix that, but I am totally open to better ways.

web-api/src/business/useCases/caseConsolidation/removeConsolidatedCasesInteractor.ts
Mwindo: Ack, thank you!

web-api/src/business/useCases/courtIssuedDocument/fileAndServeCourtIssuedDocumentInteractor.consolidatedCases.test.ts
TomElliottFlexion: No this looks good, I just wanted to learn, thank you!

PR #6191, opened 2025-06-25
10286 service parties column should wrap long text intermediate branch to test 1750875204

AnthonyLoeraTaxCourt:
https://github.com/orgs/flexion/projects/11/views/1?pane=issue&itemId=76509477&issue=flexion%7Cef-cms%7C10286

### DIFF
--- a/web-client/src/views/CourtIssuedDocketEntry/CourtIssuedDocketEntry.tsx
+++ b/web-client/src/views/CourtIssuedDocketEntry/CourtIssuedDocketEntry.tsx
@@ -236,7 +236,7 @@ export const CourtIssuedDocketEntry = connect(
                             <div className="service-party-name">
                               {party.displayName}
                             </div>
-                            <div>
+                            <div className="service-party-indicator">
                               <b>Service: </b>
                               {party.serviceIndicator}
                             </div>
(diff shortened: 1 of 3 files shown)


web-client/src/views/CourtIssuedDocketEntry/CourtIssuedDocketEntry.tsx
```diff
@@ -230,11 +230,13 @@ export const CourtIssuedDocketEntry = connect(
                       {addCourtIssuedDocketEntryHelper.serviceParties.map(
                         party => (
                           <div
-                            className="margin-bottom-2"
```
Mwindo, line 233: Running this locally, I don't think it is quite what we want. I suspect we always want the `Service` info on the right, but here in the screen shot it wraps below and to the left.

Screenshot 2025-06-25 at 2 45 31 PM

Mwindo (approved):
Thanks @AnthonyLoeraTaxCourt! This seems like it works.

I have a few questions, but I'll leave it up to you how far you want to address them:

- Is this styling actually for a form? If not, it might be good to move these classes somewhere else.
- Why do we have 600px hard-coded? Can we use an existing constant? Is there a reason we cannot use $medium-screen (example usage is in this file)?
- Similar questions about the hard-coded 150px, 400px.
- Is there a way to avoid !important and maybe to simplify? For instance, I don't see a difference in behavior it I get rid of the flex definition in service-party-name (which addresses the bullet above). That said, I only looked this cursorily.
- You might consider nesting `service-party-indicator` and `service-party-name` under `service-party` if they should only ever be used in a nested context.

Again, these are just questions that come to mind as I look at things, and I leave it to you to decide whether to address any.

PR #6197, opened 2025-06-26
10584 pdf icon is invisible when in mobile intermediate branch to test 1750962823

AnthonyLoeraTaxCourt:
https://github.com/orgs/flexion/projects/11/views/1?pane=issue&itemId=94607946&issue=flexion%7Cef-cms%7C10584


web-client/src/styles/custom.scss
```diff
@@ -859,7 +859,7 @@ textarea#case-caption {
 }
 
 .view-pdf-button svg {
-  color: $color-white;
+  color: #005ea2;
```
Mwindo: It looks like we use `#005ea2` in a few places. Is it worth making styling constant?

Mwindo (approved):

PR #6198, opened 2025-06-26
10614 sesssion copy trial status counters to test

alexdziarn:
https://github.com/orgs/flexion/projects/11/views/1?pane=issue&itemId=103073796&issue=flexion%7Cef-cms%7C10614


cypress/local-only/tests/integration/trialSessionWorkingCopy/trial-session-working-copy-filter.cy.ts
```diff
@@ -0,0 +1,38 @@
+import { loginAsColvin } from "cypress/helpers/authentication/login-as-helpers"
+
+
+describe('trials session working copies filtering', () => {
```
Mwindo, line 8: Thank you @alexdziarn!

A few notes:

- Typically, we try to have each it block be self contained so that we can run tests in isolation. (That way you could it.only and debug, for instance. Check out project documentation [here](https://github.com/ustaxcourt/ef-cms/blob/staging/cypress/CYPRESS-README.md).) My recommendation would be to put all of these `it`s into a single test. (You could do them each separately, but it would make the CI pipeline a little slower.)
- Although some Cypress tests do rely on seed data, it is better to create the data you need in the test.* Rather than relying on the 3 cases in the seed data, it would probably be better to create the cases you need via the helper functions. If you look at other Cypress tests, like (selected at random) [this one](https://github.com/ustaxcourt/ef-cms/blob/staging/cypress/local-only/tests/integration/caseDetail/docketRecord/courtIssuedFiling/clerk-of-court-stamp-order.cy.ts), you can see some examples using the helpers from `petitioner-creates-electronic-case`. 

* Why? 1) This means our seed data becomes less flexible and 2) our tests can interfere with each other (test 1 changes X seed data and test 2 relied on unchanged X). Both of these can lead to confusing test failures.

cypress/local-only/tests/integration/trialSessionWorkingCopy/trial-session-working-copy-filter.cy.ts
```diff
@@ -0,0 +1,38 @@
...
+	it('logins as judge colvin', () => {
+		loginAsColvin();
+	});
+	it('should have all docket numbers set to statusUnassigned on initial load', () => {
+		cy.get('[data-testid="trial-session-link"]').click();
+		cy.get('[data-testid="trial-location-link-959c4338-0fac-42eb-b0eb-d53b8d0195cc"]').click();
+		cy.get('[data-testid="trial-session-working-copy-filter-statusUnassigned').find('span').invoke('text').then((text) => {
+			expect(text).to.equal(' (3)');
+		});
```
Mwindo: It feels like we are testing the style when we check for the preceding space. I would just expect the text to contain `(3)`. Not sure it is a big deal though.

ChristieLaferriereTaxCourt (approved):

PR #6204, opened 2025-06-27
Dependency updates 6-26-25

zacharybrown204:


zacharybrown204:
Dependencies successfully built and deployed on exp1
https://app.circleci.com/pipelines/github/ustaxcourt/ef-cms/10879/workflows/686495ce-5bc4-464c-86bb-109475884445

web-api/terraform/modules/batch/docker-image/package.json
```diff
@@ -10,14 +10,14 @@
   "license": "ISC",
   "description": "",
   "devDependencies": {
-    "@types/node": "22.15.30"
+    "@types/node": "24.0.4"
```
jimlerza: Do we want to lock this along with our app's node version?

jimlerza (approved):

web-api/terraform/modules/batch/docker-image/package.json
TomElliottFlexion: Ditto

zacharybrown204: ahh good catch, yeah I think we should. I'll go revert that one

web-api/terraform/modules/batch/batch.tf
```diff
@@ -103,7 +103,7 @@ EOF
 }
 
 resource "aws_batch_compute_environment" "aws_batch_compute_environment" {
-  compute_environment_name = "compute_environment_${var.environment}_${var.current_color}_${var.region}"
+  name = "compute_environment_${var.environment}_${var.current_color}_${var.region}"
```
TomElliottFlexion, line 106: Good find

TomElliottFlexion (approved):
Nice job!

web-api/terraform/modules/batch/docker-image/package.json
zacharybrown204: Okay I've reverted that change. Since it was just a dev dependency in the docker image I don't think we'll need to retest the deployment.

PR #6206, opened 2025-06-27
Devex: Standardize Postgres app code and enforce types

Mwindo:
The goal of this PR is to polish up some existing Postgres work to make the transition to Gunnison easier.

- Ensures that every table so far has a mapper file with toDatabase/fromDatabase functions.
- Ensures that our mapper functions are typed accurately.
- Tries to enforce some consistent nomenclature.

Notes:

- I removed `updateMessage` and relied on `upsertMessages` instead, because it was really doing an upsert.


web-api/src/persistence/postgres/caseCorrespondences/schema.ts
```diff
@@ -1,13 +1,13 @@
 import { CaseTable } from '@web-api/persistence/postgres/cases/schema';
-import { Selectable, Insertable, Updateable } from 'kysely';
+import { Selectable, Insertable } from 'kysely';
 
 const DEFAULT = {};
 
 export const caseCorrespondenceTableDefinition = {
-  archived: DEFAULT as boolean | undefined,
+  archived: DEFAULT as boolean | null,
```
Mwindo, line 7: `undefined` is not a Postgres type. While there are Kysely-specific reasons we could use `undefined`, I do not believe we are leveraging them, and I think it is probably better to use `null`.

web-api/src/persistence/postgres/caseCorrespondences/schema.ts
```diff
@@ -21,5 +21,3 @@ export const DW_CASE_CORRESPONDENCE_COLUMNS = Object.keys(
 
 export type CaseCorrespondenceKysely = Selectable<CaseCorrespondenceTable>;
 export type NewCaseCorrespondenceKysely = Insertable<CaseCorrespondenceTable>;
-export type UpdateCaseCorrespondenceKysely =
```
Mwindo, line 24: In general, I think it is safer for us to specify the exact fields we want to update--which is already type-safe (example [here](https://github.com/ustaxcourt/ef-cms/blob/staging/web-api/src/persistence/postgres/messages/markMessageThreadRepliedTo.ts))--than to have these updateable types and any mappers for them (e.g., `toUpdateCaseCorrespondenceKysely`).

Suppose you want to update table X and you pass in `{this: true, that: undefined}` to `toUpdateXKysely`. In our existing codebase, how should `toUpdateXKysely` handle that `undefined`? There is no way to say for certain:
- Maybe you _do not_ want to update `that`. Then `toUpdateXKysely` should keep `that` as `undefined` (since `undefined` values leave the existing data untouched).
- Maybe you _want_ to update `that` to be "unset" (`null`). Then `toUpdateXKysely` should map `undefined` to `null` (since `null` will overwrite the data in the db).

Our codebase in not in a state to be able to arbitrate between these two cases effectively.

PR #6208, opened 2025-06-27
10670 update browser modal alert test

JayFlexy:
See [10670](https://github.com/orgs/flexion/projects/11/views/1?pane=issue&itemId=112767198&issue=flexion%7Cef-cms%7C10670)

### DIFF
--- a/web-client/src/views/FileHandlingHelpers/pdfValidation.ts
+++ b/web-client/src/views/FileHandlingHelpers/pdfValidation.ts
@@ -8,6 +8,11 @@ import {
 } from '@web-client/views/FileHandlingHelpers/pdfValidationHelpers';
 import { getPdfJs } from '@shared/business/utilities/pdfs/getPdfJs';
 
+export const UNSUPPORTED_BROWSER_ERROR_MESSAGE =
+  'We noticed you are on an older or unsupported browser. For security reasons, this request could not be completed. \
+  You can try updating the browser or using a different browser or machine. If you need help uploading \
+  this document, contact DAWSON support.';
+
 export const PDF_PASSWORD_PROTECTED_ERROR_MESSAGE =
   'The file is encrypted or password protected. Remove encryption or password protection and try again.';
 export const PDF_CORRUPTED_ERROR_MESSAGE =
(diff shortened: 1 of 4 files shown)


web-client/src/views/FileHandlingHelpers/pdfValidation.ts
```diff
@@ -21,6 +26,63 @@ export const validatePdf = ({
 }: {
   file: File;
 }): Promise<FileValidationResponse> => {
+  const isBrowserSupported = (): boolean => {
```
Mwindo: I love the idea here, but I don't like this solution.

- We are relying on the User Agent string and regex parsing, which is thorny and brittle.
- Some older browsers might, in fact, work, even if we think they shouldn't. In other words, we are aborting early before confirming there is actually a problem.
- We have to keep these user agents up-to-date.

I have two suggestions:

- Use feature detection rather than user agent parsing. See the commented out code in https://github.com/ustaxcourt/ef-cms/blob/staging/shared/src/business/utilities/pdfs/getPdfJs.ts for how we did this before.
- Rather than aborting early, let the user at least try to do the thing. On failure, check if the browser is old via that feature detection. If so, message accordingly.

JayFlexy: @Mwindo Going to head over to discuss

web-client/src/views/FileHandlingHelpers/pdfValidation.ts
```diff
@@ -21,6 +29,20 @@ export const validatePdf = ({
 }: {
   file: File;
 }): Promise<FileValidationResponse> => {
+  if (!clientSupportsES2022()) {
```
Mwindo, line 32: I do not think we should show this error _until_ the pdf process itself fails. As is, this says, "Don't even try to do the thing, it's an old browser." I think a better user experience is, "Try to do the thing, but if it fails, check if it is an old browser and message accordingly." Does that make sense? But I could be missing something.

Mwindo: Also, is it worth caching `clientSupportsES2022`? (Not sure! Just wondering.)

JayFlexy: > I do not think we should show this error _until_ the pdf process itself fails. As is, this says, "Don't even try to do the thing, it's an old browser." I think a better user experience is, "Try to do the thing, but if it fails, check if it is an old browser and message accordingly." Does that make sense? But I could be missing something.

So we attempted to put this in many different places downstream after the file read attempt. However, it never caught it properly, or at all really, the same way as putting at the top. I see what you're saying here and if this was any other situation other than dealing with known and documented breakage of safari for pdf uploads we would definitely spend more time checking downstream relative to the read. 

Since, we are unsure of how this breaks older Safari versions, I'd rather not waste browser resources parsing a potentially faulty or large PDF and risk having the tab hang (which did happen on occasion with the older safari browsers). This was truly an odd bug for Safari, imo.

Mwindo: I think I need to understand this better. The current code, while it prevents Safari, could also prevent any other browsers that might happen to work.

web-client/src/views/FileHandlingHelpers/pdfValidation.ts
```diff
@@ -29,14 +51,29 @@ export const validatePdf = ({
...
       if (!result || typeof result === 'string') {
-        resolve({
-          errorInformation: {
-            errorMessageToDisplay: GENERIC_FILE_ERROR_MESSAGE,
-            errorMessageToLog: `${GENERIC_FILE_ERROR_MESSAGE} (Failed to read file as ArrayBuffer.)`,
-            errorType: ErrorTypes.UNKNOWN,
-          },
-          isValid: false,
-        });
+        if (!clientSupportsES2022()) {
```
Mwindo, line 54: If it turns out we have to put `clientSupportsES2022` above, is putting here too redundant?

JayFlexy: This was fixed/removed. In the new PR [here](https://github.com/ustaxcourt/ef-cms/pull/6243).

PR #6219, opened 2025-06-30
Dependencies 6 26 25

zacharybrown204:


package.json
```diff
@@ -54,11 +54,11 @@
     "aws-xray-sdk": "3.10.3",
     "axios": "1.10.0",
     "broadcast-channel": "7.1.0",
-    "canvas": "3.1.0",
+    "canvas": "3.1.2",
     "cerebral": "5.2.1",
     "classnames": "2.5.1",
     "cookie": "1.0.2",
-    "core-js": "3.42.0",
+    "core-js": "^3.43.0",
```
TomElliottFlexion: Probably doesn't matter in exp1, but we want to pin dependencies so that the exact same version is deployed each time

zacharybrown204: Good call I had missed that when going back over some trouble children in the dependencies. I've removed them.

TomElliottFlexion (approved):

PR #6231, opened 2025-07-01
BUG: OpenSearch queries that sort on `sortableDocketNumber` stopped working

jimlerza:


zachrog (approved):

web-api/src/persistence/elasticsearch/searchClient.ts
```diff
@@ -117,52 +119,52 @@ export const searchAll = async ({
...
-      index,
-      size,
     });
     const hits = get(chunk, 'body.hits.hits', []);
 
     if (hits.length > 0) {
       results = results.concat(hits);
       search_after = hits[hits.length - 1].sort;
     }
+    // i += hits.length;
```
TomElliottFlexion, line 167: Not needed?

TomElliottFlexion (approved):

web-api/src/persistence/elasticsearch/searchClient.ts
jimlerza: Good question. I wrote this code originally, and all the way back then, @mmarcotte had advocated for the version that is commented out, and I opted to play it safe in fear of infinite loops.

But now I think an infinite loop would've exposed the problem I was debugging a lot sooner.

web-api/src/persistence/elasticsearch/searchClient.ts
```diff
@@ -117,52 +119,52 @@ export const searchAll = async ({
...
         query,
       },
       index,
-    });
-  } catch (searchError) {
-    applicationContext.logger.error(searchError);
-    throw new Error('Search client encountered an error.');
-  }
+    },
+  });
```
jimlerza, line 138: The `count()` method will wrap the query in a try-catch so doing it here is not necessary.

web-api/src/persistence/elasticsearch/searchClient.ts
jimlerza, line 1: These changes add some more types to get rid of squigglies in our search queries and hopefully make it easier to remove `applicationContext` from this file someday.

PR #6232, opened 2025-07-01
BUG: OpenSearch queries that sort on `sortableDocketNumber` stopped working (to test)

jimlerza:


web-api/src/persistence/elasticsearch/searchClient.ts
```diff
@@ -3,13 +3,18 @@ import { get } from 'lodash';
 import { getIndexNameFromAlias } from '../../../elasticsearch/elasticsearch-aliases';
 import { unmarshall } from '@aws-sdk/util-dynamodb';
 import { updateIndex } from '@web-api/persistence/elasticsearch/helpers/getIndexName';
-import { ServerApplicationContext } from '@web-api/applicationContext';
+import { type ServerApplicationContext } from '@web-api/applicationContext';
```
akuny: Using [this import method](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export) means that ServerApplicationContext actually won't be pulled into this file at runtime, correct? If so, maybe we should consider using this import syntax elsewhere as well, down the road.

akuny (approved):

zachrog (approved):

TomElliottFlexion (approved):

web-api/src/persistence/elasticsearch/searchClient.ts
jimlerza: Excellent suggestion!

PR #6236, opened 2025-07-01
BUG: OpenSearch queries that sort on `sortableDocketNumber` stopped working (to test)

jimlerza:


jimlerza:
This [super helpful suggestion](https://github.com/ustaxcourt/ef-cms/pull/6232#discussion_r2178031242) from @akuny in a previous PR was implemented in [this commit](https://github.com/ustaxcourt/ef-cms/pull/6236/commits/287cbd2b1643990a9a9aebe17c40b091fe0001a3).

PR #6241, opened 2025-07-01
10684 bug: Large Case Data Causing 502

Mwindo:
**TL;DR:**

- Put a bandaid on https://github.com/flexion/ef-cms/issues/10684
- Fix the issue where we nest docket entries on work items and vice-versa (the affectionately named "spiderman" problem)

bekphnqftcb41

---

**Details:**

There are all sorts of ways to fix this. This PR represents one, very biased approach: We will fix the the large case problem by fixing the vexed relationship between work items and docket entries, which has led to issue like https://github.com/flexion/ef-cms/issues/10683. Instead of returning a bunch of work item data per docket entry, we will return only a few attributes we need for the UI to function properly, with the side effect that our response size for cases will go down.

The general approach:
- Normalize docket entry data. If you need docket entry data, you get it from `dwDocketEntry`. This means no more nested `docketEntry` on `dwWorkItem`.
- Only get work item data when you need it. Do not default to always getting it in `getCaseByDocketNumber`. Likewise, only attach the relevant work item data to docket entries as needed. This now happens in `getCaseInteractor`.
- Update some types. This is mostly around the types for `Case.docketEntries`. They are, admittedly, not the most _amazing_ types ... but this is just a result of our current types. They are at least better: for a `Case`, we now know when we have `DocketEntry` vs. `RawDocketEntry` instead of `any[]`.

Note: see https://github.com/ustaxcourt/ef-cms/pull/6362 for notes about updates to setting the `section` for work items.


shared/src/business/entities/cases/Case.ts
```diff
@@ -990,14 +990,16 @@ export class Case extends JoiValidationEntity {
     });
   }
 
-  toRawObject(processPendingItems = true) {
+  //@ts-ignore
```
Mwindo, line 981: What is the idea here?

Sometimes, we are using a `RawCase`. Sometimes, we are using a full `Case` entity.

In the former situation, we don't want to do `aRawCase.docketEntries[0].someMethod()` -- it will fail because `docketEntries[0]` is a `RawDocketEntry`! -- so it seems dangerous to type `docketEntries` as `DocketEntry`.

In the latter case, we don't want to do new `DocketEntry(aCaseEntity, {authorizedUser}).someMethod()` -- `aCaseEntity` has `docketEntries` that are already `DocketEntry` types, so this is redundant! -- so it seems better to type `docketEntries` as `DocketEntry`.

Shucks!

We could do a type like `(DocketEntry | RawDocketEntry)[]`, but now we never really know what the type is.

So my thought was to dynamically type `docketEntries`: if we have a `Case` entity, the `docketEntries` are of type `DocketEntry`, and if we have a `RawCase`, the docketEntries are of type `RawDocketEntry`. To do this really well would involve a bunch of type wrangling, and, frankly, I didn't think it was worth it. It seemed more important to get the typing of `docketEntries` correct than to mess with all the `toRawObject` code and make something complex.

I can try to do this better if someone really wants to. The basic issue is `RawCase` is no longer quite `ExcludeMethods<Case>`, which is what `toRawObject` wants.

shared/src/business/entities/DocketEntry.ts
```diff
@@ -703,7 +715,7 @@ export class DocketEntry extends JoiValidationEntity {
   ): boolean => {
     if (!entry.isFileAttached) return false;
 
-    const petitionDocketEntry = getPetitionDocketEntry(rawCase);
+    const petitionDocketEntry = getPetitionDocketEntry(rawCase)!;
```
Mwindo, line 692: This exclamation mark is a white lie because _technically_ we could return `undefined`. That said, throwing a not found error below this caused a whole lot of tests to fail because data wasn't mocked well, and I preferred to tell a white lie than spend hours fixing it.

web-api/src/persistence/postgres/cases/getCaseByDocketNumber.ts
```diff
@@ -32,13 +32,11 @@ export const getCaseByDocketNumber = async ({
     throw new NotFoundError(`Case ${docketNumber} not found`);
   }
 
-  const [caseCorrespondences, workItems, caseItemsRaw] = await Promise.all([
+  const [caseCorrespondences, docketEntries, caseItemsRaw] = await Promise.all([
     getCaseCorrespondenceByDocketNumber({
       docketNumber,
     }),
-    getWorkItemsByDocketNumber({
```
Mwindo, line 39: The only reason we fetched work items here was to attach some information to docket entries. The approach in this PR is to _not_ fetch work items unless we need them, and attach the information we need when we need to.

shared/src/business/entities/DocketEntry.ts
```diff
@@ -475,6 +474,19 @@ export class DocketEntry extends JoiValidationEntity {
     return DocketEntry.isCourtIssued({ eventCode: this.eventCode });
   }
 
+  static attachWorkItemInfoForUI(
```
Mwindo: Yuck!

I agree. What is the purpose of this?

The `DocketEntry` entity is shared on the front end and back end. This is a double-edge sword: consistency is great, but coupling logic between front end and back end is less great.

On the UI, we associate some data with docket entries to determine, e.g., what icons need to display. This really feels like it "belongs" on the UI docket entry: we are saying things like, "This docket entry has been qced." However, on the back end, we don't care about this: this is all work-item land. We really want to say, "Hey, front end, your docket entries should have this extra data that is gotten from work items; but back end, you don't need to worry about it." I don't know of a great, simple way to do this super well, and I think the current implementation here is at least better than it was before (the spiderman problem).

There are alternatives, e.g.:
- We could _always_ attach work item info. This is less efficient.
- Rather than attaching them on the back end response object, we could return work items separately and stitch them to the docket entries on the front end. There are a few ways to do this, but they feel worse: again, this data really feels like it "belongs" the docket entry.

Happy to be persuaded to a different approach.

shared/src/business/useCases/getCaseInteractor.ts
```diff
@@ -7,23 +7,31 @@ import {
...
 import { getCaseByDocketNumber } from '@web-api/persistence/postgres/cases/getCaseByDocketNumber';
+import { getWorkItemsByDocketNumber } from '@web-api/persistence/postgres/workitems/getWorkItemsByDocketNumber';
+import { UIDocketEntry } from '@shared/business/entities/DocketEntry';
 
 export const getCaseInteractor = async (
   applicationContext: ServerApplicationContext,
   { docketNumber }: { docketNumber: string },
   authorizedUser: UnknownAuthUser,
-) => {
+): Promise<UICase> => {
```
Mwindo: The type here is a bit of a lie since it could be a `RawPublicCase` or `RawRestrictedCase` ... but it's no more of a lie than we already have given that we were saying that the client-side `caseDetail` was always a `RawCase`. (Now we say it is a `UICase`.)

The idea with `UIDocketEntry` and `UICase` is the following: on the UI, we need certain work item info attached to docket entries. I thought about a variety of approaches, including adding these attributes to `DocketEntry` directly or returning work items separately to stitch together the data on the front end. The approach I settled on: attach the work item data we need to an instance of a new type, a `UIDocketEntry`, which represents data that we need on the UI but not the back end. `UICase` is just `RawCase` with differently type docket entries (`UIDocketEntry`).

web-client/integration-tests/sentWorkItemsExpireAfterNDays.test.ts
```diff
@@ -60,6 +61,21 @@ describe('verify old sent work items do not show up in the outbox', () => {
     workItemIdN = applicationContext.getUniqueId();
     workItemIdNPlus1 = applicationContext.getUniqueId();
 
+    const docketEntry: RawDocketEntry = {
```
Mwindo, line 64: Before, we relied on the nested docket entry saved as part of work item: we did not have to have a "real" docket entry saved into the db. Now, because our query must join `dwWorkItem` with `dwDocketEntry` to get the docket entry data, we need a "real" docket entry saved in the db. The same goes for the test `servedWorkItemsExpireAfter7Days`.

shared/src/business/entities/DocketEntry.ts
```diff
@@ -185,7 +188,11 @@ export class DocketEntry extends JoiValidationEntity {
   public signedJudgeUserId?: string;
   public strickenBy?: string;
   public strickenByUserId?: string;
-  public workItem?: any;
+
+  // These are optional fields set for the UI
+  public qcComplete?: boolean;
+  public qcViewed?: boolean;
+  public workItemId?: string;
```
Mwindo, line 192: I would love to just remove these fields from `DocketEntry.ts` altogether ... but `genericHandler` makes this difficult.

web-api/src/persistence/postgres/cases/getCaseByDocketNumber.ts
```diff
@@ -16,12 +13,9 @@ export const getCaseByDocketNumber = async ({
   includeConsolidatedCases?: boolean;
   user?: UnknownAuthUser;
 }): Promise<RawCase> => {
-  const [caseData, workItems] = await settlePromises([
-    getCasesByDocketNumbers({
-      docketNumbers: [docketNumber],
-    }),
-    getWorkItemsByDocketNumber({ docketNumber }),
```
Mwindo, line 23: The only reason we fetched work items here was to attach some information to docket entries. The approach in this PR is to _not_ fetch work items unless we need them, and attach the information we need when we need to.

TomElliottFlexion (approved):
Approval based mostly on my review when this went to test.
Nice job @Mwindo!

jimlerza (approved):

PR #6247, opened 2025-07-03
10684 bug: Large Case Data Causing 502

Mwindo:
Holding off on merging for now since we have a lot in test.

**TL;DR:**

- Put a bandaid on https://github.com/flexion/ef-cms/issues/10684
- Fix the issue where we nest docket entries on work items and vice-versa (the affectionately named "spiderman" problem)

bekphnqftcb41

---

**Details:**

There are all sorts of ways to fix the issue. This PR represents one, very biased approach: We will mitigate the large case problem by fixing the vexed relationship between work items and docket entries, which has led to issues like https://github.com/flexion/ef-cms/issues/10683. Instead of returning a bunch of work item data per docket entry, we will return only a few attributes we need for the UI to function properly, with the side effect that our response size for cases will go down (rough calculations suggest ~.4-.5 MB).

The general approach:
- Normalize docket entry data. If you need docket entry data, you get it from `dwDocketEntry`. This means no more nested `docketEntry` on `dwWorkItem`.
- Only get work item data when you need it. Do not default to always getting it in `getCaseByDocketNumber`. Likewise, only attach the relevant work item data to docket entries as needed. This now happens in `getCaseInteractor`.
- Update some types. This is mostly around the types for `Case.docketEntries`. They are, admittedly, not the most _amazing_ types ... but this is just a result of our current types. They are at least better: for a `Case`, we now know when we have `DocketEntry` vs. `RawDocketEntry` instead of `any[]`. I have tried to fix the type errors that resulted from going to `any`, but in cases that were non-trivial and might necessitate changes that lead to different app behavior, I have used `@ts-ignore`.

Currently based off of 10494


shared/src/business/entities/cases/Case.ts
```diff
@@ -993,14 +993,16 @@ export class Case extends JoiValidationEntity {
     });
   }
 
-  toRawObject(processPendingItems = true) {
+  //@ts-ignore
```
Mwindo, line 996: What is the idea here?

Sometimes, we are using a `RawCase`. Sometimes, we are using a full `Case` entity.

In the former situation, we don't want to do `aRawCase.docketEntries[0].someMethod()` -- it will fail because `docketEntries[0]` is a `RawDocketEntry`! -- so it seems dangerous to type `docketEntries` as `DocketEntry`.

In the latter case, we don't want to do new `DocketEntry(aCaseEntity, {authorizedUser}).someMethod()` -- `aCaseEntity` has `docketEntries` that are already `DocketEntry` types, so this is redundant! -- so it seems better to type `docketEntries` as `DocketEntry`.

Shucks!

We could do a type like `(DocketEntry | RawDocketEntry)[]`, but now we never really know what the type is.

So my thought was to dynamically type `docketEntries`: if we have a `Case` entity, the `docketEntries` are of type `DocketEntry`, and if we have a `RawCase`, the docketEntries are of type `RawDocketEntry`. To do this really well would involve a bunch of type wrangling, and, frankly, I didn't think it was worth it. It seemed more important to get the typing of `docketEntries` correct than to mess with all the `toRawObject` code and make something complex.

I can try to do this better if someone really wants to. The basic issue is `RawCase` is no longer quite `ExcludeMethods<Case>`, which is what `toRawObject` wants.

web-client/integration-tests/sentWorkItemsExpireAfterNDays.test.ts
```diff
@@ -60,6 +61,21 @@ describe('verify old sent work items do not show up in the outbox', () => {
     workItemIdN = applicationContext.getUniqueId();
     workItemIdNPlus1 = applicationContext.getUniqueId();
 
+    const docketEntry: RawDocketEntry = {
```
Mwindo, line 64: The type here is a bit of a lie since it could be a `RawPublicCase` or `RawRestrictedCase` ... but it's no more of a lie than we already have given that we were saying that the client-side `caseDetail` was always a `RawCase`. (Now we say it is a `UICase`.)

The idea with `UIDocketEntry` and `UICase` is the following: on the UI, we need certain work item info attached to docket entries. I thought about a variety of approaches, including adding these attributes to `DocketEntry` directly or returning work items separately to stitch together the data on the front end. The approach I settled on: attach the work item data we need to an instance of a new type, a `UIDocketEntry`, which represents data that we need on the UI but not the back end. `UICase` is just `RawCase` with differently type docket entries (`UIDocketEntry`).

shared/src/business/entities/DocketEntry.ts
```diff
@@ -895,3 +886,9 @@ export const getServedPartiesCode = (servedParties?: any[]) => {
 declare global {
   type RawDocketEntry = ExcludeMethods<DocketEntry>;
 }
+
+export type UIDocketEntry = DocketEntry & {
+  workItemId?: string;
+  qcComplete?: boolean;
+  qcViewed?: boolean;
+};
```
Mwindo: This is the minimum set of work item data per docket entry that we need on the UI.

shared/src/business/entities/DocketEntry.ts
```diff
@@ -189,7 +188,11 @@ export class DocketEntry extends JoiValidationEntity {
   public signedJudgeUserId?: string;
   public strickenBy?: string;
   public strickenByUserId?: string;
-  public workItem?: any;
+
+  // These are optional fields set solely for the UI in certain cases.
+  public qcComplete?: boolean;
+  public qcViewed?: boolean;
+  public workItemId?: string;
```
Mwindo, line 195: This represents the minimum amount of data the UI needs associated with a docket entry.

I wanted to make an entirely new type `UIDocketEntry` that was `RawDocketEntry & {these fields}` to avoid further cluttering `DocketEntry`, but `genericHandler`, which reconstructs entities, made this difficult: without the fields here on the entity, they were wiped away when sending data back to the client (since `genericHandler` said, "Oh look, a `Case`, let me reconstruct it from scratch--oops, dropped those custom fields of yours when rebuilding the docket entries!"). With the fields thus needing to be here in `DocketEntry` anyway (short of refactoring `genericHandler`--and I feel like I am already taking a big enough detour for 10684 in my "fix the spiderman problem approach"), it felt useless to have the new type.

shared/src/business/entities/WorkItem.ts
```diff
@@ -14,7 +13,7 @@ export class WorkItem extends JoiValidationEntity {
   public completedByUserId?: string;
   public completedMessage?: string;
   public createdAt: string;
-  public docketEntry: any;
```
Mwindo, line 17: We will remove all uses of `WorkItem` `docketEntry` in app code and drop the column in a followup PR.

shared/src/business/entities/DocketEntry.ts
```diff
@@ -705,7 +705,7 @@ export class DocketEntry extends JoiValidationEntity {
   ): boolean => {
     if (!entry.isFileAttached) return false;
 
-    const petitionDocketEntry = getPetitionDocketEntry(rawCase);
+    const petitionDocketEntry = getPetitionDocketEntry(rawCase)!;
```
Mwindo, line 708: The exclamation point is a white lie to avoid fixing a whole bunch of tests with poorly mocked data.

web-api/src/persistence/postgres/workitems/getDocumentQCInboxForUser.ts
```diff
@@ -1,37 +1,57 @@
...
+export type WorkItemWithCaseInfo = Omit<RawWorkItem, 'docketEntry'> & {
   caseTitle?: string;
   caseStatus?: string;
   leadDocketNumber?: string;
   trialDate?: string;
   trialLocation?: string;
+  docketEntry: RawDocketEntry;
+};
+
+export const workItemQCQueryBase = (dbReader: Kysely<Database>) => {
```
Mwindo, line 33: All of our queries for document qc were very similar, so I extracted them here.

web-api/src/persistence/postgres/workitems/getDocumentQCInboxForUser.ts
```diff
@@ -1,37 +1,57 @@
...
+    )
+    .selectAll('w')
+    .select(eb => [
+      jsonObjectFrom(
+        eb
+          .selectFrom('dwDocketEntry as de')
+          .selectAll()
+          .whereRef('de.docketEntryId', '=', 'w.docketEntryId')
+          .whereRef('de.docketNumber', '=', 'w.docketNumber'),
+      ).as('docketEntry'),
```
Mwindo, line 50: This `jsonObjectFrom` says "I want to join on `dwDocketEntry`, but give me the `dwDocketEntry` stuff as a type-safe json object version of `dwDocketEntry`."

web-client/src/presenter/actions/EditDocketRecordEntry/setDocketEntryMetaFormForEditAction.ts
```diff
@@ -22,12 +22,16 @@ export const setDocketEntryMetaFormForEditAction = ({
...
 
+  if (!documentDetail) {
+    throw new Error(
+      `Could not find docket entry with index ${docketRecordIndex}`,
+    );
+  }
+
   store.set(state.docketRecordIndex, docketRecordIndex);
 
-  documentDetail.filersMap = {};
```
Mwindo, line 27: `documentDetail` is a `DocketEntry`. We were just sticking on this `filersMap` to it for convenience since `DocketEntry` has no `filersMap`.

shared/src/business/useCases/getCaseInteractor.ts
```diff
@@ -29,5 +36,27 @@ export const getCaseInteractor = async (
...
+      ...docketEntry,
+      qcComplete: !!workItem?.completedAt,
+      qcViewed: !!workItem?.isRead,
+      workItemId: workItem?.workItemId,
+    };
+  });
+  return {
+    ...theCase.toRawObject(),
+    docketEntries: docketEntriesWithUIInfo,
+  } as RawCase;
```
Mwindo, line 61: As far as I could tell, the UI only needs the work item data when calling this interactor (which corresponds to `getCaseAction` on the client).

web-api/src/persistence/postgres/workitems/getDocumentQCInboxForUser.ts
```diff
@@ -1,37 +1,57 @@
...
   trialDate?: string;
   trialLocation?: string;
+  docketEntry: RawDocketEntry;
+};
+
+export const workItemQCQueryBase = (dbReader: Kysely<Database>) => {
+  return dbReader
+    .selectFrom('dwWorkItem as w')
+    .leftJoin('dwCase as c', 'c.docketNumber', 'w.docketNumber')
+    .innerJoin('dwDocketEntry as d', join =>
```
Mwindo, line 37: Inner join because if you don't got a docket entry, somethin's amiss.

shared/src/business/useCases/saveSignedDocumentInteractor.ts
```diff
@@ -94,13 +94,13 @@ export const saveSignedDocumentInteractor = async (
...
-    docketEntry => docketEntry.docketEntryId === originalDocketEntryId,
-  );
+  const originalDocketEntryEntity = caseEntity.getDocketEntryById({
+    docketEntryId: originalDocketEntryId,
+  });
 
   let signedDocketEntryEntity;
   if (
-    originalDocketEntryEntity.documentType === 'Proposed Stipulated Decision'
+    originalDocketEntryEntity?.documentType === 'Proposed Stipulated Decision'
```
akuny: We have an opportunity to swap `TRACKED_DOCUMENT_TYPES.proposedStipulatedDecision.documentType` for the magic string here.

shared/src/business/useCases/getCaseInteractor.ts
akuny: This chunk of code seems important enough to be under unit test in and of itself, maybe extracted as a "use case helper" in DAWSON terminology?

shared/src/business/entities/cases/Case.ts
```diff
@@ -136,7 +136,7 @@ export class Case extends JoiValidationEntity {
   public canAllowPrintableDocketRecord?: boolean;
   public canDojPractitionersRepresentParty?: boolean;
   public archivedDocketEntries?: RawDocketEntry[];
-  public docketEntries: any[];
+  public docketEntries: DocketEntry[];
```
akuny, line 139: This in itself is a big win!

web-api/src/business/useCases/docketEntry/addPaperFilingInteractor.ts
```diff
@@ -174,7 +169,7 @@ export const addPaperFiling = async (
       docketEntryId,
     });
     const electronicParties =
-      currentDocketEntry.eventCode === 'ATP' ? [] : undefined;
+      currentDocketEntry?.eventCode === 'ATP' ? [] : undefined;
```
akuny: `INITIAL_DOCUMENT_TYPES.attachmentToPetition.eventCode`?

web-client/src/presenter/actions/saveCaseDetailInternalEditAction.ts
```diff
@@ -26,6 +26,12 @@ export const saveCaseDetailInternalEditAction = async ({
             document.eventCode === INITIAL_DOCUMENT_TYPES.petition.eventCode,
         );
 
+        if (!oldPetitionDocument) {
```
akuny, line 29: Does this guard clause merit an update to this file's unit tests? Applies elsewhere in Cerebral actions too.

akuny (approved):
Outstanding!

shared/src/business/useCases/saveSignedDocumentInteractor.ts
Mwindo: Thank you!

web-client/src/presenter/actions/saveCaseDetailInternalEditAction.ts
Mwindo: My gut reaction: Noooo!

My more measured response: It's probably the _right_ thing to do.

My opinion after thinking further: We are already breaking in these situations, we are just pretending we aren't because of the lack of types. The guard here is a way of at least failing in a more "graceful" way. My worry is that adding these tests everywhere contributes to making the system less flexible. (Maybe `getDocketEntryById` could throw an error so it could be tested in one place? But I think returning `undefined` makes more sense.) If we don't find these docket entries, we will throw an error one way or another; I am not sure it is worth crystallizing the exact error message. 

I guess it does contribute to code coverage though :/

I'm thinking about this. I won't merge it for a couple of days at least since `test` is a bit clogged atm.

shared/src/business/useCases/saveSignedDocumentInteractor.ts
Mwindo: I went with a different existing constant since this one wasn't exported :)

shared/src/business/useCases/getCaseInteractor.ts
Mwindo: Good call--I have added a unit test for the interactor. I don't want to consecrate it with much more than this just yet because, frankly, I don't love the solution, it's just the best I could think of at the moment. So I've kept it constrained to the interactor and not a helper for now.

docs/postgres/connect-to-deployed-db.md
```diff
@@ -29,7 +29,7 @@ Note: run all commands from the root of the `ef-cms` directory.
       ```bash
       scripts/postgres/generate-token.sh
       ```
-   - To generate a token for the writeable endpoint:
+   - To generate a token for the writeable endpoint:set-env.zs
```
TomElliottFlexion: Nit

shared/src/business/entities/EntityValidationConstants.ts
```diff
@@ -442,7 +442,6 @@ export const DOCKET_ENTRY_VALIDATION_RULE_KEYS = {
     otherwise: joi.optional(),
     then: joi.required(),
   }),
-  workItem: joi.object().optional(),
```
TomElliottFlexion, line 445: I know the new fields are optional, so I can see both sides: is it worth adding `qcComplete`, `qcViewed` and/or `workItemId` to the the joi validation?

web-api/src/business/useCases/docketEntry/addPaperFilingInteractor.ts
```diff
@@ -174,7 +170,10 @@ export const addPaperFiling = async (
       docketEntryId,
     });
     const electronicParties =
-      currentDocketEntry.eventCode === 'ATP' ? [] : undefined;
+      currentDocketEntry?.eventCode ===
+      INITIAL_DOCUMENT_TYPES.attachmentToPetition.eventCode
```
TomElliottFlexion, line 174: Nice

web-api/src/business/useCases/document/batchDownloadDocketEntriesInteractor.ts
```diff
@@ -34,23 +35,25 @@ export const batchDownloadDocketEntriesInteractor = async (
...
-    await applicationContext.getNotificationGateway().sendNotificationToUser({
-      applicationContext,
-      clientConnectionId,
-      message: {
-        action: 'batch_download_error',
-        error,
-      },
-      userId,
-    });
+    if (isAuthUser(authorizedUser)) {
```
TomElliottFlexion, line 45: Tiny nitpick: would we rather check the authuser before calling `batchDownloadDocketEntriesHelper` than before sending the error message? Might be missing something

web-api/src/business/useCases/externalDocument/fileExternalDocumentInteractor.test.ts
```diff
@@ -411,7 +411,6 @@ describe('fileExternalDocumentInteractor', () => {
     expect(
       applicationContext.getUseCaseHelpers().sendServedPartiesEmails,
     ).not.toHaveBeenCalled();
-    expect(updatedCase!.docketEntries[3].status).toBeUndefined();
```
TomElliottFlexion, line 414: Woah

web-api/src/business/useCases/saveCaseDetailInternalEditInteractor.ts
```diff
@@ -144,18 +146,30 @@ export const saveCaseDetailInternalEdit = async (
...
-      inProgress: true,
+    const petitionWorkItem = await getWorkItemByDocketNumberAndDocketEntryId({
+      docketNumber,
+      docketEntryId: petitionDocketEntry?.docketEntryId,
     });
 
-    await upsertWorkItems({
-      workItems: [workItemEntity.validate().toRawObject()],
-    });
+    if (petitionWorkItem) {
```
TomElliottFlexion: Should we throw an error if no petitionWorkItem exists?

web-api/src/persistence/postgres/cases/getCaseByDocketNumber.ts
```diff
@@ -43,31 +37,10 @@ export const getCaseByDocketNumber = async ({
...
     ...theCase,
     consolidatedCases: consolidatedCases.map(
       c => new ConsolidatedCaseSummary(c),
     ),
   };
 };
-
-// This function is super stupid. We need it because certain spots
-// in the code expect fields from work items on the docket entry.
-function attachWorkItemsToDocketEntries({
```
TomElliottFlexion, line 58: Feels good to see this go

web-api/src/persistence/postgres/workitems/mapper.ts
```diff
@@ -14,6 +14,7 @@ function pickFields(workItem) {
     completedMessage: workItem.completedMessage,
     createdAt: workItem.createdAt,
     docketEntry: JSON.stringify(workItem.docketEntry),
```
TomElliottFlexion, line 16: I don't believe we need this anymore

web-api/src/persistence/postgres/workitems/mapper.ts
```diff
@@ -80,6 +80,7 @@ export function toWorkItemWithCaseInfo(dbWorkItem): WorkItemWithCaseInfo {
     leadDocketNumber: dbWorkItem?.leadDocketNumber || undefined,
     trialDate: dbWorkItem?.trialDate?.toISOString(),
     trialLocation: dbWorkItem?.trialLocation || undefined,
+    docketEntry: dbWorkItem.docketEntry,
```
TomElliottFlexion, line 83: ''

TomElliottFlexion (approved):
This is a feat! Well done

shared/src/business/entities/EntityValidationConstants.ts
Mwindo: Honestly, I really hate these fields, so I just want to pretend like they don't exist so someone can find a better solution after this one.

web-api/src/business/useCases/document/batchDownloadDocketEntriesInteractor.ts
Mwindo: Ah, good point. The reason for this current change was to make a type happy (no use sending a notification if we do not have a userId!). But it's no use doing _anything_ at all if you aren't an auth user.

web-api/src/business/useCases/saveCaseDetailInternalEditInteractor.ts
Mwindo: I will look into this tomorrow.

web-api/src/persistence/postgres/workitems/mapper.ts
Mwindo: Fantastic catch--thank you!

docs/postgres/connect-to-deployed-db.md
Mwindo: 🤦

web-api/src/business/useCases/saveCaseDetailInternalEditInteractor.ts
Mwindo: Narrator: He didn't.

Looking into now.

Mwindo: Yes, we should. Thank you!

web-api/src/business/useCases/document/batchDownloadDocketEntriesInteractor.ts
Mwindo: Actually, I think this is ok as is. The `batchDownloadDocketEntriesHelper` does an authorization check first thing and will throw an error in this case. We can check authuser, but it would be duplicative.

web-api/src/persistence/postgres/workitems/mapper.ts
Mwindo: This one we actually do need: the object we pass in (which will be typed better once [this PR](https://github.com/ustaxcourt/ef-cms/pull/6206) is merged) is based on a join with the docket entry table. `WorkItemWithCaseInfo` needs docket entry info as well.

PR #6250, opened 2025-07-03
10574 add deadline to sro to test

alexdziarn:
https://github.com/orgs/flexion/projects/11/views/1?filterQuery=10574&pane=issue&itemId=91400993&issue=flexion%7Cef-cms%7C10574


web-api/src/business/useCases/courtIssuedDocument/fileAndServeCourtIssuedDocumentInteractor.ts
```diff
@@ -188,12 +188,25 @@ export const fileAndServeCourtIssuedDocument = async (
...
+            'statusReportStipulatedDecision': 'Status Report or Proposed Stipulated Decision Due'
+          };
+          /* 
+            getAutoGeneratedDeadlineDescription gets the description based on docketEntryEntity.eventCode, but 
+            Status Report Orders has an eventCode of 'O' which is very generic.
+          */
+          const description = (form.dueDate && form.orderType) 
+            ? orderDescriptions[form.orderType]
+            : docketEntryEntity.getAutoGeneratedDeadlineDescription();
+
```
Mwindo: This technically works, but it is adding a new control flow that exists parallel to our current control flow and muddies the waters. What do I mean?

- Before this PR, we have `shouldAutoGenerateDeadline`, which, if true, then gets the description via `getAutoGeneratedDeadlineDescription`. (In fact, I am not sure I see the necessity of both of these--why not just the latter, where a non-empty string means `shouldAutoGenerateDeadline` and an empty string means `shouldNotAutoGenerateDeadline`? But that's a slight digression.)
- After this PR, we now do that, but also we say, "Even if `shouldAutoGenerateDeadline()` is `false`, maybe you should still _actually_ auto-generate a deadline." And if so, that description still isn't from `getAutoGeneratedDeadlineDescription`, but from this one-off workaround.

I think the better way to handle this is to update `shouldAutoGenerateDeadline` and `getAutoGeneratedDeadlineDescription`.

Also, I would use our constants for the `statusReport` and `statusReportStipulatedDecision` strings (see [here](https://github.com/ustaxcourt/ef-cms/blob/d758d9a5a7c018a06f84813a6a4978ceedf15097/shared/src/business/entities/EntityConstants.ts#L1819-L1832)). It would probably be best to extend that object with these deadline descriptions.

web-api/src/business/useCases/courtIssuedDocument/fileAndServeCourtIssuedDocumentInteractor.ts
```diff
@@ -155,7 +155,7 @@ export const fileAndServeCourtIssuedDocument = async (
           {
             ...omit(docketEntryToServe, 'filedBy'),
             attachments: form.attachments,
-            date: form.date,
+            date: form.date || createISODateString(form.dueDate),
```
Mwindo: Are we sure that `date` is always meant to be a due date? This is changing persistence, which makes me nervous given what this ticket is doing.

web-api/src/business/useCases/courtIssuedDocument/fileAndServeCourtIssuedDocumentInteractor.ts
Mwindo: (To make this work, you'd have to pass in the `orderType` argument since, as you note, there isn't another way to see if this is a status report order ... which is annoying.)

web-api/src/business/useCases/courtIssuedOrder/fileCourtIssuedOrderInteractor.ts
```diff
@@ -99,6 +99,11 @@ export const fileCourtIssuedOrder = async (
   const docketEntryEntity = new DocketEntry(
     {
       ...documentMetadata,
+      draftOrderState: {
+        ...documentMetadata.draftOrderState,
+        dueDate: documentMetadata.dueDate,
+        orderType: documentMetadata.orderType,
```
Mwindo: 👍

web-api/src/business/useCases/courtIssuedDocument/fileAndServeCourtIssuedDocumentInteractor.ts
Mwindo: Another option to consider in the future: it is weird that we have no way to distinguish a status order. It would be great to be able to call, e.g., `docketEntry.isStatusOrder()`. Are there other orders like this, i.e., ones that have no special event code but are nonetheless "special?" If so, maybe `orderType` or something should be a field on the docket entry table. Again, just something to keep in the back of your mind for the future.

PR #6261, opened 2025-07-07
10589 unable to edit notice of receipt of petition intermediate branch to test 1751915647

AnthonyLoeraTaxCourt:
https://github.com/orgs/flexion/projects/11/views/1?pane=issue&itemId=95708770&issue=flexion%7Cef-cms%7C10589


Mwindo:
This is the behavior I get locally pulling in your branch when I try to update a Notice of Receipt for the first time.

https://github.com/user-attachments/assets/22b18dd1-8cbd-4ab7-8849-e96948e7b41e

web-api/src/business/useCases/docketEntry/updateDocketEntryMetaInteractor.ts
```diff
@@ -123,15 +123,22 @@ export const updateDocketEntryMeta = async (
...
+      shouldGenerateCoversheet = shouldGenerateCoversheetForDocketEntry({
+        certificateOfServiceUpdated,
+        documentTitleUpdated,
+        entryRequiresCoverSheet,
+        filingDateUpdated,
+        originalDocketEntry,
+        servedAtUpdated,
+        shouldAddNewCoverSheet,
+      });
+    }
```
Mwindo: I don't think this is a good idea. I want to try to walk you through my thought process when reviewing this in case it is helpful, because otherwise it might just seem like I have "magic" context.

Me: "This approach ... should work, I think. But are we sure that system-generated orders should never have a cover sheet? That's a question for Chris Holly and Tenille. Ok, fair enough. Let's just assume that's correct for now.

"But _something_ about this gives me pause. Let me think more carefully. What are we trying to solve here? Oh yeah, we were generating a cover sheet when we shouldn't. Why was that happening? Something to do with the document title not being passed through/generated correctly from the form to this wep-api function: the document title was previously `Notice of Receipt of Petition` or something, and now it is `NULL`. Because the document title had 'changed,' we were generating a cover sheet. Ok. So this solution here will technically fix it.

"Well ... wait, why _don't_ we have a title? _Shouldn't_ the document have a title? And come to think of it, won't this change the state of the data? We're in a web-api function after all. Hmm ... I bet if I check what happens to the database after saving I'll see `NULL` overwriting the actual document title. Ah, yep. That's no good."

Screenshot 2025-07-15 at 5 12 38 PM

Mwindo (changes requested):
Thanks, @AnthonyLoeraTaxCourt! I've requested some changes. The solution in the most recent commit will fix the _specific_ issue, but it will lead to other problems. I tried to leave a helpful PR comment explaining how I arrived at this so it seems less like "I have a lot of context that you don't" and more "here are the sort of questions that are useful to ask." If it isn't helpful, feel free to ignore :)

web-client/src/business/useCases/externalDocument/generateExternalDocumentTitle.ts
```diff
@@ -14,6 +15,11 @@ export const generateExternalDocumentTitle = (
   applicationContext: ClientApplicationContext,
   { documentMetadata },
 ) => {
+  const entry = Object.values(SYSTEM_GENERATED_DOCUMENT_TYPES).find(
+    entry => entry.eventCode === documentMetadata.eventCode,
+  );
+  if (entry) return entry.documentTitle;
```
Mwindo, line 21: 👍 This seems better. I would do this and get rid of the changes to `web-api/src/business/useCases/docketEntry/updateDocketEntryMetaInteractor.ts`

web-client/src/views/EditDocketEntry/EditDocketEntryMetaFormDocument.tsx
```diff
@@ -88,24 +89,18 @@ export const EditDocketEntryMetaFormDocument = connect(
...
-                      internalTypesHelper.internalDocumentTypesForSelectWithLegacySorted,
-                    selectedEventCode: form.eventCode,
-                  })
-            }
-            isDisabled={form.eventCode === 'NOTR'}
+            value={reactSelectValue({
+              documentTypes: MERGED_DOCUMENT_TYPES,
+              selectedEventCode: form.eventCode,
+            })}
+            isDisabled={form.filedByRole === 'System'}
```
Mwindo: Can we used the `EntityConstants.ts` `SYSTEM_ROLE` constant?

shared/src/business/entities/EntityConstants.ts
```diff
@@ -909,6 +909,19 @@ export const SYSTEM_GENERATED_DOCUMENT_TYPES = {
   ...AUTO_GENERATED_DEADLINE_DOCUMENT_TYPES_WITH_NAMES,
 };
 
+export const MERGED_DOCUMENT_TYPES = [
```
Mwindo: This name is better than before, but it still feels off. Maybe this is `COURT_ISSUED_DOCUMENT_TYPES`? Not a big deal though.

Mwindo (approved):
Left a couple of comments, but looks good overall. Thank you @AnthonyLoeraTaxCourt!

PR #6265, opened 2025-07-08
devex: remove one-time test

TomElliottFlexion:


.circleci/config.yml
```diff
@@ -1078,9 +1059,6 @@ workflows:
           type: approval
           requires:
             - migrate
-      - index-docket-entries-once-test:
-          requires:
-            - wait-for-migration
```
jimlerza: Rather than delete this, you can fix it by adding the missing config:

```yaml
      - index-docket-entries-once-test:
          <<: *only-deployed-lower-environments
          requires:
            - wait-for-migration
```

PR #6283, opened 2025-07-11
9009 dxox: Transaction support (formerly flexion#10705)

Mwindo:

### Details:

1. We say, `getConnection` and `establishConnection` ... but these are not getting a connection or establishing a connection. They are returning a kysely wrapper object around a client pool. Behind the scenes, kysely is creating/acquiring/releasing/removing connections when we call, e.g., `.execute`.
2. ~We have two connection pools, one for `getConnection` and one for `getLockingConnection`. For the latter especially, we abuse the pool: we say, "Make sure there is only ever one connection in your pool so that we don't accidentally create a lock on one connection and try to remove it with another, which won't work!" But this is not what a connection pool is for. Instead, we should have one connection pool and get an honest-to-goodness connection _from_ the pool and use _that_ throughout the lifetime of the lock.~ Nope. This is complicated.

2. ~We have _one_ connection pool. The equivalent of `getLockingConnection`, `acquireOneDbConnection`, now simply gets a particular connection from the existing pool and uses it throughout the lifetime of the locking code.~ Nope.

Implementation details:


web-api/src/persistence/postgres/utils/transactions.ts
```diff
@@ -0,0 +1,88 @@
...
+    store.onCommitCallbacks.push(cb);
+  } else {
+    throw new Error(
+      'onTransactionCommit was called without an ongoing transaction',
+    );
+  }
+}
+
+// A wrapper to keep everything within it scoped to a transaction
+export async function withTransaction<T>(fn: () => Promise<T>): Promise<T> {
```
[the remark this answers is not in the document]
Mwindo: All parent level transactions are coordinated via a `pg_advisory_lock` so only one transaction is running at a time. Any nested transactions "know" they are nested thanks to `AsyncLocalStorage` and "plug into" the already running transaction.

web-api/src/getConnection.ts
```diff
@@ -1,38 +1,73 @@
-import { CamelCasePlugin, Kysely, PostgresDialect } from 'kysely';
```
Mwindo, line 1: The basic idea of the changes in this file:

1. We only ever want one main connection per lambda, so why are we handling token refresh logic at all? Offload that to the pool since the pool config accepts `async () => Promise<string>`. When the pool connects, it will get a fresh token. Otherwise, we do not need to reconnect.
2. Implement a health check before running a query. If we do not have an expiration mechanism, we need some way to say, "The connection failed, so re-connect." My research suggests that a connection from Lambda to Aurora in the same AZ should be very fast, so I don't think this would measurably impact performance.

web-api/src/persistence/postgres/utils/transactions.ts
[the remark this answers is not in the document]
Mwindo: On failure (at top level or nested), the entire transaction rolls back. Otherwise, once the top level finishes, the transaction is committed.

web-api/src/persistence/postgres/utils/transactions.ts
```diff
@@ -0,0 +1,86 @@
...
+}
+
+// Whether or not the caller is currently part of a transaction
+export function inTransaction() {
+  const store = als.getStore();
+  return !!(store && store.inTransaction);
+}
+
+// Callbacks to run once the commit is successful
+export function onTransactionCommit(cb: () => Promise<void>) {
```
Mwindo, line 51: We do not want to update OpenSearch when we make a query unless we commit the transaction. `onTransactionCommit` allows us to say, "Ok, do X--like update OpenSearch--once the transaction commits."

Mwindo: One thing to test: if we queue up a ton of OpenSearch updates, will we run into memory limits?

web-api/src/persistence/postgres/utils/transactions.ts
```diff
@@ -0,0 +1,46 @@
+import { settlePromises } from '@web-api/utilities/settlePromises';
+import { ConnectionStore, getConnection } from '@web-api/getConnection';
+
+export async function withTransaction<T>(fn: () => Promise<T>): Promise<T> {
+  // If we are already in a transaction, continue like normal.
+  if (inTransaction()) return fn();
```
Mwindo: If we have `withTransaction(() => withTransaction(some callback))`, we only want _one_ transaction, not two: the inner transaction is meaningless in this context (it is already part of an outer transaction). Therefore, we execute as normal in a nested `withTransaction`; the outer `withTransaction` will handle commit/rollback.

web-api/src/persistence/postgres/utils/transactions.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  const db = await getConnection();
+
+  // Otherwise, we need to start a transaction. We let kysely take care of the begin/commit/rollback details
+  return db.transaction().execute(async trx => {
```
Mwindo: If we are not a nested withTransaction, then go ahead and use the kysely API to checkout a connection and start a transaction. We'll then run the callback in a context that tells it what connection it is using, that it is in a transaction, etc.

web-api/src/databaseConnection.ts
```diff
@@ -0,0 +1,156 @@
+import {
```
Mwindo, line 1: I have renamed `getConnection.ts` to `databaseConnection.ts`, and renamed/revised functions within it. Summary of changes:

1. First and foremost, `getConnection` is a confusing name: we are _not_ returning a connection; instead, we are returning a Kysely wrapper around a full connection pool, with Kysely handling client.connect/client.release, etc. behind the scenes. I therefore have renamed this. I think we can get into all sorts of trouble if we think that `getConnection` is actually giving us a connection.
2. Refactor to allow us to get the `Kysely<Database>` without running a callback.
3. Token caching/generation logic is now separate from pool generation. In other words, I made `getToken` handle all the token stuff. Then in our pool we can just say password is `getToken`.
4. Implement a health check before running a query. If we do not have an expiration mechanism, we need some way to say, "The connection failed, so re-connect." My research suggests that a connection from Lambda to Aurora in the same AZ should be very fast, so I don't think this would measurably impact performance.

web-api/src/persistence/postgres/utils/transactions.ts
Mwindo: TODO: investigate whether or not .transaction already handles this by default.

Mwindo: It does not. Trying to call `.transaction` on a transaction instance will throw an error.

web-api/src/persistence/postgres/utils/transactions.ts
Mwindo: Note that kysely is handling the "locking" for us to avoid a single connection trying to spin up two transactions at once. Process 2 will wait until process 1 has completed its transaction. In other words, `.transaction` implements a mutex lock on the connection behind the scenes.

web-api/src/persistence/postgres/databaseConnection.ts
```diff
@@ -0,0 +1,156 @@
+import {
```
Mwindo: I have renamed `getConnection.ts` to `databaseConnection.ts`, and renamed/revised functions within it. Summary of changes:

1. First and foremost, `getConnection` is a confusing name: we are _not_ returning a connection; instead, we are returning a Kysely wrapper around a full connection pool, with Kysely handling client.connect/client.release, etc. behind the scenes. I therefore have renamed this. I think we can get into all sorts of trouble if we think that `getConnection` is actually giving us a connection.
2. Refactor to allow us to get the `Kysely<Database>` without running a callback.
3. Token caching/generation logic is now separate from pool generation. In other words, I made `getToken` handle all the token stuff. Then in our pool config we can just say password is `getToken` (since password accepts a function that returns a string).
4. Implement a health check before running a query. If we do not have an expiration mechanism, we need some way to say, "The connection failed, so re-connect." My research suggests that a connection from Lambda to Aurora in the same AZ should be very fast, so I don't think this would measurably impact performance.

Mwindo: I am not dead-set on 4, but I think it is probably a good idea.

web-api/src/persistence/postgres/utils/transactions.ts
Mwindo: Other potential use cases for `onTransactionCommit`: do not generate these PDFs or send these emails until we've updated the DB.

web-api/src/persistence/postgres/utils/transactions.ts
```diff
@@ -0,0 +1,46 @@
...
+      currentConnection: trx,
+      onCommitCallbacks: [] as Array<() => Promise<void>>,
+    };
+
+    // Then we pass in information to any nested processes. If any fail, the whole transaction will fail.
+    return ConnectionStore.run(store, async () => {
+      const result = await fn();
+
+      // At the end, we run all the things we need to do on commit, like indexing in OpenSearch.
+      await settlePromises(store.onCommitCallbacks.map(cb => cb()));
```
Mwindo: This should probably go outside of the transaction. Will update.

web-api/src/persistence/postgres/databaseConnection.ts
Mwindo: After discussion, no need for a health check. The pool should do that for us.

web-api/src/persistence/postgres/utils/transactions.ts
```diff
@@ -0,0 +1,55 @@
...
+    return fn();
+  }
+
+  const db = await getDb();
+
+  // We'll capture the store here so it's available after commit.
+  let transactionStore: ConnectionInfo = {} as ConnectionInfo;
+
+  // Start the transaction; Kysely handles BEGIN/COMMIT/ROLLBACK.
+  const result = await db.transaction().execute(async trx => {
```
Mwindo, line 21: Note that `kysely.transaction` 1) guarantees the same connection throughout the transaction lifecycle and 2) handles the mutex lock so that no two transactions will ever run on the same connection.

codyseibert:
this has successfully been deployed to exp8 without error https://app.circleci.com/pipelines/github/ustaxcourt/ef-cms/16910/workflows/cfa2326f-9e4a-4d4a-89bf-ccced17d69cd

nechamakrigsman:
PR to staging here https://github.com/ustaxcourt/ef-cms/pull/10072

PR #6288, opened 2025-07-14
10603 deadline report usability to test

alexdziarn:
https://github.com/orgs/flexion/projects/11/views/1?filterQuery=assignee%3Aalexdziarn&pane=issue&itemId=100304578&issue=flexion%7Cef-cms%7C10603

- made fixes to the ui
- resovle merge conflict with 10469, caseDealines.tsx, caseDeadlinesHelper.ts, and DateRangePickerComponent.tsx changes are made in this branch.


web-client/src/views/CaseDeadlines/CaseDeadlines.tsx
```diff
@@ -72,11 +71,35 @@ export const CaseDeadlines = connect(
...
+                );
+                if (caseDeadlinesStartFG) {
+                  caseDeadlinesStartFG.style.marginBottom = '0';
+                }
+                const caseDeadlinesEndFG = window.document.getElementById(
+                  'case-deadlines-end-formgroup',
+                );
+                if (caseDeadlinesEndFG) {
+                  caseDeadlinesEndFG.style.marginBottom = '0';
+                }
```
Mwindo: This `window.document.getElementById` feels like a code smell to me. I would need to understand why we couldn't 1) use CSS or (if we can't) 2) use `useRef`.

web-client/src/views/CaseDeadlines/CaseDeadlines.tsx
```diff
@@ -72,11 +71,35 @@ export const CaseDeadlines = connect(
...
+                const caseDeadlinesDP = window.document.getElementById(
+                  'case-deadlines-datepicker',
+                );
+                if (caseDeadlinesDP) {
+                  caseDeadlinesDP.style.marginBottom = '0';
+                }
+                const caseDeadlinesStartFG = window.document.getElementById(
+                  'case-deadlines-start-formgroup',
+                );
+                if (caseDeadlinesStartFG) {
```
zachrog: Rather than handling styling choices in javascript I would suggest adding uswds conditional styles. If you have used tailwind they are very similar to conditional styles. If you look at the example below this is giving css styles depending on how large the screen is.

image
image

PR #6291, opened 2025-07-14
10617 Removing High Priority for calendaring

zacharybrown204:


Mwindo (commented):
Thanks @zacharybrown204! This is looking good! I wonder if the following are needed?

- https://github.com/ustaxcourt/ef-cms/blob/f55e0a2b168fe9fc274562ec88356d25461900f6/web-client/src/presenter/computeds/formattedEligibleCasesHelper.ts#L17 and related
- https://github.com/ustaxcourt/ef-cms/blob/f55e0a2b168fe9fc274562ec88356d25461900f6/web-api/src/persistence/postgres/cases/schema.ts#L32-L33 (note this is only the schema in app code; it is not making actual changes to the database)

Also, it might be good to write a comment pointing to a followup PR that creates a migration (you can use `npm run migration:generate:postgres`) to drop the columns from the database.

zachrog (approved):

zachrog:
> Thanks @zacharybrown204! This is looking good! I wonder if the following are needed?
> 
> * https://github.com/ustaxcourt/ef-cms/blob/f55e0a2b168fe9fc274562ec88356d25461900f6/web-client/src/presenter/computeds/formattedEligibleCasesHelper.ts#L17
>    and related
> * https://github.com/ustaxcourt/ef-cms/blob/f55e0a2b168fe9fc274562ec88356d25461900f6/web-api/src/persistence/postgres/cases/schema.ts#L32-L33
>    (note this is only the schema in app code; it is not making actual changes to the database)
> 
> Also, it might be good to write a comment pointing to a followup PR that creates a migration (you can use `npm run migration:generate:postgres`) to drop the columns from the database.

I agree with Chris. I just want to make sure there is a follow up PR dropping these columns.

Nice job Zach!

zacharybrown204:
Good catch on the spot in formattedEligibleCaseHandler I had missed that one. Just pushed a commit to handle that one.
I would prefer take out the schema change when I make the migration removing it. It might be a preference but I really don't like when a schema definition and migrations don't match.

I don't have the PR to remove the DB portion yet but I did create a devex ticket to follow up on this. https://github.com/flexion/ef-cms/issues/10712

Mwindo (approved):

TomElliottFlexion (approved):

PR #6292, opened 2025-07-14
10428 consolidate two date range picker components intermediate branch to test 1752524500

AnthonyLoeraTaxCourt:
https://github.com/orgs/flexion/projects/11/views/1?pane=issue&itemId=76509488&issue=flexion%7Cef-cms%7C10428


web-client/src/views/AdvancedSearch/OpinionSearchForm.tsx
```diff
@@ -38,13 +38,7 @@ export const OpinionSearchForm = connect(
   }) {
     return (
       <>
-        <form
-          data-testid="opinion-search-container"
-          onSubmit={e => {
-            e.preventDefault();
-            submitAdvancedSearchSequence();
-          }}
-        >
```
Mwindo, line 48: Just for others following along: if we set the validation on the form like this and call `onSubmit` in the buttons, USWDS errors will prevent the form submission, which also prevents us from running our own validation sequence (and we don't want that!).

Mwindo (approved):
Thanks @AnthonyLoeraTaxCourt!

PR #6294, opened 2025-07-15
10539 document view displaying document when shouldnt be intermediate branch to test 1752593867

AnthonyLoeraTaxCourt:
https://github.com/orgs/flexion/projects/11/views/1?pane=issue&itemId=84528704&issue=flexion%7Cef-cms%7C10539


web-client/src/views/DocketRecord/DocumentViewer.tsx
```diff
@@ -45,6 +45,23 @@ export const DocumentViewer = connect(
       blueHeader?.scrollIntoView();
     }, []);
 
+    useEffect(() => {
```
Mwindo, line 48: I haven't looked at this deeply, but in general my first thought when using `useEffect` is "[Do I need useEffect](https://react.dev/learn/you-might-not-need-an-effect)?" Maybe here you do, and if so, great! However, if you haven't considered alternatives--e.g., maybe setting up the state you need in the sequence that got you here--I would recommend trying out such alternatives.

Mwindo (approved):

PR #6299, opened 2025-07-15
10574 add deadline to sro to test

alexdziarn:
https://github.com/orgs/flexion/projects/11/views/1?filterQuery=assignee%3Aalexdziarn&pane=issue&itemId=91400993&issue=flexion%7Cef-cms%7C10574

ustaxcourt/ef-cms#7228


shared/src/business/entities/DocketEntry.ts
```diff
@@ -457,6 +462,11 @@ export class DocketEntry extends JoiValidationEntity {
     return DocketEntry.isCourtIssued({ eventCode: this.eventCode });
   }
 
+  isStatusReport(): boolean {
+    if(!this.draftOrderState) return false;
+    return this.draftOrderState.orderType === 'statusReport' || this.draftOrderState?.orderType === 'statusReportStipulatedDecision';
```
Mwindo: We should be able to get rid of `?` in the second disjunct.

web-api/src/business/useCases/courtIssuedDocument/fileAndServeCourtIssuedDocumentInteractor.ts
```diff
@@ -183,6 +183,9 @@ export const fileAndServeCourtIssuedDocument = async (
           { authorizedUser },
         );
 
+        docketEntryEntity.setDraftOrderState({
+          orderType: form.orderType,
+        });
```
Mwindo: Why are we doing this here rather than in the constructor above? I don't necessarily think it is bad to have this function, but I do think it is probably worse to set `draftOrderState` in different ways. Now we have some places that do `draftOrderState: {` in the constructor and this place that does `setDraftOrderState`; it would be better to be consistent one way or the other. I would default to just using the constructor.

shared/src/business/entities/DocketEntry.ts
```diff
@@ -457,6 +462,11 @@ export class DocketEntry extends JoiValidationEntity {
     return DocketEntry.isCourtIssued({ eventCode: this.eventCode });
   }
 
+  isStatusReport(): boolean {
```
Mwindo, line 465: Good idea. The only thing I am concerned about is that I believe `draftOrderState` is cleared when you serve the document, so this function really only works until then. If `draftOrderState` is in fact cleared when the document is served, maybe a better name for this method would be `isDraftStatusReport`.

Mwindo (approved):
I left a few notes/questions, but otherwise this is looking good. Thanks @alexdziarn!

PR #6307, opened 2025-07-16
10652 status report order enhancement, grey text if disabled

alexdziarn:
https://github.com/orgs/flexion/projects/11/views/1?filterQuery=assignee%3Aalexdziarn&pane=issue&itemId=109321136&issue=flexion%7Cef-cms%7C10652

need to add grey text if the input is disabled. Have to use inline styling because '.usa-checkboxinput:disabled + .usa-checkboxlabel' overides the classes

### DIFF
--- a/web-client/src/views/StatusReportOrder.tsx
+++ b/web-client/src/views/StatusReportOrder.tsx
@@ -261,6 +261,7 @@ export const StatusReportOrder = connect(
                         className="usa-checkbox__label"
                         htmlFor="stricken-from-trial-sessions"
                         id="stricken-from-trial-sessions-label"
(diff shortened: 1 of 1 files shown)


web-client/src/views/StatusReportOrder.tsx
```diff
@@ -261,6 +261,7 @@ export const StatusReportOrder = connect(
                         className="usa-checkbox__label"
                         htmlFor="stricken-from-trial-sessions"
                         id="stricken-from-trial-sessions-label"
+                        style={!statusReportOrderHelper.isCalendared ? { color: '#757575' } : {}}
```
Mwindo, line 264: I don't fully understand why we can't use classes here, but I'll trust the PR comment :)

Mwindo (approved):

PR #6308, opened 2025-07-16
 Devex: Standardize Postgres mapper functions and schemas, and improve types

Mwindo:
This is a "let's go back and clean up some of our Postgres stuff" PR. The more we can make the code self-documenting--consistent file structures and clear types--the better off Gunnison will be.

This PR:
- Changes schema definitions from `| undefined` to `| null` (undefined is not a Postgres type and should probably not be in our schemas).
- Makes sure every table has a `mapper.ts` file with `toKyselyNewX` and `fromKyselyX` functions for consistency.
- Adds better typing to the mappers, fixing any issues that were caught in the process.
- Removes `updateMessage` and relies on `upsertMessages` instead, because `updateMessage` was really trying to do an upsert all along.


web-api/src/persistence/postgres/workitems/getDocumentQCInboxForUser.ts
```diff
@@ -1,57 +1,77 @@
...
     .selectFrom('dwWorkItem as w')
     .leftJoin('dwCase as c', 'c.docketNumber', 'w.docketNumber')
-    .innerJoin('dwDocketEntry as d', join =>
-      join
-        .onRef('d.docketEntryId', '=', 'w.docketEntryId')
-        .onRef('d.docketNumber', '=', 'w.docketNumber'),
-    )
     .selectAll('w')
-    .select(eb => [
-      jsonObjectFrom(
```
Mwindo, line 44: This `jsonObjectFrom` lies about the type: it thinks it is a `DocketEntryKysely` ... but it ends up being a serialized version of `DocketEntryKysely` (so, e.g., `Date` is actually `string`). I tried a few different ways of getting this to work before finally giving up and saying, "We're just going to go get the docket entry data separately and stitch it together in app code." (Thanks to @AnthonyLoeraTaxCourt for talking through this.) It means one more query, but it makes the typing much easier, and I'm not really sure the extra query will be less performant vs. the join anyway.

web-api/src/persistence/postgres/workitems/getDocumentQCInboxForUser.ts
```diff
@@ -1,57 +1,77 @@
...
+    entryById.set(docketEntry.docketEntryId, docketEntry);
+  }
+
+  const workItemsWithDocketEntries = workItems.map(w => {
+    const docketEntry = entryById.get(w.docketEntryId);
+    if (!docketEntry) {
+      throw new NotFoundError(
+        `Could not find docket entry for work item ${w.docketNumber}, docket entry id ${w.docketEntryId}`,
+      );
+    }
```
Mwindo: I'm not actually sure we want this. Might be safer to have {} as RawDocketEntry.

web-api/src/persistence/postgres/workitems/schema.ts
```diff
@@ -28,6 +31,23 @@ export const DW_WORK_ITEM_COLUMNS = Object.keys(
...
+  caseTitle?: string;
+  caseStatus?: string;
+  leadDocketNumber?: string;
+  trialDate?: string;
+  trialLocation?: string;
+};
+
+export type RawWorkItemWithCaseAndDocketEntryInfo = RawWorkItemWithCaseInfo & {
+  docketEntry: RawDocketEntry;
+};
```
Mwindo, line 50: I am not in love with these types, but they are at least telling the truth. Here is how they evolved:
- Let's get all the work item, case, and docket entry data in one query! Then we'll have `WorkItemWithExtraDataKysely` (db representation, rename) and `RawWorkItemWithExtraData` (app-code representation, rename) types.
- Whoa, this makes getting accurate types for docket entry surprisingly tricky. Ok, let's just make a separate query for docket entry info and stitch it together in app code.
- Ok, so now we have a query that gets the work item and case data, and then we need to augment it with the docket entry data.
- The docket entry data we get is already RawDocketEntry. So I guess we have `WorkItemWithCaseInfoKysely` (db representation) that needs to be translated to `RawWorkItemWithCaseInfo` (app-code representation) to which we will append the `RawDocketEntry`, which would give us the `RawWorkItemWithCaseAndDocketEntryInfo` we want.

web-api/src/persistence/postgres/utils/typeHelpers.ts
```diff
@@ -0,0 +1,6 @@
+// Pick fields, but make them nullable
+// Useful in cases where we left join table X with Y: we want the fields from Y,
+// but with a left join they might all be null (if nothing in Y matches with X)
+export type NullablePick<T, K extends string> = {
```
TomElliottFlexion, line 4: Very cool

web-api/src/persistence/postgres/utils/seed/fixtures/workItems.ts
```diff
@@ -2131,26 +2131,6 @@ export const workItems: NewWorkItemKysely[] = [
...
-  {
-    createdAt: calculateDate({ dateString: '2020-06-25T20:58:55.415Z' }),
-    docketEntryId: '',
-    docketNumber: '120-20',
-    section: 'petitions',
-    sentBy: 'Test Petitioner',
-    sentByUserId: '7805d1ab-18d0-43ec-bafb-654e83405416',
-    updatedAt: calculateDate({ dateString: '2020-06-25T20:58:55.415Z' }),
-    workItemId: 'd9edf05a-7b27-447c-aaf9-f12157823bc6',
-  },
```
TomElliottFlexion, line 2153: Bad data?

TomElliottFlexion (approved):
Thank you @Mwindo!

jimlerza (approved):

web-api/src/persistence/postgres/utils/seed/fixtures/workItems.ts
Mwindo: Yes--there is no associated docket entry, which--unless I am missing something--is nonsense.

TomElliottFlexion: Nonsense!

PR #6320, opened 2025-07-18
10539 document view displaying document when shouldnt be intermediate branch to test 1752852809

AnthonyLoeraTaxCourt:


web-client/src/views/DocketRecord/DocketRecordHeader.tsx
```diff
@@ -179,7 +183,33 @@ export const DocketRecordHeader = connect<
     openDownloadDocketEntriesModalSequence,
     showModal,
     sortTableSequence,
+    setViewerDocumentToDisplaySequence,
+    docketRecordFilter,
+    viewerDocumentToDisplay,
   }) {
+    const isFirstRender = useRef(true);
+
+    useEffect(() => {
```
zachrog: Moving control flow logic + rendering logic to using cerebral + hooks will make debugging particularly difficult in this area. I would recommend trying to make this work in Cerebral first before doing a useRef() + useEffect().

I think using hooks is fine, and I think using cerebral is fine, but using both at the same time can and has made it very difficult to understand rendering logic.

zachrog (dismissed):

TomElliottFlexion (approved):
Nice job! If you wanted and it was convenient to extend another test, you could add an e2e/integration test expectation around this, but I wouldn't insist on it. Definitely not before going to test!

PR #6331, opened 2025-07-21
10637 dxox postgres test

pavanyenduri:


jimlerza:
How do you intend to get the existing practitioner documents out of dynamodb and into postgres?

cruzjone-flexion (approved):

pavanyenduri:
> How do you intend to get the existing practitioner documents out of dynamodb and into postgres?

We did not do it for change-of-address or other postgres migration story, i was under assumption that there was a stream already created which takes care of migrating existing data to postgres. Let me check with Kaitlyn or Chris.

jimlerza:
> > How do you intend to get the existing practitioner documents out of dynamodb and into postgres?
> 
> We did not do it for change-of-address or other postgres migration story, i was under assumption that there was a stream already created which takes care of migrating existing data to postgres. Let me check with Kaitlyn or Chris.

The purpose of the streams is to index dynamodb data in opensearch. We "hooked in" to this process to get data migrated into postgres _for data that was already being indexed in opensearch_ because it was the easiest way to get that data into postgres.

We should not be using the streams for data that is not being indexed in opensearch. You will need to write a script to migrate this data.

pavanyenduri:
> > > How do you intend to get the existing practitioner documents out of dynamodb and into postgres?
> > 
> > 
> > We did not do it for change-of-address or other postgres migration story, i was under assumption that there was a stream already created which takes care of migrating existing data to postgres. Let me check with Kaitlyn or Chris.
> 
> The purpose of the streams is to index dynamodb data in opensearch. We "hooked in" to this process to get data migrated into postgres _for data that was already being indexed in opensearch_ because it was the easiest way to get that data into postgres.
> 
> We should not be using the streams for data that is not being indexed in opensearch. You will need to write a script to migrate this data.

Will write a script which will write existing data into postgres.

jimlerza:
You will also need to update `scripts/upload-practitioner-application-packages.md` and it appears `scripts/upload-practitioner-application-packages.ts` will need a substantial rewrite in the areas where it inserts data into dynamodb.

pixiwyn:
Regarding the script to move the records into Postgres, after discussing with Zach, and based on prior experience, this isn't necessary. We have set it up to use process stream like with Document QC, Messages, Case Notes, etc, etc. Process stream is the best way to ensure we get the data into Postgres without blocking or losing data during a deployment. If you disagree, let's talk @jimlerza. Users was a special case. It was the only deployment that hasn't used process stream to date. The stream doesn't care what is in opensearch or not. It all about the entityName, pk, and sk (thus matching the correct documents).

PR #6332, opened 2025-07-22
10686 todays orders sorting to test 1753194548

AbelTaxCourt:
https://github.com/orgs/flexion/projects/11/views/1?pane=issue&itemId=116650799&issue=flexion%7Cef-cms%7C10686


web-client/src/presenter/computeds/Public/todaysOrdersHelper.ts
```diff
@@ -51,15 +48,43 @@ export const todaysOrdersHelper = (
...
-  const showLoadMoreButton = formattedOrders.length < totalCount;
+  const sortedFormattedOrders = formattedOrders.sort((orderA, orderB) => {
+    let sortNumber = 0;
+    const compare1 = tableSort?.sortOrder === DESCENDING ? orderB : orderA;
+    const compare2 = tableSort?.sortOrder === DESCENDING ? orderA : orderB;
+
+    if (!tableSort) {
+      sortNumber = compare1.createdAt.localeCompare(compare2.createdAt);
+    } else if (
+      SUPPORTED_SORT_FIELDS_FOR_TODAYS_ORDERS.includes(tableSort.sortField)
```
Mwindo, line 61: I would recommend updating the name `SUPPORTED_SORT_FIELDS_FOR_TODAYS_ORDERS` or else including docket number in it and updating the implementation here accordingly. We _do_ support sorting by docket number, it just needs to be sorted differently.

Mwindo (approved):
This looks good to me overall! Thanks @AbelTaxCourt and @alexdziarn! Two quick style comments:

The hover style looks a little odd, but maybe that is intentional:
Screenshot 2025-07-22 at 1 26 20 PM

The table cannot scroll horizontally when it is between mobile-view size and full size. I am not sure if that is necessary or not, and maybe it was already this way.
Screenshot 2025-07-22 at 1 24 24 PM

PR #6349, opened 2025-07-25
10705 dxox: Transaction support

Mwindo:

### Details:

1. We say, `getConnection`, `getLockingConnection`, `establishConnection` ... but these are not getting a connection or establishing a connection. They are returning a kysely wrapper object around a client pool. Behind the scenes, kysely is creating/acquiring/releasing/removing connections when we call, e.g., `.execute`.
2. We have two connection pools, one for `getConnection` and one for `getLockingConnection`. For the latter especially, we abuse the pool: we say, "Make sure there is only ever one connection in your pool so that we don't accidentally create a lock on one connection and try to remove it with another, which won't work!" But this is not what a connection pool is for. Instead, we should have one connection pool and get an honest-to-goodness connection _from_ the pool and use _that_ throughout the lifetime of the lock.

1. Rename `getConnection`, `getLockingConnection`, `establishConnection` etc. If we think of these objects as connections, we will confuse things.
2. We have _one_ connection pool. The equivalent of `getLockingConnection`, `acquireOneDbConnection`, now simply gets a particular connection from the existing pool and uses it throughout the lifetime of the locking code.

Implementation details:

- Revision: `getLockingConnection.ts` is gone. It is replaced with `acquireOneDbConnection`.


web-api/src/persistence/postgres/databaseConnection.ts
```diff
@@ -0,0 +1,154 @@
+import { CamelCasePlugin, Kysely, PostgresDialect, Transaction } from 'kysely';
```
Mwindo, line 1: I have renamed `getConnection.ts` to `databaseConnection.ts`, and renamed/revised functions within it. Summary of changes:

1. First and foremost, `getConnection` is a confusing name: we are _not_ returning a connection; instead, we are returning a Kysely wrapper around a full connection pool, with Kysely handling client.connect/client.release, etc. behind the scenes. I therefore have renamed this. I think we can get into all sorts of trouble if we think that `getConnection` is actually giving us a connection.
2. Refactor to allow us to get the `Kysely<Database>` without running a callback.
3. Token caching/generation logic is now separate from pool generation. In other words, I made `getToken` handle all the token stuff. Then in our pool config we can just say password is `getToken` (since password accepts a function that returns a string).
4. Incorporate https://github.com/ustaxcourt/ef-cms/pull/6342 logic.

web-api/src/persistence/postgres/messages/markMessageThreadRepliedTo.ts
```diff
@@ -12,14 +13,16 @@ export const markMessageThreadRepliedTo = async ({
...
+      await settlePromises(
+        messages.map(async message => {
+          await pgUpdateTable({
+            table: 'dwMessage',
+            values: { isRepliedTo: true },
+            where: cb => cb.where('messageId', '=', message.messageId),
+          });
+        }),
+      );
+    });
```
Mwindo, line 26: Testing it out with a minor use case first, then can expand.

cypress/helpers/cypressTasks/postgres/getCypressPostgresDb.ts
```diff
@@ -1,5 +1,5 @@
 import { Signer } from '@aws-sdk/rds-signer';
-import type { Database } from '@web-api/database-schema';
+import type { Database } from '@web-api/persistence/postgres/database-schema';
```
Mwindo, line 2: I moved our postgres stuff into the postgres directory rather than the root.

web-api/src/persistence/postgres/acquireOneDbConnection.ts
```diff
@@ -0,0 +1,11 @@
+import { ConnectionBuilder } from 'kysely';
```
Mwindo, line 1: This file:

"Hey, I need a particular connection from the connection pool."

Why might you want this?

- A `pg_advisory_lock` needs to be opened/closed on the same connection, so trying to call pool.connect when setting a lock and then again on releasing a lock could cause an issue since, in theory, you could get one connection the first time and another the second.
- Because transactions are "one per connection," a different or further implementation of transactions might want to keep track of a specific transaction. (In this PR, however, it is handled by kysely.transacation.)
- Etc.

web-api/src/persistence/postgres/utils/operation/tryReleaseLocks.ts
```diff
@@ -0,0 +1,25 @@
+import { Database } from '@web-api/persistence/postgres/database-schema';
```
Mwindo, line 1: In our previous implementation, we we destroying the connection, which destroyed the lock. Now, we are no longer destroying the connection, so we actually need to run a SQL query to release the lock.

web-api/src/persistence/postgres/utils/transactions.ts
```diff
@@ -0,0 +1,55 @@
+import { settlePromises } from '@web-api/utilities/settlePromises';
+import {
+  ConnectionInfo,
+  ConnectionStore,
+  getDb,
+} from '@web-api/persistence/postgres/databaseConnection';
+
+export async function withTransaction<T>(fn: () => Promise<T>): Promise<T> {
```
Mwindo, line 8: I considered two approaches (thank you @zachrog and @TomElliottFlexion for the great discussions!):

1. A "magic" wrapper function, `withTransaction`, which is the implementation we here.
2. A less "magic" approach with `startTransaction` and `endTransaction` functions.

1 works via some `AsyncLocalStorage` magic. We pass around context (via `ConnectionStore.run` below) to children node processes.

Main pros:

- Easy to use
- You can't mess it up by forgetting to call `endTransaction`, or accidentally missing that a transaction is nested and should not call `BEGIN` transaction.

Main cons:

- The `AysncLocalStorage` context is just sort of magic, and this could make things harder to debug/understand.

A sketch of 2 can be seen [here](https://github.com/ustaxcourt/ef-cms/tree/10705-dxox-alt) and works like this:

`startTransaction(transactionId?: string)` starts a transaction (puts some data in a global mapping object), unless you pass in a transactionId (= "I have a parent transaction") and returns a uuid. `endTransaction(transactionId)` does the commit/OpenSearch stuff and then removes the data from the global mapping object.

Main pros:

- More explicit: there is no `AsyncLocalStorage` doing magic behind the scenes. (The equivalent is now a JS object holding data about transactions.) Debugging is probably easier.
- You could pass a `transactionId` into functions and follow a transaction in code by following this parameter.

Main cons:

- A `startTransaction` and `endTransaction` are decoupled. A developer might forget to pass in transactionId to a nested transaction or forget to call endTransaction.
- Passing `transactionId` into functions could get unwieldy and couples application code to database code.

Because we worried about the cons of 2, and because the "magic" of 1 is pretty self-contained, we decided to go with 1.

web-api/src/persistence/postgres/acquireOneDbConnection.ts
Mwindo: TODO: Will this actually persist the connection after a query is executed?

Mwindo: Sadly, as I learn more about this, I think this is actually not quite right. My initial understanding was that this connection would persist ... but it does not. See, e.g., https://github.com/tediousjs/node-mssql/issues/1517. So I will need to reassess the way the connection for withLocking works.

PR #6351, opened 2025-07-25
10687 story to test

rgf314159:
10687-story-to-test
https://github.com/flexion/ef-cms/issues/10687


shared/src/business/utilities/formatAttachments.ts
```diff
@@ -27,6 +27,7 @@ export const formatAttachments = ({
         archived: !!doc.archived,
         documentId,
         documentTitle: generatedDocumentTitle || doc.documentType,
+        index: doc.index,
```
TomElliottFlexion, line 30: Doesn't need to happen before going to `test`, but we'll probably need updated tests on this utility and the action below

web-client/src/views/Messages/MessageModalAttachments.tsx
```diff
@@ -15,6 +15,7 @@ const getDocumentOption = doc => {
       key={doc.docketEntryId}
       value={`${doc.docketEntryId}`}
     >
+      {doc.index != null ? `${doc.index} - ` : ''}
```
Mwindo, line 18: I think we'd want to ensure index is not `undefined` too.

web-client/src/views/Messages/MessageModalAttachments.tsx
```diff
@@ -67,23 +68,29 @@ export const MessageModalAttachments = connect(
               return (
                 <div className="margin-top-1" key={doc.documentId}>
                   <div className="grid-row">
-                    <div className="grid-col-9">{doc.documentTitle}</div>
+                    <div className="grid-col-9">
+                      {doc.index ? doc.index + ' - ' : ''}
```
Mwindo: This wouldn't work for index 0. Not sure if we are starting at 1 or 0. If the former, then this should be ok.

Mwindo: Same comment where applicable below

Mwindo (approved):
Left a few comments to look into, and I agree with @TomElliottFlexion. Thanks @rgf314159 and @AnthonyLoeraTaxCourt.

web-client/src/views/Messages/MessageModalAttachments.tsx
rgf314159: Sanity check: 

**{doc.index != null ? `${doc.index} - ` : ''}**

the != operator (not strict) checks for both null and undefined.
doc.index != null is true only if doc.index is not null and not undefined.

This is our current understanding. ????

web-client/src/views/Messages/MessageModalAttachments.tsx
rgf314159: Why would it not work for 0?

rgf314159: This should work if 0 is present.

TomElliottFlexion: 0 will evaluate to falsy

PR #6355, opened 2025-07-28
10684 Large Case Followup (dropping a table)

Mwindo:
In https://github.com/ustaxcourt/ef-cms/pull/6241, we stopped using the `docketEntry` column of work items and instead used the `dwDocketEntry` table. Here, we finish the job and drop the `docketEntry` column entirely. (This has to be done as a follow-up so that we do not drop the `docketEntry` column while the active color is still using it.)

### DIFF
--- a/web-api/src/persistence/postgres/utils/migrate/migrations/2025-07-26T23_34_53Z-drop-docketEntry-from-dwWorkItem.ts
+++ b/web-api/src/persistence/postgres/utils/migrate/migrations/2025-07-26T23_34_53Z-drop-docketEntry-from-dwWorkItem.ts
@@ -1,0 +1,13 @@
+import { Kysely } from 'kysely';
+
+export async function up(db: Kysely<any>): Promise<void> {
+  await db.schema.alterTable('dwWorkItem').dropColumn('docketEntry').execute();
+}
+
+export async function down(db: Kysely<any>): Promise<void> {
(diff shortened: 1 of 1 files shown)


web-api/src/persistence/postgres/utils/migrate/migrations/2025-07-26T23_34_53Z-drop-docketEntry-from-dwWorkItem.ts
```diff
@@ -0,0 +1,28 @@
...
+    .addColumn('docketEntry', 'jsonb', col =>
+      col.notNull().defaultTo(JSON.stringify({})),
+    )
+    .execute();
+  await attachDocketEntryDataToWorkItems(db);
+}
+
+async function attachDocketEntryDataToWorkItems(db: Kysely<Database>) {
+  // Easier to do this in raw SQL rather than fiddling with kysely, and we don't really need the types anyway
+  return db.executeQuery(
```
Mwindo, line 20: This query takes about 3 mins.

Mwindo: (But hopefully it never has to be run.)

codyseibert (approved):

PR #6362, opened 2025-07-29
10684 Large Cases Bug (An Update to Test)

Mwindo:
Thank you to @zachrog and @codyseibert for catching this!

**What is the bug?**

In some spots we upsert a work item with `section` set to `user.section` (or some derivative thereof). For sections `caseServicesSupervisor` and `clerkofcourt` (or, more accurately, any section that is not `petitions` or `docket`), we should translate the section to `petitions` for a petition and `docket` for anything else. Before 10684, we used the `documentTitle` of the associated docket entry to determine if the document was a petition. 10684 removed that logic, which means that the work items for documents made by, or the work items assigned to, `caseServicesSupervisor` and `clerkofcourt` users now all get assigned to the `petition` section. Whoops. Luckily 10684 has not been merged to test.

**What is the fix?**

Before, we re-calculated the section _every time we saved a work item_. That's silly. The current fix says, "Hey, when you need to set/change the section in an interactor, make sure you get the appropriate one."

TODO:

- [x] **Write a test.** I opted for a few unit tests and better typing. There are several pathways in which we could set the wrong section, and e2e tests either seemed overkill or arbitrary.
- [x] **Can we determine the section directly from the document without having to think about the user at all?** (Answer: I don't believe so.)
- [x] **Check that this actually fixes the thing and does not cause more issues**

### DIFF
--- a/shared/src/business/entities/WorkItem.ts
+++ b/shared/src/business/entities/WorkItem.ts
@@ -22,7 +22,7 @@ export class WorkItem extends JoiValidationEntity {
   public docketNumber: string;
   public inProgress?: boolean;
   public isRead?: boolean;
-  public section: string;
+  public section: typeof PETITIONS_SECTION | typeof DOCKET_SECTION;
   public sentBy: string;
   public sentBySection?: string;
   public sentByUserId?: string;
--- a/shared/src/business/entities/WorkItem.ts
+++ b/shared/src/business/entities/WorkItem.ts
@@ -92,7 +92,7 @@ export class WorkItem extends JoiValidationEntity {
   }: {
     assigneeId?: string;
     assigneeName?: string;
-    section: string;
+    section: typeof PETITIONS_SECTION | typeof DOCKET_SECTION;
     sentBy: string;
     sentBySection?: string;
     sentByUserId?: string;
--- a/shared/src/business/entities/WorkItem.ts
+++ b/shared/src/business/entities/WorkItem.ts
@@ -2,7 +2,12 @@ import { JoiValidationEntity } from '@shared/business/entities/JoiValidationEnti
 import { createISODateString } from '../utilities/DateHandler';
 import { getUniqueId } from '@shared/sharedAppContext';
 import { JoiValidationConstants } from '@shared/business/entities/JoiValidationConstants';
-import { CASE_STATUS_TYPES } from '@shared/business/entities/EntityConstants';
+import {
+  CASE_STATUS_TYPES,
+  DOCKET_SECTION,
+  INITIAL_DOCUMENT_TYPES,
+  PETITIONS_SECTION,
+} from '@shared/business/entities/EntityConstants';
 import joi from 'joi';
 
 export class WorkItem extends JoiValidationEntity {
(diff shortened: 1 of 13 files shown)


shared/src/business/entities/WorkItem.ts
```diff
@@ -132,6 +132,28 @@ export class WorkItem extends JoiValidationEntity {
   getValidationRules() {
     return WorkItem.VALIDATION_RULES;
   }
+
+  static getWorkItemSectionFromUserSection({
```
Mwindo, line 141: Moved out of the work items mapper file.

web-api/src/business/useCases/workItems/assignWorkItemsInteractor.ts
```diff
@@ -71,6 +70,23 @@ export const assignWorkItemsInteractor = async (
     workItemEntity = new WorkItem(workItem);
   }
 
+  const docketEntry = (
```
Mwindo, line 73: The only reason we call this is to be able to pass in the document title. How necessary is this? Can we know that the user being assigned to is always petitions or docket? If so, we can get rid of this.

Mwindo: I suppose, in principle, we cannot assume that, based on the signature of the function :/

shared/src/business/entities/WorkItem.ts
```diff
@@ -132,6 +132,28 @@ export class WorkItem extends JoiValidationEntity {
...
+    documentTitle: string;
+  }) {
+    if (!section) {
+      return undefined;
+    }
+    // We have sections for caseServicesSupervisor and clerkofcourt, but as far as we can tell, they aren't used.
+    // Instead, we need to translate these into either the petitions section or the docket section depending
+    // on the document type.
+    if (!['caseServicesSupervisor', 'clerkofcourt'].includes(section)) {
+      return section;
```
Mwindo: Better: return `petitions` or `docket`.

shared/src/test/mockAuthUsers.ts
```diff
@@ -91,3 +91,10 @@ export const mockCaseServicesSupervisorUser: AuthUser = {
   role: ROLES.caseServicesSupervisor,
   userId: '97451b05-ae9c-46d5-9074-1ac6ef12cfc6',
 };
+
+export const mockClerkOfTheCourtUser: AuthUser = {
+  email: 'mcokClerkOfTheCourt@example.com',
```
Mwindo: Darnit! I will fix this typo before deploying to test.

shared/src/business/entities/WorkItem.ts
Mwindo: I also refactored this/retyped this. Before, we said:

1. If you are not a clerk of the court or case supervisors section, return your section.
2. If you are, then we will assign work items for petitions to the petitions section and other work items to the docket section.

Now, we say:

1. If you are a petitions or docket section, return your section. (Because we should only ever be assigning to the petition or docket section!)
2. If you aren't, then we will assign work items for petitions to the petitions section and other work items to the docket section.

shared/src/business/entities/WorkItem.ts
```diff
@@ -132,6 +137,31 @@ export class WorkItem extends JoiValidationEntity {
...
+    }
+    // A work item is assigned to either the petitions section or the docket section.
+    // If the section we are passing in is either of those, just return it.
+    if ([DOCKET_SECTION, PETITIONS_SECTION].includes(section)) {
+      return section as typeof PETITIONS_SECTION | typeof DOCKET_SECTION;
+    }
+    // Otherwise, we will assign a petition to the petitions section and anything else to the docket section.
+    if (
+      documentTitle?.toLocaleLowerCase() ==
+      INITIAL_DOCUMENT_TYPES.petition.documentTitle.toLocaleLowerCase()
```
Mwindo, line 156: Probably don't need to lowercase it, but can't hurt.

shared/src/business/entities/WorkItem.ts
```diff
@@ -132,6 +137,31 @@ export class WorkItem extends JoiValidationEntity {
...
   }
+
+  static getWorkItemSectionFromUserSection({
+    section,
+    documentTitle,
+  }: {
+    section?: string;
+    documentTitle: string;
+  }): typeof PETITIONS_SECTION | typeof DOCKET_SECTION | undefined {
+    if (!section) {
```
Mwindo: Why is this optional?? Because `user.section` is `string | undefined`, and that's often what we pass in. There are probably more elegant ways to fix this, but I'm hoping to get this out in time for this week's release.

web-api/src/business/useCaseHelper/docketEntry/fileAndServeDocumentOnOneCase.ts
```diff
@@ -87,11 +88,18 @@ export const fileAndServeDocumentOnOneCase = async ({
...
+  docketEntryEntity,
+}) => {
   workItemToUpdate.assignToUser({
     assigneeId: user.userId,
     assigneeName: user.name,
-    section: user.section,
+    section: WorkItem.getWorkItemSectionFromUserSection({
+      section: user.section,
+      documentTitle: docketEntryEntity.documentTitle,
+    }),
```
Mwindo, line 102: The basic idea: anytime section depends on user.section, we need to call this function. I _could_ put this logic directly in `assignToUser` and/or in the constructor, but it felt messy passing in a `docketEntry` field like that. I would rather put the onus on the person constructing/calling the method and use better typing to enforce which sections are allowed.

codyseibert (approved):

zachrog (approved):

PR #6364, opened 2025-07-30
10692 minute sheet update package

alexdziarn:
https://github.com/orgs/flexion/projects/11/views/1?pane=issue&itemId=117981398&issue=flexion%7Cef-cms%7C10692


cypress/local-only/tests/integration/minuteSheet/minute-sheet.cy.ts
```diff
@@ -0,0 +1,96 @@
...
+        scheduleTrialSession(docketNumber, trialSessionId);
+      });
+    });
+  });
+  beforeEach(() => {
+    loginAsColvinChambers();
+    cy.get('[data-testid="trial-session-link"]').click();
+    cy.get(`[data-testid="trial-location-link-${ts}"]`).click();
+    cy.get(`[data-testid="minute-sheet-button-${dn}"]`)
+      .invoke('removeAttr', 'target')
```
zachrog, line 37: I do not think you need the .invoke(). I think you can just do cy.get().click()

cypress/local-only/tests/integration/minuteSheet/minute-sheet.cy.ts
```diff
@@ -0,0 +1,96 @@
...
+} from 'cypress/helpers/authentication/login-as-helpers';
+import { updateCaseStatus } from 'cypress/helpers/caseDetail/caseInformation/update-case-status';
+import { goToCase } from 'cypress/helpers/caseDetail/go-to-case';
+import { createAndServePaperPetition } from 'cypress/helpers/fileAPetition/create-and-serve-paper-petition';
+import { calendarTrialSession } from 'cypress/helpers/trialSession/calendar-trial-session';
+import { createTrialSession } from 'cypress/helpers/trialSession/create-trial-session';
+import { scheduleTrialSession } from 'cypress/helpers/trialSession/schedule-trial-session';
+
+describe('user opens a minute sheet', () => {
+  let ts: string;
```
zachrog: I would remove these let variable assignments from your file as cypress does not handle variable assignments like a traditional test runner like jest. Becuase of all the async nature of cypress we have had problems with flakey tests creating let variables, assigning the values in a before block, then trying to use them in a test. Cypress even recommends not doing this (although I wish we could).

https://docs.cypress.io/app/core-concepts/variables-and-aliases

web-api/src/business/useCaseHelper/trialSessionMinutes/formatMinuteSheet.ts
```diff
@@ -205,13 +205,26 @@ export const getBriefDetails = (briefDetails: BriefDetailsType) => {
 
 export const getConsolidatedDocketNumbers = (aCase: RawCase): string => {
   if (aCase.consolidatedCases.length === 0) {
-    return aCase.docketNumber;
+    return aCase.docketNumberWithSuffix;
   }
   return aCase.consolidatedCases
-    .map(consolidatedCase => consolidatedCase.docketNumber)
+    .map(consolidatedCase => consolidatedCase.docketNumberWithSuffix)
```
zachrog: I think the complexity of this can be reduced if you were to sort before you map. Currently you are adding docketNumber with suffix, then stripping it out with a regex, then sorting based on no letters.

My suggestion is to sort first based on docket number which is very common in the app, then map into docketNumberWithSuffix

cypress/local-only/tests/integration/minuteSheet/minute-sheet.cy.ts
alexdziarn: this is to prevent cypress from opening the minute sheet in a new tab. Cypress can't run tests in mutiple windows/tabs so its a workaournd that makes it act as a redirect instead.

zachrog (approved):
Overall looks good. I only ask that you make some updates to the cypress tests + the mpa/sort.

Nice!

jimlerza (approved):

PR #6366, opened 2025-07-30
9009: Transactions (formerly flexion#10705) to test

Mwindo:

### Details:

2. Note that `getLockingConnection` is kept as is. Technically, this is also a pool; however, it is guaranteed to have only 1 client, so the name is a white lie. I thought originally, "We can just check out a connection and use that." Nope. This is hard.

Implementation details:


web-api/src/persistence/postgres/messages/markMessageThreadRepliedTo.ts
```diff
@@ -12,14 +13,16 @@ export const markMessageThreadRepliedTo = async ({
...
+      await settlePromises(
+        messages.map(async message => {
+          await pgUpdateTable({
+            table: 'dwMessage',
+            values: { isRepliedTo: true },
+            where: cb => cb.where('messageId', '=', message.messageId),
+          });
+        }),
+      );
+    });
```
Mwindo, line 26: Testing in a low-stakes place at first.

web-api/src/persistence/postgres/utils/transactions.ts
```diff
@@ -0,0 +1,55 @@
...
+    return fn();
+  }
+
+  const db = await getDb();
+
+  // We'll capture the store here so it's available after commit.
+  let transactionStore: ConnectionInfo = {} as ConnectionInfo;
+
+  // Start the transaction; Kysely handles BEGIN/COMMIT/ROLLBACK.
+  const result = await db.transaction().execute(async trx => {
```
Mwindo, line 20: Kysely guarantees that only one connection is used for the transaction, and that only one transaction is on that connection at a time.

web-api/src/persistence/postgres/getLockingDbConnection.ts
```diff
@@ -48,6 +49,13 @@ async function getToken() {
   return token;
 }
 
+// Q: "Why are we creating a pool dedicated for locking? Just grab a connection from the pool!"
+// A1: Kysely operates on pools, not connections. Without running a specific callback, there is no
+// way in kysely to say, "Give me a connection and I'll release it when I want."
+// A2: Suppose you have a single pool with 1 client and some unit of work calls withLocking.
+// You will get a connection. Then you will try to run some unit of work ... and deadlock, because
+// you have acquired (and not yet released) the only connection in the pool for the lock. By induction,
+// this problem exists for any n clients in one pool.
```
Mwindo, line 58: You might also wonder, "Ok, fine. But why are we re-creating the pool every time in `getLockingDbConnection`?" It is the same problem as A2 above, but with nested `withLocking`. In other words, withLocking(withLocking(...)) would cause issues because the inner withLocking would try to grab a connection, but the outer withLocking would have claimed it.

There are a lot of gotchas here. It eventually seemed safer to leave this all alone.

web-api/src/persistence/postgres/utils/transactions.ts
Mwindo: My recommendation:

1. Test this
2. Before merging to staging, test again on a more high-stakes place, e.g., when creating a case in `createCaseInteractor` and `createCaseFromPaperInteractor`.
3. _Then_ merge and say, "It's ready to use!"

PR #6384, opened 2025-08-04
10495 - Migrate Users to Postgres

codyseibert:
## Thing to be aware of

we modified the scripts/user/setup-glued-judges.ts script, so glue judge users may not work correctly

## Manual Deploy Steps
- NOTE: This deployment CANNOT have a regular migration with it.
- Turn on maintenance mode
  - run environment switcher
  - run command: `npm run maintenance:engage`
- migrate the postgres db forward.
  - run environment switcher
  - run command: `npm run migration:postgres`
- add permissions to allow application to access users table
   - cd `./scripts/postgres` 
   - run `./create-rds-users.sh`
- move petitioners from dwCase.petitioners -> dwUser via script
  - run environment switcher
  - run script: `npx ts-node --transpile-only scripts/run-once-scripts/postgres-migration/move-petitioners-out-of-case.ts`
- Move users, privatePractitioners, irsPractitioners, pending-case from dynamo to postgres via script 
  - run environment switcher
  - run script: `npx ts-node --transpile-only scripts/run-once-scripts/postgres-migration/migrate-users-to-postgres.ts`
  - note ^ if this script fails, it might be because a user is both a privatePractitioner and irsPractitioner on a case 973-24 userid e6bb4f9b-5b9b-48b9-83e8-8a19165dba5a which you need to remove the privatePractitioner mapping record
- Update accountStatus in Postgres via script 
  - run environment switcher
  - run script: `npx ts-node --transpile-only scripts/run-once-scripts/postgres-migration/add-account-status-to-users.ts`
- Deploy + switch color (the smoketests might fail.. so you might have to turn off maintenance mode to get them to pass or comment the smoke tests out)
- Turn off maintenance mode
  - run environment switcher
  - run command: `npm run maintenance:disengage`
- Run readonly smoketests against prod to test deployment worked
  - run environment switcher
  - enable smoketest user:  `./scripts/user/create-and-enable-smoketest-user.ts`
  - run readonly smoketests: `./scripts/run-cypress.sh -rc`
  - disable smoketests: `./scripts/user/disable-smoketest-user.ts`
  - run public readonly smoketests: `./scripts/run-cypress.sh -rpc`

- Zendesk specific
  - run environment switcher
  - run zendesk user setup: `./scripts/user/setup-zendesk-user.ts` 
  - merge https://github.com/ustaxcourt/ustc-zendesk/pull/317
- AT SOME POINT LATER: Delete all user data from dynamo
  - run environment switcher
  - run script: `npx ts-node --transpile-only scripts/run-once-scripts/postgres-migration/delete-users.ts`


cypress/helpers/fileAPetition/create-and-serve-paper-petition.ts
```diff
@@ -302,4 +302,4 @@ export function createAndServePaperPetitionMyselfAndSpouse(
 
       return cy.wrap({ docketNumber: docketNumber! });
     });
-}
+}
```
jimlerza: 👎

cypress/local-only/tests/integration/practitionerEdit/admissions-clerk-changes-irs-practitioner-to-private.cy.ts
jimlerza, line 1: > it('should retain the users original role as an irsPractitioner on a case when they change from an irsPractitioner to a privatePractitioner'

Should it though?

scripts/archived/run-once/fix-cognito-users-without-ids.ts
jimlerza, line 1: This file can be deleted if we don't want to fix it

scripts/archived/run-once/fix-cognito-users-without-ids.ts
```diff
@@ -94,11 +94,11 @@ const setUserAttributes = async ({
...
-          user: omit(user, 'userId'),
-          userId: user.userId,
-        });
+        // await createPetitionerUserRecords({
+        //   applicationContext,
+        //   user: omit(user, 'userId'),
+        //   userId: user.userId,
+        // });
         console.log(`Created petitioner entity in Dynamo for ${user.email}.`);
         totals.userEntitiesCreatedInDynamo++;
```
jimlerza: ...or at least comment these lines too! lol

scripts/judge/set-judge-title.ts
```diff
@@ -0,0 +1,61 @@
...
+
+const { userId, judgeTitle } = parseArgsAndEnvVars(scriptConfig) as {
+  userId: string;
+  judgeTitle: JudgeTitle;
+};
+
+if (JUDGE_TITLES.includes(judgeTitle)) {
+  console.error(`Judge title must be valid`);
+  process.exit(1);
+}
```
jimlerza: Is this condition correct?

scripts/run-once-scripts/readBigFile.ts
jimlerza, line 1: Should we clean this up after?

scripts/user/become-user.ts
jimlerza, line 1: We run this _all the time_. It needs to work.

I even have a `tenille-become` alias for making (prod) Tenille become other users.

scripts/user/offboard-judge-user.sh
```diff
@@ -1,5 +1,9 @@
 #!/bin/bash
 
+# ⚠️ WARNING ⚠️
+# THIS SCRIPT IS BROKEN AFTER THE POSTGRES REFACTORING.  YOU WILL NEED TO UPDATE IT IF YOU PLAN TO RUN IT
+# ⚠️ WARNING ⚠️
```
jimlerza: ok so this one isn't run daily or anything but I still think we ought to update it.

scripts/user/revoke-e-access.ts
jimlerza, line 1: We use this script too. Can't delete this

scripts/user/rotate-environment-secrets.ts
```diff
@@ -75,16 +75,64 @@ const rotateSecrets = async (environmentName: string): Promise<void> => {
...
-  const putSecretValueCommand = new PutSecretValueCommand({
-    SecretId: `${env}_deploy`,
-    SecretString: JSON.stringify({
-      ...secrets,
-      DEFAULT_ACCOUNT_PASS,
-      USTC_ADMIN_PASS,
-    }),
-  });
-  await secretsClient.send(putSecretValueCommand);
-  console.log('✅ Secrets updated');
```
jimlerza: There's a USTC admin password in every environment, regardless of whether that environment is "zendesk-compatible." It needs to be rotated regardless.

Also, the term we've used for that everywhere else is "prodlike environment," for better or worse.

web-api/src/getConnection3.test.ts
```diff
@@ -25,7 +25,7 @@ import { environment } from '@web-api/environment';
 import { sleep } from '@shared/tools/helpers';
 
 describe('getConnection', () => {
-  environment.stage = 'testing';
+  environment.stage = 'testing'
```
jimlerza: but why?

web-api/src/getLockingConnection.ts
```diff
@@ -41,7 +41,7 @@ async function generateRDSAuthToken() {
 
 async function getToken() {
   const token =
-    environment.nodeEnv !== 'production'
+    environment.stage === 'local'
```
jimlerza, line 44: 🤩 

thanks for this

web-api/src/getPersistenceGateway.ts
jimlerza, line 1: 💯

web-api/src/persistence/postgres/utils/migrate/migrations/2025-06-24T17_12_01Z-user-table.ts
```diff
@@ -0,0 +1,126 @@
...
+  await db.schema.dropIndex('idx_user_name').execute();
+  await db.schema.dropIndex('idx_user_role').execute();
+  await db.schema.dropIndex('idx_user_bar_number').execute();
+  await db.schema.dropIndex('idx_user_section').execute();
+
+  // Drop tables
+  await db.schema.dropTable('dwUserOnCasePending').execute();
+  await db.schema.dropTable('dwUserConfirmationCode').execute();
+  await db.schema.dropTable('dwUserOnCase').execute();
+  await db.schema.dropTable('dwUser').execute();
```
jimlerza: Ok...so we're not gonna have to roll this back, because we Always Fail Forward. ™️ 

But let's say IRL we really actually need to run this `down` due to catastrophic unforeseen circumstances. That's truly all it's here for, right? So in that incredibly unlikely event, shouldn't we rename these tables so we don't lose the data?

web-api/src/persistence/postgres/cases/pendingCases/schema.ts
```diff
@@ -0,0 +1,26 @@
...
+const DEFAULT = {};
+
+// TODO: consider redesiging "dwUserOnCasePending" feature. As-is,
+// when a practitioner submits a document that requires court approval before
+// they are associated with the case, the system creates a "pending" case
+// association. This "pending" association is never deleted: it is instead
+// orphaned and effectvely overridden by the existence of a record in userOnCase
+// that has the same userId-docketNumber combination. What we have below is
+// effectively a one-to-one reimlementation of how these associations were
+// stored in dynamodb.
```
jimlerza, line 12: Excellent point. Should we put in a devex card for this?

web-api/src/persistence/postgres/cases/userOnCase/mapper.ts
```diff
@@ -0,0 +1,18 @@
+import {
+  NewUserOnCaseKysely,
+  UserOnCaseAssociation,
+} from '@web-api/persistence/postgres/cases/userOnCase/schema';
+
+export function toKyselyNewUserOnCase(
+  association: UserOnCaseAssociation,
+): NewUserOnCaseKysely {
```
jimlerza, line 8: Yay for embracing the recently-established `toKysely[Thing]` paradigm!

web-api/src/persistence/postgres/users/getInternalUsers.ts
```diff
@@ -0,0 +1,19 @@
...
+  DOCKET_SECTION,
+  PETITIONS_SECTION,
+} from '../../../../../shared/src/business/entities/EntityConstants';
+import { DbUser, rawUser } from '@web-api/persistence/postgres/users/mapper';
+
+export const getInternalUsers = async (): Promise<DbUser[]> => {
+  const users = await getDbReader(db =>
+    db
+      .selectFrom('dwUser')
+      .where('section', 'in', [DOCKET_SECTION, PETITIONS_SECTION, ADC_SECTION])
```
jimlerza: I believe the "general" role is an "internal" user. And possibly (probably?) chambers as well.

web-api/src/persistence/postgres/users/mapper.ts
```diff
@@ -0,0 +1,97 @@
...
+export function rawUser(user: UserKysely) {
+  return transformNullToUndefined({
+    ...user,
+    admissionsDate: user.admissionsDate
+      ? formatDateString(user.admissionsDate?.toISOString(), FORMATS.YYYYMMDD)
+      : null,
+    pendingEmailVerificationTokenTimestamp:
+      user.pendingEmailVerificationTokenTimestamp?.toISOString(),
+  });
+}
```
jimlerza: Should this be called `fromKyselyUser` to match convention?

jimlerza (changes requested):
Fantastic work!

I have identified just a few things that need attention. Nothing major.

jimlerza:
I don't see manual steps listed for generating the initial values for the zendesk user credentials or for creating the new zendesk user. Does that happen magically?

Also need to include these values in `scripts/secrets/create-env-secrets.ts` as well.

jimlerza:
Why are we still indexing users in Opensearch? I know we had to keep cases and docket entries in Opensearch because we didn't want to reimplement the advanced search operators in postgres. User search doesn't use those.

Are we planning a subsequent cleanup effort to remove the `efcms-user` index and move user-related queries to postgres?

scripts/judge/set-judge-title.ts
codyseibert: fixing now

scripts/archived/run-once/fix-cognito-users-without-ids.ts
codyseibert: deleting the file

scripts/archived/run-once/fix-cognito-users-without-ids.ts
codyseibert: deleting the file

scripts/user/become-user.ts
codyseibert: fixing

scripts/user/offboard-judge-user.sh
codyseibert: fixing

scripts/user/revoke-e-access.ts
codyseibert: adding back

web-api/src/persistence/postgres/utils/migrate/migrations/2025-06-24T17_12_01Z-user-table.ts
codyseibert: good point, I can rename it instead of dropping.  we do drop in all the other scripts though, so maybe that's a new pattern we should discuss going forward

web-api/src/persistence/postgres/users/mapper.ts
codyseibert: fixing

cypress/local-only/tests/integration/practitionerEdit/admissions-clerk-changes-irs-practitioner-to-private.cy.ts
codyseibert: from our understanding, yes, because a user might be on a bunch of closed cases as the irsPractitioner, and if you change to privatePractitioner we do not want to change the actingAsRole field on those cases.  Recently there was logic added to the frontend to prevent someone changing their role if they have open cases, so in no scenario would a irsPRactitioner be on an open case and then need to switch to a privatePractitioner, at least our system will not allow it.

web-api/src/persistence/postgres/cases/pendingCases/schema.ts
codyseibert: devex card added https://github.com/orgs/flexion/projects/11/views/1?pane=issue&itemId=125361990&issue=flexion%7Cef-cms%7C10751

codyseibert:
> Why are we still indexing users in Opensearch? I know we had to keep cases and docket entries in Opensearch because we didn't want to reimplement the advanced search operators in postgres. User search doesn't use those.
> 
> Are we planning a subsequent cleanup effort to remove the `efcms-user` index and move user-related queries to postgres?

I think we are keeping this around because we still use practitioner search which does some of those special elasticsearch queries when searching for practitioner name.  I think we could potentially move that to just do a postgres search, but I think we deferred when we discussed it.  I will confirm with Zach.

scripts/user/rotate-environment-secrets.ts
TomElliottFlexion: I believe the USTC admin password will be rotated either way, on line 124.

web-api/src/persistence/postgres/users/getInternalUsers.ts
codyseibert: this entire method is actually unused now, so we are removing it.

TomElliottFlexion:
> I don't see manual steps listed for generating the initial values for the zendesk user credentials or for creating the new zendesk user. Does that happen magically?
> 
> Also need to include these values in `scripts/secrets/create-env-secrets.ts` as well.

I've added manual steps to merge the zendesk pr and run `./scripts/user/setup-zendesk-user.ts`
Do we want this in `create-env-secrets.ts`? That script neither runs on prod nor updates the Zendesk secrets. Happy to add it, but it sounds like this is more Zendesk-specific than environment-specific

jimlerza:
> > I don't see manual steps listed for generating the initial values for the zendesk user credentials or for creating the new zendesk user. Does that happen magically?
> > Also need to include these values in `scripts/secrets/create-env-secrets.ts` as well.
> 
> I've added manual steps to merge the zendesk pr and run `./scripts/user/setup-zendesk-user.ts` Do we want this in `create-env-secrets.ts`? That script neither runs on prod nor updates the Zendesk secrets. Happy to add it, but it sounds like this is more Zendesk-specific than environment-specific

Dynamsoft is only enabled on test and prod, but the dynamsoft code and infrastructure is still deployed everywhere. Similarly, I'd rather have the dawson-side zendesk infra in place everywhere so all we have to do in the future to "turn on" zendesk in a given environment is deploy the zendesk lambdas (and secrets).

This should actually simplify the changes you made to the secrets rotation script. The admin and zendesk users should always exist and their secrets should always be rotated.

scripts/user/rotate-environment-secrets.ts
jimlerza: Oh I see. Perhaps we should check for the existence of the zendesk secrets to determine if a given environment is zendesk-compatible?

TomElliottFlexion: That sounds reasonable given that we won't necessarily be limiting zendesk to dev and prod anymore

jimlerza (approved):
🔥

TomElliottFlexion (approved):

cypress/local-only/tests/integration/statusReportOrder/edit-status-report-order.cy.ts
```diff
@@ -136,7 +136,7 @@ describe('edit status report order', () => {
       cy.get('#additional-order-text').should('contain', 'Test');
     });
 
-    it('should load existing signed order', () => {
+    it.only('should load existing signed order', () => {
```
zacharybrown204, line 139: an it.only snuck in

TomElliottFlexion: Sacrilege!

PR #6385, opened 2025-08-05
10576: Handle Lead Cases being closed;

cruzjone-flexion:


ChristieLaferriereTaxCourt (approved):

jimlerza (approved):

shared/src/business/useCases/updateCaseContextInteractor.ts
```diff
@@ -106,13 +108,36 @@ export const updateCaseContext = async (
...
+            LEAD_CASE_DEADLINES,
+            oldCase.leadDocketNumber,
+          );
+
+        DEADLINE_TASKS.push(
+          ...CHILDREN_DEADLINES.map(async childCaseDeadline => {
+            return upsertCaseDeadlines([
+              {
+                ...childCaseDeadline,
+                consolidatedCaseDeadlineId: undefined,
```
TomElliottFlexion, line 133: Just double checking, will this remove the deadline? Does it need to be null?

cruzjone-flexion: This will remove the parent (lead case) deadline association from the children deadlines;

PR #6386, opened 2025-08-06
8990: Display recently filed documents (formerly flexion#10685)

AnthonyLoeraTaxCourt:
#8990


shared/src/business/useCases/getRecentFilingsForUserInteractor.ts
```diff
@@ -0,0 +1,138 @@
+import { ServerApplicationContext } from '@web-api/applicationContext';
```
Mwindo: I know this is a draft, but just a quick note: if we can do this in Postgres after 10495, that would probably be best! We are trying to use OpenSearch only as really needed, and I don't see anything in here that Postgres couldn't handle. If that seems right, it might be worth merging 10495 into this ticket, but I would discuss this with Chris/Tenille/Jim.

PR #6388, opened 2025-08-06
Devex: Remove processDocketEntries

Mwindo:
Also remove `deleteRecord`, which is apparently unused.

We should not be using `processDocketEntries` anymore since the docket entry data is already in Postgres.


web-api/src/business/useCases/processStreamRecords/processStreamRecordsInteractor.test.ts
```diff
@@ -18,22 +16,14 @@ const errorSpy = jest.spyOn(logger, 'error');
...
-      caseDeadlineRecords: [],
-      caseEntityRecords: [],
-      caseWorksheetRecords: [],
-      docketEntryRecords: [],
       irsPractitionerMappingRecords: [],
       otherRecords: [],
       privatePractitionerMappingRecords: [],
       removeRecords: [],
-      userCaseNoteRecords: [],
-      workItemRecords: [],
```
Mwindo, line 36: Most of these guys have actually been gone for a while.

jimlerza (approved):

AbelTaxCourt (approved):

PR #6407, opened 2025-08-13
OpEx: Re-Populate the docket_entry_id column in the dw_work_item table in environments in which it has been lost

jimlerza:
When they were separated by a whole month of near-daily deployments, the `2025-06-26T16_31_56Z-normalize-docket-entry` and `2025-07-26T23_34_53Z-drop-docketEntry-from-dwWorkItem` postgres migrations did not conflict. But when run concurrently (during the last glue job), they resulted in an empty `docket_entry_id` column in the `dw_work_item` table. Oops!

Here's how we're addressing this:

* added a new `restoreTableFromSource` script that can restore a single table from prod into a lower environment, with all the same obfuscation goodies that the `restoreDbFromSource` script uses.
* manual step: we'll run that script to pull down a copy of prod's `dw_work_item` table to a new `dw_work_item_prod` table in test
* added a new `populate-dw_work_item-docket_entry_id-column` **run-once** script that will read the missing data out of the `dw_work_item_prod` table and write it to `dw_work_item`'s `docket_entry_id` column
* manual step: we'll run the aforementioned run-once script to restore the lost data in test
* we won't run another glue job while `prod` and `test` diverge so significantly

Ok, that makes sense. But why is this PR going to staging if it's just fixing an oopsie that will eventually correct itself with the next glue job? Answer: we imagine the ability to copy a single prod table down to a lower environment without clobbering the existing table will probably be worth keeping for future debugging purposes.

## Manual Steps !!!Test environment only!!!

- [x] Run script to copy prod `dw_work_item` table to a new `dw_work_item_prod` table in `test`
   ```bash
   . scripts/env/set-env.zsh ustc-prod
   export TARGET_ACCOUNT_ID="<test account id>"
   export TARGET_ENV=test
   scripts/postgres/restoreTableFromSource.ts dw_work_item dw_work_item_prod
   ```
- [x] Run script to repopulate the `docket_entry_id` column:
   ```bash
   . scripts/env/set-env.zsh ustc-test
   export PGPASSWORD=$(scripts/postgres/generate-token.sh -t -w)
   psql \
     --host=$DB_HOST \
     --username=$DB_USER \
     --dbname=$DB_NAME \
     --port=5432 \
     --file=scripts/run-once-scripts/populate-dw_work_item-docket_entry_id-column.sql \
     --echo-errors
   ```


scripts/postgres/restoreTableFromSource.ts
```diff
@@ -0,0 +1,139 @@
...
+      dbName: sourceDbname,
+      host: sourceHost,
+      port: sourcePort,
+      sourceTableName,
+      username: sourceUsername,
+    });
+
+    if (sourceTableName !== targetTableName) {
+      const sourceIndexPrefix = sourceTableName.replace('dw', 'idx');
+      const targetIndexPrefix = targetTableName.replace('dw', 'idx');
```
jimlerza, line 109: All of the indices in the `dw_work_item` table were prefixed with `idx_work_item`.

This code assumes that all indices are named following this convention. Therefore, indices that don't follow this naming convention will not be replaced by the `sed` call that follows.

scripts/run-once-scripts/populate-dw_work_item-docket_entry_id-column.sql
jimlerza, line 1: This script doesn't need to be run again. We would only be keeping it for reference. Perhaps it should go directly to the `archive` directory?

zachrog: If you want a reference I think you can add it to the archive but I am okay with deleting the file

scripts/postgres/restoreTableFromSource.ts
zachrog: So for the index name you can name it whatever you want to. We just happen to commonly use ''idx_${whatever}'.

Because this is a script that is only called as like a useful tool I am okay not enforcing this in any way but if we want to integrate this script somewhere or rely on it for deployments or something I would want to come up with a better solution for handling any index name.

zachrog (approved):

TomElliottFlexion (approved):

scripts/run-once-scripts/populate-dw_work_item-docket_entry_id-column.sql
jimlerza: Moved to archive

PR #6410, opened 2025-08-13
10685 clean up test to test

AnthonyLoeraTaxCourt:


web-client/src/presenter/computeds/recentFilingsHelper.ts
```diff
@@ -40,6 +57,44 @@ const checkDocumentAccess = (
...
+    'TCRP',
+    'SPOS',
+    'SPTO',
+    'SPTN',
+    'NTD',
+    'NORP',
+    'NOIP',
+    'NCTL',
+    'NODC',
+  ];
```
alexdziarn: I would add all these hard coded values to the EntitiyConstants file

web-client/src/presenter/computeds/recentFilingsHelper.ts
```diff
@@ -40,6 +57,44 @@ const checkDocumentAccess = (
...
+      'NOTR',
+      'NTD',
+      'SPOS',
+      'SPTO',
+      'TCRP',
+      'NORP',
+      'NOIP',
+      'NCTL',
+      'NODC',
+    ];
```
alexdziarn: same as above

alexdziarn (approved):

PR #6440, opened 2025-08-19
9036 part 2: Update and Configurable Engine Version (formerly flexion#10737)

codyseibert:
Adding changes to allow the engine_version of the rds instance to be configurable. Upgrading Postgres will cause some downtime, but the maintenance mode should automatically kick in when RDS is not accessible.

#### Before deployment

- [ ] Set the value of the `RDS_ENGINE_VERSION` secret in Secrets Manager to `17.5`
   ```bash
   scripts/secrets/update-secret.ts --key "RDS_ENGINE_VERSION" --value "17.5"
   ```


AnthonyLoeraTaxCourt (approved):

paulo-filho-ctr (approved):

jimlerza (approved):

scripts/postgres/rds-upgrade-global-engine.sh
jimlerza, line 1: Should we put this in `scripts/archived`?

jimlerza (approved):

PR #6445, opened 2025-08-19
8590 case title consolidated case (formerly flexion#10220)

rgf314159:
10220-case-title-consolidated-case

#8590


shared/src/business/entities/cases/Case.ts
```diff
@@ -1550,10 +1550,27 @@ export class Case extends JoiValidationEntity {
...
 
+  setCaseCaptionInConsolidatedCases() {
+
+    const currentCase = this.consolidatedCases.find((c) => c.docketNumber === this.docketNumber);
+
+    if (currentCase) {
+      currentCase.caseCaption = this.caseCaption;
+    }
+
+  }
```
jimlerza: I don't think I agree that this method "sets case caption in consolidated cases."

It finds a given case within an array of consolidated cases and sets the case caption on only the target case. As such, it needs a more descriptive name.

shared/src/business/entities/cases/Case.setCaseCaption.test.ts
```diff
@@ -17,4 +17,34 @@ describe('setCaseCaption', () => {
 
     expect(updatedCase.caseCaption).toEqual('A whole new caption');
   });
+
+  it('should set the case caption on the consolidated case array', () => {
```
jimlerza: This test asserts that `setCaseCaptionInConsolidatedCases` sets the case caption for a _single_ case in a consolidated group. The `it` declaration should better reflect that.

shared/src/business/entities/cases/Case.ts
```diff
@@ -1550,10 +1550,27 @@ export class Case extends JoiValidationEntity {
    * @returns {Case} the updated case entity
    */
   setCaseCaption(caseCaption) {
+    
     this.caseCaption = caseCaption;
+    
+    // New method to find case in case array and update that caption as well
```
jimlerza: We love comments that give additional context or explain things that aren't readily apparent from otherwise self-descriptive code. This comment, however, is unnecessary.

jimlerza (changes requested):

jimlerza (approved):

TomElliottFlexion (approved):

PR #6448, opened 2025-08-19
10712 high priority followup

zacharybrown204:


TomElliottFlexion:
Looking good! I suspect that there are more references to cases being `highPriority` in the codebase, in helpers and tests. Not sure if this is the place to remove them, but we'll want to make sure they are removed

jimlerza:
> Looking good! I suspect that there are more references to cases being `highPriority` in the codebase, in helpers and tests. Not sure if this is the place to remove them, but we'll want to make sure they are removed

I agree. There are still more instances of `highPriority` and `high priority` in the repo. If we don't want to squash them here, do we need a follow-up to the follow-up?

zacharybrown204:
Yeah I see them now, I missed them on the first pass. Taking care of them now.
Some of them are still valid, the high priority work item stuff is still in and cases still have high priority suffixes, but high priority cases need to be removed.
The only places I see it now are a few references in some unit tests

zacharybrown204:
Okay scanning through again I found and removed a few more references to case high priority. 
One in the old swagger doc, one in the elastic definitions, and one in a unit test.
Everything else that I saw seemed to be either a high priority suffix or for work items.

TomElliottFlexion (approved):

PR #6450, opened 2025-08-19
Bug Fix - Setting Section after Role

codyseibert:
@ttlenard found a bug related to switching a practitioner's practiceType and it not updating their section correctly.


jimlerza (approved):

TomElliottFlexion (approved):

jimlerza:
I think the test that is failing was probably taking advantage of this bug

PR #6469, opened 2025-08-21
10749 BUG: ignore suffixes when sorting docket numbers

TomElliottFlexion:


shared/src/business/utilities/sortRecentFilings.test.ts
```diff
@@ -43,15 +43,6 @@ describe('sortRecentFilings', () => {
...
 
-  it('should handle docket number sorting with invalid formats', () => {
-    const filings = [
-      createFiling({ docketNumber: 'invalid' }),
-      createFiling({ docketNumber: '101-20' }),
-    ];
-    const result = sortRecentFilings(filings, 'docketNumber', 'asc');
-    expect(result[0].docketNumber).toBe('invalid');
-  });
-
```
TomElliottFlexion, line 54: Unclear that we should be enforcing that invalid comes before a valid docket number

pixiwyn (approved):

PR #6471, opened 2025-08-22
9045 bugfix (formerly flexion#10748)

AnthonyLoeraTaxCourt:
#9045


shared/src/business/entities/Practitioner.ts
```diff
@@ -199,12 +199,14 @@ export class Practitioner extends User {
     return Practitioner.VALIDATION_RULES as any;
   }
 
-  toRawObject() {
+  toRawObject(removeValidationProperties: boolean = true) {
```
pixiwyn: Consider removing the boolean param altogether (or replace with an options object) to avoid boolean-trap mistakes in the future.

shared/src/business/entities/Practitioner.ts
```diff
@@ -199,12 +199,15 @@ export class Practitioner extends User {
...
     // We don't want to persist these values as they are only used for validation
-    result.confirmEmail = undefined;
-    result.updatedEmail = undefined;
+    if (removeValidationProperties) {
+      result.confirmEmail = undefined;
+      result.updatedEmail = undefined;
+    }
 
     return result;
   }
```
TomElliottFlexion, line 213: Thought about this for a while, and really the idea of tying `toRawObject` to database code is silly and outdated. I like this without the boolean trap

web-client/src/presenter/actions/updatePractitionerUserAction.ts
```diff
@@ -13,7 +14,10 @@ export const updatePractitionerUserAction = async ({
   get,
   path,
 }: ActionProps) => {
-  const user = get(state.form);
+  const formState = get(state.form);
+  const user = new Practitioner(formState, {
+    applicationContext,
+  }).toRawObject({ removeValidationProperties: false });
```
TomElliottFlexion, line 20: For posterity: this is being used to strip out `openCaseInfo` and `closedCaseInfo` from being passed to the backend, solving the bug

TomElliottFlexion (approved):

shared/src/business/entities/Practitioner.ts
TomElliottFlexion: Especially since we have mapper functions to deal with this type of problem now

shared/src/business/entities/cases/Case.ts
```diff
@@ -982,7 +982,8 @@ export class Case extends JoiValidationEntity {
   }
 
   //@ts-ignore
-  toRawObject(processPendingItems = true): RawCase {
+  toRawObject(options: { processPendingItems?: boolean } = {}): RawCase {
+    const { processPendingItems = true } = options;
     const result = this.toRawObjectFromJoi();
```
TomElliottFlexion, line 987: Good call

jimlerza (approved):

PR #9061, opened 2025-08-27
9619 sealing email address to test

alexdziarn:
Added tests and included conditional styling for participants in cases


web-api/src/business/useCaseHelper/service/createChangeItems.ts
```diff
@@ -66,7 +66,8 @@ const createDocketEntryForChange = async ({
 
   const isAddressSealed =
     caseEntity.petitioners.find(
-      petitioner => petitioner.email === newData.email,
+      petitioner =>
+        petitioner.email === newData.email && petitioner.name === newData.name,
```
TomElliottFlexion, line 70: Mostly out of curiosity, what prompted this change? If there's a chance of ambiguity here, we should probably compare by `userId`, even if it means passing more data around

alexdziarn: its super wierd, if the they have the same email, they have the same userId. I talked to Tennile and Chris, and they said in practice, they shouldn't even have the same email, but can happen if there is human error or if its a really old case. I added name here just in case they do have the same email since name and address are the only differentiating fields.

TomElliottFlexion: Ah gotcha thanks

web-client/src/views/CaseDetail/ParticipantsAndCounsel.tsx
```diff
@@ -56,8 +57,17 @@ export const ParticipantsAndCounsel = connect(
...
+                    className={classNames({
+                      'margin-left-205': petitioner.isAddressSealed,
+                      'sealed-address': petitioner.isAddressSealed,
+                      'address-line': true,
+                    })}
+                    data-testid={`participant-email`}
+                  >
+                    {!petitioner.sealedAndUnavailable && (
+                      <span>{petitioner.formattedEmail}</span>
+                    )}
```
TomElliottFlexion, line 70: Just to be 100% sure, this email is unavailable to the user even without this check, right? Ie it shouldn't have come back on the network request

TomElliottFlexion (approved):

web-client/src/views/CaseDetail/ParticipantsAndCounsel.tsx
alexdziarn: yep i see the issue, I moved the conditional render up one component

TomElliottFlexion: Oh good catch! That wasn't even what I was asking- I just wanted to make sure that we weren't relying on the tsx to hide this sealed email. The email shouldn't even come back to the user's browser via request if it was sealed to them

alexdziarn: The email isn't returned in the request, but the contactid is. This could be related, but if a user is able to match the contactid from another case whose address isn't sealed could get the email that way.

TomElliottFlexion: Interesting...I think that's probably fine, but it could be worth double checking w/ Chris/Tenille

alexdziarn: This issue will be taken care of in a seperate ticket according to Chris

Issue #9004, opened 2025-07-08
Pro Se Checklist


cholly75:
As a pro se petitioner, so that I can be prepared for the next steps in my case, I need the Pro Se Checklist appended to the Notice of Receipt of Petition.

The Court's Pro Se Committee has determined that a pro se checklist will be added to the NOTR package that gets sent out to pro se petitioners. The attached checklist should be reproduced in full as the page directly after the NOTR and before any appended clinic letters.

## Acceptance Criteria
* When generating a NOTR, if the petitioner for which the NOTR is generated is pro se (i.e., not represented by counsel at the time of generation), append the Pro Se Checklist to the NOTR.
* Pro Se Checklist is appended to the NOTR prior to any clinic letters that may be appended
* Pro Se Checklist resides in S3 and can be updated w/ new verbiage/copy by replacing said file in S3

## Notes
Pro Se Checklist: 

[Pro Se Petitioner(s) Checklist - Revised 08-07-2025.pdf](https://github.com/user-attachments/files/21736419/Pro.Se.Petitioner.s.Checklist.-.Revised.08-07-2025.pdf)

## Test Cases

Test Rail test cases are located here:

[10700.xlsx](https://github.com/user-attachments/files/21990522/10700.xlsx)

## Story Definition of Ready (updated on 12/23/22)
The following criteria must be met in order for the user story to be picked up by the Flexion development team.
The user story must: 
- [ ] Is framed in business/user need, the value has been addressed.
- [ ] Includes acceptance criteria
- [ ] Has been refined
- [ ] Pre conditions have been satisfied. 

**Process:** 
Flexion developers and designers will test if the story meets acceptance criteria and test cases in Flexion dev and staging environments (“standard testing”). If additional acceptance criteria or testing scenarios are discovered while the story is in progress, a new story should be created, added to the backlog and prioritized by the product owner. 

## Definition of Done (Updated 5-19-22)
**Product Owner**
 - [ ]  Acceptance criteria have been met and validated on the Court's migration environment
 - [ ] Add scenario to testing document, if applicable (https://docs.google.com/spreadsheets/d/1FUHKC_YrT-PosaWD5gRVmsDzI1HS_U-8CyMIb-qX9EA/edit?usp=sharing)

**UX**
 - [ ] Business test scenarios have been refined to meet all acceptance criteria 
 - [ ] Usability has been validated
 - [ ] Wiki has been updated (if applicable) 
[...]

PR #9075, opened 2025-09-02
9004: adds prosechecklist to notr

alexdziarn:
https://github.com/ustaxcourt/ef-cms/issues/9004

#### Before deployment

- [ ] Copy the pro-se checklist file to the documents bucket
   ```bash
   aws s3 cp ~/Desktop/pro-se-checklist "s3://${EFCMS_DOMAIN}-documents-${ENV}-us-east-1"
   ```


TomElliottFlexion:
Looking great! One suggestion, take it or leave it, is to add a script for adding the new checklist to s3 the first time. Doing it manually isn't really a problem, but making it repeatable makes it easier to stand up new environments

scripts/s3/upload-documents-to-s3.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash -e
+
```
codyseibert: usually our .sh scripts have a "how to use" section in comments.  I'd suggest adding one which also includes an intent of this script, when would we use it, why would we use it.  Is the purpose of this script just to upload the prose.pdf to the s3 bucket from a directory?

alexdziarn: right now, we upload all the clinic letters(~50 pdfs) and the pro se checklist to s3 manually. This is just an easier way upload them.

scripts/s3/documents/pro-se-checklist
jimlerza, line 1: I don't think this file belongs in the repo

scripts/s3/upload-documents-to-s3.sh
jimlerza, line 1: I realize this script was added in response to Tom's feedback, but I don't think it has any real-life use case, at least not as-is.

1. this would be better as a typescript script - the majority of the lines in this file are the type of boilerplate stuff that we've been getting away from in our typescript scripts
2. Tenille is the person who does this IRL and this script does not help Tenille. Her workstation is a Windows laptop.

If the problem we're trying to solve is giving Tenille a solution for bulk uploading lots of documents to the prod documents bucket, this missed the mark.

If the problem we're trying to solve is a mechanism for uploading the pro-se checklist file to the prod docs bucket when this story deploys to prod, I think we'd be better served including that as a Manual Step:

```
. scripts/env/set-env.zsh prod
aws s3 cp /path/to/pro-se-checklist "s3://${EFCMS_DOMAIN}-documents-${ENV}-us-east-1"
```

scripts/s3/upload-documents-to-s3.sh
```diff
@@ -0,0 +1,78 @@
...
+# uploads or overwrites existing files in s3 documents directory for the respective aws environment
+
+# Usage
+#   . scripts/env/set-env.zsh {YOUR ENV HERE}
+# After setting the environment, use one of these commands
+#   zsh scripts/s3/upload-documents-to-s3.sh
+#   bash scripts/s3/upload-documents-to-s3.sh
```
jimlerza: I think the usage that is most important to explain is "there is a magic directory that you need to put files in first" and that isn't mentioned at all.

cypress/helpers/fileAPetition/petitioner-creates-electronic-case.ts
```diff
@@ -155,6 +155,69 @@ export function petitionerCreatesElectronicCaseWithSpouse(
     });
 }
 
+export function externalUserCreatesElectronicCaseWithSpouseDifferentAddress(
+  primaryFilerName: string = 'John',
+  secondaryFilerName: string = 'Sally',
```
jimlerza: Should we use faker to randomize the personal info in this test?

jimlerza (dismissed):

codyseibert (dismissed):

cypress/helpers/fileAPetition/petitioner-creates-electronic-case.ts
```diff
@@ -155,6 +155,69 @@ export function petitionerCreatesElectronicCaseWithSpouse(
...
+) {
+  cy.get('[data-testid="file-a-petition"]').click();
+  cy.get('[data-testid="go-to-step-1"]').click();
+
+  cy.get('[data-testid="filing-type-1"]').click();
+  cy.get('[data-testid="is-spouse-deceased-1"]').click();
+  cy.get('body').then($body => {
+    if ($body.find('[data-testid="have-spouse-consent-label"]').length > 0) {
```
codyseibert, line 169: I wouldn't use cy.get('body'). if we expect the label to always exist, just find the label and click it.  I don't think we need to first get the body then find the element in the body.  Cypress will find the data-testid="have-spouse-consent-label and click it, if it never shows up, then the test is failing for the right reason.

cypress/local-only/tests/integration/proSeChecklist/pro-se-checklist.cy.ts
```diff
@@ -0,0 +1,131 @@
...
+  });
+  it('appends pro se checklist for petitioner when petition filed by petitions clerk', () => {
+    createAndServePaperPetition().then(({ docketNumber }) => {
+      goToCase(docketNumber);
+      cy.get(
+        `[data-testid="docket-entry-numberOfPages-${docketNumber}"]`,
+      ).contains(2);
+    });
+  });
+  it('appends pro se checklist for petitioner and spouse when petition filed by peititions clerk', () => {
```
pixiwyn: typo on peititions should be petitions

cypress/local-only/tests/integration/proSeChecklist/pro-se-checklist.cy.ts
```diff
@@ -0,0 +1,131 @@
...
+    createAndServePaperPetitionMyselfAndSpouse().then(({ docketNumber }) => {
+      goToCase(docketNumber);
+      cy.get(
+        `[data-testid="docket-entry-numberOfPages-${docketNumber}"]`,
+      ).contains(2);
+    });
+  });
+  it('appends pro se checklist for each petitioner and spouse of different addresses when petition is filed by peitions clerk', () => {
```
pixiwyn: typo on peititions should be petitions

cypress/helpers/fileAPetition/petitioner-creates-electronic-case.ts
alexdziarn: so this is a function that should work for both private practitioners and petitioners when creating a case. The checkbox exists for petitioners, but not private practitioners. They follow the same exact steps other than this one checkbox

TomElliottFlexion (dismissed):

jimlerza (approved):

jimlerza (approved):

Issue #8700, opened 2024-04-26
BUG: Data in the C/S, Attachments, Additional info 2, and selections for if there are objections for motions are duplicated on the printable docket record (formerly flexion#10367)


ttlenard:
**Describe the Bug**
If Certificate of Service or Attachments is checked, or if text is added to the "Additional info 2" text box in the document metadata, this text is duplicated on the Printable docket record.

For motions, the objection information that is selected when filing the document is duplicated also on the Printable docket record

**Business Impact/Reason for Severity**
Low

**In which environment did you see this bug?**
Prod, Test

**Who were you logged in as?**
Any user

**What were you doing when you discovered this bug? (Using the application, demoing, smoke tests, testing other functionality, etc.)**
Testing other functionality

**To Reproduce**
1. Log in to DAWSON as a docket clerk
2. Navigate to a docket record that has at least one motion filed already
3. For any motion document, click on the edit link
4. Check the boxes for certificate of Service and Attachments
5. Add in text to the additional info 2 text box
6. Click save
7. Notice that on the docket record, this text is displayed appropriately (one time)
8. Next, click on the printable docket record link
9. Notice that on the printable docket record, the Certificate of Service, Attachments and the text you added in the additional info 2 text box is now duplicated.  The objection information for the motion is also duplicated (nothing will display if objection is unknown).

The certificate of Service, Attachments, additional info 2 and motion objection info should not be getting duplicated on the printable docket record

**Actual Behavior**
A clear and concise description of what actually happened.
Any data in the Certificate of Service, Attachments, additional info 2 or objection info is getting duplicated on the printable docket record.

**Screenshots**

This image shows the duplication in Certificate of Service, Attachments, Motion objections?, and Additional Info 2 text when viewing the printable docket record.

<img width="805" height="788" alt="Image" src="https://github.com/user-attachments/assets/e4aa8892-76d2-4695-a248-d92559ead131" />

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Cause of Bug, If Known**

**Process for Logging a Bug:**
* Complete the above information
* Add a severity tag (Critical, High Severity, Medium Severity or Low Severity). See below for priority definition. 

**Severity Definition:**
* Critical Defect
Blocks entire system's or module’s functionality
No workarounds available
Testing cannot proceed further without bug being fixed.

* High-severity Defect
Affects key functionality of an application
There's a workaround, but not obvious or easy
[...]

PR #9077, opened 2025-09-03
8700: duplicated additional info in printable docket (formerly flexion#10367)

jdr81394:
#8700 

Prevents duplicate additional info from unnessarily showing up in the "additional info" column of the docket record.


shared/src/business/utilities/documentGenerators/docketRecord.test.ts
```diff
@@ -1,3 +1,19 @@
...
+ *
+ * The test data includes:
+ * - Case caption and title
+ * - Docket number and suffix
+ * - Petitioner and practitioner details
+ * - A sample docket entry with formatted dates and served status
+ *
+ * @see docketRecord
+ * @see generateAndVerifyPdfDiff
+ */
```
TomElliottFlexion, line 16: Don't see the harm in adding this, but it should be unnecessary if we've written our tests well

TomElliottFlexion (approved):

shared/src/business/utilities/documentGenerators/docketRecord.test.ts
jimlerza: ➕

jimlerza (approved):

alexdziarn (approved):

Issue #9028, opened 2025-07-25
v2 API: return isSealed boolean in case and docket entry entities


jimlerza:
As an IRS Superuser, so that I can display the sealed status of cases and docket entries in my eCite and eTCS applications, I need `Case` and `DocketEntry` entities returned by the [/v2/cases/{docketNumber}](https://github.com/ustaxcourt/ef-cms/blob/staging/docs/api/v2.yaml#L20) and [/v2/reconciliation-report/{date}](https://github.com/ustaxcourt/ef-cms/blob/staging/docs/api/v2.yaml#L86) endpoints to contain the `isSealed` flag.

## Acceptance Criteria

- [ ] case json contains `isSealed`
- [ ] docket entry json contains `isSealed`
- [ ] Swagger docs updated

## Notes

- See [irs-super-user.md](https://github.com/ustaxcourt/ef-cms/blob/staging/docs/additional-resources/irs-super-user.md) for instructions on testing V2 endpoints as the IRS superuser

## Test Cases

- https://github.com/flexion/ef-cms/issues/10728#issuecomment-3208500882

## Story Definition of Ready (updated on 12/23/22)
The following criteria must be met in order for the user story to be picked up by the Flexion development team.
The user story must: 
- [ ] Is framed in business/user need, the value has been addressed.
- [ ] Includes acceptance criteria
- [ ] Has been refined
- [ ] Pre conditions have been satisfied. 

**Process:** 
Flexion developers and designers will test if the story meets acceptance criteria and test cases in Flexion dev and staging environments (“standard testing”). If additional acceptance criteria or testing scenarios are discovered while the story is in progress, a new story should be created, added to the backlog and prioritized by the product owner. 

## Definition of Done (Updated 5-19-22)
**Product Owner**
 - [ ]  Acceptance criteria have been met and validated on the Court's migration environment
 - [ ] Add scenario to testing document, if applicable (https://docs.google.com/spreadsheets/d/1FUHKC_YrT-PosaWD5gRVmsDzI1HS_U-8CyMIb-qX9EA/edit?usp=sharing)

**UX**
 - [ ] Business test scenarios have been refined to meet all acceptance criteria 
 - [ ] Usability has been validated
 - [ ] Wiki has been updated (if applicable) 
 - [ ] Story has been tested on a mobile device (for external users only)

**Engineering**
 - [ ] Automated test scripts have been written, including visual tests for newly added PDFs.
[...]

PR #9083, opened 2025-09-03
9028 - Adding isSealed to a couple v2 endpoints

zacharybrown204:
#9028


codyseibert (approved):

jimlerza (approved):

shared/src/business/entities/ReconciliationReportEntry.ts
```diff
@@ -29,6 +30,7 @@ export class ReconciliationReportEntry extends JoiValidationEntity {
     this.index = rawDocketEntry.index;
     this.servedAt = rawDocketEntry.servedAt;
     this.servedPartiesCode = rawDocketEntry.servedPartiesCode;
+    this.isSealed = rawDocketEntry.isSealed ?? false;
```
ChristieLaferriereTaxCourt, line 33: Awesome use of the nullish coalescing operator here! The current `??` approach does gracefully handle `null/undefined` cases. However, it might still allow problematic values like string `"false"` if it was parsed from an API (which is truthy) or inconsistent handling of numeric `0` vs string `"0"`.

Also, if `rawDocketEntry.isSealed` evaluates to `false` (since its type is a boolean), we're essentially doing `false ?? false`, which just assigns `false` in either case, potentially making the `?? false` redundant in that case. 

To be more defensive and robust, could we maybe implement strict equivalence here or something else that makes this more conservative while not being redundant?

ChristieLaferriereTaxCourt: Screenshot 2025-09-27 at 7 06 43 PM

PR #9096, opened 2025-09-05
9071: second deployment for 8840 (formerly flexion#10507) to test

jimlerza:


jimlerza:
I suspect we will need a script to delete all of the records out of the deploy table before trying to destroy it in terraform. I also suspect we will need to remove deletion protection on the table outside of terraform as well.

jimlerza:
> I suspect we will need a script to delete all of the records out of the deploy table before trying to destroy it in terraform. I also suspect we will need to remove deletion protection on the table outside of terraform as well.

Update: [we don't](https://app.circleci.com/pipelines/github/ustaxcourt/ef-cms/13314/workflows/626876bb-e702-4dc2-9dfe-a5b88730464e/jobs/81792#step-109-25062_72)!

Issue #8857, opened 2024-10-10
BUG: Judge/Chambers user can access New trial session pages if they know the URL (formerly flexion#10528)


ttlenard:
**Describe the Bug**

Judges and Chambers users do not see the New tab when viewing the Trial Sessions page; however, if they know the URL of a specific New session, they can paste this URL into the address bar and view the New trial session details.  In this story (https://github.com/flexion/ef-cms/issues/9481), we removed the New tab from view, but if Judge/Chambers know the URL, they can still access the new sessions.

Since we removed the New tab for these users, we should probably remove the ability for Judge/Chambers users from still accessing the data that is on the New tab.

**Business Impact/Reason for Severity**
Low

**In which environment did you see this bug?**
Test/Staging

**Who were you logged in as?**
Judge, Chambers, Petitions Clerk

**What were you doing when you discovered this bug? (Using the application, demoing, smoke tests, testing other functionality, etc.)**
Testing other functionality

**To Reproduce**
1. Log in as a Petitions Clerk
2. Click on Trial Sessions Tab
3. Click on the new tab
4. Click on the City/State link for a new trial session
5. Copy the URL
6. Log out as the Petitions Clerk
7. Log in as a Judge or Chambers
8. Paste the URL into your address bar
9. Notice that you now have access to the New trial session details.

**Expected Behavior**
Judge/Chambers users shouldn't have access to New trial Sessions

**Actual Behavior**
A clear and concise description of what actually happened.
Judge/chambers users have access to view the trial session details for a New Trial Sessions

**Screenshots**
This is a Chambers user that copied a new session URL. They can see the new session:

![Image](https://github.com/user-attachments/assets/c5aa3831-ddbf-4942-a0bb-fb4f6de2966c)

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Cause of Bug, If Known**

**Process for Logging a Bug:**
* Complete the above information
* Add a severity tag (Critical, High Severity, Medium Severity or Low Severity). See below for priority definition. 

**Severity Definition:**
* Critical Defect
Blocks entire system's or module’s functionality
No workarounds available
Testing cannot proceed further without bug being fixed.

* High-severity Defect
Affects key functionality of an application
There's a workaround, but not obvious or easy
App behaves in a way that is strongly different from the one stated in the requirements

* Medium-severity Defect
A minor function does not behave in a way stated in the requirements.
Workaround is available and easy

* Low-severity Defect
[...]

PR #9099, opened 2025-09-05
8857 (formerly 10528 )  judge chambers can access urls

jdr81394:
#8857


jimlerza (approved):

codyseibert (approved):

codyseibert:
is there any requirement to prevent the backend use case from being able to be ran in this scenario, or are we just hiding the url?

jdr81394:
> is there any requirement to prevent the backend use case from being able to be ran in this scenario, or are we just hiding the url?

Hey Cody, nope , there is no requirement to prevent the backend use case. We're simply trying to hide the URL on the front end to prevent the judges from getting distracted.

PR #9115, opened 2025-09-09
Changed text to match desired values

jdr81394:


zacharybrown204 (approved):

AbelTaxCourt (approved):

codyseibert (approved):

shared/src/business/entities/docketEntry/externalFilingEvents.ts
```diff
@@ -415,8 +415,8 @@ export const EXTERNAL_FILING_EVENTS: AllExternalFilingEvents = {
       ordinalField: '',
     },
     {
-      documentTitle: 'Report',
-      documentType: 'Report',
+      documentTitle: 'Expert Report',
+      documentType: 'Expert Report',
```
TomElliottFlexion, line 419: Do we need to migrate existing data?

TomElliottFlexion: Looks like no based on the story, but I'm not 100% sure off the top of my head whether this will continue to validate correctly

jdr81394: From story it looked like it didn't need to be migrated

PR #9117, opened 2025-09-09
8727: PO feedback private practioners order search fix to test

AbelTaxCourt:


ChristieLaferriereTaxCourt (dismissed):

shared/src/business/entities/EntityConstants.ts
```diff
@@ -1779,6 +1779,9 @@ export const MAX_ELASTICSEARCH_PAGINATION = 10000;
 export const MAX_SEARCH_CLIENT_RESULTS = 10000;
 export const MAX_SEARCH_RESULTS = 10000;
 
+export const RAW_BATCH_SIZE = 1500;
```
codyseibert: this name could probably be more descriptive for what feature this is related to.  RAW_BATCH_SIZE doesn't imply opensearch, order search or anything to me from initially reading it.   Same with MAX_FETCH_BATCHES

web-api/src/business/useCases/public/orderPublicSearchInteractor.ts
```diff
@@ -38,32 +44,65 @@ export const orderPublicSearchInteractor = async (
...
-      documentEventCodes: ORDER_EVENT_CODES,
-      omitSealed: true,
-      ...rawSearch,
-      from,
-      overrideResultSize: limit,
-    });
+  const detectionCeiling = Math.min(limit + 1, MAX_SEARCH_RESULTS + 1);
+  const desired = Math.min(limit, MAX_SEARCH_RESULTS);
+  const accumulated: any[] = [];
+  let searchAfter: any[] | undefined = undefined;
```
codyseibert: this pattern could be put in a util function, it's the same pattern used in 3 different interactors

codyseibert (dismissed):

shared/src/business/entities/EntityConstants.ts
AbelTaxCourt: I agree, will rename those two constants to improve code readability. Thanks!

jimlerza (dismissed):

web-api/src/business/useCases/public/orderPublicSearchInteractor.ts
AbelTaxCourt: completed, ty!

AnthonyLoeraTaxCourt (approved):

ChristieLaferriereTaxCourt (approved):

PR #9147, opened 2025-09-15
8764 to test

TomElliottFlexion:
- Change `Report` documentType to `Expert Report`
- Will update with expand/contract documentation if this works


jdr81394 (approved):

AbelTaxCourt (approved):

zacharybrown204 (approved):

scripts/run-once-scripts/postgres-migration/update-document-type.ts
jimlerza, line 1: This should be implemented as a migration, with up/down.

The up should do `Report` -> `Expert Report` and the down should do `Expert Report` -> `Report`.

Additionally, you should either:

A. implement app code that can handle both "Report" and "Expert Report" without error and deploy that _first_, so it is already the active code when this migration runs

or

B. implement this as an expand-and-contract migration does the following in separate deployments:
   1. creates a new column, copies the data from the old column and transforms it in the new column; app code deployed in this deployment knows to read from new column
   2. deletes all data in old column, copies the properly-transformed data from the new column to the old column; app code deployed in this deployment knows to read from the old column once again
   3. deletes the new column, which is no longer needed

jimlerza (changes requested):

jdr81394 (approved):

web-api/src/persistence/postgres/cases/reports/fetchPendingItems.ts
```diff
@@ -57,7 +57,8 @@ export const fetchPendingItems = async ({
         'c.trialDate',
         'c.trialLocation',
         'd.docketEntryId',
-        'd.documentType',
+        // 8764: will need to be reverted to documentType
+        'd.documentTypeUpdated',
```
TomElliottFlexion, line 61: I don't like this one bit. Open to better ideas/a whole new paradigm.

TomElliottFlexion:
Deployed both steps to exp3 env

jimlerza (approved):

Issue #9012, opened 2025-07-11
Remove Pending Item When Stricken


cholly75:
As a docket clerk, so that I do not have to remember to remove Pending Items from a case, I need DAWSON to automatically remove a Pending Item if the docket entry for that item is stricken.

When the Court deems an entry stricken, it means the item in question is considered to have never been entered in the first place. Accordingly, when striking items the docket clerk must remember to go back and remove the associated Pending Item if there is one associated with the stricken entry.

DAWSON should just do that for them.

## Acceptance Criteria
* When a docket entry is stricken, if it has an associated entry on the Tracked Items/Pending Report, that entry is removed as if a user had removed it.

## Test Cases

Test cases are in Test Rail and also available here:

[9012.xlsx](https://github.com/user-attachments/files/22125653/9012.xlsx)

## Story Definition of Ready (updated on 12/23/22)
The following criteria must be met in order for the user story to be picked up by the Flexion development team.
The user story must: 
- [ ] Is framed in business/user need, the value has been addressed.
- [ ] Includes acceptance criteria
- [ ] Has been refined
- [ ] Pre conditions have been satisfied. 

**Process:** 
Flexion developers and designers will test if the story meets acceptance criteria and test cases in Flexion dev and staging environments (“standard testing”). If additional acceptance criteria or testing scenarios are discovered while the story is in progress, a new story should be created, added to the backlog and prioritized by the product owner. 

## Definition of Done (Updated 5-19-22)
**Product Owner**
 - [ ]  Acceptance criteria have been met and validated on the Court's migration environment
 - [ ] Add scenario to testing document, if applicable (https://docs.google.com/spreadsheets/d/1FUHKC_YrT-PosaWD5gRVmsDzI1HS_U-8CyMIb-qX9EA/edit?usp=sharing)

**UX**
 - [ ] Business test scenarios have been refined to meet all acceptance criteria 
 - [ ] Usability has been validated
 - [ ] Wiki has been updated (if applicable) 
 - [ ] Story has been tested on a mobile device (for external users only)

**Engineering**
 - [ ] Automated test scripts have been written, including visual tests for newly added PDFs.
 - [ ] Field level and page level validation errors (front-end and server-side) integrated and functioning.
 - [ ] Verify that language for docket record for internal users and external users is identical.
 - [ ] New screens have been added to pa11y scripts.
[...]

PR #9154, opened 2025-09-16
9012 remove pending item

zacharybrown204:
#9012


jimlerza (approved):

jimlerza:
Should we add a cypress test for this?

Something like

1. unblocks case when the doc corresponding to the only pending item is stricken
   1. create case w/ 1 pending item
   2. assert case is blocked
   3. strike doc w/ pending item
   4. assert that case is no longer blocked
2. does not unblock case when there are more than one pending item and one is stricken
   1. create case w/ 2 pending items
   2. assert case is blocked
   3. strike doc w/ pending item
   4. assert that case is still blocked

zacharybrown204:
Can do, I'll add a cypress test for that.

web-api/src/business/useCases/docketEntry/strikeDocketEntryInteractor.ts
```diff
@@ -29,7 +30,7 @@ export const strikeDocketEntryInteractor = async (
     docketNumber,
   });
 
-  const caseEntity = new Case(caseToUpdate, { authorizedUser });
+  let caseEntity = new Case(caseToUpdate, { authorizedUser });
```
JayFlexy, line 33: Just curious, why opt for let over const here?

JayFlexy: For additional context, we could opt for `const updatedCaseEntity` over reassigning to the same variable. Admittedly, I have a preference for immutability over variables that can change at some point for debugging purposes. @zacharybrown204

zacharybrown204: Honestly I don't have a strong preference for either const or let here. I used let because I thought it looked a bit cleaner and the code I was referencing in removeCasePendingInteractor.ts also used a let here. 
I initialy did write it with a const but there but I felt it was a bit messy immediately having to rename it when there wasn't any big transformation on the object, just some data changing. 

But again I don't have a strong preference, if you want a const I can do that.

JayFlexy: Understood, and that makes sense. No need to change; just a nit on my end.

JayFlexy (approved):

PR #9170, opened 2025-09-18
8813 - Github action workflows caching part 2

zacharybrown204:
Updating our github actions to check for a cache. If there is no cache then do a clean install. At the end if we did a clean install then store our dependencies in the cache.

All in all the cache size is now about 700MB, we have 10GB of storage for these caches so we can hold aroundabout 13 caches at a time. If we go over this 10GB limit then github will delete the oldest caches until we are back under the limit.
Each pull request will have it's own cache that will expire after 7 days or if it is the oldest cache when our storage goes over 10GB.

These caches are keyed by a hash of our package lock, so they should be pretty static for most PRs.


jimlerza (approved):

paulo-filho-ctr (approved):

jimlerza:
For a moment I thought I had a major lightbulb:

"At least half of the actions don't need chromium, and chromium is a significant portion of the package size. Could we save space in the cache by tossing it (just for those actions) somehow?"

But then I realized, that'd just be 2 separate package caches per ref and that's worse.

PR #9200, opened 2025-09-23
9195 hearings is removed to test 1758645743

zacharybrown204:


TomElliottFlexion (approved):

web-api/src/persistence/postgres/cases/getCasesByDocketNumbers.ts
```diff
@@ -343,6 +343,7 @@ async function getHearings(
       .selectAll('ts')
       .where('ch.docketNumber', 'in', docketNumbers)
       .where('ch.isHearing', 'is', true)
+      .where('ch.removedFromTrial', 'is', false)
```
TomElliottFlexion, line 346: Might want to consider "is not true". We _should_ be initializing these all to false...

AnthonyLoeraTaxCourt (approved):

web-api/src/persistence/postgres/cases/getCasesByDocketNumbers.ts
zacharybrown204: The column on the DB is set to not null and we default it to false, and it's defaulted to false when we put it into the DB. Since this is on the query to the DB there should be no chance of any errors comparing to null.
I'd prefer keep it as "is false" since using "is not true" on a "isRemoved" kinda column feels like a double negative, but that's not a strong feeling, I could go either way.

TomElliottFlexion: Your reasoning is solid, sounds good to me!

PR #9230, opened 2025-10-01
opex:  adjust timing of s3 sync during glue jobs

jimlerza:
The S3 document sync is now the single longest-running aspect of a "glue job" (we're not using AWS Glue for this anymore, so "glue job" is a colloquialism). It takes at least 90 minutes (and longer depending on how many docs are out of sync). It is entirely independent of the postgres data restoration and the indexing of cases and users in opensearch, though, so by starting it sooner we can reduce the total duration of the glue job.

I believe this to be a good change in practice, because this will save us about half an hour on each glue job. But with this change the code is now relying upon the S3 sync to take at least as long as the ~30m it takes to destroy and recreate the OpenSearch cluster. This is definitely the case for DAWSON, but would not be the case for significantly smaller data sets. As such, I have introduced the parameter `sync_early` and set it to `false` by default. When set to `true`, we will start the S3 sync before destroying and recreating the OpenSearch cluster, but will start it afterwards otherwise.

I've also modified the `glue-to-test` alias commensurately in the ustc-devops repo.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,6 +37,10 @@ parameters:
     default: ''
     type: string
 
+  sync_early:
+    default: false
+    type: boolean
+
   run_sync_s3_to_lower_env:
     default: false
     type: boolean
@@ -508,6 +512,8 @@ jobs:
           aws_access_key_id: $AWS_ACCESS_KEY_ID
           aws_secret_access_key: $AWS_SECRET_ACCESS_KEY
     resource_class: medium+
+    environment:
+      SYNC_EARLY: << pipeline.parameters.sync_early >>
     steps:
       - git-shallow-clone/checkout
       - npm-install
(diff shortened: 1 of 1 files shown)


jimlerza:
An over-engineered compromise could be to add a `sync_early` parameter, default it to `false`, but change our `glue-to-test` alias to set that parameter to `true`. Then if that param is `true`, we kick off the sync before doing the destroy/recreate of the opensearch cluster and otherwise we kick it off after.

TomElliottFlexion (approved):
You are DAWSON, no more, no less

jimlerza:
> An over-engineered compromise could be to add a `sync_early` parameter, default it to `false`, but change our `glue-to-test` alias to set that parameter to `true`. Then if that param is `true`, we kick off the sync before doing the destroy/recreate of the opensearch cluster and otherwise we kick it off after.

This wound up being necessary once I discovered that the S3 sync needs to have completed before we can start indexing docket entries.

paulo-filho-ctr (approved):

Issue #9060, opened 2025-08-27
Add Sealed Lock Icon to Document View


cholly75:
As a Court user, so that I know whether a document I'm reading is sealed or not, I need the sealed lock icon to display in the Document View index for sealed documents.

Currently the index displayed in the Document View mode index (the left nav displaying the docket entries) does not display whether a particular document is sealed. We should change that.

## Acceptance Criteria
* On the Document view index, if a document is sealed then the red lock icon is displayed for that index entry.

## Test Cases
The Test Case is ready: 

[TC_9060.xlsx](https://github.com/user-attachments/files/22646513/TC_9060.xlsx)

## Definition of Done (Updated 01-16-24)
**Product Owner**
 - [ ] Acceptance criteria have been met and validated on the Court's test environment
 - [ ] Add scenario to testing document, if applicable (https://docs.google.com/spreadsheets/d/1FUHKC_YrT-PosaWD5gRVmsDzI1HS_U-8CyMIb-qX9EA/edit?usp=sharing)

**Engineering**
 - [ ] Automated test scripts have been written, including visual tests for newly added PDFs.
 - [ ] Field level and page level validation errors (front-end and server-side) integrated and functioning.
 - [ ] New screens have been added to cypress accessibility axe
 - [ ] All new functionality verified to work with keyboard and macOS voiceover https://www.apple.com/voiceover/info/guide/_1124.html.
 - [ ] Swagger docs have been updated if API endpoints have been added or updated.
 - [ ] UI should be touch optimized and responsive for external users.
 - [ ] Interactors should validate entities before calling persistence methods.
 - [ ] Features have been optimized where possible to reduce response times. For example, reducing api response times, parallelizing client network calls, optimizing database reads, etc.
 - [ ] Types have been added to all added and updated functions.
 - [ ] Code refactored for clarity and to remove any known technical debt.
 - [ ] Acceptance criteria for the story has been met.
 - [ ] If there are special instructions in order to deploy into the next environment, add them as a comment in the story.
 - [ ] Code that resides in the shared folder that only runs on the API or browser has been moved to either /web-client or /web-api.

Evelyn-Durbin:
Design Guidance
https://www.figma.com/design/H4TO88zDgRWStFi0jYzynw/9060-Sealed-Lock-Icon-in-Document-View?node-id=8-11&t=Z0q2Od9LuRYhF3zo-1

[...]

PR #9241, opened 2025-10-06
9060 add sealed lock icon to test 1759766921

paulo-filho-ctr:
https://github.com/ustaxcourt/ef-cms/issues/9060

State bug: star icon was being removed after sealing/unsealing a document. Fixed.


ChristieLaferriereTaxCourt:
Had an in-depth discussion with @AnthonyLoeraTaxCourt and @paulo-filho-ctr. They explained that this was a combo ticket where lock icon was updated and also stumbled upon a bug that was not updating a state of document correctly. Both are addressed in this ticket. Will approve

Editing to add although docketEntryEntity validtion was rmeoved, it was replaced with docketEntry instantiation.

ChristieLaferriereTaxCourt (approved):

web-client/src/presenter/computeds/formattedDocketEntries.ts
```diff
@@ -39,6 +39,7 @@ export const setupIconsToDisplay = ({ formattedResult, isExternalUser }) => {
     });
   } else if (formattedResult.qcNeeded) {
     iconsToDisplay.push({
+      className: 'fa-icon-red',
       icon: ['fa', 'star'],
       title: 'Is untouched',
```
AbelTaxCourt, line 44: Consider whether 'Is untouched' should also come from a constants file, maybe it doesn't need to be. Is there another name that "Is untouched" can be converted to? Maybe something more descriptive.

AbelTaxCourt (approved):

web-client/src/presenter/computeds/formattedDocketEntries.ts
AnthonyLoeraTaxCourt: This is converting the color to red. Since we use utility classes.

Issue #8546, opened 2023-03-27
Docket Clerk: Connect Order to Motion - One to One


cholly75:
As a party to a case, so that I can quickly tell the disposition of a particular motion, I need motions to indicate their disposition and connection to the dispositive Order.

Currently when an order granting or denying a motion is docketed, there is no indication on the motion docket entry or in the motion data of how it has been disposed. This means that to find out whether a particular motion has been granted or denied the docket record and orders must be scanned and parsed individually for the order doing so. This makes it difficult to see at a glance the disposition of the motion for parties to the case and Court personnel.

Additionally, this makes it difficult to report out on operational Court data regarding types and behaviors w/r/t motions. For example, the Court is interested in monitoring how many Motions for Summary Judgment are granted vs. denied, but because there's no explicit connection between the motion itself and the order disposing of it, these statistics are difficult to collect with any degree of accuracy.

This story will address an order disposing of a single motion.

We would like to keep this behind a feature flag pending the full rollout of use cases. A future story will extend this functionality to orders disposing of multiple motions.

## Acceptance Criteria
* When adding an order to the docket record, if an order is disposing of a motion, the clerk can indicate which motion on the docket record the order is disposing of
* When adding an order disposing of a motion to the docket record, the clerk must indicate whether the motion is Granted, Granted in Part, or Denied
* The motion disposition is indicated on the docket record entry for the **motion** by appending " - [DISPOSITION] #[ENTRY]" to the current description in the "Filings and Proceedings" column of the appropriate motion
  - [DISPOSITION] is either "GRANTED", "GRANTED IN PART" or "DENIED" based on the selection by the docket clerk
  - [ENTRY] is a hyperlink to the **docket entry for the order** disposing of the motion (this means that for Court users and parties to the case with visibility into the Motion docket entry there will be two different hyperlinks present on the entry - one for the original motion and one linking to the order disposing of the motion. Both links should obey all visibility rules.)
  - This does **not** trigger a Notice of Docket Change entry on the record
* If the order disposing of the motion is sealed, the [ENTRY] hyperlink displays the red lock icon as well
* If the docket entry for an order is edited, the motion disposition association can be edited as well
* If the order is in a consolidated group, all appropriate motion records in each case in the group are edited in the manner above upon multi-docketing of the order

## Notes
* We would like the motion disposition to be stored as data in some way on the motion itself to enable reporting

## Story Definition of Ready (updated on 12/23/22)
[...]

PR #9242, opened 2025-10-06
8546 connect order to motion revamped

JayFlexy:
See #8546


shared/src/business/utilities/getFormattedCaseDetail.ts
```diff
@@ -50,7 +50,7 @@ export const computeIsNotServedDocument = ({ formattedEntry }) => {
       !DocketEntry.isMinuteEntry(formattedEntry))
   );
 };
-
+// TODO (#8546): update disposed motions here
```
codyseibert: left over comments?

web-api/src/business/useCases/courtIssuedDocument/fileAndServeCourtIssuedDocumentInteractor.ts
```diff
@@ -69,6 +73,11 @@ export const fileAndServeCourtIssuedDocument = async (
     docketNumber: subjectCaseDocketNumber,
   });
 
+  console.log(`!!!!!!!!!!!!!!!!!!!!!!`);
+  console.log('subjectCase:', subjectCase);
```
codyseibert: left over logs

shared/src/business/entities/DocketEntry.ts
```diff
@@ -186,6 +187,10 @@ export class DocketEntry extends JoiValidationEntity {
   public signedJudgeName?: string;
   public strickenBy?: string;
   public strickenByUserId?: string;
+  // public orderDocketEntryId?: string;
+  // public motionDisposition?: string; // 'GRANTED' | 'DENTIED' | 'PARTIAL';
+  public affectedDocketEntries?: DocketEntryRelation[];
```
codyseibert, line 190: I'm not sure I understand the term "affected".  what is the intent of these new fields?  affected by what?

web-client/src/views/DocketRecord/FilingsAndProceedings.tsx
```diff
@@ -143,6 +156,39 @@ export const FilingsAndProceedings = connect<
               )}
               {entry.descriptionDisplay}
             </Button>
+            {_.concat(
+              entry.affectedByDocketEntries ?? [],
+              entry.affectedDocketEntries ?? [],
+            ).map(entry => {
+              return (
+                <>
+                  <br></br>
```
codyseibert: I'm not sure the indent of the <br> but I'd recommend using gap or margin or padding if the goal is to space things out instead of just br tags

web-api/src/persistence/postgres/cases/getCasesByDocketNumbers.ts
```diff
@@ -309,9 +378,77 @@ export async function getDocketEntriesOnCases(
 ): Promise<DocketEntryKysely[]> {
   return getDbReader(reader =>
     reader
-      .selectFrom('dwDocketEntry')
+      .with('affectedDocketEntries', db =>
+        db
+          .selectFrom('dwDocketEntryRelatedDocketEntry')
```
codyseibert: Just curious, but has an EXPLAIN statement been executd on this query to check it's performance and verify no indexes are needed.  The where('served', 'is', true) seems to be used in this CTE but won't that loop over a ton of docket entries? I probably don't understand the intent.  I would think the CTEs should at least do an IN clause by docket number so that it only has to fetch the related cases instead of needing to grab all the rows in the docket entry table, which is I think 10 million last time I remembered.

.with('affectedDocketEntries', db =>
        db
          .selectFrom('dwDocketEntryRelatedDocketEntry')
          .where('served', 'is', true)
          .where('docketNumber', 'in', docketNumbers) // ✅ Add filter here!
          
          
additionally, the 
join
          .onRef(
            eb => eb.cast('affectedByDocketEntries.docketEntryId', 'varchar'),
claude is hinting that this cast might prevent the db engine from using the index correctly which might also cause some performance issues.

shared/src/business/entities/DocketEntry.ts
JayFlexy: The new field `affectedDocketEntries` refer to docket entries that are being affected by a different docket entry. So in this case we create an order that affects a motion. Orders and motions are but one type of inter-relation amongst docket entries but we were attempting to make the code generic so that future relations between docket entries are not restricted to order/motion verbiage.

web-client/src/views/DocketRecord/FilingsAndProceedings.tsx
JayFlexy: Will review this to see if there are better options than `<br>` tag. I do vaguely remember this being a finicky part of the UI where margin/padding tinkering never really gave what we were intending.

codyseibert (approved):

web-api/src/business/useCases/docketEntry/updateDocketEntryMetaInteractor.ts
```diff
@@ -209,3 +222,36 @@ export const updateDocketEntryMetaInteractor = withLocking(
...
+    const { added, updated, removed } = diff(
+      originalDocketEntry.affectedDocketEntries,
+      docketEntryMeta.affectedDocketEntries,
+      'docketEntryId',
+    );
+
+    if (added.length > 0 || updated.length > 0) {
+      await upsertDocketEntryRelatedEntries({
+        orderDocketEntry: docketEntryMeta,
+        motionDocketEntries: _.concat(added ?? [], updated ?? []).map(m => ({
```
jimlerza: Can you import just `concat` instead of importing all of lodash in this interactor?

web-api/src/persistence/postgres/docketEntries/commonQueries.ts
jimlerza, line 1: This is a great idea! ❤️

web-client/src/views/CourtIssuedDocketEntry/CourtIssuedDocketEntry.tsx
```diff
@@ -18,6 +18,9 @@ import { reactSelectValue } from '@web-client/ustc-ui/Utils/documentTypeSelectHe
 import { sequences } from '@web-client/presenter/app.cerebral';
 import { state } from '@web-client/presenter/app.cerebral';
 import React from 'react';
+import _ from 'lodash';
```
jimlerza: Here's another instance where we can probably import just the methods we need.

web-client/src/views/DocketRecord/FilingsAndProceedings.tsx
```diff
@@ -6,12 +6,20 @@ import { sequences } from '@web-client/presenter/app.cerebral';
 import { state } from '@web-client/presenter/app.cerebral';
 import React from 'react';
 import classNames from 'classnames';
+import _ from 'lodash';
```
jimlerza: Editor tells me this one is unused.

web-client/src/views/ModalDialog.tsx
```diff
@@ -81,6 +81,7 @@ export const ModalDialog = ({
   };
 
   const runConfirmSequence = evt => {
+    console.log(`runConfirmSequence was triggered`);
```
jimlerza: Do we want to keep this or was this for debugging?

jimlerza (changes requested):
Looks great! I left a couple of nits related to how lodash is being imported.

jimlerza (approved):

PR #9246, opened 2025-10-07
9219 download all cases trial session to test 1759862337

alexdziarn:


web-api/src/business/useCases/trialSessions/batchDownloadTrialSessionInteractor.ts
```diff
@@ -126,6 +127,8 @@ export const batchDownloadTrialSessionInteractorHelper = async (
   let numberOfDocketRecordsGenerated = 0;
   const numberOfDocketRecordsToGenerate = batchableSessionCases.length;
 
+  applicationContext.logger.info('Generating file');
```
AbelTaxCourt, line 130: Is this API logger intended to be kept?

zacharybrown204 (approved):

AbelTaxCourt (approved):
Looks good

web-api/src/business/useCases/trialSessions/batchDownloadTrialSessionInteractor.ts
alexdziarn: There is a bigger problem happening elsewhere, and we need this log to see how far it gets in this file.

PR #9251, opened 2025-10-08
Update docs

JayFlexy:
Adds container runtime instructions as well as updates to experimental environment links


paulo-filho-ctr (approved):

alexdziarn (approved):

docs/getting-started.md
```diff
@@ -37,13 +37,13 @@ Here is a list of the **Tax Court**'s deployed Dawson instances:
...
-- [https://exp1.ustc-case-mgmt.flexion.us/](https://exp1.ustc-case-mgmt.flexion.us/) (flexion experimental1)
-- [https://exp2.ustc-case-mgmt.flexion.us/](https://exp2.ustc-case-mgmt.flexion.us/) (flexion experimental2)
-- [https://exp3.ustc-case-mgmt.flexion.us/](https://exp3.ustc-case-mgmt.flexion.us/) (flexion experimental3)
-- [https://exp4.ustc-case-mgmt.flexion.us/](https://exp4.ustc-case-mgmt.flexion.us/) (flexion experimental4)
-- [https://exp5.ustc-case-mgmt.flexion.us/](https://exp5.ustc-case-mgmt.flexion.us/) (flexion experimental5)
+- [https://exp1.ef-cms.ustaxcourt.gov/](https://exp1.ef-cms.ustaxcourt.gov/) (ustc experimental1)
+- [https://exp2.ef-cms.ustaxcourt.gov/](https://exp2.ef-cms.ustaxcourt.gov/) (ustc experimental2)
+- [https://exp3.ef-cms.ustaxcourt.gov/](https://exp3.ef-cms.ustaxcourt.gov/) (ustc experimental3)
+- [https://exp4.ef-cms.ustaxcourt.gov/](https://exp4.ef-cms.ustaxcourt.gov/) (ustc experimental4)
+- [https://exp5.ef-cms.ustaxcourt.gov/](https://exp5.ef-cms.ustaxcourt.gov/) (ustc experimental5)
```
jimlerza, line 46: There are 9 of these now.

docs/running-locally.md
jimlerza, line 1: So what's the intention here? This reads as though this is absolutely required and not optional, and the instructions that now follow this new Colima block don't make sense anymore because of this new tangent.

I thought this was an additional option for flexion laptops and/or Docker haters. If this is the case, it should come after all the required stuff and be clearly marked as optional.

jimlerza (changes requested):

docs/running-locally.md
JayFlexy: This is seen as a universal option that runs better on MacOS and Linux as well as more open licensing. Will go back and re-work so that it's more cohesive with the entire page. Also, colima is less resource intensive on MacOS which could definitely affect the developer experience if MacOS is the default on this project. Can come over and have a sync chat as well though.

codyseibert (approved):

jimlerza (approved):

PR #9254, opened 2025-10-09
9219 download all cases trial session

alexdziarn:


web-api/src/dispatchers/batch/pollAWSBatchProgress.ts
```diff
@@ -0,0 +1,114 @@
...
+  pollInterval?: number;
+  timeout?: number;
+  onProgress: (progressData: ProgressData) => Promise<void>;
+}
+
+export const pollAWSBatchProgress = async ({
+  applicationContext,
+  jobId,
+  pollInterval = 5000,
+  timeout = 600000,
```
jimlerza: Should we use constants for these values instead of magic integers?

web-api/src/dispatchers/batch/pollAWSBatchProgress.ts
```diff
@@ -0,0 +1,114 @@
...
+  onProgress,
+}: PingCloudWatchParams) => {
+  const startTime = Date.now();
```
jimlerza: Did we recently drop the "no JS `Date`s" rule and nobody told me? I think we're supposed to use luxon (via the date helper) for stuff like this.

jimlerza (changes requested):

jimlerza (approved):

web-api/src/dispatchers/batch/pollAWSBatchProgress.ts
TomElliottFlexion: As it turns out, the rule is configured to alert on `new Date` rather than just `Date`. I _think_ the idea was to avoid date comparisons over avoiding timers entirely, but a consistent approach probably makes sense: here is a PR to expand the rule: https://github.com/ustaxcourt/ef-cms/pull/9263

TomElliottFlexion (approved):

AbelTaxCourt (approved):

Issue #8576, opened 2023-11-27
Track Granted MOTR


cholly75:
As a Court, so that I know whether a particular case has been granted a motion to proceed remotely, I need to be able to indicate this on the Case Information screen.

The Court would like to start tracking cases that have been granted permission to have a remote trial session, to better assist us in solving scheduling problems and making sure that these cases do not get scheduled for an in-person trial session. Currently these cases are being tracked manually offline. 

We would like to A) display this information on the Case Information screen in the Trial card, and B) give Docket clerks a place to adjust this data.

Future stories will be setting this flag automatically upon granting of the motion and using this flag to control display of cases on eligible case lists and reports.

## Acceptance Criteria
* On the Case Information screen/Trial card, data is displayed indicating whether a case has been granted permission to proceed remotely (a simply yes/no or on/off flag will do) along with a date stamp indicating the date permission was granted.
* Users in the Docket Clerk, Case Services Supervisor, and Clerk of the Court role can edit this data (both the flag and the date).

## Notes

<img width="1434" height="502" alt="Image" src="https://github.com/user-attachments/assets/4db35fec-4760-4b94-aac5-598a81e4c9ad" />

## Test Cases
Test cases are in TestRail and are also added here:

[8576.xlsx](https://github.com/user-attachments/files/22669964/8576.xlsx)

## Story Definition of Ready (updated on 12/23/22)
The following criteria must be met in order for the user story to be picked up by the Flexion development team.
The user story must: 
- [ ] Is framed in business/user need, the value has been addressed.
- [ ] Includes acceptance criteria
- [ ] Has been refined
- [ ] Pre conditions have been satisfied. 

**Process:** 
Flexion developers and designers will test if the story meets acceptance criteria and test cases in Flexion dev and staging environments (“standard testing”). If additional acceptance criteria or testing scenarios are discovered while the story is in progress, a new story should be created, added to the backlog and prioritized by the product owner. 

## Definition of Done (Updated 5-19-22)
**Product Owner**
 - [ ]  Acceptance criteria have been met and validated on the Court's migration environment
 - [ ] Add scenario to testing document, if applicable (https://docs.google.com/spreadsheets/d/1FUHKC_YrT-PosaWD5gRVmsDzI1HS_U-8CyMIb-qX9EA/edit?usp=sharing)

**UX**
 - [ ] Business test scenarios have been refined to meet all acceptance criteria 
 - [ ] Usability has been validated
[...]

PR #9269, opened 2025-10-14
8576 track granted motr to test 1760468941

paulo-filho-ctr:
8576 https://github.com/ustaxcourt/ef-cms/issues/8576

On the Case Information screen/Trial card, data is displayed indicating whether a case has been granted permission to proceed remotely (a simply yes/no or on/off flag will do) along with a date stamp indicating the date permission was granted.

The table `dw_case` requires two new columns `remote_trial_granted` and `remote_trial_granted_date`.
So, it is necessary to run 

``````
npm run migration:postgres

### CODE REVIEW

File: cypress/local-only/tests/integration/caseDetail/caseInformation/edit-remote-status.cy.ts [hunk 1]
```diff
@@ -0,0 +1,240 @@
...
+        getCaseDetailTab('case-information').click();
+
+        cy.get('[data-testid="edit-remote-status"]').should('exist');
+        cy.get('[data-testid="edit-remote-status"]').click();
+
+        cy.get('.modal-header').should('contain', 'Edit Remote Status');
+
+        const today = DateTime.now();
+        const dateString = today.toFormat('MM/dd/yyyy');
+
```
TomElliottFlexion: This is not super important for tests, but the emphasis is on using the DateHandler over using luxon specifically as opposed to JS date objects; you _could_ use the `formatNow` function

web-api/src/business/useCases/editRemoteStatusInteractor.ts
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * used for setting a date to indicate when a motr was granted
+ * @param {object} applicationContext the application context
+ * @param {object} providers the providers object
+ * @param {string} providers.timestamp timestamp when the motr was granted
+ * @param {string} providers.docketNumber the docket number
+ * @returns {object} the case data
+ */
+export const editRemoteStatus = async (
```
TomElliottFlexion: Tiny nit: we shouldn't export this, instead only export the interactor function at the bottom to force future callers to use it and not bypass the locking

web-api/src/business/useCases/editRemoteStatusInteractor.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  const updatedCase = await updateCaseAndAssociations({
+    authorizedUser,
+    caseToUpdate: caseEntity,
+  });
+
+  return new Case(updatedCase, { authorizedUser }).validate().toRawObject();
+};
+
+export const editRemoteStatusInteractor = withLocking(
```
TomElliottFlexion: We don't seem to be using this interactor, instead updating the case on the front end and then using `updateCaseDetailsInteractor`. (If we did use this interactor, it wouldn't work since it does not currently update the `remoteTrialGranted` or `remoteTrialGrantedDate` fields on the case)

web-api/elasticsearch/efcms-case-mappings.ts
```diff
@@ -110,6 +110,12 @@ export const efcmsCaseMappings: Record<string, Object> = {
     'receivedAt.S': {
       type: 'date',
     },
+    'remoteTrialGranted.BOOL': {
+      type: 'boolean',
+    },
+    'remoteTrialGrantedDate.S': {
+      type: 'date',
+    },
```
TomElliottFlexion, line 118: Do these need to be indexed in OpenSearch?

web-client/src/presenter/actions/CaseDetail/editRemoteStatusAction.ts
```diff
@@ -0,0 +1,41 @@
...
+  const hasDate = Boolean(
+    remoteTrialGrantedDate && remoteTrialGrantedDate.trim() !== '',
+  );
+
+  const currentCase = get(state.caseDetail);
+  const caseDetails = {
+    ...currentCase,
+    remoteTrialGranted: hasDate,
+    remoteTrialGrantedDate: hasDate ? remoteTrialGrantedDate : null,
+  };
```
TomElliottFlexion: Take it or leave it, but I'd recommend adding an entity function on Case to set these values, simply for repeatability and clarity

web-client/src/presenter/actions/CaseDetail/editRemoteStatusAction.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  const caseDetail = await applicationContext
+    .getUseCases()
+    .updateCaseDetailsInteractor(applicationContext, {
```
TomElliottFlexion, line 25: Do we want to be calling `editRemoteStatusInteractor` here instead?

web-client/src/presenter/sequences/submitEditRemoteTrialModalSequence.ts
```diff
@@ -0,0 +1,24 @@
...
+import { clearModalStateAction } from '../actions/clearModalStateAction';
+import { setAlertSuccessAction } from '../actions/setAlertSuccessAction';
+import { setCaseAction } from '../actions/setCaseAction';
+import { setValidationErrorsAction } from '../actions/setValidationErrorsAction';
+import { showProgressSequenceDecorator } from '../utilities/showProgressSequenceDecorator';
+import { startShowValidationAction } from '../actions/startShowValidationAction';
+import { updateRemoteTrialPermissionAction } from '../actions/CaseDetail/updateRemoteTrialPermissionAction';
+import { validateRemoteTrialPermissionAction } from '../actions/CaseDetail/validateRemoteTrialPermissionAction';
+
+export const submitEditRemoteTrialModalSequence = [
```
TomElliottFlexion: This sequence is not used

web-client/src/presenter/actions/CaseDetail/updateRemoteTrialPermissionAction.ts
```diff
@@ -0,0 +1,58 @@
+import { state } from '@web-client/presenter/app.cerebral';
+
+/**
+ * Updates the remote trial permission for a case
+ * @param {object} providers the providers object
+ * @param {object} providers.applicationContext the application context
+ * @param {Function} providers.get the cerebral get function
+ * @returns {object} the alertSuccess and updated caseDetail object
+ */
+export const updateRemoteTrialPermissionAction = async ({
```
TomElliottFlexion: This action is not used

web-api/src/business/useCases/editRemoteStatusInteractor.test.ts
```diff
@@ -0,0 +1,181 @@
...
+import { updateCaseAndAssociations as updateCaseAndAssociationsMock } from '@web-api/business/useCaseHelper/caseAssociation/updateCaseAndAssociations';
+
+describe('editRemoteStatusInteractor', () => {
+  const getCaseByDocketNumber = jest.mocked(getCaseByDocketNumberMock);
+  const updateCaseAndAssociations = jest.mocked(updateCaseAndAssociationsMock);
+  const tryGetLocks = jest.mocked(tryGetLocksMock);
+
+  const mockRemoteTrialGrantedDate = '2023-10-14T00:00:00.000Z';
+
+  beforeEach(() => {
```
jdr81394: Consider adding a jest.clearAllMocks() or the jest.resetAllMocks() method call so you can avoid any issues related to cross-test contamination in the future.

web-api/src/business/useCases/editRemoteStatusInteractor.ts
paulo-filho-ctr: Excellent point. We will update to force the use of locking.

web-client/src/presenter/actions/CaseDetail/editRemoteStatusAction.ts
paulo-filho-ctr: I agree on repeatability but, I think it is quite clear it is setting the remote trial date. Do we repeat the setting remote trial granted anywhere? If so, yes, I agree we should move to Case, otherwise, I think it is simpler to keep it here. Thought?

web-api/elasticsearch/efcms-case-mappings.ts
paulo-filho-ctr: I'm not sure. @jimlerza Do we need to index these?

cypress/local-only/tests/integration/caseDetail/caseInformation/edit-remote-status.cy.ts
```diff
@@ -0,0 +1,235 @@
...
+describe('Edit Remote Status', () => {
+  describe('As a docket clerk', () => {
+    beforeEach(() => {
+      loginAsDocketClerk();
+    });
+
+    it('should allow docket clerk to edit remote status and add a granted date', () => {
+      createAndServePaperPetition().then(({ docketNumber }) => {
+        cy.visit(`/case-detail/${docketNumber}`);
```
alexdziarn: this should be replaced with the goToCase(docketNumber) function.

alexdziarn: there are multiple instances cy.visit is being used throughout this pr, so just hcnage as you see.

cypress/local-only/tests/integration/caseDetail/caseInformation/edit-remote-status.cy.ts
```diff
@@ -0,0 +1,235 @@
...
+    });
+
+    it('should not allow future dates for remote trial granted date', () => {
+      createAndServePaperPetition().then(({ docketNumber }) => {
+        cy.visit(`/case-detail/${docketNumber}`);
+        getCaseDetailTab('case-information').click();
+
+        cy.get('[data-testid="edit-remote-status"]').click();
+
+        const futureDate = DateTime.now().plus({ days: 30 });
```
alexdziarn: we should be using the DateHandler library instead of DateTime due to luxon leaks

cypress/local-only/tests/integration/caseDetail/caseInformation/edit-remote-status.cy.ts
```diff
@@ -0,0 +1,235 @@
...
+
+        cy.get('[data-testid="edit-remote-status"]').click();
+        const initialDateString = formatNow(FORMATS.MMDDYYYY);
+        cy.get('#remote-trial-granted-date-picker').type(initialDateString);
+        cy.get('.modal-button-confirm').click();
+
+        cy.get('.usa-alert--success').should('exist');
+
+        cy.get('[data-testid="edit-remote-status"]').click();
+        const newDate = DateTime.now().minus({ days: 7 });
```
alexdziarn: replace using the DateHandler library

shared/src/business/entities/cases/Case.ts
```diff
@@ -1921,6 +1937,22 @@ export class Case extends JoiValidationEntity {
...
+  setRemoteTrialGrantedDate(remoteTrialGrantedDate?: string | null) {
+    const hasDate = Boolean(
+      remoteTrialGrantedDate &&
+        typeof remoteTrialGrantedDate === 'string' &&
+        remoteTrialGrantedDate.trim() !== '',
+    );
+    this.remoteTrialGranted = hasDate;
+    this.remoteTrialGrantedDate = hasDate ? remoteTrialGrantedDate : null;
+    return this;
+  }
```
zacharybrown204: Since we have a known return type, maybe have an explicit declaration instead of letting typescript infer it? IMO it's fine either way but I like explicit types in the method signature.

```suggestion
  setRemoteTrialGrantedDate(remoteTrialGrantedDate?: string | null): Case {
    const hasDate = Boolean(
      remoteTrialGrantedDate &&
        typeof remoteTrialGrantedDate === 'string' &&
        remoteTrialGrantedDate.trim() !== '',
    );
    this.remoteTrialGranted = hasDate;
    this.remoteTrialGrantedDate = hasDate ? remoteTrialGrantedDate : null;
    return this;
  }
```

web-client/src/presenter/actions/CaseDetail/editRemoteStatusAction.ts
```diff
@@ -0,0 +1,41 @@
...
+ * @returns {object} the alertSuccess and updated caseDetail object
+ */
+export const editRemoteStatusAction = async ({
+  applicationContext,
+  get,
+}: ActionProps) => {
+  const docketNumber = get(state.caseDetail.docketNumber);
+  const modalState = get(state.modal);
+  const { remoteTrialGrantedDate } = modalState;
+  const rawCase = get(state.caseDetail);
```
zacharybrown204, line 17: Could cut out a const by directly deconstructing the modal state thats returned from this get call. Not important, but it might help readability on this variable block.

```suggestion
  const docketNumber = get(state.caseDetail.docketNumber);
  const { remoteTrialGrantedDate } = get(state.modal);
  const rawCase = get(state.caseDetail);
```

web-client/src/presenter/actions/CaseDetail/editRemoteStatusAction.ts
```diff
@@ -0,0 +1,41 @@
...
+  const user = get(state.user);
+  const currentCase = new Case(rawCase, { 
+    authorizedUser: user,
+    filtered: false,
+    isNewCase: false 
+  });
+  
+  currentCase.setRemoteTrialGrantedDate(remoteTrialGrantedDate);
```
zacharybrown204, line 21: The filtered and isNewCase parameters on the case both default to false, so we could omit them and make this a one line call. 

```suggestion
  const user = get(state.user);
  const currentCase = new Case(rawCase, { authorizedUser: user });

  currentCase.setRemoteTrialGrantedDate(remoteTrialGrantedDate);
```

zacharybrown204 (commented):
Saw a few things that might help readability, though none of these are important. Implement or ignore at your discretion.
Though I do agree with Alex's comments above and that we probably should address those.

cypress/local-only/tests/integration/caseDetail/caseInformation/edit-remote-status.cy.ts
paulo-filho-ctr: For sure, we will replace all usage of `cy.visit(``/case-detail/${docketNumber}``)` with `goToCase(docketNumber)`.

cypress/local-only/tests/integration/caseDetail/caseInformation/edit-remote-status.cy.ts
paulo-filho-ctr: Good catch.

cypress/local-only/tests/integration/caseDetail/caseInformation/edit-remote-status.cy.ts
paulo-filho-ctr: Nice, we will use the DateHandler library.

shared/src/business/entities/cases/Case.ts
paulo-filho-ctr: Thank you @zacharybrown204 I completely agree. I all for readability and and explicit declarations. We will return the type for the function.

web-client/src/presenter/actions/CaseDetail/editRemoteStatusAction.ts
paulo-filho-ctr: For sure, I think it reads much better now. Thank you!

web-client/src/presenter/actions/CaseDetail/editRemoteStatusAction.ts
paulo-filho-ctr: I like the suggestion. It improves readability.

alexdziarn (approved):

Issue #8922, opened 2025-02-18
Generate login credentials for some real users in test


jimlerza:
As a product specialist, so I can create accurate test cases involving real data, I need to be able to log into `test` as certain prolific or noteworthy real-life (`prod`) users. We are already doing this for Judges, perhaps we can do this similarly for other users.

## Acceptance Criteria

- Able to log in as specific users in test with the default password.

## Pain Avoided/Frustration Saved

- Allows us to test new features as real users with real data.

## Notes
  
- Is it possible to somehow programatically determine "prolific" users so we don't have to hard-code them? Would we want to?
- The script that does this for judges is `setup-glued-judges.ts`

Ruleset for how these real test users were created
- All Active Court users
- Private Practitioners with 50 cases or more
- IRS Practitioners with 50 cases or more
- All DOJ practitioners
- Petitioners with 5 cases or more
- There is one inactive practitioner that has an account created 

To find these users, go to Cognito and search by email address.  To find users, start typing [role].[firstnamelastname]@example.com. 

examples: 
privatepractitioner.johndoe@example.com
irspractitioner.johndoe@example.com
admissionsclerk.johndoe@example.com

jimlerza:
The existing script retrieves all Judge users from the db and makes new cognito accounts for all of them (as necessary).

What if, instead of a list of specific users, we do this for all users with an "internal" role?

Mwindo:
Currently, we can "fake" a given user in test (e.g., "make docketclerk1 be so-and-so") by manipulating the Cognito record to point to a different ID. The goal of _this_ ticket is to 1) automate this better and 2) get more realistic user data into our test environment for non-judge users in general. Ideally, we want to do this without having to hard-code certain users in Secrets Manager or something (since users might change). Some open questions to investigate:

- Do we need both internal _and_ external roles (e.g., private practitioners)?
- Do we want to do _all_ users, either by syncing Cognito as part of the glue job (how difficult is this?) or by using some custom logic? Do we want to do some subset of users? Or do we want to do something more role-based, with prototypical or extreme users for each role?
- Depending on the above choices, how can we map a user to a unique but predictable email? Currently, we map judges to emails, but users in general could have naming conflicts (e.g., "John Smith" x 2--how do we assign emails?) 
[...]

PR #9277, opened 2025-10-17
8922 generate login credentials

alexdziarn:
#8922


scripts/user/setup-glued-users.ts
```diff
@@ -0,0 +1,362 @@
...
+      where: db => db.where('userId', '=', bulkImportedUserId),
+    });
+  } catch (err) {
+    console.error(`ERROR deleting duplicate ${name}:`, err);
+  }
+  console.log(`Deleted duplicate ${name}`);
+};
+
+const getPractitionerUsers = async (): Promise<any[]> => {
+  // 1) Get userIds that have count > 500 in dwUserOnCase
```
paulo-filho-ctr: I think maybe we can make without the comments?

scripts/user/setup-glued-users.ts
```diff
@@ -0,0 +1,362 @@
...
+      .where('serviceIndicator', '=', 'Electronic')
+      .groupBy('userId')
+      .having(sql<boolean>`count(*) > ${500}`)
+      .execute(),
+  );
+
+  const userIds = userIdRows.map(r => r.userId).filter(Boolean) as string[];
+  if (!userIds.length) return [];
+
+  // 2) Fetch users from dwUser where userId in (list)
```
paulo-filho-ctr: I think comments are not necessary here? I think the intent is self-explanatory.

paulo-filho-ctr (commented):
Looks awesome, my general recommendation is to remove the lines with comments.

codyseibert (approved):

.circleci/config.yml
```diff
@@ -196,7 +196,8 @@ jobs:
                 --query 'DomainStatus.Endpoint' \
                 --output text)
               export ELASTICSEARCH_ENDPOINT="$ELASTICSEARCH_ENDPOINT"
```
jimlerza: I believe we can remove these lines since the `setup-glued-users` script doesn't use OpenSearch.

jimlerza (approved):

Issue #9037, opened 2025-08-04
Indicate "Aged" Cases On Eligible Case Lists


cholly75:
As a calendaring clerk, so that I can appropriately prioritize aged cases when calendaring, I need to be able to quickly recognize aged cases on the Eligible case lists.

Every month the Court publishes a report of DAWSON data indicating cases which have had no activity on the docket record for the past 365 days. Many of these cases are simply waiting in line to be scheduled for trial at the appropriate location.

We would like to increase visibility into these specific cases when viewing the Eligible case lists in DAWSON.

These "aged" cases are defined as:
* There has been no docket filing on the case in the last 365 days
* The case Status is NOT "Closed", "Closed-Dismissed", or "On Appeal"

## Acceptance Criteria
* When viewing the Eligible Case list on the Location screen for a city, indicate the Eligible cases that have had no docket activity for the last 365 days with an icon or other visual indicator
* When viewing the Eligible Case list on the Trial Session screen for a New Trial Session, indicate the Eligible cases that have had no docket activity for the last 365 days with an icon or other visual indicator
* When viewing the Eligible Case list on the Trial Session screen for a New Trial Session, Eligible cases that have had no docket activity for the last 365 days are displayed at the top of the list
  - Note that this is a change to the current order of cases displayed; this criteria takes precedence over the existing priorities, which remain the same
* When viewing the Eligible Case list on the Location screen, Eligible cases that have had no docket activity for the last 365 days are displayed at the top of the list
  - Note that this is a change to the current order of cases displayed; this criteria takes precedence over the existing priorities, which remain the same

## Notes

- Existing report is [stale-cases.ts](https://github.com/ustaxcourt/ef-cms/blob/staging/scripts/reports/stale-cases.ts), which was written before cases and docket entries were available in postgres.

## Test Cases
Test cases are listed in the spreadsheet below as well as in Test Rail.

[9037.xlsx](https://github.com/user-attachments/files/22408816/9037.xlsx)

## Story Definition of Ready (updated on 12/23/22)
The following criteria must be met in order for the user story to be picked up by the Flexion development team.
The user story must: 
- [ ] Is framed in business/user need, the value has been addressed.
- [ ] Includes acceptance criteria
- [ ] Has been refined
- [ ] Pre conditions have been satisfied. 

**Process:** 
[...]

PR #9286, opened 2025-10-20
9037 indicate aged cases to test 1760988667

devinganey613:
#9037


shared/src/business/useCases/getEligibleCasesForCityInteractor.ts
```diff
@@ -9,6 +9,49 @@ import {
...
+import {
+  dateStringsCompared,
+  createISODateString,
+  calculateISODate,
+} from '../utilities/DateHandler';
+
+export const setIsAgedCaseForEligibleCases = (
+  eligibleCases: Omit<RawCase, 'consolidatedCases'>[],
+): Omit<RawCase, 'consolidatedCases'>[] => {
+  console.log(eligibleCases)
```
paulo-filho-ctr: Do we want to keep `console.log(eligibleCases)`?

shared/src/business/useCases/getEligibleCasesForCityInteractor.ts
```diff
@@ -9,6 +9,49 @@ import {
...
+  return eligibleCases.map(eligibleCase => {
+    console.log(eligibleCase.docketEntries)
```
paulo-filho-ctr: Do we want to keep `console.log(eligibleCase.docketEntries)`?

web-client/src/presenter/computeds/formattedEligibleCasesHelper.ts
```diff
@@ -43,41 +43,91 @@ const getSortableDocketNumber = docketNumber => {
...
-  if (!Array.isArray(cases)) {
-    return '';
-  }
-  const leadCase = cases.find(
-    aCase => aCase.docketNumber === theCase.leadDocketNumber,
-  );
-
-  const isLeadInEligible = !!theCase.leadDocketNumber && !!leadCase;
+export const sortEligbleCases = (cases, formattedCases) =>
+  // Sort by default group (lowest priority)
```
paulo-filho-ctr: While nice, I think it would be ok to remove any comments from the code.

web-client/src/presenter/computeds/formattedEligibleCasesHelper.ts
```diff
@@ -43,41 +43,91 @@ const getSortableDocketNumber = docketNumber => {
...
+    .sort((a, b) => {
+      if (a[groupKeySymbol] === 'default' && b[groupKeySymbol] === 'default') {
+        const aSortable = getSortableDocketNumber(a.docketNumber);
+        const bSortable = getSortableDocketNumber(b.docketNumber);
+        return aSortable.localeCompare(bSortable);
+      }
+      return 0;
+    })
+    // Sort by suffixHighPriority group (medium priority)
+    .sort((a, b) => {
```
paulo-filho-ctr: Since we have this complex `if` statements, maybe naming the `.sort` callback function can serve as documentation and intent of the sorting and to remove the comments.

web-client/src/presenter/computeds/formattedEligibleCasesHelper.ts
```diff
@@ -43,41 +43,91 @@ const getSortableDocketNumber = docketNumber => {
...
+      ) {
+        const aSortable = getSortableDocketNumber(a.docketNumber);
+        const bSortable = getSortableDocketNumber(b.docketNumber);
+        return aSortable.localeCompare(bSortable);
+      }
+      if (a[groupKeySymbol] === 'suffixHighPriority') return -1;
+      if (b[groupKeySymbol] === 'suffixHighPriority') return 1;
+      return 0;
+    })
+    // Sort by manuallyAdded group (highest priority)
```
paulo-filho-ctr: We can take or leave it, but I think we can move the `.sort` to a named function.

web-api/src/business/useCases/trialSessions/getEligibleCasesForTrialSessionInteractor.test.ts
```diff
@@ -113,6 +113,7 @@ describe('getEligibleCasesForTrialSessionInteractor', () => {
       error = e;
     }
 
+    console.log(result)
```
paulo-filho-ctr: Maybe we can remove `console.log(result)`?

web-api/src/business/useCases/trialSessions/getEligibleCasesForTrialSessionInteractor.ts
```diff
@@ -56,8 +57,11 @@ export const getEligibleCasesForTrialSessionInteractor = async (
       trialCity: trialSessionEntity.trialLocation!,
     });
 
+  const eligibleCasesWithIsAgedCase =
+    setIsAgedCaseForEligibleCases(eligibleCases);
+
   const eligibleCasesFiltered = calendaredCases
-    .concat(eligibleCases)
+    .concat(eligibleCasesWithIsAgedCase)
```
paulo-filho-ctr, line 64: I like the use of named function as callback. Kudos!

paulo-filho-ctr (commented):
It looks great! I think most of my suggestions reside in `web-client/src/presenter/computeds/formattedEligibleCasesHelper.ts`

shared/src/business/useCases/getEligibleCasesForCityInteractor.ts
```diff
@@ -9,6 +9,47 @@ import {
...
+import { CASE_STATUS_TYPES } from '../entities/EntityConstants';
+import {
+  dateStringsCompared,
+  createISODateString,
+  calculateISODate,
+} from '../utilities/DateHandler';
+
+export const setIsAgedCaseForEligibleCases = (
+  eligibleCases: Omit<RawCase, 'consolidatedCases'>[],
+): Omit<RawCase, 'consolidatedCases'>[] => {
```
TomElliottFlexion: I believe this logic could use a unit test

web-client/src/styles/custom.scss
```diff
@@ -1829,6 +1829,10 @@ button.change-scanner-button {
   color: color($theme-color-secondary-dark);
 }
 
+.aged-case-flag {
+  color: rgb(216 57 51 / 100%);
```
TomElliottFlexion: Are/can these colors (be) included as sass `$theme-color` variables?

shared/src/business/useCases/getEligibleCasesForCityInteractor.ts
devinganey613: Added tests

shared/src/business/utilities/trialSession/getFormattedTrialSessionDetails.ts
```diff
@@ -75,7 +78,7 @@ export const formatCaseForTrialSession = ({
   eligibleCases = [],
   setFilingPartiesCode = false,
 }: {
-  applicationContext: IApplicationContext;
+  applicationContext: ClientApplicationContext;
```
jimlerza: If this method is no longer shared and is intended to only be used by the client, should this file be moved into the `web-client` package?

web-client/src/presenter/computeds/formattedEligibleCasesHelper.ts
```diff
@@ -43,42 +44,73 @@ const getSortableDocketNumber = docketNumber => {
...
-    const aSortString = getFullSortString(a, groups[a[groupKeySymbol]]);
-    const bSortString = getFullSortString(b, groups[b[groupKeySymbol]]);
+    const aSortString = getFullSortString(a);
+    const bSortString = getFullSortString(b);
     return aSortString.localeCompare(bSortString);
   };
 };
 
+export const getFormattedEligibleCases = formattedCases => {
+  const memberCasesObj = {} as Record<string, any[]>;
```
jimlerza: I believe we know the type here. Should be `RawEligibleCase` or `Omit<RawEligibleCase, 'whatever' | 'was' | 'omitted'>`

jimlerza (approved):

shared/src/business/utilities/trialSession/getFormattedTrialSessionDetails.ts
devinganey613: Switched it back, this is being used by the api as well

web-client/src/presenter/computeds/formattedEligibleCasesHelper.ts
```diff
@@ -43,42 +44,73 @@ const getSortableDocketNumber = docketNumber => {
...
+      caseItem.memberCases = memberCasesObj[caseItem.leadDocketNumber];
+      caseItem.memberCases.sort((a, b) => {
+        const aSortString = getFullSortString(a);
+        const bSortString = getFullSortString(b);
+        return aSortString.localeCompare(bSortString);
+      });
+    }
+  });
+
+  // loop through formatted cases
```
akuny: Remove this comment.

shared/src/business/useCases/getEligibleCasesForCityInteractor.ts
```diff
@@ -9,6 +9,47 @@ import {
...
   RawEligibleCase,
 } from '@shared/business/entities/cases/EligibleCase';
+import { CASE_STATUS_TYPES } from '../entities/EntityConstants';
+import {
+  dateStringsCompared,
+  createISODateString,
+  calculateISODate,
+} from '../utilities/DateHandler';
+
+export const setIsAgedCaseForEligibleCases = (
```
akuny: Since this function is used both in this interactor and in `getEligibleCasesForTrialSessionInteractor`, I recommend moving it to the use case helpers layer and putting it directly under unit test.

cypress/local-only/tests/integration/trialSession/trial-session-aged-eligible-cases.cy.ts
```diff
@@ -0,0 +1,417 @@
...
+  loginAsCaseServicesSupervisor,
+  loginAsDocketClerk,
+  loginAsPetitionsClerk1,
+} from 'cypress/helpers/authentication/login-as-helpers';
+import { updateCaseStatus } from 'cypress/helpers/caseDetail/caseInformation/update-case-status';
+import { goToCase } from 'cypress/helpers/caseDetail/go-to-case';
+import { createAndServePaperPetition } from 'cypress/helpers/fileAPetition/create-and-serve-paper-petition';
+import { createTrialSession } from 'cypress/helpers/trialSession/create-trial-session';
+
+describe('Trial Session Aged Eligible Cases', () => {
```
akuny, line 16: The tests in this suite fail locally when attempting to set `eligibleCase.isAgedCase` to `true` in the `req.continue` callback.

akuny (changes requested):

paulo-filho-ctr (approved):

cypress/helpers/trialSession/create-trial-session.ts
```diff
@@ -39,6 +42,12 @@ export function createTrialSession(
   cy.get('#start-date-picker').type(startDate);
   cy.get('#estimated-end-date-picker').type(endDate);
   cy.get(`[data-testid="session-type-${sessionType}"]`).click();
+
+  if (associatedSwingTrialSessionId) {
```
paulo-filho-ctr, line 46: It is best to keep logical blocks from the test case body. I think we should move the `if` statement out of the test, and create an entire new `it` case block to test the specific expectation.

paulo-filho-ctr (changes requested):
I think it would be better to keep logical blocks outside from the test case body. I think we should move the if statement out of the test, and create an entire new it case block to test the specific expectation.

web-client/src/presenter/computeds/formattedEligibleCasesHelper.ts
```diff
@@ -43,42 +44,73 @@ const getSortableDocketNumber = docketNumber => {
...
   }
 
-  return `${priorityPrefix}_${getSortableDocketNumber(
-    isLeadInEligible
-      ? theCase.docketNumber === theCase.leadDocketNumber
-        ? theCase.docketNumber
-        : theCase.leadDocketNumber
-      : theCase.docketNumber,
-  )}-${getSortableDocketNumber(theCase.docketNumber)}`;
+  priorityPrefix += 'D';
```
paulo-filho-ctr, line 60: It seems that `priorityPrefix += 'D';` is not inside an `if` statement block. Is that the case that every `priorityPrefix` will contain a `D`? 

If so, maybe the default value of `priorityPrefix` should be `D` opposed to `''`?

web-client/src/presenter/computeds/formattedEligibleCasesHelper.ts
```diff
@@ -43,42 +44,73 @@ const getSortableDocketNumber = docketNumber => {
...
+      memberCasesObj[caseItem.leadDocketNumber]
+    ) {
+      caseItem.memberCases = memberCasesObj[caseItem.leadDocketNumber];
+      caseItem.memberCases.sort((a, b) => {
+        const aSortString = getFullSortString(a);
+        const bSortString = getFullSortString(b);
+        return aSortString.localeCompare(bSortString);
+      });
+    }
+  });
```
paulo-filho-ctr, line 104: Maybe we can combine the operations inside the `forEach` loop? Since both loop over formattedCases items?

cypress/local-only/tests/integration/trialSession/trial-session-aged-eligible-cases.cy.ts
alexdziarn: Ive been having trouble because it is passing locally, its just failing in github actions.

web-client/src/presenter/computeds/formattedEligibleCasesHelper.ts
jimlerza: I had a similar thought when I saw this code, but I believe that would only work if the `formattedCases` iterable had been sorted such that _all_ of the not-lead cases come before _all_ of the lead cases.

codyseibert (approved):

cypress/helpers/trialSession/create-trial-session.ts
alexdziarn: this is a helper method to create a trial session. its not supposed to be testing anything

web-client/src/presenter/computeds/formattedEligibleCasesHelper.ts
alexdziarn: I can refactor this, but I think this is the best solution in terms of logic and readability.

paulo-filho-ctr (approved):

akuny (approved):

PR #9290, opened 2025-10-22
8922 generate login credentials to test 1761154796

kampofo:
- Added IRS practitioners w/ DOJ practice_type
- Fixed user creation for users w/ mixed case emails


scripts/user/setup-glued-users.ts
```diff
@@ -152,7 +152,23 @@ const getPractitionerUsers = async (): Promise<any[]> => {
       .execute(),
   );
 
-  return users;
+  // 2b) Also include users who are irsPractitioner with practice_type = 'DOJ'
```
paulo-filho-ctr: Are the comments necessary or do you think we can remove them?

scripts/user/setup-glued-users.ts
```diff
@@ -152,7 +152,23 @@ const getPractitionerUsers = async (): Promise<any[]> => {
...
+  const dojIrsPractitioners = await getDbReader(reader =>
+    reader
+      .selectFrom('dwUser')
+      .selectAll()
+      .where('role', '=', 'irsPractitioner')
+      .where('practiceType', '=', 'DOJ') // adjust column name if your schema differs
+      .execute(),
+  );
+
+  // Merge and deduplicate by userId
```
paulo-filho-ctr, line 163: I think that comments can be safely removed, what you think?

scripts/user/setup-glued-users.ts
```diff
@@ -152,7 +152,23 @@ const getPractitionerUsers = async (): Promise<any[]> => {
...
   );
 
-  return users;
+  // 2b) Also include users who are irsPractitioner with practice_type = 'DOJ'
+  const dojIrsPractitioners = await getDbReader(reader =>
+    reader
+      .selectFrom('dwUser')
+      .selectAll()
+      .where('role', '=', 'irsPractitioner')
+      .where('practiceType', '=', 'DOJ') // adjust column name if your schema differs
```
paulo-filho-ctr: I think we can skip the comments here.

scripts/user/setup-glued-users.ts
kampofo: i think we can remove them.

scripts/user/setup-glued-users.ts
kampofo: I would like to keep this comment and the comment at the top of the getPetitionerUsers function, if that's ok. The logic is a bit funky (at least for me ) and it gives me a clearer understanding of whats happening there.

scripts/user/setup-glued-users.ts
jimlerza, line 18:
```suggestion
    region: 'REGION',
```

jimlerza:
```suggestion
const { Password, UserPoolId, region } = parseArgsAndEnvVars(scriptConfig) as {
```

jimlerza:
```suggestion
const cognito = new CognitoIdentityProvider({ region });
```

scripts/user/setup-glued-users.ts
```diff
@@ -152,7 +150,22 @@ const getPractitionerUsers = async (): Promise<any[]> => {
...
   );
 
-  return users;
+  const dojIrsPractitioners = await getDbReader(reader =>
+    reader
+      .selectFrom('dwUser')
+      .selectAll()
+      .where('role', '=', 'irsPractitioner')
+      .where('practiceType', '=', 'DOJ')
+      .execute(),
```
jimlerza, line 160: Do we need all of them? Should we add a limit or something?

```suggestion
      .where('practiceType', '=', 'DOJ')
      .limit(10)
      .execute(),
```

jimlerza (approved):
You don't have to take the suggestion to "use the region from for environment" but that env var is set to `us-east-1` and this is an open-source repo after all, so it's good to avoid hard-coding when we can.

scripts/user/setup-glued-users.ts
alexdziarn: there are only 15 dojpractitioners so its not too much to add them all

kampofo: should i leave this as is for now? or do we predict this number growing substantially over the life of the application?

jimlerza: Should be fine for now. We don't anticipate this number growing substantially.

scripts/user/setup-glued-users.ts
paulo-filho-ctr: I think if the code is too dense for clarity maybe rewrite to be a bit more explicit and add proper unit test to cover functionality cases?

AnthonyLoeraTaxCourt (approved):

PR #9291, opened 2025-10-22
8783 ts2339 clean up

AnthonyLoeraTaxCourt:
https://github.com/orgs/ustaxcourt/projects/9/views/1?pane=issue&itemId=126059121&issue=ustaxcourt%7Cef-cms%7C8783


docs/additional-resources/DAWSON-team-charter.md
jimlerza, line 1: Can this be in a separate PR?

shared/src/business/entities/docketEntry/DocketEntryFactory.ts
```diff
@@ -193,7 +193,8 @@ export class DocketEntryFactory extends JoiValidationEntity {
 
     const exDoc = ExternalDocumentFactory(this);
     const rules = exDoc.getValidationRules();
-    const docketEntryExternalDocumentSchema = rules.validate
+    const docketEntryExternalDocumentSchema = (rules as { validate?: unknown })
+      .validate
```
jimlerza: I am curious. What's the difference between doing a type assertion:

```
const rules = exDoc.getValidationRules();
const docketEntryExternalDocumentSchema = (rules as { validate?: unknown }).validate
```

and a type annotation:

```
const rules: { validate?: unknown } = exDoc.getValidationRules();
const docketEntryExternalDocumentSchema = rules.validate
```

here? If it's all the same, I find annotations (the latter) to be more readable and less heavy-handed.

shared/src/business/entities/externalDocument/SecondaryDocumentInformationFactory.test.ts
```diff
@@ -13,9 +13,10 @@ describe('SecondaryDocumentInformationFactory', () => {
         documentType: 'Motion for a New Trial',
         secondaryDocumentFile: {},
       });
-      expect(extDoc.getValidationErrors()?.objections).toEqual(
-        'Enter selection for Objections.',
-      );
```
jimlerza, line 18: If we're expecting `getValidationErrors().objections` to be populated, wouldn't swapping the `?` for `!` achieve the same thing?

web-api/src/lambdas/v1/marshallers/marshallCase.test.ts
```diff
@@ -87,7 +87,9 @@ describe('marshallCase (which fails if version increase is needed, DO NOT CHANGE
     expect(mock.sortableDocketNumber).toBeDefined();
     expect(mock.status).toBeDefined();
     expect(mock.trialLocation).toBeDefined();
+    // @ts-ignore
     expect(mock.contactPrimary).not.toBeDefined();
+    // @ts-ignore
     expect(mock.contactSecondary).not.toBeDefined();
```
jimlerza: What are we ignoring here?

web-api/src/lambdas/v1/v1ApiWrapper.test.ts
```diff
@@ -1,9 +1,13 @@
 /* eslint-disable jest/no-conditional-expect */
 import { v1ApiWrapper } from './v1ApiWrapper';
 
+interface ErrorWithStatusCode extends Error {
+  statusCode?: number;
+}
```
jimlerza: Why is this an interface when the file this is testing does the same thing as a type?

web-api/src/lambdas/v1/v1ApiWrapper.ts
```diff
@@ -11,19 +11,19 @@
...
-      toString() {
-        return this.message;
-      },
-    };
+    this.statusCode =
+      error.statusCode && V1_API_ERROR_STATUS_CODES.includes(error.statusCode)
+        ? error.statusCode
+        : 500;
+    const msg = error.message || 'An unexpected error occurred';
+    this.message = { message: msg, toString: () => msg } as unknown as string;
```
jimlerza: This type assertion doesn't pass the sniff test for me.

web-api/src/lambdas/v2/marshallers/marshallCase.test.ts
```diff
@@ -86,7 +86,9 @@ describe('marshallCase (which fails if version increase is needed, DO NOT CHANGE
     expect(mock.status).toBeDefined();
     expect(mock.trialDate).toBeDefined();
     expect(mock.trialLocation).toBeDefined();
+    // @ts-ignore
     expect(mock.contactPrimary).not.toBeDefined();
+    // @ts-ignore
     expect(mock.contactSecondary).not.toBeDefined();
```
jimlerza: Surely we can do better than `ts-ignore` here

web-api/src/lambdas/v2/v2ApiWrapper.ts
```diff
@@ -11,19 +11,19 @@
  * 500: returned by wrapping any Error which doesn't specify a statusCode.
  */
 const V2_API_ERROR_STATUS_CODES = [401, 403, 404, 500];
+type ErrorWithStatusCode = Error & { statusCode?: number };
```
jimlerza: I keep seeing this line. Perhaps we should export this type so it can be reused?

web-api/src/createLogger.test.ts
```diff
@@ -67,7 +90,9 @@ describe('redact', () => {
   };
 
   it('should remove sensitive information and preserve important information from a log entry', () => {
-    const redactedLogEntry = redact().transform(mockLogEntry);
+    const redactedLogEntry = redact().transform(
+      mockLogEntry,
+    ) as LogEntryWithRequest;
```
jimlerza, line 96: Here's another place where you might be able to use type annotation instead of type assertion.

web-client/integration-tests/sptoSposRequiresJudgeJourney.test.ts
```diff
@@ -131,12 +131,16 @@ describe('SPTO/SPOS requires a judge', () => {
...
+        formattedDocketEntries: RawDocketEntry[];
+      };
+
+      const caseDetailFormatted = (await runCompute(
         withAppContextDecorator(formattedCaseDetail),
         {
           state: cerebralTest.getState(),
         },
-      );
+      )) as FormattedCaseDetail;
```
jimlerza: Here's another instance where an annotation is sufficient:

```typescript
const caseDetailFormatted: FormattedCaseDetail = runCompute(
  withAppContextDecorator(formattedCaseDetail),
  {
    state: cerebralTest.getState(),
  },
);
```

web-client/src/persistence/dynamsoft/getScannerInterface.ts
```diff
@@ -1,6 +1,27 @@
+// @ts-nocheck - Legacy Dynamsoft scanner integration with extensive type safety issues
```
jimlerza: I think we can do better here!

web-client/src/persistence/dynamsoft/getScannerInterface.ts
```diff
@@ -1,6 +1,27 @@
...
+          onSuccess: (obj: unknown) => void,
+          onError: (error: unknown) => void,
+        ) => void;
+        EnumDWT_ImageType: { IT_JPG: number; IT_PNG: number };
+        EnumDWT_PixelType: { TWPT_RGB: number };
+        EnumDWT_CapSupportedSizes: { TWSS_A4: number };
+      };
+    };
+  }
+}
```
jimlerza, line 14: The `dwt` package has exported types that you can use instead:

```typescript
import Dynamsoft from 'dwt';

declare global {
  interface Window {
    Dynamsoft?: typeof Dynamsoft;
  }
}
```

web-client/src/presenter/actions/CaseDetail/completeMessageAction.ts
```diff
@@ -21,7 +21,8 @@ export const completeMessageAction = async ({
     .completeMessageInteractor(applicationContext, {
       messages: [
         {
-          messageBody: form?.message,
+          messageBody: ((form as { message?: string })?.message ||
+            '') as string,
```
jimlerza: The `as string` assertion shouldn't be necessary here.

You could use annotation instead:

```typescript
const form: { message?: string } = get(state.modal.form);
```
and then just
```typescript
messageBody: form?.message || '',
```

web-client/src/presenter/actions/CaseDetail/validateAddToTrialSessionAction.ts
```diff
@@ -9,9 +9,12 @@ import { state } from '@web-client/presenter/app.cerebral';
...
  */
 export const validateAddToTrialSessionAction = ({ get, path }: ActionProps) => {
-  const { calendarNotes, trialSessionId } = get(state.modal);
+  const { calendarNotes, trialSessionId } = get(state.modal) as {
+    calendarNotes?: string;
+    trialSessionId?: string;
+  };
 
-  const errors = {};
+  const errors = {} as { trialSessionId?: string; calendarNotes?: string };
```
jimlerza: Another instance where annotation is sufficient.

web-client/src/presenter/actions/TrialSession/extractUserNotesFromCalendaredCasesAction.ts
```diff
@@ -16,7 +16,7 @@ export const extractUserNotesFromCalendaredCasesAction = ({
   const userNotes = [];
 
   for (const calendaredCase of calendaredCases) {
-    userNotes.push(calendaredCase.notes);
+    userNotes.push((calendaredCase as { notes?: unknown }).notes);
```
jimlerza: This still results in a TS2345 error. I think the correct fix is to annotate the correct type for `userNotes` on line 16. If you do so, you won't need this new assertion on line 19 either.

web-client/src/presenter/actions/TrialSessionWorkingCopy/updateCalendaredCaseUserNoteAction.ts
```diff
@@ -16,11 +16,11 @@ export const updateCalendaredCaseUserNoteAction = ({
   const caseToUpdate = calendaredCases.find(
     aCase => aCase.docketNumber === userNote.docketNumber,
   );
-  caseToUpdate.notes = userNote;
+  (caseToUpdate as { notes?: unknown }).notes = userNote;
 
   const userNotes = [];
   for (const calendaredCase of calendaredCases) {
-    userNotes.push(calendaredCase.notes);
+    userNotes.push((calendaredCase as { notes?: unknown }).notes);
```
jimlerza: So, I get that `RawCase` doesn't have a `notes` prop, but I think this assertion is a little heavy-handed and is actually causing more headache for you in `extractUserNotesFromCalendaredCaseAction` later on. Extending the `RawCase` type would allow you to use a type annotation once rather than lots of assertions everywhere this is touched.

web-client/src/presenter/actions/validatePetitionFromPaperAction.ts
```diff
@@ -45,9 +51,13 @@ export const aggregateStatisticsErrors = ({ errors, get }: ActionProps) => {
...
 
-export const aggregatePetitionerErrors = ({ errors }) => {
+export const aggregatePetitionerErrors = ({
+  errors,
+}: {
+  errors: Record<string, any>;
+}) => {
   if (errors?.petitioners) {
-    errors.petitioners.forEach(e => {
+    (errors.petitioners as { index: number }[]).forEach(e => {
```
jimlerza: You wouldn't need this assertion if you included this in the type annotation on line 57.

web-client/src/presenter/computeds/caseDetailPractitionerSearchHelper.ts
```diff
@@ -32,19 +33,24 @@ export const caseDetailPractitionerSearchHelper = (get: Get): any => {
   const practitionerMatchesFormatted = formatMatches('practitionerMatches');
   const respondentMatchesFormatted = formatMatches('respondentMatches');
 
+  const modalStateTyped = modalState as {
+    practitionerMatches?: unknown[];
+    respondentMatches?: unknown[];
+  };
```
jimlerza: Annotation should be sufficient here.

jimlerza (changes requested):
I've gotten through about half of the files in this PR and decided to go ahead and post all of the comments that I've come up with so far.

The biggest feedback I have is that we should use _type annotation_ wherever possible. There is a **ton** of _type assertion_ going on in this PR.

Sometimes this is unavoidable, but assertion seems to have been preferred over annotation and I strongly believe it should be used sparingly.

> Type annotations declare the type, while type assertions override the inferred type. Annotations strengthen type checking, while assertions can weaken it if used incorrectly, as they bypass the compiler's checks. Use type annotations for general type safety and clarity, and reserve type assertions for cases where you have explicit knowledge of a value's type that the compiler lacks.

shared/src/business/utilities/documentGenerators/noticeOfChangeOfTrialJudge.test.ts
```diff
@@ -20,7 +20,7 @@ describe('noticeOfChangeOfTrialJudge', () => {
           nameOfClerk: 'Stephanie A. Servoss',
           titleOfClerk: 'Clerk of the Court',
           trialInfo: {
-            caseProcedureType: PROCEDURE_TYPES.SMALL,
+            caseProcedureType: PROCEDURE_TYPES['small'],
```
TomElliottFlexion:
```suggestion
            caseProcedureType: PROCEDURE_TYPES_MAP.small,
```

web-api/src/createLogger.test.ts
AnthonyLoeraTaxCourt: Even though LogEntryWithRequest extends TransformableInfo, we can't use type annotation because transform() returns boolean | TransformableInfo

web-api/src/lambdas/v1/marshallers/marshallCase.test.ts
AnthonyLoeraTaxCourt: I was just worried abut modifying the test. As it says `DO NOT CHANGE TESTS` in both v1 and v2 versions

paulo-filho-ctr (approved):

paulo-filho-ctr (approved):

jimlerza (approved):

Issue #8783, opened 2024-08-06
TS Error Cleanup: Property Not Found on Given Type


akuny:
**TS2339: `Property '<property>' does not exist on type '<type>'.`**

- [x] shared/src/business/entities/cases/Case.toRawObject.test.ts
- [x] shared/src/business/entities/docketEntry/DocketEntryFactory.ts
- [x] shared/src/business/entities/externalDocument/SecondaryDocumentInformationFactory.test.ts
- [x] shared/src/business/test/getFakeFile.ts
- [x] shared/src/business/useCases/getCaseInteractor.test.ts
- [x] shared/src/business/useCases/getReconciliationReportInteractor.ts
- [x] shared/src/business/useCases/uploadDocumentAndMakeSafeInteractor.test.ts
- [x] shared/src/business/useCases/validatePetitionerInformationFormInteractor.ts
- [x] shared/src/business/utilities/caseFilter.test.ts
- [x] shared/src/business/utilities/documentGenerators/noticeOfChangeOfTrialJudge.test.ts
- [x] shared/src/business/utilities/getFormattedCaseDetail.test.ts
- [x] shared/src/business/utilities/getJudgeWithTitle.test.ts
- [x] shared/src/business/utilities/pdfGenerator/documentTemplates/NoticeOfChangeOfTrialJudge.tsx
- [x] shared/src/business/utilities/shouldAppendClinicLetter.test.ts
- [x] shared/src/tools/generateCategories.ts
- [x] web-api/src/business/useCaseHelper/initialFilingDocuments/updateInitialFilingDocuments.ts
- [x] web-api/src/business/useCaseHelper/trialSessions/setNoticeOfChangeOfTrialJudge.test.ts
- [x] web-api/src/business/useCases/trialSessions/generateNoticeOfChangeOfTrialJudgeInteractor.test.ts
- [x] web-api/src/business/useCases/trialSessions/generateNoticesForCaseTrialSessionCalendarInteractor.test.ts
- [x] web-api/src/business/useCases/trialSessions/updateTrialSessionInteractor.ts
- [x] web-api/src/business/useCases/workItems/setWorkItemAsReadInteractor.ts
- [x] web-api/src/createLogger.test.ts
- [x] web-api/src/createLogger.ts
- [x] web-api/src/genericHandler.test.ts
- [x] web-api/src/genericHandler.ts
- [x] web-api/src/lambdas/cognitoAuthorizer/websocket-authorizer.test.ts
[...]

PR #9294, opened 2025-10-24
8783 ts2339 clean up to test 1761335521

AnthonyLoeraTaxCourt:
https://github.com/ustaxcourt/ef-cms/issues/8783


web-client/src/presenter/actions/TrialSession/getCreateTrialSessionAlertSuccessAction.test.ts
```diff
@@ -20,6 +20,9 @@ describe('getCreateTrialSessionAlertSuccessAction', () => {
         trialSession: '123',
       },
     });
-    expect(result.state.lastCreatedTrialSessionId).toEqual('123');
+    expect(
+      (result.state as { lastCreatedTrialSessionId?: string })
```
akuny: Instead of using type assertion here, we could instead phrase the expectation as:

`expect(result.state).toHaveProperty('lastCreatedTrialSessionId', '123');`

That way we don't get the type error, and we also don't have to do any type manipulation. This comment applies broadly speaking to the many instances where we assert type in this manner in this PR.

For context, `toHaveProperty` is a runtime check that doesn't require TypeScript to know about the property. It should be safe to use in this context even though it behaves differently from `toEqual`.

web-client/src/presenter/actions/AdvancedSearch/validateCaseDocketNumberSearchAction.ts
```diff
@@ -15,7 +15,7 @@ export const validateCaseDocketNumberSearchAction = ({
   const { docketNumber } = get(
     state.advancedSearchForm.caseSearchByDocketNumber,
   );
-  const errors = {};
+  const errors = {} as { docketNumber: string };
```
akuny: Since it's possible that `errors.docketNumber` is not assigned in the if block below, should `docketNumber` be an optional property in this type? This comment applies to similar instances in this PR.

web-client/src/views/Public/TodaysOpinions.tsx
```diff
@@ -15,7 +15,7 @@ export const TodaysOpinions = connect(
   function TodaysOpinions({
     openCaseDocumentDownloadUrlSequence,
     todaysOpinionsHelper,
-  }) {
+  }: any) {
```
akuny: Instead of using `any` here, we could update `app-public.cerebral` with:

```ts
  todaysOpinionsHelper: todaysOpinionsHelper as unknown as ReturnType<
    typeof todaysOpinionsHelper
  >,
```

Then, we could update `present-public` with:

```ts
  openCaseDocumentDownloadUrlSequence:
    openCaseDocumentDownloadUrlSequence as unknown as Function,
```

This comment applies generally speaking to the untyped sequences and computeds in `app-public.cerebral` and `present-public`. It's not the most graceful solution, but at least it adheres to the conventions we use in the main app presenter.

We want to avoid adding any `any` assertions in this PR.

akuny (changes requested):
There's a lot of good work here, but we do not want additional `any` type assertions in this PR, and there are several different uses of it throughout the code.

paulo-filho-ctr (approved):

web-client/src/presenter/actions/updateMessageModalAttachmentsAction.ts
```diff
@@ -25,31 +33,31 @@ export const updateMessageModalAttachmentsAction = ({
 
     const isSubjectEmpty = !subject || subject.trim() === '';
 
+  if (
+    isSubjectEmpty &&
+    (attachments?.length || 0) + (draftAttachments?.length || 0) === 0
```
codyseibert: for readability of this, I'd recommend pulling out some of this to variables then pass into the if statement

```
const attachmentCount = attachments?.length || 0;
const draftAttachmentCount = draftAttachments?.length || 0;
const hasNoAttachments = (attachmentCount + draftAttachmentCount) === 0;
```

web-client/src/presenter/actions/updateMessageModalAttachmentsAction.ts
```diff
@@ -25,31 +33,31 @@ export const updateMessageModalAttachmentsAction = ({
...
+  ) {
+    store.set(state.modal.form.subject, documentTitle.slice(0, 250));
```
codyseibert: I think moving that 250 to the ui constants would be better

codyseibert (approved):

akuny (approved):

PR #9295, opened 2025-10-24
Update to Team Charter

AnthonyLoeraTaxCourt:
Updating Team Charter


paulo-filho-ctr (approved):

alexdziarn (approved):

jimlerza:
Did the entire team get together and come up with this while I was away?

This document was originally written collaboratively and its contents were arrived upon as a consensus.

If you want to change the names and roles, that's fine, but where did the rest of this come from?

Issue #9258, opened 2025-10-09
BUG: Case is not automatically blocked if an APW is served at the time of Petition QC


ttlenard:
**Describe the Bug**
Currently, if a Petitions clerk is QCing a Paper petition, and Petitioner included an Application of Waiver of Filing Fee (APW), upon service of the Petition, the APW is added to the Pending items on the case, but no Blocked Label displays on the Case header, and the case is not on the Blocked Cases Report.

Any Case that has a Pending item, needs to be blocked from trial.

**Business Impact/Reason for Severity**
Med

**In which environment did you see this bug?**
Test

**Who were you logged in as?**
Petitions Clerk

**What were you doing when you discovered this bug? (Using the application, demoing, smoke tests, testing other functionality, etc.)**
Zack Brown found this bug while working on something else. Thanks Zack!

**To Reproduce**
1. Log in as a Petitions Clerk
2. Click on the Document QC Tab
3. Click start a new petition
4. Fill out all the required fields for a new petition
5. Be sure that you upload an APW as well as a Petition document (skip an RQT for now)
6. Click Review 
7. Click Serve to IRS
8. Click on Served in your QC box
9. Click on the Case docket number
10. Click on the Tracked items tab
11. Notice that you have a Pending item for the APW (Good!)
12. Notice that there is NOT a blocked label on the case (Bad!)
13. Run a Pending Report - case is on the pending report (Good!)
14. Run a Blocked cases report - case is NOT on the blocked cases report (but not sure if in this scenario it EVER would be? There isn't yet a designated place of trial)
15. Repeat steps 1-4, and the on step 5, include an RQT.
16. Repeat steps 6-14 (case is still not on the blocked cases report, even with a Trial location)
17. Next, go into one of the cases you just created
18. Add in a deadline (now the blocked label will display)
19. If you look at the case JSON now, it will say "automatic blocked reason: Deadline and Pending item"
20. The blocked label only displays if you add in a deadline first, if you delete the deadline, the blocked label stays, and the automatic block reason in the JSON now says pending item.

**Expected Behavior**
The blocked label needs to be added to the case for any pending item on the case, and in this situation an APW was included in the Paper petition, so a blocked label needs to be added to the case

**Actual Behavior**
A clear and concise description of what actually happened.
The blocked label isn't getting added to the case when an APW is included in a Paper Petition filing, and the case does not display on the blocked cases report.

**Screenshots**

[...]

PR #9297, opened 2025-10-27
9258 case is automatically blocked to test 1761588067

paulo-filho-ctr:
https://github.com/ustaxcourt/ef-cms/issues/9258
Fix the bug where the case is not automatically blocked if an APW is served at the time of Petition QC


AnthonyLoeraTaxCourt (approved):

cypress/local-only/tests/integration/trialSession/trial-session-blocked-cases.cy.ts
```diff
@@ -0,0 +1,66 @@
...
+      yearReceived: '2019',
+      includeApwDocument: true,
+    }).then(({ docketNumber }) => {
+
+      goToCase(docketNumber);
+      
+      cy.get('[data-testid="tab-tracked-items"]').click();
+      cy.get('[data-testid="pending-report-tab"]').click();
+      cy.get('#pending-items').should('exist');
+      cy.get('#pending-items').should('contain', 'Application for Waiver of Filing Fee');
```
jdr81394: Why are we not using the data-testid to select instead of the pending-items id?

paulo-filho-ctr: Because in this case, we do want to check for the contents, since a `data-testid` would not be sufficiently unique to capture only the value we want to assert, in this case "Application for Waiver of Filing Fee".

Screenshot 2025-10-28 at 1 15 42 PM

codyseibert: I would just check if there was maybe already a data-testid on the filing and proceedings link.  I'd rather it run a `cy.get('[data-testid="105-23_Application for Waiver of Filing Fee"]').should('exist') which you could add to the jsx when the table is rendering the link.  at lesat checking that a specific docket number has that document instead of just checking the full table that one exists would be a more reliable test imo.

TomElliottFlexion (approved):

PR #9304, opened 2025-10-29
8922 generate login credentials to test 1761752024

alexdziarn:
- only add accounts that are active
- little refactoring


kampofo (approved):
Looks good to me!

devinganey613 (approved):

paulo-filho-ctr (approved):

PR #9312, opened 2025-10-30
8546 Connect Order to Motion to test 1761846877

zacharybrown204:
For #8546 
Adding new fields to the file court issued docket entry view. 
- Adds a checkbox only for orders
- - When checked it will display pairs of dropdowns to select an order and a disposition

Added validation for above fields. 
- Note that our code processing validation errors was condensing any errors for arrays of complex objects to show as a single error for any object in that array. For example if I have an array like [{a: 1, b: 2, c:3}, {a:1, b:2, c:'error value'}], the validation errors would only say that "c" as an error and not "c" at index 1. 
We updated that validation to now report that "array-1-c" has an error. The validation errors are now returning the full path of the error so that we can pinpoint the exact spot the error is occuring in the form. This will allow us to have more accurate validation on dynamic forms. 

Added new table to hold relationship between docket entries. We tried to make it more generic in the case that there are any more relationships between docket entries in the future.


shared/src/business/entities/JoiValidationEntity.ts
```diff
@@ -104,7 +105,9 @@ export abstract class JoiValidationEntity {
     if (!error) return null;
     const errors = {};
     error.details.forEach(detail => {
-      if (!Number.isInteger(detail.context.key)) {
+      if (detail.path.length > 1) {
+        errors[_.join(detail.path, '-')] = detail.message;
```
akuny: Extreme nitpick: we tend to import lodash utility functions by name specifically (see line 2 in this file).

web-api/src/business/useCases/courtIssuedDocument/fileAndServeCourtIssuedDocumentInteractor.consolidatedCases.test.ts
```diff
@@ -265,4 +278,63 @@ describe('consolidated cases', () => {
...
+      documentTitle: '',
+      affectedDocketEntries: [
+        {
+          disposition: 'GRANTED',
+          docketEntryId: '7f61161c-ede8-43ba-8fab-69e15d057012',
+        },
+      ],
+    };
+
+    MOCK_LEAD_CASE_WITH_PAPER_SERVICE.docketEntries.push(mockOrderDocketEntry);
```
akuny: Do we want to mutate test fixture data as such? I think we could achieve the same thing with `leadCaseDocketEntries.push(mockOrderDocketEntry);` since `leadCaseDocketEntries` is created fresh each test run: that way we aren't touching the fixture data, and there's no risk of test order dependency.

web-api/src/business/useCases/docketEntry/fileCourtIssuedDocketEntryInteractor.ts
```diff
@@ -176,6 +176,28 @@ export const fileCourtIssuedDocketEntry = async (
     }),
   );
 
+  if (documentMeta.affectedDocketEntries) {
```
akuny, line 179: The code wrapped by this if statement is very similar to code in `fileAndServeCourtIssuedDocumentInteractor`: this could be extracted and placed in the use case helpers layer to reduce duplication.

web-api/src/business/useCases/docketEntry/updateDocketEntryMetaInteractor.ts
```diff
@@ -61,6 +64,16 @@ export const updateDocketEntryMeta = async (
     throw new Error('Unable to update unserved docket entry.');
   }
 
+  if (
```
akuny, line 67: Is it worth updating the accompanying unit test suite to cover this path?

web-client/src/presenter/computeds/Public/publicCaseDetailHelper.ts
```diff
@@ -77,7 +77,7 @@ export const formatDocketEntryOnDocketRecord = (
 
   entry.filingsAndProceedings = getFilingsAndProceedings(entry);
 
-  const canPublicUserSeeLink = DocketEntry.isDownloadable(entry, {
+  const canPublicUserSeeLink = DocketEntry.isDownloadable(entry, { // NOTE (#8546): may use similar to determine if link should be active for relations
```
akuny: Are these comments still necessary?

web-client/src/views/CourtIssuedDocketEntry/CourtIssuedDocketEntry.tsx
```diff
@@ -154,6 +157,177 @@ export const CourtIssuedDocketEntry = connect(
...
+                          className="usa-checkbox__label"
+                          htmlFor="dispositionOrder"
+                        >
+                          This order acts on at least one motion
+                        </label>
+                      </div>
+                    </fieldset>
+                  </FormGroup>
+                )}
+                {/* TODO (#8546): Fix this */}
```
akuny: Is there still a fix required somewhere here?

web-client/src/views/DocketRecord/FilingsAndProceedings.tsx
```diff
@@ -143,6 +156,39 @@ export const FilingsAndProceedings = connect<
...
+              return (
+                <>
+                  <br></br>
+                  <span className="display-inline-block">
+                    <span> --- </span>
+                    <Button
+                      link
+                      className={classNames(
+                        'text-right',
+                        // entry.isStricken && 'stricken-docket-record', // NOTE (#8546): This may be unnecessary
```
akuny: Is this comment necessary?

web-client/src/presenter/computeds/addCourtIssuedDocketEntryHelper.ts
```diff
@@ -77,6 +99,8 @@ export const addCourtIssuedDocketEntryHelper = (
   return {
     documentTypes,
     formattedDocumentTitle,
+    caseMotions,
```
akuny, line 102: Is it worth updating the unit tests for this computed to cover these two new things?

cypress/local-only/tests/integration/caseDetail/docketRecord/courtIssuedFiling/docket-clerk-applies-order-to-lead-case-motion.cy.ts
```diff
@@ -0,0 +1,59 @@
+import { goToCase } from '../../../../../../helpers/caseDetail/go-to-case';
```
akuny, line 1: When this test is run multiple times locally, it creates new test motions and test orders each run, but every order is associated with the initially created motion (see screenshot below). Is this intentional? If not, could we update this test such that documents are associated as expected over the course of multiple runs?

Screenshot 2025-10-31 at 10 28 52 AM

akuny (changes requested):
This looks good. Just a few areas where we might want to tighten up.

cypress/local-only/tests/integration/caseDetail/docketRecord/courtIssuedFiling/docket-clerk-applies-order-to-lead-case-motion.cy.ts
zacharybrown204: Updated test to be more idempotent. We will now capture the docket entry id of our new motion and select that one when creating a new order.

web-client/src/views/DocketRecord/FilingsAndProceedings.tsx
zacharybrown204: Removed

web-client/src/views/CourtIssuedDocketEntry/CourtIssuedDocketEntry.tsx
zacharybrown204: Comment removed

web-client/src/presenter/computeds/Public/publicCaseDetailHelper.ts
zacharybrown204: Comment Removed

web-api/src/business/useCases/docketEntry/updateDocketEntryMetaInteractor.ts
zacharybrown204: Added 3 new tests to cover updating docket entries with affectedDocketEntries

web-api/src/business/useCases/docketEntry/fileCourtIssuedDocketEntryInteractor.ts
zacharybrown204: Extracted code out to a helper function and used that instead. I know we have a bunch of helper functions going through app context and that function to get helpers, but I also know we want to get away from app context so I had just direct imported the helper.

web-api/src/business/useCases/courtIssuedDocument/fileAndServeCourtIssuedDocumentInteractor.consolidatedCases.test.ts
zacharybrown204: Fixed, turns out this test didn't even need that mock docket entry pushed.
I could have sworn I copied that line from another test, but I couldn't find where it was to check if it was a problem there too. Might've just picked a bad replacement for a missing reference on the copied line or something.

shared/src/business/entities/JoiValidationEntity.ts
zacharybrown204: Duely noted, fixed

web-api/src/business/useCases/docketEntry/fileCourtIssuedDocketEntryInteractor.ts
akuny: Thanks for doing that--direct importing is the way to go.

akuny (approved):

paulo-filho-ctr (approved):

shared/src/business/entities/DocketEntry.ts
```diff
@@ -186,6 +187,10 @@ export class DocketEntry extends JoiValidationEntity {
   public signedJudgeName?: string;
   public strickenBy?: string;
   public strickenByUserId?: string;
+  // public orderDocketEntryId?: string;
+  // public motionDisposition?: string; // 'GRANTED' | 'DENTIED' | 'PARTIAL';
```
TomElliottFlexion: Intentional?

shared/src/business/entities/EntityConstants.ts
```diff
@@ -88,7 +87,17 @@ export const JURISDICTIONAL_OPTIONS = {
   undersigned: 'Jurisdiction is retained by the undersigned',
 };
 
-export const MOTION_DISPOSITIONS = { DENIED: 'Denied', GRANTED: 'Granted' };
+export type DocketEntryRelation = {
+  disposition: string;
+  docketEntryId: string;
+  // docketEntryType: string;
```
TomElliottFlexion: Still needed?

shared/src/business/entities/JoiValidationEntity.ts
```diff
@@ -104,7 +104,9 @@ export abstract class JoiValidationEntity {
     if (!error) return null;
     const errors = {};
     error.details.forEach(detail => {
-      if (!Number.isInteger(detail.context.key)) {
+      if (detail.path.length > 1) {
+        errors[join(detail.path, '-')] = detail.message;
+      } else if (!Number.isInteger(detail.context.key)) {
```
TomElliottFlexion, line 109: Nice!

shared/src/business/entities/courtIssuedDocument/CourtIssuedDocumentTypeH.ts
```diff
@@ -1,14 +1,13 @@
...
 import { JoiValidationConstants } from '../JoiValidationConstants';
 import { replaceBracketed } from '../../utilities/replaceBracketed';
 
 export class CourtIssuedDocumentTypeH extends CourtIssuedDocument {
   public attachments: boolean;
-  public documentTitle?: string;
+  public documentTitle: string;
   public documentType: string;
-  public eventCode?: string;
+  public eventCode: string;
```
TomElliottFlexion, line 10: Really appreciate these fixes

shared/src/business/entities/EntityConstants.ts
JayFlexy: Nope, good catch

web-api/src/business/useCaseHelper/docketEntry/addAssociatedDocketEntries.ts
```diff
@@ -0,0 +1,30 @@
+import { CourtIssuedDocumentAnyType } from '@shared/business/entities/courtIssuedDocument/CourtIssuedDocumentConstants';
+import { DocketEntry } from '@shared/business/entities/DocketEntry';
+import { upsertDocketEntryRelatedEntries } from '@web-api/persistence/postgres/docketEntries/upsertDocketEntryRelatedEntries';
+
+export async function addAssociatedDocketEntries(
+  casesToUpdate: Omit<RawCase, 'consolidatedCases'>[],
```
TomElliottFlexion, line 6: Perhaps a unit test for this one?

zacharybrown204: Added unit test for this helper

JayFlexy (approved):
LGTM

JayFlexy (approved):
🥇

Issue #9308, opened 2025-10-29
BUG: When the document type "Notice" is selected for a party filed document, it is automatically generating the title "Notice of Attachments in the Nature of Evidence"


ttlenard:
**Describe the Bug**
This issue was reported to us via DAWSON Support helpdesk ticket.  A docket clerk added Notice as a paper filing to a case but noticed after it was filed on the docket record that it was given the title Notice of Attachments in the nature of evidence.  The document that she added was a generic Notice with free text added, NOT a Notice of Attachments in the Nature of Evidence.  The clerk was able to edit the document title after this happened but was not able to edit or add a new coversheet to the document to ensure the coversheet had the correct document name.

After troubleshooting, it appears that any time a generic Notice is selected, either when adding a paper filing, or if a clerk were to edit an existing document on the docket record, instead of displaying the generic "Notice" as the document title, it switches to Notice of Attachments in the nature of evidence.  This should not be happening.

Additional information: I reviewed the DAWSON Document type JSON's (courtIssuedEventCodes, internalFilingEvents) and didn't see Notice of Attachments in the Nature of Evidence as a document type in either of the lists. Not sure where this is coming from?

UPDATED: NOT ADRESSING THIS ISSUE WITH THIS BUG: Another thing: Once the Notice is docketed, it appears that it is "confused" as to whether or not the document is an internal filed document or a court issued document.  For example, if the clerk edits the notice after docketing the paper filing, the document list is all court issued documents, rather than the internal filing event codes.  AND, there is no opportunity to add additional info 1, 2, or check the box to add to coversheet because of this.

**Business Impact/Reason for Severity**
medium

**In which environment did you see this bug?**
Production, Test

**Who were you logged in as?**
Docket Clerk

**What were you doing when you discovered this bug? (Using the application, demoing, smoke tests, testing other functionality, etc.)**
This issue was reported to us by a docket clerk

**To Reproduce**
1. Log in to DAWSON as a docket clerk
2. Navigate to a case
3. Click on the Create button and select Paper filing
4. Add in the filed date, and select Notice as the document type (at this point, it will still display Notice correctly)
5. Add in some text to the "What is this notice for?" field
6. Fill in the rest of required data for the paper filing
7. click Save and Serve
8. Notice that on the "Are you ready to Initiate Service?" modal, the document title now says "Notice of Attachments in the nature of Evidence"
9. There is no way to NOT have this document title, it appears
10. click Yes, Serve
[...]

PR #9318, opened 2025-10-31
9308 bugfix auto generated title notice to test 1761938556

alexdziarn:
Talked with Tenille and Tom and decided to only address the issue regarding title name after a notice is created from a paper filing. Will address the issue with title name after editing docket entry in another ticket.
https://github.com/ustaxcourt/ef-cms/issues/9308

### DIFF
--- a/cypress/local-only/tests/integration/caseDetail/docketRecord/paperFiling/docket-clerk-adds-paper-filing.cy.ts
+++ b/cypress/local-only/tests/integration/caseDetail/docketRecord/paperFiling/docket-clerk-adds-paper-filing.cy.ts
@@ -5,6 +5,8 @@ import {
 } from '../../../../../../helpers/authentication/login-as-helpers';
 import { logout } from '../../../../../../helpers/authentication/logout';
 import { petitionsClerkServesPetition } from '../../../../../../helpers/documentQC/petitionsclerk-serves-petition';
+import { goToCase } from 'cypress/helpers/caseDetail/go-to-case';
+import { attachFile } from 'cypress/helpers/file/upload-file';
 
 /**
  * Given a case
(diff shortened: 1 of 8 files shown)


akuny (approved):
This fix works on my end. It may be worth considering if the unit tests for the interactor and the frontend action that were changed in this PR should be updated to reflect the change in functionality.

shared/src/business/entities/EntityConstants.ts
```diff
@@ -945,17 +945,17 @@ export const SYSTEM_GENERATED_DOCUMENT_TYPES = {
   ...AUTO_GENERATED_DEADLINE_DOCUMENT_TYPES_WITH_NAMES,
 };
 
-export const SYSTEM_AND_INTERNAL_DOCUMENT_TYPES = [
-  ...Object.values(SYSTEM_GENERATED_DOCUMENT_TYPES).map(doc => ({
-    ...doc,
-    label: doc.documentTitle,
-    value: doc.eventCode,
-  })),
+export const INTERNAL_DOCUMENT_TYPES_AND_NOTR = [
```
TomElliottFlexion, line 948: Doesn't need to be addressed before going to test, but how would you feel about the name `EDITABLE_DOCUMENT_TYPES`? I believe that's what this constant is meant to represent

web-client/src/business/useCases/externalDocument/generateExternalDocumentTitle.ts
```diff
@@ -15,11 +14,6 @@ export const generateExternalDocumentTitle = (
   applicationContext: ClientApplicationContext,
   { documentMetadata },
 ) => {
-  const entry = Object.values(SYSTEM_GENERATED_DOCUMENT_TYPES).find(
-    entry => entry.eventCode === documentMetadata.eventCode,
-  );
-  if (entry) return entry.documentTitle;
-
```
TomElliottFlexion, line 22: Just to be 100% sure, does removing this line affect the ability of users to edit an NOTR?

TomElliottFlexion (approved):

shared/src/business/entities/EntityConstants.ts
alexdziarn: This constant is used when creating a paper filing, I think its fine the way it is since it not only used when editing documents.

web-client/src/business/useCases/externalDocument/generateExternalDocumentTitle.ts
alexdziarn: I had no problem with editing a NOTR

cypress/local-only/tests/integration/caseDetail/docketRecord/paperFiling/docket-clerk-adds-paper-filing.cy.ts
```diff
@@ -30,4 +32,50 @@ describe('Docket clerk adding a paper filing', () => {
       'Please correct the following errors on the page',
     );
   });
+  it.only('should set title properly for a notice', () => {
```
akuny: We probably don't want this only in this test.

akuny (changes requested):

cypress/local-only/tests/integration/caseDetail/docketRecord/paperFiling/docket-clerk-adds-paper-filing.cy.ts
alexdziarn: you're right good catch

akuny (approved):

PR #9319, opened 2025-10-31
Fixed login.cy.ts bug due to new button implementation

jdr81394:


jimlerza (approved):

jimlerza:
Note to self: run this against prod (or check prod's cognito for the admissions clerk user that this uses)

alexdziarn (approved):

cypress/readonly/integration/authentication/login.cy.ts
```diff
@@ -7,21 +8,14 @@ describe('Given a user with a DAWSON account', () => {
...
-      cy.get('[data-testid="email-input"]').type(
-        'testAdmissionsClerk@example.com',
-      );
-      cy.get('[data-testid="password-input"]').type(
-        getCypressEnv().defaultAccountPass,
-      );
-      cy.get('[data-testid="login-button"]').click();
-      cy.get('[data-testid="account-menu-button"]');
-      cy.get('[data-testid="error-alert"]').should('not.exist');
+      loginAsAdmissionsClerk();
```
jimlerza: We can't change the user this uses because `admissionsclerk1@example.com` doesn't exist on prod.

```suggestion
      loginAsAdmissionsClerk('testAdmissionsClerk@example.com');
```

jimlerza (changes requested):

jimlerza (approved):

PR #9351, opened 2025-11-10
9265 minute sheet updates to test 1762786627

kampofo:


cypress/local-only/tests/integration/trialSessionMinutes/create-and-print-minute-sheet.cy.ts
```diff
@@ -148,20 +148,38 @@ describe('Create a minute sheet, fill out sections of the form, navigate away an
         });
       });
 
-      it('Can fill out Trial/Hearing date, type, and note', function () {
-        cy.get('#trialHearingTranscriptOrdered').check({
+      it('Can fill out Trial date, type, and note', function () {
+        cy.get('#trialTranscriptOrdered').check({
```
jdr81394, line 152: We should be using the "data-testid" attribute. Is there a reason we are not?

cypress/local-only/tests/integration/trialSessionMinutes/create-and-print-minute-sheet.cy.ts
```diff
@@ -148,20 +148,38 @@ describe('Create a minute sheet, fill out sections of the form, navigate away an
...
+        cy.get('#trialNote').should('have.value', 'Trial note');
+        cy.get('#trialTranscriptOrdered').should('be.checked');
 
-        cy.get('#trialHearingNote').clear();
-        cy.get('#trialHearingType').select('');
-        cy.get('#trialHearingTranscriptOrdered').uncheck({
+        cy.get('#trialNote').clear();
+        cy.get('#trialType').select('');
+        cy.get('#trialTranscriptOrdered').uncheck({
+          force: true,
```
jdr81394, line 165: Why is force being used? It should only really be used as a last resort. For example you want to deliberately test hidden or disabled elements.

cypress/local-only/tests/integration/trialSessionMinutes/create-and-print-minute-sheet.cy.ts
kampofo: that was how it was being targeted before. I just updated the id to match the the new field.

cypress/local-only/tests/integration/trialSessionMinutes/create-and-print-minute-sheet.cy.ts
kampofo: This was also an edit of an existing test that was broken. That was already there.

jdr81394 (approved):

web-client/src/views/TrialSessionMinutes/CaseMetadataFieldset.tsx
```diff
@@ -510,15 +640,15 @@ export const CaseMetadataFieldset = ({
             <div className="usa-checkbox">
               <input
                 aria-describedby="representing-legend"
-                checked={caseMetadataFormState.trialHearing.transcriptOrdered}
+                checked={caseMetadataFormState.hearing.transcriptOrdered} // add state for hearing seperately?
```
AnthonyLoeraTaxCourt, line 643: Is this comment neccesary?

AnthonyLoeraTaxCourt (approved):

PR #9375, opened 2025-11-14
9329 Zendesk Reg Status Automation Bugfix to Test

devinganey613:
The previous query to cognito for an email was not paginating the results, which resulted in emails with common names not being found because not all the results were returned. There was also an issue where short emails (e.g. ab@example.com) were not being found, possibly because they also returned a lot of results and needed to be paged through. As a fail safe, I added a second query to cognito to use the exact email passed if the user was not found by the first query.
Also addresses #9324


AnthonyLoeraTaxCourt (approved):

shared/src/business/useCases/automations/getRegStatusInteractor.test.ts
```diff
@@ -148,6 +148,9 @@ describe('getRegStatusInteractor', () => {
     getCognito.mockReturnValueOnce({
```
akuny, line 150: I think comment(s) explaining the intent behind the duplicated calls of

```ts
getCognito.mockReturnValueOnce({
      send: jest.fn().mockResolvedValue({ Users: [] }),
    });
```

would be helpful, since on the face of it the intent isn't immediately apparent.

akuny (changes requested):
Just one spot where a comment might be helpful for future developers making sense of this code.

akuny (approved):

PR #9381, opened 2025-11-17
9333 replace favicon to test 1763405365

favourustc:


akuny (approved):

AnthonyLoeraTaxCourt (approved):

jimlerza (approved):
Question for @ttlenard and @cholly75 - is this the best resolution we can do for this?

Can we track down an SVG from the resources that were delivered when this logo was originally designed?

PR #9384, opened 2025-11-17
9265: Remove minute sheet permissions for chambers and trial clerks

akuny:
This PR removes minute sheet permissions for chambers and trial clerk users, effectively deactivating the minute sheet feature so that we can run a data transformation migration.


jimlerza (approved):
With the cypress tests removed, is it possible to once again remove the permission? I think that they may still be able to access minute sheets via bookmark or hyperlink if we don't do so.

AnthonyLoeraTaxCourt (approved):

jimlerza:
This was released to prod for 30 minutes, but shouldn't actually be merged to staging because it deletes things!

Issue #9284, opened 2025-10-20
Consolidate lower environments' `info` cluster to a single instance


jimlerza:
As a wallet, so I might stay fat...

OpenSearch is expensive. We can score some nice savings by consolidating all of the lower environments' logs into a single `info` cluster.

## Pre-Conditions
  - None
 
## Acceptance Criteria
- `info` cluster's terraform resources are not removed from the repo
- environments create the `info` cluster resources conditionally
- the `LogsToElasticSearch_info` lambda is updated to post logs to the consolidated `info` cluster's endpoint

## Pain Avoided/Frustration Saved
- Initially setting up kibana credentials on 13 environments is tedious
- OpenSearch costs about ~$50/month per lower environment (except `test`, which costs more)

## Breadth/Pervasiveness of Problem
  - None
  
## Complexity of Problem (Low, Medium, High) and Why it's Complex
  - low
     - plenty of examples of using `count` to conditionally create resources
     - plenty of examples of using Secrets Manager to store and retrieve shared constants
  
## Notes
  - terraform's [`count` meta-argument](https://spacelift.io/blog/terraform-count) can be utilized to conditionally create resources
  -  we have an active reservation for a 2-node `m5.large.search` cluster that can be earmarked for this purpose

jimlerza:
Question: where should this consolidated cluster live? 🤷

jdr81394:
Decision was taken to put it into the test account because it's an M5.large.

PR #9394, opened 2025-11-19
9284 consolidate lower env info cluster paulo to test 1763585172

paulo-filho-ctr:
[9284](https://github.com/ustaxcourt/ef-cms/issues/9284)
### The Pul Request contains:
- The addition of new variables to control if an info cluster should be deployed for that specific account/env
- The output of a deployed info cluster that includes the arn/endpoint of the deployed info cluster
- Consumers of the info cluster depends on new Secrets Manager keys:
  - `ES_INFO_CLUSTER_ARN`: stores deployed info cluster arn
  - `ES_INFO_CLUSTER_ENDPOINT`: stores deployed info cluster endpoint
  -  `ES_INFO_CLUSTER_LOWER_ENVIRONMENT_ACCOUNT_IDS`: stores lower environments account ids, to set access permissions to the host info cluster
 - The addition of `null_resources` so consumers of the info cluster can make a backup the info cluster log bucket, copy all the contents to the backup bucket, empty the original info cluster bucket; delete the original info cluster bucket for consumers


web-api/terraform/modules/kibana/kibana.tf
```diff
@@ -177,31 +202,55 @@ CONFIG
...
+  )
+
+  consumer_account_roots = [
+    for account_id in var.es_info_cluster_lower_environment_account_ids :
+    "arn:aws:iam::${account_id}:root"
+  ]
+  all_account_roots = concat(
+    ["arn:aws:iam::${data.aws_caller_identity.current.account_id}:root"],
+    local.consumer_account_roots
+  )
```
paulo-filho-ctr: @jimlerza Can you double check if the permission are set correctly here? We are using :root but is there another permission that should be set here, with less privileges?

jimlerza (approved):

TomElliottFlexion (approved):

akuny (approved):

Issue #8615, opened 2024-02-08
BUG: Some documents on a docket record display "0" pages


ttlenard:
**Describe the Bug**
A Court user reported that multiple documents on the docket record of the Facebook Case (21959-16) are displaying "0" pages.  One example document is index 829, which has 132,013 pages.  Not sure if there are other cases that this is occurring in?

At first, I thought that it may have to do with the large number of pages, and that maybe we had a digit limit on the docket record, but that doesn't seem to be the case (Thanks for checking Tom!).  

After doing research for bug ustaxcourt/ef-cms#8616, I'm pretty sure that the 0 pages thing is the side effect of the 504's that occurred for the big documents.  Maybe during the upload PDF process, if it bonked out somewhere in this sequence, does it mess up the document page count step?  Is this why it is reporting 0 pages?  I also note that (some? all?) these documents do not have a coversheet on them.  Does something in the add coversheet step also count out the page numbers? If it bonked on the coversheet step, does it skip the page count?  Just throwing out ideas at this point.

**Business Impact/Reason for Severity**
low

**In which environment did you see this bug?**
Prod, Test

**Who were you logged in as?**
Any DAWSON user

**What were you doing when you discovered this bug? (Using the application, demoing, smoke tests, testing other functionality, etc.)**
This was reported to us by a Court user.

**To Reproduce**
1. Go to docket 21959-16
2. Scroll down the docket record, notice that there are multiple Exhibit documents that display 0 pages.  
3. I did download index 829, and this document is 132,013 pages in length.

**Expected Behavior**
Documents need to reflect the appropriate page number.

**Actual Behavior**
A clear and concise description of what actually happened.
Some documents are displaying "0" pages

**Screenshots**
<img width="1234" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMDFDQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--0acbd7e5c253928085e125f2f1a2b6b2c050a9ac/image.png" alt="image.png" />

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Cause of Bug, If Known**

**Process for Logging a Bug:**
* Complete the above information
* Add a severity tag (Critical, High Severity, Medium Severity or Low Severity). See below for priority definition. 

**Severity Definition:**
* Critical Defect
Blocks entire system's or module’s functionality
[...]

PR #9421, opened 2025-11-24
8615 script to fix null page counts

devinganey613:
#8615 
Created a script to populate number of pages column in the database for served docket entries where the column is null.
Also updated process to save external docket entries to populate the number of pages column on initial insert so it is never null.

### Manual Deploy Steps

- [ ] Run the following script on production to fix the page counts in the database where needed. Will output a list of docket entry ids that it could not update. Based on running the script on test, it should only be about 14 documents from 2020-2021 where a locked PDF was uploaded.
   ```bash
   scripts/run-once-scripts/fix-null-page-counts.ts
   ```


scripts/run-once-scripts/fix-null-page-counts.ts
```diff
@@ -0,0 +1,115 @@
...
+    env: 'ENV',
+  },
+  requireActiveAwsSession: true,
+};
+
+parseArgsAndEnvVars(scriptConfig);
+// eslint-disable-next-line @typescript-eslint/no-floating-promises
+(async () => {
+  console.log('Starting...');
+  const applicationContext = createApplicationContext({});
```
jimlerza, line 26: At one point there was a push to import `applicationContext` directly in instances like these, where it is being instantiated with an empty object in a script, instead of using `createApplicationContext`. I believe this is because you wind up with the same static result either way and calling `createApplicationContext` does that same work all over again.

Just an FYI for future reference, though; not worth changing in a script that will only run once.

jimlerza (approved):

kampofo (approved):
Looks good to me!

PR #9422, opened 2025-11-24
9234 past sessions to test 1764019328

jdr81394:


web-client/src/views/Public/TrialSessions/PublicTrialSessionsTable.tsx
```diff
@@ -40,8 +42,9 @@ export const PublicTrialSessionsTable = connect<
               <th className="width-mobile">Judge</th>
             </tr>
           </thead>
-          {publicTrialSessionsHelper.trialSessionRows.map(row => {
+          {filteredTrialSessionRows.map(row => {
             if (isTrialSessionWeek(row)) {
+              console.log('row: ', row);
```
AnthonyLoeraTaxCourt, line 47: Is ths Console.log() needed?

AnthonyLoeraTaxCourt (approved):

zacharybrown204 (approved):

web-client/src/views/Public/TrialSessions/PublicTrialSessionsTable.tsx
paulo-filho-ctr, line 47: I think we should remove the console.log. No?

PR #9423, opened 2025-11-25
9414: remove dynamo related code

codyseibert:
These changes were successfully deployed and tested on exp8, https://app.circleci.com/pipelines/github/ustaxcourt/ef-cms/16546/workflows/ffe595b1-824f-4979-a2ef-b0a597a635c1

### Local

- [ ] Remove the local dynamodb installation:
   ```bash
   rm -rf .dynamodb
   ```


.circleci/config.yml
jimlerza, line 1: None of the changes in this file are functional, right? Chris and Tenille want to keep calling this "glue" even if we don't use AWS Glue to do it.

docs/operations/runbooks/create-new-lower-environment.md
```diff
@@ -370,9 +364,9 @@ This runbook describes the process of creating a new DAWSON lower environment in
    ```bash
    cd ./scripts/postgres && ./create-rds-users.sh && cd ../..
    ```
-1. ⚖️ If this lower environment is to have prod-like data, kick off a glue job now:
-   1. In [CircleCI](https://app.circleci.com/pipelines/github/ustaxcourt/ef-cms), trigger a glue job, choosing the following settings:
+1. ⚖️ If this lower environment is to have prod-like data, kick off a back migration job now:
+   1. In [CircleCI](https://app.circleci.com/pipelines/github/ustaxcourt/ef-cms), trigger a back migration job, choosing the following settings:
       1. `run_build_and_deploy`: `false`
-      1. `run_glue_to_[ENV]`: `true`
+      1. `run_back_migrated_to_[ENV]`: `true`
```
jimlerza: We're still gonna call this glue.

docs/dependency-updates.md
```diff
@@ -115,8 +115,6 @@ If the `Dockerfile` has changed, you will need to build a new docker image and p
 Check if there is an update to the Terraform AWS provider and update our `.tf` files to use the [latest version](https://registry.terraform.io/providers/hashicorp/aws/latest) of the provider.
 
 1. Search the entire project for `source  = "hashicorp/aws"` and make sure it's set to the latest version.  For example, some of these files have the AWS provider defined:
-   - `./shared/admin-tools/glue/glue_migrations/main.tf`
-   - `./shared/admin-tools/glue/remote_role/main.tf`
```
jimlerza, line 120: I think we should either find new examples to put here or reword the sentence above the lines that have been removed.

scripts/circleci/setup-ssm-for-glue-job.sh
jimlerza, line 1: This can still be called "glue"

scripts/circleci/setup-ssm-for-back-migration-job.sh
```diff
@@ -6,11 +6,14 @@
...
 echo "setting migrate flag to false"
 aws ssm put-parameter --region us-east-1 --name "/DAWSON/${ENV}/migrate" --value "false" --type "String" --overwrite
 
+# TODO: I'm not sure we need this anymore
 echo "setting source-table-version to alpha"
 aws ssm put-parameter --region us-east-1 --name "/DAWSON/${ENV}/source-table-version" --value "alpha" --type "String" --overwrite
 
+# TODO: I'm not sure we need this anymore
 echo "setting destination-table-version to alpha"
 aws ssm put-parameter --region us-east-1 --name "/DAWSON/${ENV}/destination-table-version" --value "alpha" --type "String" --overwrite
```
jimlerza: We can only remove this if there's no code anywhere that changes the version to `beta` for any reason. Because you're right - `test` should already be on alpha (considering we run this script once a month and nothing switches from alpha to beta anymore) by now.

scripts/circleci/setup-ssm-for-back-migration-job.sh
```diff
@@ -6,11 +6,14 @@
   "AWS_ACCESS_KEY_ID" \
   "AWS_SECRET_ACCESS_KEY"
 
+# TODO: we may still need this
 echo "setting migrate flag to false"
 aws ssm put-parameter --region us-east-1 --name "/DAWSON/${ENV}/migrate" --value "false" --type "String" --overwrite
```
jimlerza: We do still need this, although we could probably rename it to "reindex_data" or something, because now all it's used for is telling a build to reindex all data.

scripts/judge/change-chief-judge.sh
```diff
@@ -40,7 +40,7 @@ echo "Updating judge with last name: ${NEW_JUDGE_LAST_NAME} to Chief Judge"
 # update the judge's names in cognito
 USER_POOL_ID=$(aws cognito-idp list-user-pools --query "UserPools[?Name == 'efcms-${ENV}'].Id | [0]" --max-results 30 --region "${REGION}" --output text)
 
-# will not work on lower envs (emails are sanitized during glue job; email differs in cognito vs persistence)
+# will not work on lower envs (emails are sanitized during back migration job; email differs in cognito vs persistence)
```
jimlerza: Long live glue

scripts/user/add-judge.ts
```diff
@@ -38,11 +38,6 @@ import { getNewPasswordForEnvironment } from './make-new-password';
 const scriptConfig: ScriptConfig = {
   description:
     'add-judge - Creates a new Judge user in a deployed environment.',
-  environment: {
-    dynamoDbTableName: 'DYNAMODB_TABLE_NAME',
-    env: 'ENV',
-    userPoolId: 'USER_POOL_ID',
```
jimlerza, line 44: Even though these two environment variables aren't used directly in the script, they are still a helpful check to ensure that the user's terminal session is pointed to a deployed environment. We're specifically checking the presence of the Cognito user pool id env var because the call to `createOrUpdateUser()` will use that var. (The `env` check is pretty standard just to make sure you've run the env switcher, and doesn't have to stay though)

scripts/user/add-user.ts
```diff
@@ -17,11 +17,6 @@ import { ACCOUNT_STATUS } from '@shared/business/entities/EntityConstants';
 const scriptConfig: ScriptConfig = {
   description:
     'add-user - Creates a new DAWSON user in a deployed environment.',
-  environment: {
-    dynamoDbTableName: 'DYNAMODB_TABLE_NAME',
-    env: 'ENV',
-    userPoolId: 'USER_POOL_ID',
```
jimlerza, line 23: Let's keep the check for the Cognito user pool id in this script too, since it does Cognito stuff.

scripts/user/setup-back-migrated-users.ts
jimlerza, line 1: Long live glue

scripts/download-all-case-documents.ts
```diff
@@ -15,11 +15,6 @@ import fs from 'fs';
 const scriptConfig: ScriptConfig = {
   description:
     'download-all-case-documents - Downloads all docket entries for the given docket number.',
-  environment: {
-    dynamoDbTableName: 'DYNAMODB_TABLE_NAME',
-    efcmsDomain: 'EFCMS_DOMAIN',
-    env: 'ENV',
-  },
```
jimlerza, line 22: I think we're checking `efcmsDomain` in here because the `environment` object in app context uses the domain to construct the documents bucket name (and we definitely need that). The dynamo env var can go though.

web-api/terraform/modules/everything-else-deprecated/documents-bucket.tf
jimlerza, line 1: Love live glue

web-api/terraform/modules/everything-else-deprecated/main.tf
jimlerza, line 1: Do we need to delete all data before this can run?

jimlerza (changes requested):

web-api/terraform/modules/everything-else-deprecated/main.tf
codyseibert: no, this seems to have deleted the dynamo tables without manually clearing them.m

.circleci/config.yml
codyseibert: no functional changes other than maybe refactoring a script to read from postgres instead of dynamo for testing

scripts/circleci/setup-ssm-for-back-migration-job.sh
codyseibert: I may defer that refactor as this PR is already getting pretty big, but happy to do it now if you think I should

jimlerza (approved):

scripts/circleci/setup-ssm-for-back-migration-job.sh
jimlerza: It can definitely wait!

web-api/terraform/modules/everything-else-deprecated/main.tf
jimlerza: Excellent!

TomElliottFlexion (approved):

Issue #9244, opened 2025-10-07
New Petition Fields


cholly75:
As a Court, so that we can understand any potential language barriers with a petitioner, I need new questions added to the petition e-filing/QC flow.

Many times a petitioner's primary language is not English. The Court would like to capture this information up front on the petition to be aware of any language/communication issues that may arise in the adjudication of the case. This will require us to add new questions to the petitions e-filing upload flow as well as track this information at the party level on the case.

**NOTE: We need to coordinate the deployment of this story with #9387 **

## Acceptance Criteria
*** PETITION E-FILING FLOW***

* New **optional field** added for **each** petitioner on Step 1 of the Petition e-filing flow if "Myself" or "Myself and my spouse" is selected: 
  - "If you have difficulty communicating in English, please state your preferred language:"
    - Free text field (max. 20 chars)
  - NOTE: if the petitioner has indicated that their spouse is deceased, do not display this question for the spouse petitioner party.
* New **optional field** added for **each** petitioner on Step 1 of the Petition e-filing flow if "Myself" or "Myself and my spouse" is selected: 
  - "If you are deaf or hard of hearing, please state your preferred method of communication (ASL, TTY, etc.):"
    - Free text field (max. 20 chars)
  - NOTE: if the petitioner has indicated that their spouse is deceased, do not display this question for the spouse petitioner party.
* Responses to both of the above questions are captured in the generated petition PDF form (if using the generator)
* Responses to both of the above questions are displayed on step 6 of the petition e-filing flow

## Notes
* Do we need to add these questions/displays in for other petitioner types? @cholly75 to clarify w/ Rules Cmte.
* May need to include slightly different verbiage in Practitioner e-filing flow

## Test Cases
Test Cases are in Test Rail and also on this spreadsheet. 

[9244.xlsx](https://github.com/user-attachments/files/23606554/9244.xlsx)

## Definition of Done (Updated 01-16-24)
**Product Owner**
 - [ ] Acceptance criteria have been met and validated on the Court's test environment
 - [ ] Add scenario to testing document, if applicable (https://docs.google.com/spreadsheets/d/1FUHKC_YrT-PosaWD5gRVmsDzI1HS_U-8CyMIb-qX9EA/edit?usp=sharing)

**Engineering**
 - [ ] Automated test scripts have been written, including visual tests for newly added PDFs.
 - [ ] Field level and page level validation errors (front-end and server-side) integrated and functioning.
[...]

PR #9428, opened 2025-11-26
9244 new petition fields

AnthonyLoeraTaxCourt:
https://github.com/orgs/ustaxcourt/projects/9/views/1?pane=issue&itemId=132552060&issue=ustaxcourt%7Cef-cms%7C9244

Ticket #9387: Replaced the amended petition form PDF appended to auto-generated Order for Amended Petition (OAP) and Order for Amended Petition and Filing Fee (OAPF). Updated terrafrom to use etag to update file in s3 to keep the name of the file but change the contents will trigger the update.

Ticket #9244: Added two optional fields to Step 1 of the petition e-filing flow:
- Preferred language (max 20 characters) and Preferred method of communication (max 20 characters)
Fields appear conditionally:
- Shown for "Myself" and "Myself and my spouse" filing types
- Hidden for deceased spouse
- Different labels for petitioners vs practitioners
- Fields included in generated petition PDF


cypress/local-only/tests/integration/fileAPetitionUpdated/file-a-petition-step-1-myself-and-my-spouse.cy.ts
```diff
@@ -503,7 +545,7 @@ describe('File a petition: Step 1 - Petitioner Information', () => {
       cy.get('[data-testid="filing-type-1"').click();
     });
 
-    describe('Petitioner and spouse', () => {
+      describe('Petitioner and spouse', () => {
```
jimlerza, line 548: Was this accidental?

jimlerza (changes requested):

akuny (approved):

paulo-filho-ctr (approved):

web-client/src/views/StartCase/ContactSecondaryUpdated.tsx
```diff
@@ -59,9 +67,19 @@ export const ContactSecondaryUpdated = connect<
...
         phone: string;
         paperPetitionEmail: string;
+        preferredLanguage: string;
+        preferredCommunicationMethod: string;
       };
     },
   }) {
+    const preferredLanguageLabel = isPetitioner
+      ? "If spouse has difficulty communicating in English, please state spouse's preferred language:"
+      : "If petitioner spouse has difficulty communicating in English, please state the petitioner spouse's preferred language:";
```
TomElliottFlexion, line 77: Just to be sure, is this right? I would imagine it'd be something like "If the petitioner's spouse..."

TomElliottFlexion (approved):

jimlerza (approved):

web-client/src/views/StartCase/ContactSecondaryUpdated.tsx
jimlerza: Evelyn checked with Chris Holly and this is the official text that we're going with, intentionally. Legalese is odd.

PR #9430, opened 2025-12-01
fix: pdf generation of notice of change of trial judge small tax

paulo-filho-ctr:


shared/src/business/utilities/documentGenerators/noticeOfChangeOfTrialJudge.test.ts
```diff
@@ -30,7 +28,7 @@ describe('noticeOfChangeOfTrialJudge', () => {
             trialLocation: 'Mobile, Alabama',
             trialLocationAndProceedingType: `Mobile, Alabama, ${TRIAL_SESSION_PROCEEDING_TYPES.inPerson}`,
             updatedJudgeTitleWithFullName: 'Chief Judge Lady Macbeth',
-          },
+          } as any,
```
TomElliottFlexion, line 31: I believe this should remain a complete `FormattedTrialInfoType`, just with a `caseProcedureType` that is not `SMALL`

shared/src/business/utilities/documentGenerators/noticeOfChangeOfTrialJudgeSmallTax.test.ts
```diff
@@ -0,0 +1,39 @@
...
+            chambersPhoneNumber: '1-721-740-9885 x4239',
+            docketNumber: '999-99',
+            formattedStartDate: '01/01/2001',
+            priorJudgeTitleWithFullName: 'Special Trial Judge Judifer Judy',
+            proceedingType: TRIAL_SESSION_PROCEEDING_TYPES.inPerson,
+            startDate: '2019-08-25T05:00:00.000Z',
+            trialLocation: 'Mobile, Alabama',
+            trialLocationAndProceedingType: `Mobile, Alabama, ${TRIAL_SESSION_PROCEEDING_TYPES.inPerson}`,
+            updatedJudgeTitleWithFullName: 'Chief Judge Lady Macbeth',
+          } as any,
```
TomElliottFlexion, line 33: Does this need to be an `any`? It doesn't look like `docketNumber` is needed

TomElliottFlexion: The more I look at it, the more it seems like `FormattedTrialInfoType` is just plain wrong. I realize this is just fixing broken pdf tests, but we should probably type `noticeOfChangeOfTrialJudge` more accurately

shared/src/business/entities/trialSessions/TrialSession.ts
```diff
@@ -110,6 +110,7 @@ export class TrialSession extends JoiValidationEntity {
   public trialLocation?: string;
   public trialSessionId: string;
   public paperServicePdfs: { fileId: string; title: string }[];
+  public caseProcedureType?: string;
```
TomElliottFlexion: This is a tough one: I see why you'd want to add a field to the entity but not the constructor, but it seems a little dangerous. How would you feel about either a). making it optional in the constructor so that it's set if it _is_ passed in or b). adding a `getCaseProcedureForTrial` call to the constructor?

TomElliottFlexion (approved):
Sorry for all the comments on a simple test-fix pr! Appreciate you fixing the problem

shared/src/business/utilities/documentGenerators/noticeOfChangeOfTrialJudge.test.ts
paulo-filho-ctr: It actually it needs a more flexible type, otherwise many type errors appear, due to incompatibility. If you pull the branch and run it, it's a good way to see the issue. It is not as simple as augmenting the type.

shared/src/business/utilities/documentGenerators/noticeOfChangeOfTrialJudgeSmallTax.test.ts
paulo-filho-ctr: It actually it needs a more flexible type, otherwise many type errors appear, due to incompatibility. If you pull the branch and run it, it's a good way to see the issue. It is not as simple as augmenting the type.

AnthonyLoeraTaxCourt (approved):

shared/src/business/utilities/documentGenerators/noticeOfChangeOfTrialJudgeSmallTax.test.ts
jimlerza: But does it have to be `any`?

I think Tom is right - the fact that `FormattedTrialInfoType` inherits from `RawTrialSession` is actually what is causing most of the type errors that you mentioned. This feels like a "leave it better than you found it" opportunity for us to fix this to inherit from something that has the properties that we expect.

Asserting the type `as any` doesn't fix the type, it's basically like throwing your hands up.

PR #9440, opened 2025-12-02
9244 new petition fields to test 1764703432

AnthonyLoeraTaxCourt:
https://github.com/orgs/ustaxcourt/projects/9/views/1?pane=issue&itemId=132552060&issue=ustaxcourt%7Cef-cms%7C9244

Ticket #9387: Replaced the amended petition form PDF appended to auto-generated Order for Amended Petition (OAP) and Order for Amended Petition and Filing Fee (OAPF)

Ticket #9244: Added two optional fields to Step 1 of the petition e-filing flow:
- Preferred language (max 20 characters) and Preferred method of communication (max 20 characters)
Fields appear conditionally:
- Shown for "Myself" and "Myself and my spouse" filing types
- Hidden for deceased spouse
- Different labels for petitioners vs practitioners
- Fields included in generated petition PDF


cypress/local-only/tests/accessibility/petitionGeneration/petition-generation-step-1.cy.ts
```diff
@@ -18,6 +18,21 @@ describe('Petition generation - step 1', () => {
       cy.get('[data-testid="international-country-btn"]').click();
       checkA11y();
     });
+
+    it('Myself - with language fields filled: should be free of a11y issues', () => {
```
paulo-filho-ctr: Since we have `Myself` in the describe block , the `it` case can omit the `Myself` prefix from each description. The usual practice is to have the `it` clause start with `should` so `it('should...)`.

paulo-filho-ctr (approved):

jdr81394 (approved):

Issue #8932, opened 2025-03-17
Use real user logins in automated tests


TomElliottFlexion:
## Pre-Conditions
- #8922 

## Acceptance Criteria
Automated tests run on `test` using real users created in #8922

## Pain Avoided/Frustration Saved
Would help catch edge cases that only appear in data associated with real users in prod

ttlenard:
Testing Feedback

@kampofo 

- [x]  Can the Judge logins go back to NOT having a 1 in the email? This is preferred, if possible. Thanks!

kampofo:
> Testing Feedback
> 
> [@kampofo](https://github.com/kampofo)
> 
> * [x]   Can the Judge logins go back to NOT having a 1 in the email? This is preferred, if possible. Thanks!

The setup glued users script has been updated to drop the 1 in the email. And these changes will be reflected on the users emails the next time the glue job is run and the cognito users are recreated

PR #9449, opened 2025-12-03
8932 use real users in automated tests to test 1764795208

kampofo:
https://github.com/ustaxcourt/ef-cms/issues/8932

- Updated setup glued users script to use pg email instead of custom email
- Add script to clear all users in cognito
- Move the new script and setup glued users script in circle CI to within the glue pipeline


.circleci/config.yml
```diff
@@ -181,10 +181,6 @@ jobs:
             if [ "${CIRCLE_BRANCH}" == "prod" ]; then
               echo "skipping…"
               exit 0
-            fi
-            if [ "${CIRCLE_BRANCH}" == "test" ]; then
-              export REGION="us-east-1"
-              ./scripts/user/setup-glued-users.ts
```
jimlerza, line 187:
```suggestion
            fi
            if [ "${CIRCLE_BRANCH}" != "test" ]; then
```

jimlerza: The logic that you have committed results in the `bulkImportJudgeUsers.ts` script running on test where it wasn't previously. We don't want it to run on test.

scripts/user/setup-glued-users.ts
```diff
@@ -234,34 +234,37 @@ const getUsers = async (): Promise<Users> => {
...
-
-    const fullName = user.name;
-    user.name = user.name.replace(/\s+/g, '-').replace(/[^a-zA-Z0-9-_]/g, '');
 
-    if (user.role in ['judge', 'legacyJudge']) {
+    if (['judge', 'legacyJudge'].includes(user.role)) {
       users[user.name] = {
-        email: `${
-          user.judgeTitle!.indexOf('Special Trial') !== -1 ? 'st' : ''
-        }judge.${user.name.toLowerCase()}@example.com`,
```
jimlerza, line 254: Let's not lose this email address formatting.

Regular Judges and Senior Judges are all `judge.lastname@example.com` but Special Trial Judges are `stjudge.lastname@example.com`.

jimlerza (changes requested):

PR #9464, opened 2025-12-05
8932 use real users in automated tests to test 1764964039

kampofo:


.circleci/config.yml
```diff
@@ -729,6 +730,13 @@ jobs:
             if [ "$MIGRATE_FLAG" == "true" ]; then
               NODE_ENV=production ./scripts/reindex/index-docket-entries/index-docket-entries.ts 5
             fi
+      - run:
+          name: Populate test users in Cognito
+          command: |
+            if [ "${CIRCLE_BRANCH}" == "test" ]; then
+                  export REGION="us-east-1"
+                  ./scripts/user/setup-glued-users.ts
+            fi
```
jimlerza: We only run glue jobs to test currently, but we've run glue jobs to environments other than test before.

```suggestion
            export REGION="us-east-1"
            ./scripts/user/setup-glued-users.ts
```

.circleci/config.yml
```diff
@@ -544,6 +541,10 @@ jobs:
           name: 'Delete OpenSearch alpha/beta clusters'
           command: |
             ./scripts/elasticsearch/delete-alpha-and-beta-clusters.sh
+      - run:
+          name: 'Delete Cognito users in test'
```
jimlerza, line 545:
```suggestion
          name: 'Delete Cognito users'
```

jimlerza (approved):

devinganey613 (approved):

AnthonyLoeraTaxCourt (approved):

scripts/user/clear-all-cognito-users.ts
```diff
@@ -0,0 +1,100 @@
...
+  } while (paginationToken); // Continue as long as there's a pagination token
+
+  return allUsers;
+};
+
+// eslint-disable-next-line @typescript-eslint/no-floating-promises
+(async () => {
+  const allUsers: UserType[] = await getAllCognitoUsers();
+
+  const taskFns: (() => Promise<void>)[] = allUsers.map(user => {
```
TomElliottFlexion, line 62: Out of curiosity only, was there any consideration given to just deleting the user pool? I can think of half a dozen reasons that might cause problems, but in the interest of learning, why not?

TomElliottFlexion (approved):

TomElliottFlexion (approved):
Alt Text

PR #9465, opened 2025-12-05
Dependencies 12 01 2025

devinganey613:
- Updated package.json files with latest versions of packages, including changing serverless-express from @vandia/serverless-express to @codegenie/serverless-express, and upgrading @types/node to version 24
- Updated aws-cli in dockerfile to 2.32.9
- Updated terraform AWS provider to 6.24.0
- Incremented ECS docker image to 4.3.60
- Upgraded to node 24. This required updating how the labmda handler in api-public.ts is setup, instead of using awsServerlessExpress directly, it is called with just the event context arguments and returned, similar to how it is done in api.ts. This is because the node 24 runtime in AWS no longer supports including the callback argument in the handler call, so we now call it with just event and context. 

### Node upgrade steps for local environment

- [ ] Use `nvm` to install the version of NodeJS specified in `.nvmrc` (NodeJS 24):
   ```bash
   nvm install
   ```
- [ ] Set your current shell session to use NodeJS 24:
   ```bash
   nvm use
   ```
- [ ] Set NodeJS 24 as the default node version in future shell sessions:
   ```bash
   nvm alias default "$(cat .nvmrc)"
   ```


jimlerza:
Note: the Typescript File Check and Reduce Type Errors actions always fail when we do a NodeJS upgrade.

jimlerza (approved):

PR #9474, opened 2025-12-09
8677: add post merge hook

paulo-filho-ctr:
## Summary

This PR Implements a `post-merge` git hook that alerts developers when `CHANGES.md` is updated after a merge, to address the issue where manual steps can be missed.

## Proposed Changes

- **Post-merge hook** (`.husky/post-merge`):
  - Detects changes to `CHANGES.md` after merges
  - Opens IntelliJ IDEA or; WebStorm or; VS Code automatically

## How it works

After merging, if `CHANGES.md` changed:
1. Terminal shows a formatted, colored diff
2. IntelliJ IDEA or; WebStorm or; VS Code opens a side-by-side diff view

This ensures manual steps are noticed immediately after merges.


.husky/post-merge
```diff
@@ -0,0 +1,67 @@
...
+      echo "⚠️  CHANGES.md has been updated. Please review the manual steps below:"
+      echo "━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━"
+      echo ""
+      echo "$CHANGES_DIFF" | grep -E "^[\+\-]" | grep -v "^[\+\-][\+\-][\+\-]" | grep -v "No newline" | while IFS= read -r line; do
+        clean_line=$(echo "$line" | sed 's/^[+-]//')
+        case "$line" in
+        +*) printf "  \033[32m%s\033[0m\n" "$clean_line" ;;
+        -*) printf "  \033[31m%s\033[0m\n" "$clean_line" ;;
+        esac
+      done
```
jimlerza: Does this output lines that were removed? I think this has the potential to be confusing, especially if we're chopping off the character that indicates whether a line was added or removed. If we're going to be removing that leading character, then we should probably only output the lines that were added and not output the lines that were removed.

jimlerza (changes requested):

.husky/post-merge
paulo-filho-ctr: Yes it does output the lines that were removed. For sure, I can see how it can be confusing. I will update to show only display new lines.

jimlerza (approved):

Issue #8994, opened 2025-06-25
Practitioners: Auto-Generate Notice of Withdrawal as Counsel


cholly75:
As a practitioner, so that I can easily withdraw from a case I'm associated with and prevent me from misfiling an incorrect document type, I need to be able to automatically generate a Notice of Withdrawal as Counsel.

When practitioners withdraw from a case, they may file a Notice of Withdrawal as Counsel if certain conditions are met. If these conditions are NOT met, they must instead file a Motion to Withdraw as Counsel.

In order to better enforce the rules around when practitioners may file a Notice as opposed to a Motion, we would like to implement an optional workflow that can be selected when "Notice of Withdrawal as Counsel" is selected as the document type to be filed.

This workflow/wizard option will trigger for any Practitioner (Private or IRS) upon selection of the Notice of Withdrawal as Counsel document type in the e-filing workflow, provided the case and parties meet the listed criteria below.

## Acceptance Criteria
* Upon selecting Notice of Withdrawal as Counsel and clicking Continue, if the logged in user is a Practitioner then the following case data is checked:
  - There is more than one existing counsel of record associated with the same party the logged in user is associated with (Private)
  - There is more than one existing counsel of record associated with the case as Respondent counsel (IRS)
  - The case being filed in is NOT in "Calendared" status OR the case being filed in is in Calendared status but the start date of the trial session is > 30 days from today

 * If **any** of these conditions are not met, then the user is presented with validation informing them that they must file a Motion to Withdraw as Counsel instead, along with the condition(s) that was not met.
* If **all** of these conditions are met, the case is eligible for filing the NOWAC, and the next step in the workflow will now require affirmative user input to the following items individually:
  - (1) Are you removing yourself as counsel of record for [Party/s Document Is Being Filed On Behalf Of]?
  - (2) Have all parties consented to your withdrawing as counsel?
* If the logged in user is a Private practitioner associated with one or more petitioners, the workflow will also ask for optional user input to the following item **for each petitioner** the logged in user is associated with as counsel:
  - (3) Is petitioner's contact information in the record of the case current? (Display petitioner contact information alongside this question)
    - **NOTE**: if all petitioners displayed here have **sealed contact info**, do NOT display this item at all.
* If any party on the case currently receives paper service, the workflow will also require affirmative user input to the following item:
[...]

PR #9476, opened 2025-12-09
8994 autogen notow counsel to test 1765303033

alexdziarn:
Changes:

**Document Type Selection**
- Added validation to handle Notice of Withdrawal document type

**Primary Document Filing Form** (for Practitioners)
- Added new form fields and validation
- Private practitioners see all petitioners they represent
- IRS practitioners see respondent
- Added option to file manually or use auto generated document
- Disabled consolidated group filing option

**File Document Review** (for Practitioners)
- Displays list of parties being withdrawn from
- Auto-generates Notice of Withdrawal PDF when selected
- Auto-generates Certificate of Service PDF on cases with parties that have paper service

**Docket Entry Display**
- "Filed By" column now shows the filer instead of the represented party
- "Filing and Proceedings" column now appends `(C/S MM/DD/YY)` timestamp if a Certificate of Service was generated


shared/src/business/entities/EntityConstants.ts
alexdziarn, line 1: should remove since I use generateFiledBy to override the filedBy section

PR #9479, opened 2025-12-09
9009: Transactions (formerly flexion#10705) to test 

codyseibert:

### Details:

2. Note that `getLockingConnection` is kept as is. Technically, this is also a pool; however, it is guaranteed to have only 1 client, so the name is a white lie. I thought originally, "We can just check out a connection and use that." Nope. This is hard.

Implementation details:


codyseibert:
this was successfully deployed to exp8 without issue https://app.circleci.com/pipelines/github/ustaxcourt/ef-cms/16910/workflows/cfa2326f-9e4a-4d4a-89bf-ccced17d69cd

Issue #8655, opened 2024-03-15
Allow Completion of User Verification Workflows Without Having to be Logged In


cholly75:
As a DAWSON user, so that I can easily complete my change email request, I need to be able to complete the verification workflow without having to still be logged into DAWSON.

One of the pain points associated with the email change workflow is that a user must complete the verification flow on the same device where they are already logged into DAWSON. This results in many support tickets with users confused and/or unable to complete the flow.

We should fix this.

## Acceptance Criteria
* User can complete change email flow including verification without having to be already logged into DAWSON on the same device as that which is being used to complete the verification.

## Notes
* Security concerns?
Steps to recreate the pain points:

#### Pain Point 1 - User is not logged in when they attempt to verify their new email.
- As a Petitioner, create a DAWSON account (or you can also be an admissions clerk and create a Practitioner account for a user)
- Once the account is created, you will receive a message that you need to verify your email
- Navigate to your email and then click on the Verify email button
- You will receive a message indicating that your email is verified and you can now log in
- Log in as the Petitioner
- Click on the person icon in the upper right corner of the page and then select My Account
- Click on the change email link
- Type in a new email (one that you still have access to) and then click save
- There will be a yellow banner at the top of the page that tells the user to verify the new email (nothing about requirements that they have to stay logged in, or be on the same browser, or on the same device)
- Next, log out of DAWSON
- Navigate to your email
- Click on the verify email button in the "Verify your new email" message that you received (this email isn't very helpful to the user, it doesn't tell them that they must be logged in, or using the same browser, and that they must be on the same device).  
- New tab opens to the DAWSON login page. It will flash blue (indicating that it is updating contact information), but then a red error message displays: "Unable to complete your request" (This error message isn't very helpful to the user.  It doesn't tell them that they must be logged in, or using the same browser, and that they must be on the same device).  

#### Pain Point 2 - User is logged in to DAWSON on browser A, but they click Verify Email button from their email in browser B.

- As a Petitioner, create a DAWSON account (or you can also be an admissions clerk and create a Practitioner account for a user)
- Once the account is created, you will receive a message that you need to verify your email
- Navigate to your email and then click on the Verify email button
- You will receive a message indicating that your email is verified and you can now log in
- Log in as the Petitioner
- Click on the person icon in the upper right corner of the page and then select My Account
- Click on the change email link
- Type in a new email (one that you still have access to) and then click save
[...]

PR #9484, opened 2025-12-09
8655 email verification workflow

akuny:
https://github.com/ustaxcourt/ef-cms/issues/8655

See:
- https://github.com/ustaxcourt/ef-cms/pull/9501
- https://github.com/ustaxcourt/ef-cms/pull/9532


jimlerza (approved):

web-api/src/app.ts
```diff
@@ -1071,7 +1070,6 @@ app.delete(
     lambdaWrapper(getUserPendingEmailStatusLambda),
   );
   app.put('/users/pending-email', lambdaWrapper(updateUserPendingEmailLambda));
-  app.put('/users/verify-email', lambdaWrapper(verifyUserPendingEmailLambda));
```
jimlerza, line 1097: Should we set up a proxy or redirect?

If we don't, doesn't this mean that any pending password reset or welcome emails (that are otherwise still valid) will fail?

akuny: Following up on our discussion this morning, the PUT route for verify-email is now in both app.ts and app-public.ts--both routes point to the same lambda.

devinganey613 (approved):

Issue #8782, opened 2024-08-06
TS Error Cleanup: Incompatible Types Assigned


akuny:
**TS2322: `Type '<type>' is not assignable to type '<some type>'.`**

- [x] scripts/judge/update-judge-titles.ts
- [x] scripts/user/add-user.ts
- [x] scripts/user/create-and-enable-smoketest-user-helpers.ts
- [x] scripts/user/setup-test-users.ts
- [x] shared/src/business/entities/cases/Case.ts
- [x] shared/src/business/entities/contacts/ContactFactory.test.ts
- [x] shared/src/business/entities/Message.ts
- [x] shared/src/business/useCases/externalDocument/uploadExternalDocumentsInteractor.test.ts
- [x] shared/src/business/useCases/generateSignedDocumentInteractor.ts
- [x] shared/src/business/useCases/getNotificationsInteractor.test.ts
- [x] shared/src/business/useCases/getPaperServicePdfUrlInteractor.ts
- [x] shared/src/business/useCases/getReconciliationReportInteractor.test.ts
- [x] shared/src/business/useCases/updateCaseContextInteractor.ts
- [x] shared/src/business/useCases/validateHearingNoteInteractor.test.ts
- [ ] ~~shared/src/business/useCases/validatePetitionerInformationFormInteractor.ts~~ file was removed
- [x] shared/src/business/useCases/validatePetitionerInteractor.ts
- [x] shared/src/business/utilities/DateHandler.ts
- [x] shared/src/business/utilities/documentGenerators/noticeOfChangeToRemoteProceeding.test.ts
- [x] shared/src/business/utilities/documentGenerators/noticeOfTrialIssued.test.ts
- [x] shared/src/business/utilities/documentGenerators/noticeOfTrialIssuedInPerson.test.ts
- [x] shared/src/business/utilities/getCaseDocumentsIdsFilteredByDocumentType.ts
- [x] shared/src/business/utilities/getFormattedCaseDetail.test.ts
- [x] shared/src/business/utilities/getFormattedTrialSessionDetails.test.ts
- [x] shared/src/business/utilities/getFormattedTrialSessionDetails.ts
- [x] shared/src/business/utilities/pdfGenerator/documentTemplates/DocketRecord.tsx
- [x] web-api/src/business/useCaseHelper/addDocketEntryForSystemGeneratedOrder.ts
[...]

PR #9498, opened 2025-12-10
8994 autogen notow counsel to test 1765395238

alexdziarn:
Changes:

### Document Type Selection
- Added validation to handle Notice of Withdrawal document type
- Primary Document Filing Form (for Practitioners)
### Added new form fields and validation
- Private practitioners see all petitioners they are able to be removed from
- IRS practitioners see respondent
- Added option to file manually or use auto generated document
- Disabled consolidated group filing option for notice of withdrawals
### File Document Review (for Practitioners)
- Displays list of parties being withdrawn from
- Auto-generates Notice of Withdrawal PDF when selected
- Auto-generates Certificate of Service PDF on cases with parties that have paper service
### Docket Entry Display
- "Filed By" column now shows the filer instead of the represented party
- "Filing and Proceedings" column now appends (C/S MM/DD/YY) timestamp if a Certificate of Service was generated
### Typing
- ~~edited the **shared.cerebral.ts** file to allow components to accept additional props like "bind". Edited **Country.tsx** to use the props function instead of manually casting types onto props. If there is no issue with this implementation, I would like to implement this on the rest of the components.~~ being fixed in https://github.com/ustaxcourt/ef-cms/issues/8782


paulo-filho-ctr (approved):

web-client/src/views/StartCase/Country.tsx
devinganey613, line 1: Glad someone else landed on this fix too!

devinganey613 (approved):

devinganey613 (approved):

Issue #8780, opened 2024-08-06
TS Error Cleanup: Missing Type Imports


akuny:
**TS2304: `Cannot find name '<something>'.`**
- [x] web-api/src/persistence/dynamo/practitioners/getPractitionerDocuments.ts
- [x] web-client/src/presenter/computeds/paperServiceStatusHelper.test.ts
- [ ] ~~web-api/src/persistence/dynamo/practitioners/createPractitionerDocument.ts~~ file no longer exists
- [x] web-client/src/presenter/computeds/noticeStatusHelper.test.ts
- [ ] ~~web-api/src/persistence/dynamo/cases/createCase.ts~~
- [ ] ~~web-api/src/persistence/dynamo/practitioners/editPractitionerDocument.ts~~

**TS2552: `Cannot find name '<something>'. Did you mean '<something else>'?`**
- [x] web-api/src/business/useCases/workItems/completeWorkItemInteractor.ts
- [x] shared/src/authorization/getUserPermissions.ts
- [x] web-api/src/business/useCaseHelper/generateChangeOfAddressHelper.ts

devinganey613:
Test steps:

Test as both a user that can download a docket entry and can't download a docket entry to ensure permissions are still functioning correctly.

PR #9499, opened 2025-12-10
8782 ts error cleanup incompatible types assigned to test 1765395799

devinganey613:
Fixed many typescript errors in files noted in #8782 
Also includes one change for #8780


jdr81394:
hey @devinganey613 some of your tests are failing because they are using the date of 12/12/2025 when they must be in the future, for examle this is the case with trial-session-eligible-cases.cy.ts. I have the same issue with one of my PRs and saw it on yours. Hope this helps :-)

devinganey613:
> hey @devinganey613 some of your tests are failing because they are using the date of 12/12/2025 when they must be in the future, for examle this is the case with trial-session-eligible-cases.cy.ts. I have the same issue with one of my PRs and saw it on yours. Hope this helps :-)

Oh good catch, I'll update those. Thanks!

AnthonyLoeraTaxCourt (approved):

web-client/src/views/CaseDetail/PetitionPaymentForm.tsx
```diff
@@ -24,11 +24,7 @@ interface CerebralProps {
     petitionPaymentMethod?: string;
     petitionPaymentWaivedDate?: string;
   };
-  formatAndUpdateDateFromDatePickerSequence: (args: {
-    key: string;
-    toFormat: string;
-    value: string;
-  }) => void;
+  formatAndUpdateDateFromDatePickerSequence: Function;
```
zacharybrown204, line 27: Do we want to replace all of the function types with this function interface? If we just call it a function we're losing information about the args or return value of the function.

zacharybrown204: Oh wait this was for a sequence and in presenter.ts we define all the sequences as functions. So this matches.

zacharybrown204 (approved):

PR #9502, opened 2025-12-10
8546 connect order to motion revamped to test 1765402955

zacharybrown204:
Addressing feedback from demo to court users.

- Changed wording on checkbox when docketing an order.
- Changed verbiage of dispositions on the case details page.
- Reordered motions when docketing order so that last motion is first in list.
- - Used lodash flow to chain filter -> orderBy -> map. Could probably go without it but I wanted to chain them all together and didn't see a simpler method

### DIFF
--- a/web-client/src/presenter/computeds/addCourtIssuedDocketEntryHelper.ts
+++ b/web-client/src/presenter/computeds/addCourtIssuedDocketEntryHelper.ts
@@ -40,9 +40,9 @@ export const addCourtIssuedDocketEntryHelper = (
   );
 
   const caseMotions = flow([
-    (a: RawDocketEntry[]) =>
+    (docketEntries: RawDocketEntry[]) =>
       filter(
-        a,
+        docketEntries,
         d =>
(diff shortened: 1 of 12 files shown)


web-client/src/presenter/computeds/addCourtIssuedDocketEntryHelper.ts
```diff
@@ -38,30 +38,36 @@ export const addCourtIssuedDocketEntryHelper = (
...
-          form.affectedDocketEntries ?? [],
-          am => am.docketEntryId === d.docketEntryId,
-        ),
-    )
-    .map((m: RawDocketEntry) => ({
-      label: `${m.index} - ${m.documentTitle}`,
-      value: m.docketEntryId,
-    }));
+
+  const caseMotions = flow([
```
TomElliottFlexion, line 42: COOL

TomElliottFlexion (approved):

jimlerza (approved):

TomElliottFlexion (approved):

PR #9514, opened 2025-12-12
Future date test fix

jdr81394:
Bug fix for failling tests that arose due to several tests using today's date of 12/12/2025 when they needed to be using dates in the future.


shared/src/business/utilities/DateHandler.ts
```diff
@@ -781,3 +781,7 @@ export const roundDateDownToNearestHour = (isoDateString: string) => {
   formattedDate.setMilliseconds(0);
   return formattedDate;
 };
+
+export const getCurrentDateTimeInMillis = (): number => {
+  return DateTime.now().toMillis();
+};
```
jimlerza: We use `Number(formatNow(FORMATS.UNIX_TIMESTAMP_MS))` to do this elsewhere in the app.

```suggestion
export const getCurrentDateTimeInMillis = (): number => {
  return Number(formatNow(FORMATS.UNIX_TIMESTAMP_MS));
};

```

jimlerza (approved):

alexdziarn (approved):

paulo-filho-ctr (approved):

jimlerza (approved):

PR #9529, opened 2025-12-16
Fix Dependency Issue 12-16

devinganey613:
Added a version override for the command package used by s3rver to prevent an issue where the test server was not starting up.


paulo-filho-ctr (approved):

jimlerza:
You'll need to generate a new `package-lock.json` and push that up as well!

devinganey613:
> You'll need to generate a new `package-lock.json` and push that up as well!

For some reason, on the staging branch the command package is still at version 12 even though the s3rver package was updated. This package seems to have been updated to version 14 in later versions of package-lock.json, such as on the one on the test branch. So running npm install with these changes against this branch does not update package-lock.

jimlerza (approved):

PR #9532, opened 2025-12-17
8655 to Test - Part 2

akuny:
https://github.com/ustaxcourt/ef-cms/issues/8655

This PR moves the email verification-related endpoints to the public API from the app's internal API.


web-client/src/routerPublic.ts
```diff
@@ -97,6 +97,16 @@ const router = {
       });
     });
 
+    route('/verify-email..', () => {
+      setPageTitle('Verify Email');
+      const { token } = route.query();
+
+      console.log('Token Grabbed from URL!: ', token);
```
TomElliottFlexion, line 104: Nit: console.log should be removed prior to staging

web-client/integration-tests/modifyPractitionerEmail.test.ts
TomElliottFlexion, line 1: Do these tests need replacements?

TomElliottFlexion (approved):

web-client/integration-tests/modifyPractitionerEmail.test.ts
akuny: I cross-referenced them with some existing Cypress tests and my take is the latter cover these scenarios well enough, but I'm open to differing opinions. It would be challenging to test this feature in the legacy Cerebral style because we'd be dealing with both the app sequences and the public sequences (I don't think there's precedent for this).

TomElliottFlexion: I'll take your word for it! The replacements would definitely need to be Cypress tests, but if it's already well covered then we're good

devinganey613 (approved):

jimlerza (approved):

web-client/src/routerPublic.ts
jimlerza: ➕

web-client/integration-tests/modifyPractitionerEmail.test.ts
jimlerza: If the logic previously covered by these removed tests is only now covered by the newly-added "real user" tests, we will need to add those "real user" tests to one of the suites that runs in github actions. Currently the "real user" tests are in a separate, optional workflow that doesn't run automatically.

PR #9534, opened 2025-12-17
opex: outstanding deficiency report

jimlerza:
Added a report that generates statistics of outstanding IRS deficiencies among open cases, aggregated by preferred trial location.


paulo-filho-ctr (approved):

kampofo (approved):
Looks Good!

PR #9537, opened 2025-12-18
8546 connect order to motion revamped to test 1766077318

zacharybrown204:
#8546 
Added the relationship text between docket entries to the docket entries on the printed pdf of a docket record.

Bug & Error fixes
- Fixed a bug where sealing a docket entry will not load relationship information. This was because interactor that seals a case was using a different persistence function to load a docket entry that wasn't getting the relationship data. This persistence function now gets that data.
- On the docket record screen there was an empty space in the table that react was throwing an error about. 
- On the docket entry screen the docket relationships were being mapped without a key and react threw an error.


paulo-filho-ctr (approved):

web-api/src/business/useCases/generateDocketRecordPdfInteractor.ts
```diff
@@ -80,10 +82,49 @@ export const generateDocketRecordPdfInteractor = async (
...
+            affectedEntry.disposition =
+              MOTION_DISPOSITION_VERBIAGE[affectedEntry.disposition].MOTION;
+
+            return affectedEntry;
+          }),
+        );
+      }
+
+      if (docketEntry.affectedDocketEntries) {
+        formattedDocketEntry.relatedDocketEntries = concat(
```
devinganey613, line 104: Could this code be put in a function to be reused?

web-api/src/business/useCases/generateDocketRecordPdfInteractor.ts
```diff
@@ -80,10 +82,49 @@ export const generateDocketRecordPdfInteractor = async (
...
 
   formattedCaseDetail.formattedDocketEntries =
-    formattedCaseDetail.formattedDocketEntries.map(docketEntry => ({
-      ...docketEntry,
-      numberOfPages: docketEntry.numberOfPages || 0,
-    }));
+    formattedCaseDetail.formattedDocketEntries.map(docketEntry => {
+      const formattedDocketEntry = {
+        ...docketEntry,
+        numberOfPages: docketEntry.numberOfPages || 0,
```
TomElliottFlexion, line 91: Not a big deal, but do you think it would be worth adding this `|| 0` fallback to  `getFormattedCaseDetail`?

TomElliottFlexion (approved):

web-api/src/persistence/postgres/docketEntries/getDocketEntriesByDocketNumberAndDocketEntryId.ts
```diff
@@ -11,18 +12,82 @@ export const getDocketEntriesByDocketNumberAndDocketEntryId = async ({
 }): Promise<RawDocketEntry[]> => {
   const dbDocketEntries = await getDbReader(reader =>
     reader
-      .selectFrom('dwDocketEntry')
+      .with('affectedDocketEntries', db =>
```
devinganey613: This is also sort of a complicated query being reused, should it be in one place? (The with statements)

web-api/src/business/useCases/generateDocketRecordPdfInteractor.ts
nechamakrigsman: ➕ to this

zacharybrown204: Refactored out to a new function and tried to clean it up as much as I could.

web-api/src/persistence/postgres/docketEntries/getDocketEntriesByDocketNumberAndDocketEntryId.ts
zacharybrown204: Good shout, we're reusing this whole query in a few places now.
I couldn't break out just the parts relating to the with statements, so I refactored the entire query except for the where block to another function.
It did end up looking a bit awkward since even if you only make a partial query, it's still a async function. So to get the partial query, then finish and execute it I had to write: ` await (await docketEntriesBaseQuery()) `

devinganey613 (approved):

PR #9544, opened 2025-12-19
8782 ts error cleanup incompatible types assigned

devinganey613:
Fixed many typescript errors in files noted in #8782 
Also includes one change for #8780


scripts/ecr/docker-to-ecr.sh
```diff
@@ -32,7 +32,7 @@ if [[ -n $LOCAL_IMAGE_EXISTS ]]; then
 fi
 if [[ "$BUILD_IMAGE" -eq 1 ]]; then
   docker build --platform=linux/amd64 --no-cache -t "$LOCAL_IMAGE_NAME" -f Dockerfile .
-  if [[ $? -ne 0 ]]; then
+  if ! docker; then
```
jimlerza: Can you explain this?

The old code means "if the return code of the previous command was anything other than 0" and you've changed it to "if running the command `docker` with no parameters is not falsy."

From context clues, I'm guessing this was an attempt to resolve [shellcheck SC2181](https://github.com/koalaman/shellcheck/wiki/SC2181).

In that case, might I suggest combining lines 34 and 35 to this:

```bash
if ! docker build --platform=linux/amd64 --no-cache -t "$LOCAL_IMAGE_NAME" -f Dockerfile . ; then
```

This will satisfy the SC2181 error while retaining the previous functionality.

jimlerza (changes requested):

paulo-filho-ctr (approved):

scripts/ecr/docker-to-ecr.sh
devinganey613: Yes it was to fix the shellcheck error. I can make this change

akuny (approved):

jimlerza (approved):

PR #9508, opened 2025-12-11
8932 use real users in automated tests to test 1765466299

kampofo:
https://github.com/ustaxcourt/ef-cms/issues/8932

- Created real-users circle ci workflow that can be triggered manually in circle ci
- Added Cypress tests for a "real user" practitioner updating contact information
- Added new npm scripts  "cypress:real-users" and "cypress:real-users:open" to run real users tests locally 
- Created clear-all-cognito-users.ts script that runs in the glue work flow and deletes all cognito users in test env
- Setup glued user moved to glue workflow as well
- Added helper function to get practitioner with the most cases
- Updated setup-glued-users.ts to use generic names for users and the obfuscated email addresses directly from pg to make linking users to their data simpler
- Fixed small bug in setup-glued-users.ts checking if role was judge or legacy judge

PR #9555, opened 2025-12-28
8932 use real users in automated tests

jimlerza:
#8932

- Created real-users circle ci workflow that can be triggered manually in circle ci
- Added Cypress tests for a "real user" practitioner updating contact information
- Added new npm scripts  "cypress:real-users" and "cypress:real-users:open" to run real users tests locally 
- Created `clear-all-cognito-users.ts` script that runs in the glue workflow and deletes all cognito users
- Setup glued user moved to glue workflow as well
- Added helper function to get practitioner with the most cases
- Updated setup-glued-users.ts to use generic names for users and the obfuscated email addresses directly from pg to make linking users to their data simpler
- Fixed small bug in setup-glued-users.ts checking if role was judge or legacy judge


jimlerza:
See comments and approvals on #9464, #9508, and #9535 PRs

PR #9560, opened 2025-12-29
9457 end date trial session to test

codyseibert:


shared/src/business/entities/trialSessions/NewTrialSession.test.ts
```diff
@@ -51,4 +51,51 @@ describe('NewTrialSession entity', () => {
...
+      expect(trialSession.getFormattedValidationErrors()).toMatchObject({
+        estimatedEndDate: 'Enter a valid estimated end date',
+      });
+    });
+
+    it('should be invalid when estimatedEndDate is before startDate', () => {
+      const trialSession = new NewTrialSession({
+        ...MOCK_NEW_TRIAL_REMOTE,
+        estimatedEndDate: '2027-11-01T00:00:00.000Z',
+        startDate: '2027-11-11T00:00:00.000Z',
```
jdr81394: When it gets to be past the start date irl, it could be possible that it will throw an error for a valid start date. Solution is to have the dates further in the future.

jdr81394 (changes requested):

shared/src/business/entities/trialSessions/NewTrialSession.test.ts
codyseibert: good catch, fixed

jdr81394 (approved):

nechamakrigsman (approved):

Issue #9305, opened 2025-10-29
Minute Sheets: Unscheduled Case Form


cholly75:
As a trial clerk, so that I can record minutes of cases that are heard ad hoc at a trial session, I need to be able to create a minute sheet for a case that has not been scheduled for the trial session.

Sometimes at a trial session, the judge will ask if anyone is present who needs to be heard regarding their case despite not having been formally scheduled to be heard at the session. In these cases, the trial clerk still needs to record minutes for the case and send those to docket in the usual manner.

Our current implementation requires a case to be added to a trial session for the clerk to invoke the minute sheets functionality; we need a way to create minute sheets for these "unscheduled" cases,

## Acceptance Criteria
* Clerk can generate a new minute sheet form for a case not associated with the current trial session
* Clerk must indicate which trial session the case is being heard at
* Minute sheet pre-populates information in the same manner as if the case were associated with the trial session
  - NOTE: we do **not** want to add this case to the trial session itself
* If a consolidated case number is selected (lead or member), only a single minute sheet is created
* All other minute sheet functionality remains intact

## Test Cases
Test Cases for this story are in TestRail

## Definition of Done (Updated 01-16-24)
**Product Owner**
 - [ ] Acceptance criteria have been met and validated on the Court's test environment
 - [ ] Add scenario to testing document, if applicable (https://docs.google.com/spreadsheets/d/1FUHKC_YrT-PosaWD5gRVmsDzI1HS_U-8CyMIb-qX9EA/edit?usp=sharing)

**Engineering**
 - [ ] Automated test scripts have been written, including visual tests for newly added PDFs.
 - [ ] Field level and page level validation errors (front-end and server-side) integrated and functioning.
 - [ ] New screens have been added to cypress accessibility axe
 - [ ] All new functionality verified to work with keyboard and macOS voiceover https://www.apple.com/voiceover/info/guide/_1124.html.
 - [ ] Swagger docs have been updated if API endpoints have been added or updated.
 - [ ] UI should be touch optimized and responsive for external users.
 - [ ] Interactors should validate entities before calling persistence methods.
 - [ ] Features have been optimized where possible to reduce response times. For example, reducing api response times, parallelizing client network calls, optimizing database reads, etc.
 - [ ] Types have been added to all added and updated functions.
 - [ ] Code refactored for clarity and to remove any known technical debt.
 - [ ] Acceptance criteria for the story has been met.
[...]

PR #9575, opened 2026-01-01
9305 minute sheets on case

codyseibert:
#9305

## Overview

Adds support for creating minute sheets for unscheduled cases. Previously, minute sheets were only available for scheduled trial sessions.

**What it does:**
- Adds functionality to create minute sheets for cases that aren't scheduled for trial
- Includes validation for docket numbers and case confirmations
- Adds tests for the new interactor and updates Cypress tests

**Key changes:**
- New interactor: `getUnscheduledMinuteSheetsInteractor`
- UI form: "New Minute Sheet for Unscheduled Cases" with validation
- Tests and bug fixes for page validation and Cypress testing

This enables court staff to create minute sheets for unscheduled cases, expanding beyond scheduled trial sessions.


shared/src/business/utilities/trialSessionMinutes/caseIsEligibleForMinuteSheet.ts
```diff
@@ -31,7 +31,22 @@ export const caseIsEligibleForMinuteSheet = (
   return true;
 };
 
-const isEligibleConsolidatedCase = (aCase: RawCase & TCaseOrder): boolean => {
+export const isEligibleUnscheduledCaseForMinuteSheet = (
```
jimlerza: I know this is a silly thing to get picky about, but the three exported functions in this file are named inconsistently.

`caseIsEligibleForMinuteSheet`
`isEligibleUnscheduledCaseForMinuteSheet`
`isEligibleConsolidatedCase`

If we wanted to follow the convention established by the previous function, then this new function should be called `unscheduledCaseIsEligibleForMinuteSheet`, or if we wanted to follow the convention of the function that follows this function, then this function should be called `isEligibleUnscheduledCase`.

Considering this new function is referenced in a lot of places, you don't have to rename it if you don't want to.

web-client/src/presenter/actions/TrialSessionMinutes/navigateToMinuteSheetForUnscheduledCaseAction.ts
```diff
@@ -0,0 +1,27 @@
...
+  // Use lead case docket number if consolidated, otherwise use the case's docket number
+  const docketNumber = caseInfo.leadDocketNumber || caseInfo.docketNumber;
+
+  const url = `/trial-session-detail/${trialSessionId}/case/${docketNumber}/minutes?isUnscheduledCase=true`;
+
+  if (window.localStorage?.getItem('__cypressMinuteSheetInSameTab')) {
+    await router.route(url);
+  } else {
+    window.open(url, '_blank');
+  }
```
jimlerza, line 26: I understand this prevents cypress from opening the minute sheet in a new tab.

Since we are storing the minute sheet data in `state.modal.form`, do we need to somehow prevent an enduser from opening two different minute sheets? Like, check for an extant minute sheet tab and open this url in that tab?

jimlerza (approved):
Looks good, just had a question on if we should explicitly try to prevent an end user from trying to open different minute sheets in separate tabs, because it appears that we don't support that scenario.

akuny (approved):
This looks really good. My only comment is that we're now splitting minute sheet-related actions between the TrialSessions and TrialSessionMinutes directories: a potential source of confusion down the road, maybe worth nipping in the bud now if it's not too big of a pain.

Issue #8953, opened 2025-04-09
BUG: Stamp Order title is displaying incorrectly for a specific motion (M126)


ttlenard:
**Describe the Bug**
A judge reported this issue to us.  If a motion is filed by a party and it requires additional text to be input by the user (ex: Motion to Dismiss for Lack of Jurisdiction as to [person, notice, or year] (M126)), when the Judge stamps the motion, the document title that gets saved in the drafts and the message attachment is displaying the generic title, rather than the title of the document that contains the text that is input by the party.

The Judge then sends the Stamped motion order to Docket, and the docket must edit the document title prior to adding the document to the docket record. If they don't edit the title, the stamp order will have [person, notice, or year] in the document title.

We would like to have the stamped motion display the document title that is saved to the docket record, rather than having docket have to re-type this information back in before adding the stamped motion order to the docket record

I went through the list of motions, and I think that this is the only document type that does this?  For whoever picks up this bug, can you please verify?

**Business Impact/Reason for Severity**
low

**In which environment did you see this bug?**
Production, Test

**Who were you logged in as?**
Judge/Docket Clerk/Party

**What were you doing when you discovered this bug? (Using the application, demoing, smoke tests, testing other functionality, etc.)**
This bug was reported to us by a Judge

**To Reproduce**
1. Log in as a Private Practitioner (or any party)
2. Click on a case docket number in your My Cases list
3. Click on File a Document button
4. Select Motion to Dismiss for Lack of Jurisdiction as to [person, notice, or year]
5. Fill in text for person, notice, or year field.
6. Click Continue
7. Upload a document, indicate no objections
8. Check the box for the parties filing the document
9. Click on Review filing button
10. Check the box for redaction and click Submit your filing
11. Log in as docket clerk and navigate to Section document QC
12. Click on the document title for the document that you just submitted
13. Complete and send message to a Judge or ADC
14. Log in as the Judge or ADC user that you just sent the message to
15. Click on the message 
16. Click on Apply Stamp. 
17. Select GRANTED and then click Save Stamp Order button
18. Note the title of the stamp order in the message attachments. It says "[person, notice, or year] instead of the text that was added in step 5.
19. Click Reply
20. Notice that the attachment in the message modal also says "[person, notice, or year] instead of the text that was added in step 5
21. Add a message and click send
22. Log in as a docket clerk and click on the message that was just sent to you
23. Notice that the attachment says "[person, notice, or year]" instead of the text that was added in step 5
[...]

PR #9576, opened 2026-01-02
8953 to Test: use correct document type for stamped docket entry

akuny:
https://github.com/ustaxcourt/ef-cms/issues/8953

When a docket entry is stamped, a new docket entry is created. Previously, this new entry's `documentTitle` defaulted to using the original entry's `documentType`. This caused problems when the document type contained placeholder content (such as event code M126: "Motion to Dismiss for Lack of Jurisdiction as to [person, notice, or year]")—the placeholder would be copied into the stamped entry's title even though the original entry's title didn't contain the placeholder.

This PR adds logic to use the `documentTitle` instead when stamping a docket entry whose `documentType` contains a placeholder.


kampofo (approved):
Looks good

AnthonyLoeraTaxCourt (approved):

PR #9587, opened 2026-01-05
Onboarding documentation update

ben-eccles-gunnison:
New Hire Technical Onboarding Update - removes slack reference (not relevant for open source, no longer used by the USTC dev team), automation script for running locally doc (might move to devops repo)


akuny (approved):

.husky/pre-commit
```diff
@@ -1,2 +1,5 @@
+#!/bin/sh
+export NVM_DIR="$HOME/.nvm"
+[ -s "/opt/homebrew/opt/nvm/nvm.sh" ] && . "/opt/homebrew/opt/nvm/nvm.sh"
```
jimlerza: This doesn't belong here.

first-time-setup.sh
jimlerza, line 1: I appreciate that you took the initiative to write this script, but I don't want to use it when onboarding new team members.

Making new team members follow what are already _very simple_ instructions gives us valuable intel on their ability to navigate the command line and thoroughly read and follow documentation, and gives them the opportunity to contribute right away by updating the documentation as necessary.

If something goes wrong when following the existing documentation, even the greenest junior developer can troubleshoot the steps in the documentation, find a fix, and make a PR updating the documentation to reflect the changes. If we make them run this script instead, and they run into an issue, they will need to spend their first day debugging a bash script, which isn't a good first-day experience.

.husky/pre-commit
jimlerza, line 3: Even if we were keeping the workstation setup script, these lines doesn't belong here.

jimlerza (changes requested):
We can keep the documentation update, but not the script.

jimlerza:
Closing due to inactivity. I won't delete the branch, so we can re-open this PR any time.

PR #9604, opened 2026-01-07
9606 - convert endpoint to async sync as it takes longer than 30 seco…

codyseibert:
…nds with large data uploads

### DIFF
--- a/web-api/src/app.ts
+++ b/web-api/src/app.ts
@@ -437,6 +437,14 @@ app.use(expressLogger);
     lambdaWrapper(fileExternalDocumentToCaseLambda),
   );
   app.post(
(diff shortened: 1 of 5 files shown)


nechamakrigsman (approved):

nechamakrigsman (approved):

web-api/src/app.ts
devinganey613: Is there a reason to keep the old non-async endpoint?

devinganey613 (approved):

web-api/src/app.ts
codyseibert: nope, good catch

devinganey613 (approved):

PR #9619, opened 2026-01-09
9266 migration pt3 rename motion to test 1767983898

jdr81394:
Step 3: Verify migration ( we did this locally )
 

Remove the .push() lines added in PR 1. Now only the new documentType is valid.

### DIFF
--- a/web-api/src/persistence/postgres/utils/migrate/migrations/2026-01-09T15_14_46Z-document-type-expand.ts
+++ b/web-api/src/persistence/postgres/utils/migrate/migrations/2026-01-09T15_14_46Z-document-type-expand.ts
@@ -1,0 +1,31 @@
+import { Kysely } from 'kysely';
+
+export async function up(db: Kysely<any>): Promise<void> {
+
+ await db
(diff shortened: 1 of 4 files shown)


kampofo (approved):

web-api/src/persistence/postgres/utils/migrate/migrations/2026-01-09T15_14_46Z-document-type-expand.ts
```diff
@@ -0,0 +1,31 @@
+import { Kysely } from 'kysely';
+
+export async function up(db: Kysely<any>): Promise<void> {
+
+ await db
+    .updateTable('dwDocketEntry')
```
zacharybrown204, line 6: Could we use a where condition checking if documentType = Motion to Withdraw Counsel (filed by petitioner)?
As this is we're "updating" records we don't need to touch (even though it is just setting a value to itself).

zacharybrown204 (approved):

paulo-filho-ctr (approved):

PR #9620, opened 2026-01-09
added payment-portal as devDependency

a4anurag1040:
**Before the change**
- ef-cms did not include ustaxcourt/payment-portal, so developers could not use the shared payment portal integration library locally without manually installing it or pulling code in other ways. 

**After the change**
- ef-cms now includes ustaxcourt/payment-portal@^0.1.3 as a devDependency (with corresponding package-lock.json updates), enabling payment portal / Pay.gov integration development and testing within ef-cms. Documentation was added to docs/running-locally.md and docs/dependency-updates.md to describe the devDependency and clarify that payment-portal only needs to be updated/published when an ef-cms Node.js upgrade falls outside the published payment-portal engines


nechamakrigsman (approved):

package.json
```diff
@@ -294,6 +294,7 @@
     "@types/react-dom": "19.2.3",
     "@types/uuid": "10.0.0",
     "@types/websocket": "1.0.10",
+    "@ustaxcourt/payment-portal": "^0.1.3",
```
jimlerza:
```suggestion
    "@ustaxcourt/payment-portal": "0.1.3",
```

We pin all package versions explicitly so that all updates are performed via our dependency updates rotation.

jimlerza (changes requested):
Thanks for updating the documentation!

jimlerza (approved):

Issue #8787, opened 2024-08-06
TS Error Cleanup: Function Call Argument Count Mismatch


akuny:
**TS2554: `Expected 1 arguments, but got 2.`**

- [x] shared/src/business/useCases/caseNote/validateNoteInteractor.ts
- [x] shared/src/business/useCases/courtIssuedOrder/validateOrderWithoutBodyInteractor.ts
- [x] shared/src/business/useCases/stampMotion/validateStampInteractor.ts
- [x] shared/src/business/useCases/validateOpinionAdvancedSearchInteractor.test.ts
- [x] shared/src/business/utilities/aggregateCommonQueryParams.test.ts
- [x] shared/src/business/utilities/formatPhoneNumber.test.ts
- [x] shared/src/business/utilities/getFormattedCaseDetail.ts
- [x] web-api/src/business/useCases/judgeActivityReport/getCasesClosedByJudgeInteractor.ts
- [ ] web-api/src/lambdas/api/api.ts // expect callback argument error here and in api-public.ts will be resolved in later version of serverlessExpress outlined [here](https://github.com/CodeGenieApp/serverless-express/issues/700)
- [x] web-api/src/lambdas/v1/v1ApiWrapper.test.ts
- [x] web-api/src/lambdas/v2/v2ApiWrapper.test.ts
- [x] web-client/integration-tests/docketClerkEditsServiceIndicators.test.ts
- [x] web-client/integration-tests/docketClerkSealsCaseContactInformation.test.ts
- [x] web-client/integration-tests/externalUserOpinionSearch.test.ts
- [x] web-client/integration-tests/helpers.ts
- [x] web-client/integration-tests/judgeMessagesJourney.test.ts
- [x] web-client/integration-tests/noticeOfChangeOfAddressQCJourney.test.ts
- [x] web-client/integration-tests/petitionerWithRepresentationChangesEmail.test.ts
- [x] web-client/src/persistence/dynamsoft/getScannerMockInterface.ts
- [x] web-client/src/presenter/actions/CaseConsolidation/setAddConsolidatedCaseSuccessMessageAction.test.ts
- [x] web-client/src/presenter/actions/CaseDetail/setCaseDetailPageTabFrozenAction.test.ts
- [x] web-client/src/presenter/actions/CaseDetail/setCaseDetailPageTabUnfrozenAction.test.ts
- [x] web-client/src/presenter/actions/editPractitionerDocumentAction.test.ts
[...]

PR #9623, opened 2026-01-09
8787 function call argument count mismatch to test 1767994122

devinganey613:
#8787 Updated several files to fix argument count mismatch typescript errors


akuny (approved):

web-client/integration-tests/externalUserOpinionSearch.test.ts
```diff
@@ -155,7 +154,16 @@ describe('verify opinion search works for external users', () => {
...
+    let getSearchParams: () => {
+      dateRange?: string;
+      keyword?: string;
+      opinionTypes?: {
+        [x: string]: boolean;
+      };
+      keywordType?: string;
+      caseTitleOrPetitioner?: string;
+      docketNumber?: string;
+    } = () => ({
```
jimlerza, line 166: I personally find this hard to read. I think this is what this syntax is doing?

```typescript
let funcVar: () => { ... } = () => ({ ... })
             |        |    |           |
             |        |    |           |
          input type  |   assignment   |
                return type     object return shorthand
```

So basically, we're setting a return type now which includes more than we're returning now so that it can be reused later?

I'm not asking for it to be changed, btw. I think the optional presence of props that we know we don't need is what is throwing me.

jimlerza (approved):

web-client/integration-tests/externalUserOpinionSearch.test.ts
devinganey613: That's right yeah. Because we didn't define a type, typescript implicitly gave it one based on what fields were present when the return object was first defined. This implied type then didn't want the present fields to be undefined, nor did it want further fields added when the variable was redefined later in the test.

jimlerza: That makes sense. Thanks for explaining it!

Issue #9593, opened 2026-01-05
BUG: Receiving 502 error in Network tab/console when attempting to make edits, QC filings in XL case


ttlenard:
**Describe the Bug**
When attempting to Edit docket entries, Upload PDF's, Create Orders, or QC filings on XL case (Amgen 16017-21) I am receiving a 502 Error in the Network tab/console. No errors are showing up in the logs.  The edits are actually being made successfully on the case, but the user only sees a spinning wheel of doom

Review the Response string in the Postgres table dw_response_string for this request
It appears that the entire case information is coming back in the response
This seems to be over the 6mb threshold, causing the 502

Potential solution discussed - trim down the response, so that it is not returning so much data

Note: Successful tests that did not result in 502:
- Private Practitioner could eFile successfully and I checked the logs for service.  
- Private Practitioner changes petitioner contact info - notice of change successfully docketed and served
- Docket clerk updates petitioner contact info - notice of change successfully docketed and served
- Edit docket entry metadata - check box to add to coversheet - received the 502, but the coversheet did get added to the document.
- Docket clerk serves draft order - success

**Business Impact/Reason for Severity**
High

**In which environment did you see this bug?**
Test

**Who were you logged in as?**
Docket Clerk

**What were you doing when you discovered this bug? (Using the application, demoing, smoke tests, testing other functionality, etc.)**
Testing other functionality

**To Reproduce**
1. Go to 16017-21 and then open up your console
2. Click on edit for a docket entry (externally filed), or upload PDF, or create order
3. add some text to the additional info 1 text box
4. click save
5. Spinning wheel, 502 error in the console

Or:
1. Log in as a party to case 16017-21
2. Make a filing
3. Log in as a docket clerk and open up your console
4. Click on Document QC, then Section QC
5. Find the filing you just made
6. Click on the document title from the QC inbox
7. Note: I have tested multiple pathways here and most of these result in a 502. I did have one success, but it seems maybe I go lucky?  I couldn't get a successful QC after the first one that I did.
8. Suggest following these steps a few times: Click Complete - receive 502, Click complete and send message - receive 502, Add in additional info 1 and check the box for add to coversheet, then click complete and send message - receive 502, Add in additional info 1 and check the box to add to coversheet, then click complete - receive 502.
9. Messages actually go through, and the document is QC'd if you look in your message outbox and QC completed boxes, the user only sees a spinning wheel.

**Expected Behavior**
[...]

PR #9631, opened 2026-01-12
9593 handle lambda size limit to test 1768251470

alexdziarn:
https://github.com/ustaxcourt/ef-cms/issues/9593
Implemented a DTO(data transfer object) class for Case, PublicCase, and RestrictedCase to remove "docketEntries.servedParties" and "docketEntries.secondaryDocuments.previousDocument.servedParties" returned in the api response


AnthonyLoeraTaxCourt (approved):
Looks good!

akuny (approved):
We'll want to regression test thoroughly to make sure everything's working as expected now that `servedParties` isn't being sent over the wire for docket entries.

web-api/src/app.ts
```diff
@@ -288,6 +288,42 @@ app.use((req, res, next) => {
...
+              docketEntries: docketEntriesWithoutServedParties,
+            }),
+          };
+        }
+      }
+    }
+    return originalSend.call(res, body);
+  };
+  next();
+});
```
jimlerza: Is this the best place to implement this? I generally dislike middleware solutions like this.

And what is the regression testing plan?

"Test every API request that returns docket entries, as well as every workflow that calls an endpoint that returns docket entries" feels like a MASSIVE regression testing effort.

jimlerza (changes requested):
Just curious.

Are we pursuing this because you have thoroughly reviewed the usage of `servedParties` all throughout the app and are certain that this middleware solution will not cause problems?

Or are you hoping to shift that burden to someone else by pushing this to test?

jimlerza:
Can we try using a DTO as an alternate approach here? We have had success using DTOs in other areas of the app where we want to reduce the amount of information returned in certain circumstances.

TomElliottFlexion:
While we may need a short-term fix for this specific bug (I second the DTO suggestion), this issue will continue to pop up until we make architectural changes. Some random options in no specific order: 
- Compressing responses
- Streaming responses via websockets when possible/necessary
- Tooling to support querying for specific fields (GraphQL, query string params, etc)
- *Changes to persistence patterns to allowing _updating_ data using partial entities rather than _upserting_ the whole entity/entities*
- Move computed fields from frontend to backend to allow passing fewer fields total

alexdziarn:
we have decided to go along with implementing the DTO solution

alexdziarn:
Successfully deployed on exp5

codyseibert (approved):

akuny (approved):
The DTO implementation looks good. Ideally we'd use the factory universally instead of instantiating in interactors, but I understand that that inconsistency was baked into the codebase already.

shared/src/business/entities/EntityValidationConstants.ts
```diff
@@ -366,11 +366,7 @@ export const DOCKET_ENTRY_VALIDATION_RULE_KEYS = {
         .optional()
         .description('Currently only required for the IRS'),
     })
-    .when('servedAt', {
-      is: joi.exist().not(null),
-      otherwise: joi.optional(),
-      then: joi.required(),
-    })
+    .optional()
```
TomElliottFlexion, line 369: Would it be possible/a good idea to make this field required when `servedPartiesCode` has not yet been computed?

shared/src/business/entities/cases/CaseFactory.ts
```diff
@@ -101,9 +130,17 @@ function constructCaseForUser({
...
+      return useDTO
+        ? new RestrictedCaseDTO(new RestrictedCase(rawCase).toRawObject())
+        : new RestrictedCase(rawCase);
+    } else {
+      return useDTO
+        ? new PublicCaseDTO(
+            new PublicCase(rawCase, { authorizedUser: user }).toRawObject(),
+          )
+        : new PublicCase(rawCase, { authorizedUser: user });
+    }
```
TomElliottFlexion, line 143: Is it safe to assume that we need to use this double-entity pattern because of logic in the `RestrictedCase` and `PublicCase` constructors, respectively? Would it be worth moving/duplicating that logic to the DTO entities?

jimlerza (approved):

Issue #9633, opened 2026-01-13
BUG: Unable to calculate penalties or add another year/period when QCing an eFiled Petition - Screen refreshes


ttlenard:
**Describe the Bug**
We received a report that Petitions clerks are unable to calculate penalties or add another year/period when QCing a Petition.  If the Petitions clerk clicks on the Calculate Penalties on IRS notice link, the modal displays, but as soon as they attempt to input an amount, the screen refreshes and takes them back to the Petition page.  The same behavior happens if the clerk clicks on the Add another year/period button.

The Clerk has reported that there is currently a work-around that they are doing so that they can input statistics on petitions. If the Clerk saves the petition for later, and then goes to the Document QC in progress box and opens up the case, they can click on Case info/statistics, and add in the deficiency and penalties successfully on that page.  Then, they can click on the petition link, click yes that an IRS notice was provided, and the statistics information populates correctly.  At this point, they can serve the petition. The other way that they found is that they can serve the petition, then go back to the case info/statistics tab and input the statistics after service of petition using the same way as explained above.

Note: This appears to be working fine when QCing paper filed petition. The problem seems to only be with eFiled petitions

**Business Impact/Reason for Severity**
High

**In which environment did you see this bug?**
Prod/Test

**Who were you logged in as?**
Petitions Clerk

**What were you doing when you discovered this bug? (Using the application, demoing, smoke tests, testing other functionality, etc.)**
This issue was reported to us by a Petitions Clerk

**To Reproduce**
1. Log in as a Petitions Clerk
2. Click on Document QC and then Section Document QC
3. Select a Petition
4. Click on the IRS Notice tab
5. Select Yes that a Notice was provided
6. Be sure that the Type of Case = Deficiency
7. Add in a Date, Year, Deficiency amount and then click on Calculate Penalties on IRS Notice link
8. Modal displays but then the screen refreshes and brings you back to the first page of the Petition screen
9. Next, Go back to the IRS Notice tab, and then click on the Add another year/period button
10. The screen refreshes and brings you back to the first page of the Petition screen.

**Expected Behavior**
The screen should not refresh when the clerk clicks calculate penalties or when the clerk attempts to add another year/period.

**Actual Behavior**
A clear and concise description of what actually happened.
The screen is refreshing when a petitions clerk attempts to calculate penalties or attempts to add another year/period.

**Screenshots**
If you preserve the logs, and be sure that All is selected, you can see where it is erroring.

[...]

PR #9634, opened 2026-01-13
9633 button refreshing to test 1768319200

zacharybrown204:
https://github.com/ustaxcourt/ef-cms/issues/9633

Fix for buttons in QC workflow causing full page refresh.
The cause is that a button without a type attribute will default to a submit button inside a form. These two buttons did not have a type defined and were being treated as submit buttons, which was leading to a page refresh. The form that these buttons are in used to be a div instead, and was changed to a form during a round of typescript fixes.
The fix was to update the button component to default the type to "button" but still accept values to be set.


TomElliottFlexion (approved):

web-client/src/views/CaseDetailEdit/PartyInformation.tsx
```diff
@@ -89,7 +89,7 @@ export const PartyInformation = connect(
               showPrimaryContact={caseDetailEditHelper.showPrimaryContact}
               showSecondaryContact={caseDetailEditHelper.showSecondaryContact}
               useSameAsPrimary={true}
-              onBlur={validateCaseDetailSequence}
+              onBlur={() => validateCaseDetailSequence()}
```
nechamakrigsman, line 92: Why did this need to change?

devinganey613: The sequence was not running correctly and an error was being thrown when you clicked out of the Petitioner Name input while QCing a petition. I copied this syntax from how the onBlur attribute is set for the Contacts component on the create petition page where it is working properly, as seen [here](https://github.com/ustaxcourt/ef-cms/blob/staging/web-client/src/views/StartCaseInternal/Parties.tsx#L96)

paulo-filho-ctr (approved):

web-client/src/views/CaseDetailEdit/PartyInformation.tsx
jimlerza: I noticed this line too! I think the shorthand passes the onBlur event into the sequence as a parameter, so this new code is actually different.

PR #9642, opened 2026-01-15
Dependency updates 01 12 2026 flaky test

ben-eccles-gunnison:
Testing a potential flaky test for one of the smoketests

NOTE: This isn't going to be merged as is, this is only meant for testing a potential solution to allow for listening to incoming websocket actions in our Cypress tests. (Makes it easier to smartly wait in tests that rely on document generation and servicing.)


cypress.config.ts
```diff
@@ -29,7 +30,73 @@ export default defineConfig({
     baseUrl: 'http://localhost:1234',
     experimentalStudio: true,
     setupNodeEvents(on) {
+      let wsClient: WebSocket | null = null;
+      const messageQueue: any[] = [];
+
       on('task', {
+        connectWebSocket({ token, clientConnectionId }: { token: string; clientConnectionId: string }) {
+          return new Promise((resolve, reject) => {
+            const wsUrl = `ws://localhost:3011/?token=${token}&clientConnectionId=${clientConnectionId}`;
```
jimlerza: This test is in `deployed-and-local` which means it needs to run on deployed environments. You'll need to pass in the actual websocket host here if you want this to work when running against a deployed environment.

cypress.config.ts
```diff
@@ -29,7 +30,73 @@ export default defineConfig({
     baseUrl: 'http://localhost:1234',
     experimentalStudio: true,
     setupNodeEvents(on) {
+      let wsClient: WebSocket | null = null;
```
jimlerza: We don't use `null` this way in this app. The pattern we follow is to leave variables undefined when instantiating them and if we need to unset them we set them to `undefined`.

cypress/deployed-and-local/integration/advancedSearch/search.cy.ts
```diff
@@ -134,13 +141,22 @@ describe('Advanced Search', () => {
...
+
         /** Act */
         cy.get('[data-testid="search-link"]').click();
         cy.get('[data-testid="order-search-tab"]').click();
         cy.get('[data-testid="keyword-search-input"]').type(orderContents);
         let count: number;
+        cy.intercept('GET', '**/order-search**').as('orderSearch');
         retry(() => {
-          cy.intercept('GET', '**/order-search**').as('orderSearch');
           cy.get(
```
jimlerza, line 160: I am curious why this was necessary. I'd think if we managed to navigate to the results but it didn't load for whatever reason, we would now be stuck in this retry because we can't start over.

jimlerza (changes requested):

cypress/deployed-and-local/integration/advancedSearch/search.cy.ts
ben-eccles-gunnison: The intercept acts as a listener and will capture the GET in all n attempts of the retry. We don't need to create a new one each time. (According to the docs intercepts live the length of the test before getting cleaned up)

PR #9645, opened 2026-01-15
8821 refactor file selection components to test 1768487133

paulo-filho-ctr:
## Summary

- Update delete pdf button after refactor

- Adds automatic "Order - " prefix to stamp disposition text when event code includes 'O', ensuring consistent formatting for order-related docket entries.

## Changes
- **New utility**: `addOrderStampPrefix()` - conditionally adds "Order - " prefix when eventCode includes 'O' and text doesn't already start with "Order"
- **Integration**: Applied to free text input in `CourtIssuedNonstandardForm.tsx` and document titles in `addCourtIssuedDocketEntryHelper.ts`

## Impact
- Order-related docket entries now consistently display with "Order - " prefix
- Backward compatible: doesn't modify text that already starts with "Order"
- Only affects entries with event codes containing 'O' (e.g., 'O', 'OAJ')


web-client/src/views/ScanBatchPreviewer.tsx
```diff
@@ -224,27 +207,23 @@ export const ScanBatchPreviewer = connect<
...
+                removeScannedPdfSequence();
                 if (validateSequence) validateSequence();
               }}
-              isPetitionFile={isPetitionFile}
               onConfirmDelete={openConfirmDeletePDFModalSequence}
-              onConfirmReplace={openConfirmReplacePetitionPdfSequence}
-              pdfPreviewUrl={pdfPreviewUrl}
-              scanOnly={scanOnly}
+              pdfPreviewUrl={scanBatchPreviewerHelper.pdfPreviewUrl}
+              scanOnly={false}
```
jimlerza: Is the `scanOnly` parameter being deprecated? If so, should we remove it here too?

jimlerza (approved):

web-client/src/views/ScanBatchPreviewer.tsx
paulo-filho-ctr: I like your suggestions. I could not find an example using `scanOnly={true}`. I'll remove it here.

devinganey613 (approved):

PR #9649, opened 2026-01-16
9606 fix issue with page count to test

codyseibert:
once again, the result set was over 6mb causes more errors because this case was too large.  This change returns on the bare minimum for the UI so it can upload the document.  It never really needed the case information in this flow anyway.


nechamakrigsman (approved):

zacharybrown204 (approved):
Nice! Glad to see us removing some unnecessary network traffic. 
At the end of submitExternalDocumentSequence we were just navigating back to case detail which would regrab the case details anyway right?

codyseibert:
> Nice! Glad to see us removing some unnecessary network traffic. At the end of submitExternalDocumentSequence we were just navigating back to case detail which would regrab the case details anyway right?

yes exactly, so no need put the full case in the results

PR #9656, opened 2026-01-20
8821-refactor-file-selection-components-to-test-1768931507

paulo-filho-ctr:
### Summary

- Includes updates preview window for uploads


jimlerza (approved):

web-client/src/views/ScanBatchPreviewer/ScanDocumentTabs.tsx
```diff
@@ -14,26 +14,31 @@ type ScanDocumentTabsProps = {
...
   onSelect,
+  bindTo,
+  tabNameKey = 'documentType',
   scanOnly = false,
 }: ScanDocumentTabsProps) => {
   if (!documentTabs || documentTabs.length <= 1) {
     return null;
   }
 
+  const bindValue =
```
akuny, line 34: Is this logic correct? As-is, `scanOnly` being true results in `bindValue` being `currentViewMetadata.documentSelectedForPreview`, which doesn't seem right on the face of it. Should scanOnly actually bind to the scan path instead?

zacharybrown204 (approved):

PR #9657, opened 2026-01-20
Deprecate 8764 migrations

TomElliottFlexion:
- Moves migrations associated with 8674 to deprecated directory
- Deprecated migrations are removed from the `kysely_migration` table, therefore are no longer required to have been run
- Deprecated migrations can be run with an optional flag in CircleCI params
- These migrations _can_ be deprecated because they are net-zero w/r/t code changes. If there is no invalid data in the database, there is no need to expand/contract as the code looks the same at the end of the expand/contract as the beginning. 
- These migrations _should_ be deprecated because all test and production data has been migrated, making the addition, removal, and copying of data between columns unnecessarily expensive and time-consuming


jimlerza (approved):

jimlerza:
Should we have done this sooner? If so, should we come up with a process that we can all own?

akuny (approved):

TomElliottFlexion:
> Should we have done this sooner? If so, should we come up with a process that we can all own?

A little sooner? Probably yes. We wanted to give enough time for exps to have their data migrated, and these migrations are harmless on envs that the code has already been deployed to. It should at minimum be the responsibility of the person writing the migrations to deprecate them, in this case me

PR #9661, opened 2026-01-20
9266 feature flag pt1 to test 1768941428

jdr81394:
As a practitioner, so that I can be sure I am filing the correct motion, I need M116 to be re-titled to "Motion to Withdraw Counsel by Party".

We are still seeing confusion on the part of practitioners when attempting to withdraw from a case --- a large number of them are filing M116 (Motion to Withdraw Counsel) instead of M112 (Motion to Withdraw As Counsel). The Court has asked to rename the official M116 motion to "Motion to Withdraw Counsel by Party" to make it clearer where the directive to remove counsel is originating from and reduce the number of misfilings.


zacharybrown204 (approved):

jimlerza (approved):

jimlerza (changes requested):
Changing my review to "request changes" to request that the setting of the `restricted-event-codes` feature flag value occur before the color switch of this first deployment for reasons discussed [here](https://github.com/ustaxcourt/ef-cms/pull/9662/changes/BASE..ce94b55323278512f176628448c0cda5d5721fec#r2714021784)

jimlerza (approved):

PR #9662, opened 2026-01-20
9266 feature flag pt2 to test 1768941517

jdr81394:
As a practitioner, so that I can be sure I am filing the correct motion, I need M116 to be re-titled to "Motion to Withdraw Counsel by Party".

We are still seeing confusion on the part of practitioners when attempting to withdraw from a case --- a large number of them are filing M116 (Motion to Withdraw Counsel) instead of M112 (Motion to Withdraw As Counsel). The Court has asked to rename the official M116 motion to "Motion to Withdraw Counsel by Party" to make it clearer where the directive to remove counsel is originating from and reduce the number of misfilings.


zacharybrown204 (approved):

.circleci/config.yml
```diff
@@ -221,6 +221,10 @@ jobs:
           name: 'Deploy - Verify Private Elasticsearch'
           command: |
             npm run verify-private-elasticsearch -- $ENV
+      - run:
+          name: Setup Restricted Event Codes
+          command: |
+            ./scripts/postgres/featureFlags/setup-restricted-event-codes.ts "M116"
```
jimlerza: This feedback probably belongs on the previous PR, but it's only just occurred to me now. Do we need to do something to update all end users' states with the new feature flag values?

All end users whose sessions are established before this runs won't have this value in their state, and with no mechanism for invalidating that state, all of the views that are expecting to be able to do `get(state.featureFlags['the new one'])` aren't going to get this value after this runs.

jimlerza (approved):

.circleci/config.yml
jdr81394: There could be a very tiny window where the user navigates to the page that is suppose to display all the documents, and the state is retrieved before this runs.

jimlerza: Can you please substantiate this claim with facts? These are the facts as I understand them.

1. A user session can last many hours as long as they keep it active by not being inactive for an hour.
1. Feature flag values are written to state [here](https://github.com/ustaxcourt/ef-cms/blob/9266-feature-flag-pt2/web-client/src/presenter/actions/setupCurrentPageAction.ts#L22) and never expire. Once they are written to state, they do not get overwritten. This means they are written to state in the user's first pageload.

I think the window is _astronomically_ bigger than "very tiny."

jimlerza: In other words, this value will only show up in state for users whose first pageload occurs after you run this script.

You probably need to run the script to populate the feature flag value **before** the color switch of the first deployment. That way, when the first deployment's color switch kicks everybody out, end users' sessions on the new color will have the correct feature flag value in state.

jimlerza (changes requested):
I think we need to set the value of the feature flag before the color switch in the first deployment.

So the requested change for this PR is to remove the call that sets the feature flag to M116, but keep the call that unsets it later.

jimlerza (approved):

TomElliottFlexion (approved):

PR #9663, opened 2026-01-20
9606 page count issue

codyseibert:
- do batching to help prevent the lambda from running out of memory
- reset the idle timer as pdfs are uploading so that users do not see the timeout modal


web-api/src/app.ts
```diff
@@ -434,6 +434,14 @@ app.use(expressLogger);
     '/case-documents/:docketNumber/external-document',
```
nechamakrigsman: Do we still need this endpoint?

nechamakrigsman (approved):

web-api/src/app.ts
nechamakrigsman: Never mind - I was looking at the first commit.

TomElliottFlexion (approved):

zacharybrown204 (approved):

PR #9670, opened 2026-01-22
opex: deficiencies report

jimlerza:
### Summary
At the request of the Clerk of the Court, we have introduced a new deficiencies report which generates a CSV of cases with IRS deficiencies, optionally filtered by any combination of city, minimum deficiency, and maximum deficiency.

This PR introduces `Inquirer.js` as a dev dependency.

### Technical Implementation Details

1. Renamed `promptUser` to `ask`.
1. Added a new `choose` helper that will prompt the user to select a choice from a list of options, utilizing `Inquirer.js`.


scripts/reports/deficiencies.ts
```diff
@@ -0,0 +1,152 @@
...
+        CASE_STATUS_TYPES.closedDismissed,
+        CASE_STATUS_TYPES.onAppeal,
+      ])
+      .groupBy([
+        'c.docketNumber',
+        'c.associatedJudge',
+        'c.caption',
+        'c.preferredTrialCity',
+        'c.status',
+      ]);
```
jimlerza: `dwCase.statistics` is a jsonb array and any one case can have multiple IRS deficiencies defined. This first CTE query sums up all of the deficiencies for each open case so the totals can be filtered on properly in the second query.

zacharybrown204 (approved):

akuny (approved):

PR #9671, opened 2026-01-23
8994 autogen notow counsel

codyseibert:
[Figma](https://www.figma.com/design/ll1G1s1yFA75ui0bXoYQfZ/8994-Auto-Generate-Notice-of-Withdrawal?node-id=152-6660)
[testrail](https://ustaxcourt.testrail.io/index.php?/runs/view/281&group_id=18&group_by=cases:section_id&group_order=asc&display=tree)
What is in this PR:
- When filing a document as an external user, added validation specifically for notice of withdrawals in `completeDocumentSelectSequence.ts`
- after selection of notice of withdrawal, edited the conditional renders in `FileDocument.tsx` to route to a newly created `NoticeOfWithdrawalForm.tsx`. There are a couple new fields like allPartiesConsent and partiesWithPaperService that were created to handle the new form. Also introduced a new functionality to update contact information for each party within this form.
- When the review button is clicked, it runs validation for each form field and populates it with the text according to the Figma
- created 2 pdf templates and the appropriate routes in the backed used only if the autogenerate option is selected in NoticeOfWithdrawalForm, `certificateOfService.ts` and `noticeOfWithdrawal.ts`
- Edited the filedBy functionality in `generateFiledBy.ts`, so that notice of withdrawals will use the user filing the document instead representing party.
- Includes icon changes to the extra items section of the file upload component
- added frontend validation when selecting 'notw' document type
- created/edited cypress helper methods for navigating to a case, adding users to a case, and updating trial session start date.


web-client/src/presenter/computeds/fileAutoGeneratedDocumentReviewHelper.ts
```diff
@@ -0,0 +1,17 @@
+import { Get } from 'cerebral';
+import { state } from '@web-client/presenter/app.cerebral';
+import { getFilerParties } from '@web-client/presenter/computeds/getFilerParties';
+
+export const fileAutoGeneratedDocumentReviewHelper = (get: Get): any => {
```
jimlerza:
```suggestion
export const fileAutoGeneratedDocumentReviewHelper = (get: Get): { formattedParties: string[] } => {
```

web-client/src/presenter/actions/validateNoticeOfWithdrawalAction.ts
```diff
@@ -0,0 +1,84 @@
...
+import {
+  CASE_STATUS_TYPES,
+  ROLES,
+} from '../../../../shared/src/business/entities/EntityConstants';
+import {
+  calculateDifferenceInDays,
+  createISODateString,
+} from '@shared/business/utilities/DateHandler';
+
+export const validateNoticeOfWithdrawalAction = ({ get, path }) => {
```
jimlerza: Can add input and return types to this func.

web-client/src/presenter/actions/validateNoticeOfWithdrawalAction.ts
```diff
@@ -0,0 +1,84 @@
...
+        title: 'Cannot file Notice of Withdrawal as Counsel',
+        messages: errors,
+      },
+    });
+  }
+};
+
+export const getPartiesToWithdrawFrom = (
+  caseDetail: RawCase,
+  user,
```
jimlerza: Can add a type here too

shared/src/business/entities/User.ts
```diff
@@ -21,7 +22,7 @@ export type UserContact = {
   address3?: string;
   city: string;
   country?: string;
-  countryType: string;
+  countryType: CountryTypes;
```
jimlerza, line 25: Is this going to cause any type issues for existing usages of this type?

shared/src/business/utilities/documentGenerators/certificateOfService.ts
```diff
@@ -0,0 +1,50 @@
...
+import { generateHTMLTemplateForPDF } from '../generateHTMLTemplateForPDF/generateHTMLTemplateForPDF';
+
+export const certificateOfService = async ({
+  applicationContext,
+  data,
+}: {
+  applicationContext: ServerApplicationContext;
+  data: {
+    partyInformation: any;
+    practitionerInformation: any;
```
jimlerza: Surely we can do better than `any` here

web-client/src/presenter/actions/isNoticeOfWithdrawalAction.ts
```diff
@@ -0,0 +1,10 @@
+import { state } from '@web-client/presenter/app.cerebral';
+
+export const isNoticeOfWithdrawalAction = ({ get, path }) => {
```
jimlerza: I think we can add type definitions here too

web-client/src/presenter/actions/setCustomValidationAlertErrorsFileDocumentAction.ts
```diff
@@ -0,0 +1,23 @@
+import { ROLES } from '@shared/business/entities/EntityConstants';
+import { state } from '@web-client/presenter/app.cerebral';
+export const setCustomValidationAlertErrorsFileDocumentAction = ({
+  get,
+  store,
+}) => {
```
jimlerza:
```suggestion
export const setCustomValidationAlertErrorsFileDocumentAction = ({
  get,
  store,
}: ActionProps) => {
```

web-client/src/presenter/actions/updateContactInModalAction.ts
```diff
@@ -0,0 +1,26 @@
+import { state } from '@web-client/presenter/app.cerebral';
+
+/**
+ * updates contact information
+ * @param {object} providers the providers object
+ * @param {object} providers.applicationContext the application context used for getting the getUser use case
+ * @param {object} providers.get the cerebral store used for getting state.form
+ * @returns {object} alertSuccess, docketNumber
+ */
```
jimlerza:
```suggestion
```

We probably don't need this JSDoc

package-lock.json
jimlerza, line 1: But why?

jimlerza (changes requested):

package-lock.json
alexdziarn: idk why, but sometimes when you run "npm i" it edits the package.lock.json

jimlerza (approved):

package-lock.json
jimlerza: That's right, `npm i` regenerates the `package-lock.json`. That's why only the person doing dependency updates each week should be running `npm i`, and everyone else should be running `npm ci`, which does not regenerate `package-lock.json`.

shared/src/business/entities/User.ts
alexdziarn: should be fine, I don't see any rows of data in test that are not one of international or domestic

akuny (approved):

Issue #8789, opened 2024-08-06
TS Error Cleanup: Property Not Found on Type


akuny:
**TS2551: `Property '<property>' does not exist on type '<type>'. Did you mean '<other property>'?`**

- [x] shared/src/business/useCases/uploadDocumentAndMakeSafeInteractor.test.ts
- [x] web-api/src/business/useCases/caseInventoryReport/getCaseInventoryReportInteractor.test.ts
- [x] web-api/src/business/useCases/generateCoverSheetData.test.ts
- [x] web-client/src/presenter/actions/CaseDetail/addCaseToTrialSessionAction.ts
- [x] web-client/src/presenter/actions/getDefaultAttachmentViewerDocumentToDisplayAction.ts
- [x] web-client/src/presenter/actions/getShouldMarkReadAction.test.ts
- [x] web-client/src/presenter/actions/getShouldMarkReadAction.ts
- [x] web-client/src/presenter/actions/handleInvalidScannerSourceAction.test.ts
- [x] web-client/src/presenter/actions/handleInvalidScannerSourceAction.ts
- [x] web-client/src/presenter/actions/loadCachedScannerSourceAction.test.ts
- [x] web-client/src/presenter/actions/loadCachedScannerSourceAction.ts
- [x] web-client/src/presenter/actions/Maintenance/getMaintenanceModeForPublicAction.ts
- [x] web-client/src/presenter/actions/primeScannerSelectModalAction.ts
- [x] web-client/src/presenter/actions/removeSignatureAction.ts
- [x] web-client/src/presenter/actions/setCaseTypeToDisplayAction.ts
- [x] web-client/src/presenter/actions/setDefaultCaseTypeToDisplayAction.test.ts
- [x] web-client/src/presenter/actions/setDefaultCaseTypeToDisplayAction.ts
- [x] web-client/src/presenter/actions/setDocketEntrySelectedFromMessageAction.test.ts
- [x] web-client/src/presenter/actions/setDocketEntrySelectedFromMessageAction.ts
- [x] web-client/src/presenter/actions/setMessageDetailViewerDocumentToDisplayAction.test.ts
- [x] web-client/src/presenter/actions/setMessageDetailViewerDocumentToDisplayAction.ts
- [x] web-client/src/presenter/actions/setScannerSourceAction.ts
- [x] web-client/src/presenter/actions/setViewerDraftDocumentToDisplayAction.test.ts
[...]

PR #9672, opened 2026-01-23
8789 ts cleanup property not found type to test 1769204148

kampofo:
https://github.com/ustaxcourt/ef-cms/issues/8789

Fixed error not found types specified in above ticket and other small issues


web-client/src/presenter/state.ts
```diff
@@ -874,7 +874,7 @@ export const baseState = {
   openCases: [] as TAssociatedCase[],
   openCasesCurrentPage: undefined as number | undefined,
   openClosedCases: {
-    caseType: undefined as string | undefined
+    caseType: undefined as string | undefined,
```
paulo-filho-ctr, line 877: I don't understand `undefined as string`?

web-client/src/presenter/state.ts
```diff
@@ -1044,6 +1045,7 @@ export const baseState = {
   viewerDraftDocumentToDisplay: undefined as unknown as ViewerDocument,
   wizardStep: undefined as string | undefined,
   workItem: {},
+  workItemId: undefined as string | undefined,
```
paulo-filho-ctr, line 1048: I don't understand `undefined as string`? Why not only `string | undefined`?

web-client/src/presenter/state.ts
alexdziarn: I think its because we are type casting inline. So in this case we are setting the default value as undefined and saying the type can either be string or undefined.

alexdziarn (approved):

akuny (approved):

Issue #8477, opened 2023-07-17
Consolidation: Change to QC/Edit Docket Entry Process


cholly75:
As a petitioner or practitioner, so that documents are less confusing, I need the QC process to not generate unnecessary cover sheets on e-filed documents in consolidated cases.

With the implementation of multidocketed e-filings, the QC process now generates multiple unneeded cover sheets for these documents, which make them needlessly long and duplicative. This makes the cover sheets confusing and increases document length for no good purpose. Additionally, the extra work items are unnecessary since they are by definition duplicative of the work item on the lead case.

We would like to handle this process more intelligently for consolidated e-filings. This will streamline the QC/editing process for the Court, reduce overall QC volume, and maintain Court's commitment to having a single source of document truth while removing the extraneous cover sheet generation.

Buckle up. 

8477 8527 8919

## Pre-Conditions
ustaxcourt/ef-cms#8825 

## Acceptance Criteria

QC PROCESS:

* When a document is filed in a case in a consolidated group and multi-docketed across the group, a single QC work item is displayed to the user for the document
* The QC work item signifies that the document was multidocketed across the entire consolidated group
* The QC work item is displayed everywhere QC items would normally be displayed (section/personal QC display, judge/chambers QC display)
  - NOTE: the "needs QC" red star signifier is still present on all docket entries created in the multidocketing process
* Completing QC on any case work item automatically completes QC on all other docket entries created as a result of the multidocketing process
* When completing QC on the lead or member case results in a new coversheet for the document, only a single cover sheet is attached to the document. This is reflected in all the docket entries for that document.
* When completing QC, if a Notice of Docket Change would be generated, a NODC is generated for all cases in the group for which the filing is docketed
  - The NODCs contain the information specific to that case's docket entries
  - If any parties on the affected cases receive paper service, the paper NODCs requiring service are presented to the user for printing and service
* Help text is displayed on the QC screen when completing QC for the Lead or member case work item
* Table counter is displayed on the My Document QC inbox, In Progress, and Processed
* Table counters count each multidocketed filings displayed on the UI as 1 work item, rather than individual document QC work items
* When adding a paper filing documents that is going to be multidocketed, the modal confirming service includes the docket number for each petitioner receiving paper service. (The modal also includes a bulleted list that includes the case docket number and case caption for each case in the group - figure out why we added this!)

EDIT DOCKET ENTRY PROCESS 

[...]

PR #9677, opened 2026-01-26
8477 s3 refactor to test

AbelTaxCourt:
- Added documentStorageId field to the Docket Entry entity instead of using docketEntryId as the S3 key so docket entries on unconsolidated cases can update their own S3 documents independently from the consolidated group
(Enables #8477)
- This PR applies the actual code changes that use documentStorageId over docketEntryId for PDFs
- This PR includes contains the second migration that backfills the documentStorageId fields
- See https://github.com/ustaxcourt/ef-cms/pull/9676


shared/src/business/useCases/loadPDFForPreviewInteractor.ts
```diff
@@ -1,28 +1,19 @@
-/**
- * loadPDFForPreviewInteractor
- *
- * @param {string} applicationContext the application context
```
devinganey613, line 4: Why are we removing JSDoc? Is it a holdover from before the project was in typescript?

shared/src/business/useCases/generateDraftStampOrderInteractor.ts
```diff
@@ -24,24 +24,22 @@ export const generateDraftStampOrderInteractor = async (
     motionDocketEntryId,
     parentMessageId,
     stampData,
-    stampedDocketEntryId,
```
devinganey613, line 27: Some of these interactors could be moved from shared to web-api I believe.

web-api/src/business/useCaseHelper/serveDocumentAndGetPaperServicePdf.ts
```diff
@@ -38,11 +38,15 @@ export const serveDocumentAndGetPaperServicePdf = async ({
         if (stampedPdf) {
           originalPdfDoc = await PDFDocument.load(stampedPdf);
         } else {
+          const { documentStorageId } = caseEntity.docketEntries.find(de => {
```
devinganey613, line 41: You could use the new utility here.

shared/src/business/useCases/loadPDFForPreviewInteractor.ts
AbelTaxCourt: yes

jimlerza: Yes, that's it exactly. As we add types to the parameters and returns, we remove the JSDoc as the IDE can infer intellisense from the types. We did this en masse back when we initially converted to typescript and have been slowly chipping away at the stragglers ever since.

shared/src/business/useCases/generateDraftStampOrderInteractor.ts
TomElliottFlexion: Agreed! This is true of many of our interactors, I will create a devex card for it

TomElliottFlexion (approved):

web-api/src/business/useCaseHelper/serveDocumentAndGetPaperServicePdf.ts
AbelTaxCourt: Will use the utility before staging

zacharybrown204 (approved):

PR #9690, opened 2026-01-29
9593 handle lambda size limit

alexdziarn:
https://github.com/ustaxcourt/ef-cms/issues/9593
Updates:
- Created 3 new DTOs(Data transfer objects) for Cases, RestrictedCases, and Public Cases. These 3 new classes create structure in the response allowing us to edit out fields as neeeded.
- Edited the CaseFactory class to handle the new DTOs for each case type.
- Removed the servedParties field found in docketEntries and docketEntries.secondaryDocument.previousDocument to shrink the response size.
- Updated all the interactors to return the associated Case DTOs instead of RawCase


shared/src/business/dto/cases/CaseDTO.test.ts
jimlerza, line 1: Is this file in the right place? Should there be a `case` directory for these case DTOs?

~Also, this test is looking a little lean. Perhaps we could ensure that the fields we want to keep are kept, the fields we don't want to keep are filtered, and that the DTO entity is preserved when the payload passes through the `dataSecurityFilter`?~

Sorry, I wrote that before I found the CaseFactory test. Disregard!

shared/src/business/dto/docketEntries/PublicCaseDTO.ts
```diff
@@ -0,0 +1,53 @@
+import { RawPublicContact } from '@shared/business/entities/cases/PublicContact';
+import { IrsPractitioner } from '@shared/business/entities/IrsPractitioner';
+import { PrivatePractitioner } from '@shared/business/entities/PrivatePractitioner';
+import { ConsolidatedCaseSummary } from '../cases/ConsolidatedCaseSummary';
+import { removeServedParties } from '../helpers/removeServedParties';
+
+export class PublicCaseDTO {
+  public entityName: string;
```
jimlerza, line 9: The `CaseDTO` hardcoded a value here. I think we're going to need a value here too to prevent this from being reinstantiated as the `entityName` this inherits from (should be `PublicCase`) when it passes through the `dataSecurityFilter`.

shared/src/business/dto/cases/RestrictedCaseDTO.ts
```diff
@@ -0,0 +1,19 @@
+import { removeServedParties } from '../helpers/removeServedParties';
+
+export class RestrictedCaseDTO {
+  public docketNumber?: string;
```
jimlerza, line 6: Does this DTO need an `entityName`? Without one, it will get gobbled up by the `dataSecurityFilter` and not come out the other side.

jimlerza (approved):
This looks great! Excellent use of the `CaseFactory` to handle returning DTOs vs. entities based on user permissions and associations.

I have a couple of nits, but nothing requiring any refactoring or overhaul. I believe the first two nits are related and the last two nits are related.

1. Inconsistent `entityName` values across the DTOs:
   -   `CaseDTO`: hardcoded to `'CaseDTO'`
   -   `PublicCaseDTO`: inherits from the raw object (should be `'PublicCase'`)
   -   `RestrictedCaseDTO`: no `entityName` property

2. Missing Registration in `getEntityByName`:
   - Only `CaseDTO` is registered in `getEntityByName.ts`
   - Because `PublicCaseDTO` inherits `entityName: 'PublicCase'`, the `dataSecurityFilter` in the API handler will re-instantiate it using the `PublicCase` **entity** constructor rather than the DTO
   - `RestrictedCaseDTO` will not be re-instantiated by the `dataSecurityFilter` at all because it lacks an `entityName` field

3. Missing `hasIrsPractitioner` field:
   - The `Case` entity contains a `hasIrsPractitioner` field, but its constructor does not
   - `PublicCaseDTO` contains the `hasIrsPractitioner` field
   - `RestrictedCaseDTO` does not contain the `hasIrsPractitioner` field
   - The `caseDetailHeaderHelper`, which may receive a `CaseDTO`, expects to find `hasIrsPractitioner` (but it also contains fallback logic that checks the length of the `irsPractitioners` array)

4. `PublicCaseDTO` might not have been necessary:
   - As of this PR, the docket entries in `PublicCase` are now `PublicDocketEntry` entities and not `DocketEntry` entities 
   - `PublicDocketEntry` entities don't contain `servedParties` to begin with

alexdziarn:
> This looks great! Excellent use of the `CaseFactory` to handle returning DTOs vs. entities based on user permissions and associations.
> 
> I have a couple of nits, but nothing requiring any refactoring or overhaul. I believe the first two nits are related and the last two nits are related.
> 
> 1. Inconsistent `entityName` values across the DTOs:
>    
>    * `CaseDTO`: hardcoded to `'CaseDTO'`
>    * `PublicCaseDTO`: inherits from the raw object (should be `'PublicCase'`)
>    * `RestrictedCaseDTO`: no `entityName` property
> 2. Missing Registration in `getEntityByName`:
>    
>    * Only `CaseDTO` is registered in `getEntityByName.ts`
>    * Because `PublicCaseDTO` inherits `entityName: 'PublicCase'`, the `dataSecurityFilter` in the API handler will re-instantiate it using the `PublicCase` **entity** constructor rather than the DTO
>    * `RestrictedCaseDTO` will not be re-instantiated by the `dataSecurityFilter` at all because it lacks an `entityName` field
> 3. Missing `hasIrsPractitioner` field:
>    
>    * The `Case` entity contains a `hasIrsPractitioner` field, but its constructor does not
>    * `PublicCaseDTO` contains the `hasIrsPractitioner` field
>    * `RestrictedCaseDTO` does not contain the `hasIrsPractitioner` field
>    * The `caseDetailHeaderHelper`, which may receive a `CaseDTO`, expects to find `hasIrsPractitioner` (but it also contains fallback logic that checks the length of the `irsPractitioners` array)
> 4. `PublicCaseDTO` might not have been necessary:
>    
>    * As of this PR, the docket entries in `PublicCase` are now `PublicDocketEntry` entities and not `DocketEntry` entities
>    * `PublicDocketEntry` entities don't contain `servedParties` to begin with

Addressing 1,3,4:
To stay consistent, added an entityName to all DTOs and registered all DTOs in getEntityByName. 

For 2:
After doing some research `hasIrsPractitioner` in caseDetailHeaderHelper always resolves to undefined so we removed from the helper and the Case class.

shared/src/business/dto/cases/PublicCaseDTO.ts
```diff
@@ -0,0 +1,60 @@
+import { RawPublicContact } from '@shared/business/entities/cases/PublicContact';
+import { IrsPractitioner } from '@shared/business/entities/IrsPractitioner';
+import { PrivatePractitioner } from '@shared/business/entities/PrivatePractitioner';
+import { ConsolidatedCaseSummary } from '../cases/ConsolidatedCaseSummary';
+import { UnknownAuthUser } from '@shared/business/entities/authUser/AuthUser';
+import { PublicCase } from '@shared/business/entities/cases/PublicCase';
+
+export class PublicCaseDTO {
+  public entityName = 'PublicCaseDTO';
+  public canAllowDocumentService?: string;
```
TomElliottFlexion, line 10: If an entity does not extend `JoiValidationEntity`, I believe we must set `entityName` in the constructor to avoid the previously mentioned `dataSecurityFilter`

shared/src/business/entities/EntityValidationConstants.ts
```diff
@@ -366,11 +366,7 @@ export const DOCKET_ENTRY_VALIDATION_RULE_KEYS = {
         .optional()
         .description('Currently only required for the IRS'),
     })
-    .when('servedAt', {
```
TomElliottFlexion, line 369: How would you feel about making this validation dependent on `entityName` + `servedAt` instead of making it always optional?

shared/src/business/entities/cases/RestrictedCase.ts
```diff
@@ -21,7 +21,19 @@ export class RestrictedCase extends JoiValidationEntity {
     this.docketEntries = [];
   }
 
+  static VALIDATION_RULES = {
+    docketEntries: joi.array().max(0),
+  };
+
```
TomElliottFlexion, line 35: We'll still want to validate the other fields (assuming we validate this at all)

TomElliottFlexion: (Which we should probably do)

shared/src/business/dto/cases/CaseDTO.test.ts
TomElliottFlexion, line 1: I _think_ these DTOs should be in a `case` directory instead of in a `docketEntry` directory

shared/src/business/entities/EntityValidationConstants.ts
TomElliottFlexion: I suppose that would require using a new `DocketEntry` entity, so probably more effort than it's worth

alexdziarn: This occurs at the docketEntry not the Case level of validation so we can't use the entityName condition. This is optional because the frontend is doing validation on docketEntry data sent from the backend which does not include servedParties but does include servedAt.

shared/src/business/useCases/getCaseInteractor.test.ts
```diff
@@ -28,7 +28,7 @@ import { UnauthorizedError } from '@web-api/errors/errors';
 import { getWorkItemsByDocketNumber as getWorkItemsByDocketNumberMock } from '@web-api/persistence/postgres/workitems/getWorkItemsByDocketNumber';
 import { WorkItem } from '@shared/business/entities/WorkItem';
 import { docketClerk1User } from '@shared/test/mockUsers';
-import { MOCK_ANSWER_2 } from '@shared/test/mockDocketEntry';
+import { CaseDTO } from '../dto/docketEntries/CaseDTO';
```
TomElliottFlexion: Absolute nitpick, but ideally we'd use the `@shared`/`@web-api`/`@web-client` path aliases. These are defined in `tsconfig.json` The vsCode setting for using this in autoimports is `"typescript.preferences.importModuleSpecifier": "non-relative"`

shared/src/business/useCases/removeSignatureFromDocumentInteractor.ts
```diff
@@ -21,7 +22,7 @@ export const removeSignatureFromDocumentInteractor = async (
   applicationContext: ServerApplicationContext,
   { docketEntryId, docketNumber },
   authorizedUser: UnknownAuthUser,
-) => {
+): Promise<CaseDTO> => {
```
TomElliottFlexion, line 25: :heart:

shared/src/business/entities/cases/RestrictedCase.ts
alexdziarn: we never run validation on RestrictedCases but added here for the future

paulo-filho-ctr (approved):

web-client/integration-tests/externalUserFilesDocumentForLegacyCase.test.ts
```diff
@@ -38,13 +38,11 @@ describe('an external user files a document for their legacy case', () => {
       docketNumber: cerebralTest.docketNumber,
     });
 
-    const otherFilers = getOtherFilers(cerebralTest.getState('caseDetail'));
     const docketEntries = cerebralTest.getState('caseDetail.docketEntries');
     const lastServedDocument = docketEntries.pop();
 
-    const isOtherFilerServed = lastServedDocument.servedParties.find(
```
TomElliottFlexion, line 45: A `servedPartiesCode` of `BOTH` indicates that a document was served both to the IRS and to the petitioner. It does not indicate that a document was served to all filing parties on the case (ie intervenors or participants).

TomElliottFlexion (commented):
This was a lot of work, thank you

web-client/integration-tests/externalUserFilesDocumentForLegacyCase.test.ts
devinganey613: I don't think this test can work anymore then, since servedParties isn't returned to the frontend anymore. Should we remove the test? Or can we include a database query of some kind?
The test also doesn't seem to do what it says it does, which is verifying that the other filers received paper service. But it just checks that the other filer was served.

TomElliottFlexion: You're right. This is kind of a tough judgement call that I'll leave up to you. We absolutely could query the database (see `cypress/helpers/cypressTasks/wait-for-noce.ts`), I'm not sure if that's worth it. Since paper service is entirely dependent on a clerk printing out docs and mailing them, the only (?) other way to verify is by checking the content of the print paper service modal, but even that checks the service indicator of parties on the case, rather the the `servedParties` on the docket entry itself

TomElliottFlexion: ^ (the database querying infrastructure is currently only set up for _cypress_ tests)

devinganey613: Thanks, I revamped the test to use the paper service modal to check if the clerk is notified that the other party receives paper service.

jimlerza (approved):

TomElliottFlexion (approved):

PR #9694, opened 2026-01-30
8877 - petitioner email address changes and tests

codyseibert:
Summary                                                                                        
                                         
  Addresses ticket 8877: petitioner contact email handling.                                      
                                                                                                 
  Changes                                
                                                                                                 
  - Field rename: paperPetitionEmail → contactEmailAddress across entities, types, UI, and tests.
   The previous name implied paper-only, but the field now stores a contact email regardless of
  filing method.                                                                                 
  - PII protection: CaseFactory.constructCaseForUser now strips contactEmailAddress from
  petitioners on cases returned to associated users, so it never leaves the API in network       
  responses to petitioners/practitioners.
  - Editable by court personnel: added a "Contact email address" input on                        
  EditPetitionerInformationInternal, and wired updatePetitionerInformationInteractor to          
  accept/persist the field as an editable property (removed the previous read-only passthrough).
  - Capture on e-filing: formatPetitionAction now sets contactEmailAddress to the filer's email  
  for non-practitioner submissions, so an electronically filed petition records a contact email  
  at the time of filing.                 
  - Helper refactor: internalPetitionPartiesHelper splits the combined "show email field AND     
  consent box" flag into separate concerns (showContactEmailField / showEConsentCheckbox, plus   
  secondary equivalents), so court users can see/edit the field without rendering the e-consent
  checkbox.                                                                                      
  - UI relabeling: "Petition email address" → "Contact email address"; added a "Service email
  address" label on PetitionersAndCounsel to disambiguate from the new contact email.            
  - DB migration: simplified single-statement migration that renames paperPetitionEmail →
  contactEmailAddress inside the dw_case.petitioners JSONB (heavy/batched version dropped after  
  confirming only ~5.5k affected rows).


web-client/src/views/CaseDetail/PetitionersAndCounsel.tsx
```diff
@@ -64,24 +64,29 @@ export const PetitionersAndCounsel = connect(
...
-                          aria-label="has e-access"
-                          className="margin-left-05 fa-icon-blue"
-                          icon="flag"
-                          size="1x"
-                        />
-                      )}
-                    </span>
+                    <>
+                      <p className="semi-bold margin-bottom-0">
+                        Service email address
```
jdr81394, line 69: Hey cody, I don't remember if you were in the meeting a month or so ago but we decided that going forward, whenever we create new elements, we will be using the tailwind notation ( except for some edge cases like complex tables )

jdr81394 (changes requested):

TomElliottFlexion (approved):

jimlerza (approved):

jimlerza (changes requested):
I am concerned about what will happen in the 5-10 real minutes after this migration runs but before we've switched to the app code that uses the new field.

codyseibert:
> I am concerned about what will happen in the 5-10 real minutes after this migration runs but before we've switched to the app code that uses the new field.

I've verified the migration script made by @TomElliottFlexion using the following steps:

- create a case as a petitioner with a petitioner & spouse case type (gave spouse email)
- filed and served case (as petitionsclerk)
- ran the migration script from this branch against exp8
- edited the address of that case (as docketclerk), no errors
- verified in the database that both paperPetitionEmail and contactEmailAddress matched
- ran the contract migration and now data just has contactEmailAddress (new code path would use this)

devinganey613 (approved):

codyseibert:
we deployed the migration tom worked on to exp8 and verified together it seems to work during that 5-10 minute perioud

jimlerza (approved):

web-client/src/views/CaseDetail/PetitionersAndCounsel.tsx
codyseibert: I will address this on 9915 that I'm working on so we can move this forward

TomElliottFlexion (approved):

Issue #8784, opened 2024-08-06
TS Error Cleanup: Argument Type Mismatch in Function Call


akuny:
**TS2345: `Argument of type '<something>' is not assignable to parameter of type <something else>'.`**

- [x] scripts/elasticsearch/ready-cluster-for-migration.helpers.test.ts
- [x] shared/src/business/entities/cases/Case.ts
- [x] shared/src/business/entities/cases/Case.updateAutomaticBlocked.test.ts
- [x] shared/src/business/entities/courtIssuedDocument/CourtIssuedDocumentTypeA.ts
- [x] shared/src/business/entities/courtIssuedDocument/CourtIssuedDocumentTypeB.ts
- [x] shared/src/business/entities/courtIssuedDocument/CourtIssuedDocumentTypeC.ts
- [x] shared/src/business/entities/courtIssuedDocument/CourtIssuedDocumentTypeD.ts
- [x] shared/src/business/entities/courtIssuedDocument/CourtIssuedDocumentTypeE.ts
- [x] shared/src/business/entities/courtIssuedDocument/CourtIssuedDocumentTypeF.ts
- [x] shared/src/business/entities/courtIssuedDocument/CourtIssuedDocumentTypeG.ts
- [x] shared/src/business/entities/courtIssuedDocument/CourtIssuedDocumentTypeH.ts
- [x] shared/src/business/entities/DocketEntry.isPublic.test.ts
- [x] shared/src/business/entities/DocketEntry.ts
- [x] shared/src/business/entities/docketEntry/DocketEntryFactory.ts
- [x] shared/src/business/entities/documents/DocumentSearch.ts
- [x] shared/src/business/entities/externalDocument/ExternalDocumentNonStandardA.ts
- [x] shared/src/business/entities/externalDocument/ExternalDocumentNonStandardB.ts
- [x] shared/src/business/entities/externalDocument/ExternalDocumentNonStandardC.ts
- [x] shared/src/business/entities/externalDocument/ExternalDocumentNonStandardD.ts
- [x] shared/src/business/entities/externalDocument/ExternalDocumentNonStandardE.ts
- [x] shared/src/business/entities/externalDocument/ExternalDocumentNonStandardF.ts
- [x] shared/src/business/entities/externalDocument/ExternalDocumentNonStandardG.ts
- [x] shared/src/business/entities/externalDocument/ExternalDocumentNonStandardH.ts
[...]

PR #9701, opened 2026-02-02
8784 ts error cleanup arg type mismatch to test 1770063837

devinganey613:
#8784 Fixing typescript errors

- Added Error alert to gracefully handle document signing failure.


shared/src/business/useCases/validateHearingNoteInteractor.test.ts
```diff
@@ -1,17 +1,17 @@
-import { applicationContext } from '../test/createTestApplicationContext';
 import { validateHearingNoteInteractor } from './validateHearingNoteInteractor';
 
 describe('validateHearingNoteInteractor', () => {
-  it('returns a defined errors object if the note is null', () => {
-    const errors = validateHearingNoteInteractor(applicationContext, {
-      note: null,
+  it('returns a defined errors object if the note is greater than 200 characters', () => {
+    const errors = validateHearingNoteInteractor({
```
paulo-filho-ctr, line 5: Is the case where the defined object of the note is null is no longer needed?

paulo-filho-ctr (changes requested):
We need test coverage for the moved ` formatDateFromDatePicker` in `DateHandler.ts`.

web-api/src/business/useCaseHelper/service/sendIrsSuperuserPetitionEmail.ts
```diff
@@ -34,13 +34,9 @@ export const sendIrsSuperuserPetitionEmail = async ({
...
   (privatePractitioners ?? []).forEach(practitioner => {
-    const representingFormatted = [];
-
-    caseEntity.petitioners.forEach(p => {
-      if (practitioner.isRepresenting(p.contactId)) {
-        representingFormatted.push(p.name);
-      }
-    });
+    const representingFormatted = caseEntity.petitioners
+      .filter(p => practitioner.isRepresenting(p.contactId))
```
paulo-filho-ctr, line 38: Very nice!

web-api/src/lambdas/auth/confirmSignUpLambda.ts
```diff
@@ -8,7 +8,7 @@ export const confirmSignUpLambda = (event: APIGatewayProxyEvent) =>
       return await applicationContext
         .getUseCases()
         .confirmSignUpInteractor(applicationContext, {
-          ...JSON.parse(event.body),
+          ...JSON.parse(event.body!),
```
paulo-filho-ctr: Looks awesome, my only reservation is that runtime will still throw an error if for example when the body is undefined

web-client/src/presenter/actions/completeDocumentSigningAction.ts
```diff
@@ -27,9 +27,13 @@ export const completeDocumentSigningAction = async ({
     const windowWithPdfjs = window as Window & {
       pdfjsObj?: { getData: () => Promise<unknown> };
     };
-    const pdfjsObj: { getData: () => Promise<unknown> } =
+    const pdfjsObj: { getData: () => Promise<unknown> } | null =
       windowWithPdfjs.pdfjsObj || get(state.pdfForSigning.pdfjsObj);
 
+    if (!pdfjsObj) {
+      throw new Error('pdfjsObj is null!');
```
paulo-filho-ctr: Instead of throwing an error, which would result in users getting a 500 response status, should we log the error, and graciously skip the rest of the function?

web-client/src/presenter/errors/ActionError.ts
```diff
@@ -8,9 +8,8 @@ export class ActionError extends CerebralError {
   get className() {
     return this.constructor.name;
   }
-  constructor(message) {
-    // eslint-disable-next-line prefer-rest-params
-    super(arguments);
+  constructor(message: string) {
+    super(message);
```
paulo-filho-ctr: Nice! To consider: since that is a considerable change: before we were instantiating the class with `arguments` which is an array with all the parameters passed to the function, for example: `new ActionError('some message', true, 1, {some: 'data'})` Now we only instantiate with `message` meaning: `new ActionError('some message').`

shared/src/business/useCases/validateHearingNoteInteractor.test.ts
kampofo: Just added a test for this, because empty string or null throws a different error.

web-client/src/presenter/actions/completeDocumentSigningAction.ts
kampofo: The error gets thrown in the UI, but it does cause the sign order sequence to hang up, so were going to create a new ticket to address this to handle this case more gracefully

web-client/src/presenter/errors/ActionError.ts
kampofo: we added arguments as a second parameter in the constructor to handle the other arguments that may be passed to the constructor.

devinganey613:
> We need test coverage for the moved ` formatDateFromDatePicker` in `DateHandler.ts`.

Added tests

web-client/src/presenter/errors/ActionError.ts
paulo-filho-ctr: Awesome, thanks!

web-client/src/presenter/actions/completeDocumentSigningAction.ts
TomElliottFlexion: I don't think there's a right or wrong answer here, but imo a crash is better than an infinite spinner or some sort of silent error. If this error is likely to be thrown outside of a bug condition, we should show the user an error page/modal/alert

kampofo: So we've got a fix to gracefully handle this error as well as added a unit test to cover this. However we were unable to expose the cerebral state in cypress to get this tested there. It seems we may only be able to test this manually.

TomElliottFlexion: If there's nothing a (Cypress) user can do to create this error, then I don't think this needs to be integration tested. You would have Cerebral integration tests available to you if this needed that level of testing, but this is just null-checking something that _should_ always be there, right? Any variable could become undefined in theory, and I'm not opposed to trying to handle most errors gracefully, but we can't null-check everything.

TomElliottFlexion: This looks good to me!

shared/src/business/useCases/loadPDFForSigningInteractor.ts
```diff
@@ -23,8 +23,8 @@ export const loadPDFForSigningInteractor = async (
     const pdfData = await applicationContext
       .getPersistenceGateway()
       .getDocument({
-        applicationContext: applicationContext as any,
-        ...(docketNumber && { docketNumber }),
+        applicationContext,
+        docketNumber,
```
TomElliottFlexion, line 27: ❤️

TomElliottFlexion (approved):

paulo-filho-ctr (approved):

Issue #9693, opened 2026-01-30
BUG:  Order Description Field Forcing "Order -" Text for all Order Descriptions


cholly75:
**Describe the Bug**
When adding or editing the docket entry for an Order, if the user types anything into the Description field, the field is being populated with the text "Order -" and will not allow the user to remove it. This should not be happening. The user can type text after the "Order -" however they cannot remove the "Order -" portion of the field (it keeps reappearing in the field). 

On the Add or Edit Docket Entry screen, select any document type that is an order. Enter or modify the "What is this order for?" or Description field. Note that the "Order-" text is inserted at the beginning of the string and cannot be removed. This is also visible in the Document Preview text. 

If the document's default Description text is unmodified or appended to, it appears to retain the value without adding the "Order -" text. Interaction with the field default appears to be the trigger for this behavior.

This appears to be related to the work done in #9472 .

**Business Impact/Reason for Severity**

**In which environment did you see this bug?**
PROD, TEST

**Who were you logged in as?**

**What were you doing when you discovered this bug? (Using the application, demoing, smoke tests, testing other functionality, etc.)**

**To Reproduce**

1. Log in as a Docket clerk
2. Upload a PDF file with description "Test This". Apply and save the signature to the Draft. 
3. Click "Add Docket Entry".
4. In the Document Type field, select "Order". Note the default value "Order" populates in the "What Is This Order For?" field.
5. Click into the "What is this order for?" field and backspace out or edit the default text.
6. Note that the text "Order -" is inserted into the field and cannot be removed (it simply keeps reappearing).

-----------------

1. Log in as a Docket clerk
2. Upload a PDF file with description "Test This". Apply and save the signature to the Draft. 
3. Click "Add Docket Entry".
4. In the Document Type field, select "Standing Pretrial Order". Note the value "Order - Test This" populates in the Description field.
5. Click into the Description field and attempt to edit the text.
6. Note that the text "Order -" is inserted into the field and cannot be removed (it simply keeps reappearing).

-------------------

1. Log in as a Docket clerk
2. Upload a PDF file with description "Test This". Apply and save the signature to the Draft. 
3. Click "Add Docket Entry".
4. In the Document Type field, select "Order". Note the default value "Order" populates in the "What Is This Order For?" field.
5. Click into the "What is this order for?" field after the default "Order" text. Type in text.
6. Note that the issue does not occur.

**Expected Behavior**

**Actual Behavior**
A clear and concise description of what actually happened.
[...]

Issue #9472, opened 2025-12-08
Stamp Disposition - Add "Order -" to Docket Description


cholly75:
As a DAWSON user, so that I can easily tell that a stamp disposition order is in fact an order, I need the docket entry descriptions for stamp orders to begin with "Order -".

We have received user feedback that sometimes it is hard to distinguish between a stamp order and the motion it is acting on just from the description in the docket record, because the default description for the Order begins with the same description as the Motion.

To better signify that the order is an order, we would like to add "Order -" to the beginning of stamp order title and docket entry descriptions.

## Acceptance Criteria
* The description for stamp disposition orders is generated with the text "Order - " prepended to the rest of the generated title/description

## Notes
* No need to update existing Stamped Motions in the system

## Test Cases
Test cases for this story are in Test Rail

## Definition of Done (Updated 01-16-24)
**Product Owner**
 - [ ] Acceptance criteria have been met and validated on the Court's test environment
 - [ ] Add scenario to testing document, if applicable (https://docs.google.com/spreadsheets/d/1FUHKC_YrT-PosaWD5gRVmsDzI1HS_U-8CyMIb-qX9EA/edit?usp=sharing)

**Engineering**
 - [ ] Automated test scripts have been written, including visual tests for newly added PDFs.
 - [ ] Field level and page level validation errors (front-end and server-side) integrated and functioning.
 - [ ] New screens have been added to cypress accessibility axe
 - [ ] All new functionality verified to work with keyboard and macOS voiceover https://www.apple.com/voiceover/info/guide/_1124.html.
 - [ ] Swagger docs have been updated if API endpoints have been added or updated.
 - [ ] UI should be touch optimized and responsive for external users.
 - [ ] Interactors should validate entities before calling persistence methods.
 - [ ] Features have been optimized where possible to reduce response times. For example, reducing api response times, parallelizing client network calls, optimizing database reads, etc.
 - [ ] Types have been added to all added and updated functions.
 - [ ] Code refactored for clarity and to remove any known technical debt.
 - [ ] Acceptance criteria for the story has been met.
 - [ ] If there are special instructions in order to deploy into the next environment, add them as a comment in the story.
 - [ ] Code that resides in the shared folder that only runs on the API or browser has been moved to either /web-client or /web-api.

ttlenard:
@paulo-filho-ctr 

Testing Feedback:

- [ ] Test Run T6149 is failing at multiple steps.

[...]

PR #9704, opened 2026-02-03
9693: fix document title and description from stamp disposition to test 1770139476

paulo-filho-ctr:
## Summary

[9693](https://github.com/ustaxcourt/ef-cms/issues/9693) addressed a bug when adding a docket entry. On a previous feature [9472](https://github.com/ustaxcourt/ef-cms/issues/9472) the docket entry description would append an "Order -" prefix to stamp disposition. It created a bug where the prefix "Order -" could not be removed from the description. The original solution had parts done in the api, and some in the web app. The PR moves all the logic to the api, for more consistent writes while fixing the bug on the web app.

### Includes

- A new business utility `setDocumentTitle` and its tests


devinganey613 (approved):

devinganey613 (approved):

zacharybrown204 (approved):

shared/src/business/utilities/setDocumentTitle.ts
```diff
@@ -0,0 +1,10 @@
+import { Stamp } from '@shared/business/entities/Stamp';
+
+export function setDocumentTitle(documentTitle: string, stampData: Stamp) {
+  const prefix = 'Order - ';
+  return stampData.disposition
+    ? documentTitle.startsWith(prefix)
+      ? documentTitle
+      : `${prefix}${documentTitle}`
+    : documentTitle;
+}
```
jimlerza, line 10: This always returns a string, so we should add a return type. You don't have to update this PR with that, just as long as it makes it into the PR to staging.

jimlerza (approved):

PR #9705, opened 2026-02-03
devex: stream stdout in typescript scripts

jimlerza:
This PR enhances the `runCommand` helper utility to support streaming `stdout` directly to the parent process. This improvement provides real-time feedback for long-running or verbose commands executed by our scripts. Additionally, it introduces an option to skip capturing `stdout` in the return value when it is not required, reducing memory overhead for large outputs.

- Changes to support streaming output in the `runCommand` utility:
    - Added an optional configuration object as the fourth parameter with `captureStdout` (defaults to `true`) and `streamStdout` (defaults to `false`).
    - Implemented streaming logic using `process.stdout.write` when `streamStdout` is enabled.
- Test Coverage:
    - Created `scripts/helpers/runCommand.test.ts` with comprehensive tests for capturing, streaming, and disabling capture.
    - Updated `scripts/jest-scripts.config.ts` to include `runCommand.ts` in the coverage reports.
- Existing Usage:
    - Updated several scripts to use `streamStdout: true` and `captureStdout: false` for commands where real-time output is more valuable than capturing the result string.


TomElliottFlexion (approved):
:fire:

PR #9713, opened 2026-02-04
9710 - reduce the batch size

codyseibert:


TomElliottFlexion (approved):

zacharybrown204 (approved):

codyseibert:
I'm looking into if this will actually fix it... I'm thinking we might be returning too much data now.

Issue #9612, opened 2026-01-09
Motion/Order Link Enhancement


cholly75:
As a Court user issuing orders, so that I can correctly indicate the disposition of a motion, I need new options for Docket to select when linking an order to a motion.

Feedback from judicial officers indicates that we need new values in the disposition menu for linking orders to motions:
1) Denied in Part
2) Granted in Part and Denied in Part

## Acceptance Criteria
* Docket clerk can select "DENIED IN PART" when linking an order to motion(s).
  - Selecting "DENIED IN PART" creates a hyperlink/description for the motion(s) that reads "DENIED IN PART BY #XXX" linked to the order
  - Selecting "DENIED IN PART" creates a hyperlink/description for the order that reads "DENYING IN PART #XXX" linked to the motion(s)
* Docket clerk can select "GRANTED IN PART AND DENIED IN PART" when linking an order to motion(s)
  - Selecting "GRANTED IN PART AND DENIED IN PART" creates two hyperlink/descriptions for the motion(s):
    - "GRANTED IN PART BY #XXX" linked to the order
    - "DENIED IN PART BY #XXX" linked to the order 
    - Both of these links go to the same order
  - Selecting "GRANTED IN PART AND DENIED IN PART" creates two hyperlink/descriptions for the order:
    - "GRANTING IN PART #XXX" linked to the motion(s)
    - "DENYING IN PART #XXX" linked to the motion(s)
    - Both of these links go to the same motion(s)
* All links obey standard visibility rules for internal and external users
* Dispositions can be edited via editing of the order

## Test Cases

Available in TestRail
### Product Owner
 - [ ] Acceptance criteria have been met and validated on the Court's test environment.
 - [ ] Associated test cases defined in TestRail have been updated if necessary.
 - [ ] Successful test run is performed in TestRail.
 - [ ] User guides are updated if necessary.

### UX
 - [ ] All new functionality has been verified to work with keyboard navigation and screen reader software.
 - [ ] UI should be touch optimized and responsive for external users.

### Engineering
 - [ ] Automated test scripts have been written, including visual tests for newly added PDFs.
 - [ ] Successful test run is performed in TestRail.
 - [ ] New screens have been added to cypress accessibility axe.
 - [ ] Interactors should validate entities before calling persistence methods.
 - [ ] Types have been added to all added and updated functions.
 - [ ] Code refactored for clarity and to remove any known technical debt.
 - [ ] Acceptance criteria for the story has been met.
[...]

PR #9717, opened 2026-02-05
9612: Motion Order Link

akuny:
https://github.com/ustaxcourt/ef-cms/issues/9612

This PR required reworking how hyperlinks to linked motions and orders render in DAWSON. The requirement that the "GRANTED IN PART AND DENIED IN PART" disposition render two distinct hyperlinks made these changes necessary.

From a user's point of view, these changes are reflected in the "Docket Record" table and the printable docket record PDF.


web-client/src/presenter/computeds/Public/publicCaseDetailHelper.ts
```diff
@@ -125,12 +125,15 @@ export const formatDocketEntryOnDocketRecord = (
         isTerminalUser,
       );
 
+      const dispositionLinkText = MOTION_DISPOSITION_VERBIAGE[
+        affectedEntry.disposition
```
jdr81394, line 129: is it guaranteed that the affectedEntry will ALWAYS have a disposition? If not, we should have some kind of check.

web-client/src/presenter/computeds/formattedDocketEntries.ts
```diff
@@ -179,13 +179,16 @@ export const getFormattedDocketEntry = ({
             visibilityPolicyDateFormatted,
           );
 
+        const dispositionLinkText = MOTION_DISPOSITION_VERBIAGE[
+          affectedEntry.disposition
```
jdr81394, line 200: same here

web-client/src/views/DocketRecord/FilingsAndProceedings.tsx
```diff
@@ -274,4 +246,46 @@ export const FilingsAndProceedings = connect<
...
 
+const renderDispositionLinks = (
+  affectedEntry,
+  docketNumber,
+  changeTabSequence,
+  openDocumentDownloadSequence,
+) => {
+  return affectedEntry.dispositionLinkText.map((linkText, index) => (
+    <span
+      className="display-block"
```
jdr81394: When creating new elements, please use the tailwind equivalents of the stylings so we can begin to phase out the scss files

web-client/src/views/Public/PublicFilingsAndProceedings.tsx
```diff
@@ -112,4 +90,36 @@ export const PublicFilingsAndProceedings = connect<
...
-PublicFilingsAndProceedings.displayName = 'PublicFilingsAndProceedings';
+const renderDispositionLinks = (
+  docketEntry,
+  affectedEntry,
+  docketNumber,
+  openDocumentDownloadSequence,
+) => {
+  return affectedEntry.dispositionLinkText.map((linkText, index) => (
+    <span
+      className="display-block"
```
jdr81394: same here about the tailwind

jdr81394 (changes requested):

web-client/src/views/DocketRecord/FilingsAndProceedings.tsx
akuny: Uses Tailwind now.

web-client/src/presenter/computeds/Public/publicCaseDetailHelper.ts
akuny: `affectedDocketEntries` and `affectedByDocketEntries` will always have a disposition:

```typescript
// shared/src/business/entities/EntityConstants.ts
export type DocketEntryRelation = {
  disposition: string;
  docketEntryId: string;
};
```

web-client/src/presenter/computeds/formattedDocketEntries.ts
akuny: See above.

jimlerza (approved):

jdr81394 (approved):

Issue #9733, opened 2026-02-12
BUG: Activity Report Throws 400 Error Selecting Judge Groups


cholly75:
**Describe the Bug**
When selecting the Judge query parameters for the judge Activity Report, selecting any of the following values results in a 400 error:
* All Judges
* All Regular Judges
* All Senior Judges

Selecting any individual judge value or "All Special Trial Judges" does not throw the error.

The following errors are seen in the browser console:

<img width="567" height="117" alt="Image" src="https://github.com/user-attachments/assets/0d9cabb2-7d81-476e-be9f-db3a2b2bff05" />

The "case-worksheets" call throws the response "Invalid search terms"
The "pending-motions" call throws the response "Syntax error at or near $1"

**Business Impact/Reason for Severity**
Report cannot be used for aggregate statistics

**In which environment did you see this bug?**
TEST/STG/PROD

**Who were you logged in as?**
Judge Buch

**What were you doing when you discovered this bug? (Using the application, demoing, smoke tests, testing other functionality, etc.)**

**To Reproduce**
1. Log in as a judge user
2. Go to Reports -> Activity
3. Click the "Show Items For" dropdown menu
4. Select either All, Regular, or Senior judges 

**Expected Behavior**
Screen displays with selection and query can be run

**Actual Behavior**
400 Error is thrown

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Cause of Bug, If Known**

**Process for Logging a Bug:**
* Complete the above information
* Add a severity tag (Critical, High Severity, Medium Severity or Low Severity). See below for priority definition. 

**Severity Definition:**
* Critical Defect
Blocks entire system's or module’s functionality
No workarounds available
Testing cannot proceed further without bug being fixed.

* High-severity Defect
Affects key functionality of an application
There's a workaround, but not obvious or easy
App behaves in a way that is strongly different from the one stated in the requirements

* Medium-severity Defect
A minor function does not behave in a way stated in the requirements.
Workaround is available and easy

* Low-severity Defect
Mostly related to an application’s UI
Doesn't need a workaround, because it doesn't impact functionality

## Definition of Ready for Bugs (Created 2021-10-04)
Definition used: A failure or flaw in the system which produces an incorrect or undesired result that deviates from the expected result or behavior. (Note: Expected results are use cases that have been documented in past user stories as acceptance criteria and test cases, and do not include strange behavior unrelated to use cases.)

[...]

PR #9736, opened 2026-02-12
9733 Bug to Test

akuny:
https://github.com/ustaxcourt/ef-cms/issues/9733

The root cause of this issue is that the `isSeniorJudge` property of the `User` entity is treated as a boolean in application code, but is in fact saved as a varchar in postgres. Because of this inconsistency, strict boolean evaluations in `getPendingMotionDocketEntriesAction.ts` do not work--as a result, no user IDs are passed to the `getAllPendingMotionDocketEntriesForJudge` postgres function that ultimately fires when users select certain options with the "Show items for" dropdown menu in the judge activity report.

`getPendingMotionDocketEntriesAction.ts` is the only place in application code apart from the `User` entity itself that deals with the `isSeniorJudge` property.

The in-place migration in this PR uses postgres's [boolean casting behavior](https://www.postgresql.org/docs/current/datatype-boolean.html) to change `dw_user.is_senior_judge` from a varchar to a boolean. This is an "all or nothing" migration: if any single value in the `dw_user.is_senior_judge` column cannot be cast as a boolean (e.g., the string 'baddata'), postgres will not apply the change and will roll back the entire `ALTER TABLE` statement.

Here is a profile of the `dw_user.is_senior_judge` column in test. NULL records are ignored, so this migration would ultimately affect 91 rows in test.

| is_senior_judge | count |
|-----------------|-------|
| false | 78 |
| true | 13 |
| NULL | 224549 |

It would be prudent to profile this column in other environments before running this migration to ensure that it works.


nechamakrigsman (approved):

jimlerza (approved):

jimlerza:
Not required to merge and deploy this PR to test (since you have already done the work manually), but you could add a step to the `migrate` job in the circle config that checks for unexpected values and bails.

Pseudo-code here, but I'm thinking something like:

```yaml
      - run:
          name: Check dw_user.is_senior_judge for unexpected values
          command: |
            values="$(scripts/postgres/connect.sh --quiet --yes --command='SELECT DISTINCT is_senior_judge FROM dw_user;')"
            # split string into array of strings?
            # check if array has exactly 3 members?
            # or check the members specifically for the expected values?
            exit 1 # if there are unexpected values
```

jimlerza:
Or, actually, you could make that a "Manual Deployment Step" if you want!

You can slap the "Manual Deployment Steps Required" flag on the PR (to stg) and then put this markdown in the PR description:

```markdown
## Manual Deployment Steps

### Before deployment

- [ ] Check for unexpected values in dw_user.is_senior_judge
   \```bash
   . scripts/env/set-env.zsh [env]
   scripts/postgres/connect.sh --quiet --yes --command="SELECT DISTINCT is_senior_judge FROM dw_user;"
   \```
   - ABORT if there are any values other than [empty], true, and false
```

Issue #9522, opened 2025-12-15
Script to Generate a CSV of Geocoded Petitioner Addresses On Demand


jimlerza:
As a Court, so I can determine the optimal placement of new field Courtrooms, I need to be able to generate heat maps in QGIS utilizing petitioners' mailing addresses.

## Pre-Conditions
- #9521

## Acceptance Criteria
- Required parameters for the script:
   - Year
   - Fiscal (boolean)
- CSV should contain the following columns:
   - Date of petition
   - Regular/Small
   - Procedure type
   - Petitioner's address (lat/lng)
   - Requested place of Trial (lat/lng)
   - Actual place of trial (lat/lng)
   - is represented
 

## Pain Avoided/Frustration Saved
Tenille should be able to run the script to generate this CSV herself, on demand as needed.

## Breadth/Pervasiveness of Problem
The existing placement of our field Courtrooms is not reflective of our petition volume and is not optimized for throughput.

## Complexity of Problem (Low, Medium, High) and Why it's Complex
Low

## Notes

- See the [opex-get-case-data-for-qgis](https://github.com/ustaxcourt/ef-cms/tree/opex-get-case-data-for-qgis) branch for a previous (failed?) attempt at this
-  Handle invalid addresses that can't be geocoded

zacharybrown204:
## Testrail
Testrail execution at: https://ustaxcourt.testrail.io/index.php?/runs/view/265

## Summary
We found that by using the census bulk geocoding endpoint we were able to feasibly reduce the time it takes to geocode our data, and just geocode on demand when we run the geocode report script. In our load testing it took somewhere in the ballpark of 15 minutes to geocode all of our data since DAWSON launched (2021 onward). To reduce the complexity and impact on the application we took this approach to only geocode on demand when we run this report.


## Usage
### Basic usage:
```./scripts/reports/petitioner-geodata.ts -y 2021-2025 ```
### Flags
- '-y' :Year(s): single year, comma-separated list, range, or mix (e.g. 2019,2021,2023-2025)
- `-f`: Use fiscal year (starts 10/1) instead of calendar year
- `-b false` Whether or not we should try to populate any missing data for this date range

Issue #9521, opened 2025-12-15
Geocode Petitioner Addresses


jimlerza:
As a Court, so I can eventually determine the optimal placement of new field Courtrooms, I eventually need to be able to generate GPS heat maps utilizing petitioners' mailing addresses. Before I can do either of those things, though, I need petitioners' addresses to be geocoded.

## Pre-Conditions
- None

## Acceptance Criteria
- New field(s) for storing the lat/lng are added where appropriate
- Utilizes the U.S. Census Bureau's geocoding API
- All existing petitioners' addresses are geocoded
- New addresses are geocoded as they are entered
- Existing lat/lng pairs are overwritten when an address is changed

## Pain Avoided/Frustration Saved
This will enable us to generate GPS data for Tenille to use in QGIS.

## Breadth/Pervasiveness of Problem
It is very difficult to do this on demand in a single script because of the rate-limiting imposed by the US Census Bureau's geocoding API

## Complexity of Problem (Low, Medium, High) and Why it's Complex
Medium - will need to use a queue that obeys the US Census Bureau's geocoding API's speed limits

## Notes
- Does this mean we need to start validating new addresses as they are entered?
- If we do decide to validate addresses, how are we going to handle extant addresses that are invalid (there are many)?
- Is it easier/better to go ahead and geocode practitioners' addresses now too?

## Story Definition of Ready (updated on 12/23/22)
The following criteria must be met in order for the user story to be picked up by the Flexion development team.
The user story must: 
- [ ] Is framed in business/user need, the value has been addressed.
- [ ] Includes acceptance criteria
- [ ] Has been refined
- [ ] Pre conditions have been satisfied. 

**Process:** 
Flexion developers and designers will test if the story meets acceptance criteria and test cases in Flexion dev and staging environments (“standard testing”). If additional acceptance criteria or testing scenarios are discovered while the story is in progress, a new story should be created, added to the backlog and prioritized by the product owner. 

### Product Owner
 - [ ] Acceptance criteria have been met and validated on the Court's test environment.
 - [ ] Associated test cases defined in TestRail have been updated if necessary.
 - [ ] Successful test run is performed in TestRail.
 - [ ] User guides are updated if necessary.

### UX
 - [ ] All new functionality has been verified to work with keyboard navigation and screen reader software.
 - [ ] UI should be touch optimized and responsive for external users.

### Engineering
 - [x] Automated test scripts have been written, including visual tests for newly added PDFs.
[...]

PR #9738, opened 2026-02-12
9521 geocode petitioner addresses to test 1770926654

zacharybrown204:
#9521 #9522 

### Database
Adding a new table to hold user address geocode data. 
- Table named dw_user_contact.
- Currently it only holds PK and geocode data
- Each record corresponds to a petitioner in the dw_case petitioners column. So our PK is a docket number and a user ID.

### Geocoding
Added a script to geocode data.
- We ended up taking an on demand approach for geocoding user addresses as we need them, then storing them in our DB for future uses. If any addresses are updated they'll be removed from the table. Like invalidating records in a cache.
- We used an API provided by the US Census Beureu, and an NPM package to streamline the calls to this API.
- - https://geocoding.geo.census.gov/geocoder/
- - https://www.npmjs.com/package/us-census-geocoder
- - - If we want to avoid using this package then the alternative is to create CSV payloads and directly call the API ourselves.
- During manual testing it would take between 1 to 3 minutes to geocode 10,000 records. To geocode all of the addresses since DAWSONs release it will take around 18 calls to the geocoding API.
- Accepts parameters to geocode cases that were recieved within a date range.

### Report
Added a script that can call the geocoding script, and generates a CSV report with address data compelte with geocodes.


scripts/geocoding/backfill-user-geocodes.ts
```diff
@@ -0,0 +1,30 @@
...
+const scriptConfig: ScriptConfig = {
+  description:
+    'backfill-user-geocodes - Geocode addresses for users missing lat/lng',
+  environment: {
+    env: 'ENV',
+    region: 'REGION',
+  },
+  parameters: {
+    batchSize: { default: '10000', type: 'string' },
+    delayMs: { default: '60000', type: 'string' },
```
jimlerza: I notice you're casting these values as numbers. Passing `transform: 'number'` will ensure that doing so is safe.

```suggestion
    batchSize: { default: '10000', transform: 'number', type: 'string' },
    delayMs: { default: '60000', transform: 'number', type: 'string' },
```

scripts/helpers/backfillUserGeocodes.test.ts
```diff
@@ -0,0 +1,457 @@
...
+          ? executeTakeFirstResult
+          : executeResult,
+      ),
+  };
+  const qb = Object.assign(() => chain, { or: () => chain }) as {
+    (): typeof chain;
+    or: () => typeof chain;
+  };
+  return chain;
+};
```
jimlerza: This looks pretty reusable! Do you think this could be moved into a helper?

scripts/reports/petitioner-geodata.ts
```diff
@@ -0,0 +1,194 @@
...
+    },
+    backfillData: {
+      default: true,
+      description:
+        'Whether or not we should try to populate any missing data for this date range',
+      short: 'b',
+      type: 'boolean',
+    },
+  },
+  requireActiveAwsSession: false, // todo: put back
```
jimlerza: Nudge

scripts/reports/petitioner-geodata.ts
```diff
@@ -0,0 +1,194 @@
...
+      description: 'Range start date (YYYY-MM-DD, inclusive)',
+      short: 'f',
+      type: 'string',
+    },
+    toDate: {
+      default: todayDate,
+      description: 'Range end date (YYYY-MM-DD, exclusive)',
+      short: 't',
+      type: 'string',
+    },
```
jimlerza: I'd recommend doing what most of the other reports do for dates:

- one `year` parameter that takes a single year, a comma-separated list of years, a range of years, or a string that is a mix of comma-separated and range (ex. 2019,2021,2023-2025)
- one `fiscal` boolean parameter that determines whether a year starts on 1/1 or 10/1

scripts/reports/petitioner-geodata.ts
```diff
@@ -0,0 +1,194 @@
...
+    { header: 'docket_number_suffix', key: 'docketNumberSuffix' },
+    { header: 'received_year', key: 'receivedYear' },
+    { header: 'procedure_type', key: 'procedureType' },
+    { header: 'case_type', key: 'caseType' },
+    { header: 'party_type', key: 'partyType' },
+    { header: 'status', key: 'status' },
+    { header: 'is_paper', key: 'isPaper' },
+    { header: 'preferred_trial_city', key: 'preferredTrialCity' },
+    { header: 'remote_trial_granted', key: 'remoteTrialGranted' },
+    { header: 'address', key: 'address' },
```
jimlerza, line 187: You might need to format the address a bit by trimming whitespace and replacing newlines with a single space; CSVs don't like linebreaks inside of a cell. We have to do that with case captions - some Blackstone-era cases have linebreaks in their captions. See the `formatCaption` helper for an example.

scripts/geocoding/backfillUserGeocodes.ts
jimlerza, line 1: There are lots of other instances where we've made a helper for a big beefy script, but in all of those cases the helper lives alongside the script. This `scripts/helpers/` directory contains helpers that are generic enough to be used by all scripts. Following that system of organization, I'd think this file and its test should live in the `scripts/geocoding/` directory.

web-api/src/business/useCases/user/updatePetitionerInformationInteractor.ts
```diff
@@ -290,6 +308,11 @@ export const updatePetitionerInformation = async (
       user: authorizedUser,
     });
     serviceUrl = url;
+
+    await invalidateUserContactGeocode(
+      docketNumber,
+      updatedPetitionerData.contactId,
+    );
```
jimlerza, line 315: Nothing to change here, necessarily, but doing this kind of thing without using a transaction gives me heartburn.

web-api/src/persistence/postgres/cases/upsertCases.test.ts
```diff
@@ -0,0 +1,27 @@
...
+  it('inserts cases into dwCase table', async () => {
+    await upsertCases([MOCK_CASE]);
+
+    expect(pgInsertInto).toHaveBeenCalledTimes(1);
+    expect(pgInsertInto).toHaveBeenCalledWith({
+      table: 'dwCase',
+      values: expect.any(Array),
+      onConflictColumns: ['docketNumber'],
+    });
+  });
```
jimlerza, line 32: I realize that `upsertCases` wasn't created in this story, and I appreciate that you've added a test for it. Is this sufficient coverage though?

package.json
```diff
@@ -111,6 +111,7 @@
     "tailwind-merge": "3.4.0",
     "tmp": "0.2.5",
     "tw-animate-css": "1.4.0",
+    "us-census-geocoder": "^1.1.0",
```
jimlerza:
```suggestion
    "us-census-geocoder": "1.1.0",
```

web-api/src/database-schema.ts
jimlerza, line 1: Don't forget to update the diagram!

jimlerza (changes requested):
Looks great! I have just a few suggestions and requests.

scripts/geocoding/backfill-user-geocodes.ts
paulo-filho-ctr: Yes for sure. Much better. Thank you.

scripts/helpers/backfillUserGeocodes.test.ts
paulo-filho-ctr: Indeed, moved into a helper function along with its unit tests.

scripts/geocoding/backfillUserGeocodes.ts
paulo-filho-ctr: I agree. We moved it to `scripts/geocoding`

package.json
paulo-filho-ctr: Awesome catch, you are right, we should pin the dependency.

scripts/reports/petitioner-geodata.ts
paulo-filho-ctr: The report now is mirroring how the other reports do for dates, it now takes a year param (comma-separated) along with a fiscal boolean.

web-api/src/database-schema.ts
nechamakrigsman: Done!

web-api/src/persistence/postgres/cases/upsertCases.test.ts
zacharybrown204: We should probably have a conversation about unit tests for the postgres persistence functions. We haven't had any in the past since we need to mock the only thing those functions are actually doing.

scripts/geocoding/backfillUserGeocodes.ts
```diff
@@ -0,0 +1,171 @@
...
+      )
+      .select([
+        'c.docketNumber',
+        sql<string>`petitioner ->> 'contactId'`.as('userId'),
+        sql<string>`petitioner ->> 'address1'`.as('address1'),
+        sql<string>`petitioner ->> 'state'`.as('state'),
+        sql<string>`petitioner ->> 'city'`.as('city'),
+        sql<string>`petitioner ->> 'postalCode'`.as('zip'),
+      ])
+      .where('c.status', 'not in', ['Closed', 'Dismissed'])
```
devinganey613, line 32: Can any of this code be reused between the two queries?

web-api/src/business/useCases/geocoding/getAddressGeocode.ts
```diff
@@ -0,0 +1,96 @@
...
+      'addressFile',
+      Buffer.isBuffer(csvBuffer) ? csvBuffer : Buffer.from(csvBuffer),
+      {
+        filename: 'addresses.csv',
+        contentType: 'text/csv',
+      },
+    );
+
+    const url =
+      'https://geocoding.geo.census.gov/geocoder/locations/addressbatch';
```
devinganey613: Is this the previous way of geocoding before the us-census-geocoder package was used? Is this function used anywhere now?

devinganey613 (commented):
Cool stuff in here!

web-api/src/business/useCases/geocoding/getAddressGeocode.ts
paulo-filho-ctr: I think you are right, this file is now moot. I'll remove it along with its test file.

scripts/geocoding/backfillUserGeocodes.ts
zacharybrown204: Good shout, merged the base of these queries into one. I also looked for similarities to the query in reports/petitioner-geodata but I didn't see enough overlap there to be worth it

devinganey613 (approved):

scripts/geocoding/backfillUserGeocodes.ts
paulo-filho-ctr: Yes, we moved into its own function.

web-api/src/business/useCases/geocoding/getAddressGeocode.ts
paulo-filho-ctr: Right, great catch. We removed the file alongside with its test file.

jimlerza (approved):

PR #9741, opened 2026-02-13
9733 Bug

akuny:
#### Before deployment

- [ ] Check for unexpected values in dw_user.is_senior_judge
   ```bash
   . scripts/env/set-env.zsh [env]
   scripts/postgres/connect.sh --quiet --yes --command="SELECT DISTINCT is_senior_judge FROM dw_user;"
   ```
   - ABORT if there are any values other than [empty], true, and false

#9733

### Description

One root cause of this issue is that the `isSeniorJudge` property of the `User` entity is treated as a boolean in application code, but is in fact saved as a varchar in postgres. Because of this inconsistency, strict boolean evaluations in `getPendingMotionDocketEntriesAction.ts` do not work--as a result, no user IDs are passed to the `getAllPendingMotionDocketEntriesForJudge` postgres function that ultimately fires when users select certain options with the "Show items for" dropdown menu in the judge activity report.

`getPendingMotionDocketEntriesAction.ts` is the only place in application code apart from the `User` entity itself that deals with the `isSeniorJudge` property.

The in-place migration in this PR uses postgres's [boolean casting behavior](https://www.postgresql.org/docs/current/datatype-boolean.html) to change `dw_user.is_senior_judge` from a varchar to a boolean. This is an "all or nothing" migration: if any single value in the `dw_user.is_senior_judge` column cannot be cast as a boolean (e.g., the string 'baddata'), postgres will not apply the change and will roll back the entire `ALTER TABLE` statement.

The other root cause of this bug was the 20-item limit on array length for query string parameters that the qs library uses by default. This PR explicitly sets the maximum length of query string parameter arrays to 200. See comments in the diffs for rationale.


web-api/src/persistence/postgres/utils/migrate/migrations/2026-02-12T23_52_40Z-change-isSeniorJudge-data-type.ts
```diff
@@ -0,0 +1,24 @@
+import { Kysely, sql } from 'kysely';
+
+// This migration uses postgres's cast operator (::) to convert strings to booleans.
+// See https://www.postgresql.org/docs/current/datatype-boolean.html for additional
+// information about postgres's boolean casting behavior.
+export async function up(db: Kysely<any>): Promise<void> {
+  await sql`
+    ALTER TABLE "dw_user"
+    ALTER COLUMN "is_senior_judge" TYPE boolean
```
jimlerza, line 9: Should we update the ERD diagram?

jdr81394 (approved):

web-api/src/persistence/postgres/utils/migrate/migrations/2026-02-12T23_52_40Z-change-isSeniorJudge-data-type.ts
akuny: The current version of the mermaid ERD only includes primary and foreign keys to emphasize the relationships between the tables, so `dw_user.is_senior_judge` doesn't appear on it.

devinganey613 (approved):

paulo-filho-ctr (approved):

TomElliottFlexion (approved):

jimlerza (approved):

PR #9751, opened 2026-02-18
8784 ts error cleanup arg type mismatch

devinganey613:
Fixing files with typescript errors in #8784


paulo-filho-ctr (approved):

web-client/src/presenter/actions/EditDocketRecordEntry/updateDocketEntryMetaAction.ts
```diff
@@ -24,7 +24,7 @@ export const updateDocketEntryMetaAction = async ({
         docketNumber,
       });
     return path.success();
-  } catch (err) {
+  } catch (err: any) {
```
jimlerza, line 27: I believe `unknown` is safer than `any` here.

```suggestion
  } catch (err: unknown) {
```

jimlerza (approved):

web-client/src/presenter/actions/EditDocketRecordEntry/updateDocketEntryMetaAction.ts
devinganey613: Changing this to unknown creates an error at line 30 when it tries to access the message field that "'err' is of type 'unknown'"

jimlerza: Wah woh! Worth a try. Thanks for the update.

jimlerza:
🙌 

https://github.com/ustaxcourt/ef-cms/actions/runs/22461420246/job/65056435849?pr=9751#step:7:8

PR #9753, opened 2026-02-18
9733 Bug to Test: Explicitly Set Query String Array Length Limit

akuny:
This PR explicitly sets the maximum length of query string parameter arrays to 200 instead of the default of 20. See comments in the diffs for rationale.

This PR also reverts `getAllPendingMotionDocketEntriesForJudge` to its previous state since ultimately that function was not the culprit for the "All Judges" option in the judge activity report not working.

### DIFF
--- a/web-api/src/app-public.ts
+++ b/web-api/src/app-public.ts
@@ -5,11 +5,29 @@ import { lambdaWrapper } from './lambdaWrapper';
 import { set } from 'lodash';
 import cors from 'cors';
 import express from 'express';
+import qs from 'qs';
 
 export const app = express();
 
-// This was default in express 4.x. The default changed in express 5.x, so we have to specify it here
-app.set('query parser', 'extended');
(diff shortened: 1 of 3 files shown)


paulo-filho-ctr (approved):

devinganey613:
Do you still need the code that coerces the query string parameters into an array? From my understanding it was the qs limit that was causing the parameters to be converted to an object, which shouldn't happen any more.

nechamakrigsman (approved):

zacharybrown204 (approved):

akuny:
> Do you still need the code that coerces the query string parameters into an array? From my understanding it was the qs limit that was causing the parameters to be converted to an object, which shouldn't happen any more.

Good point, that code won't make it into staging.

devinganey613 (approved):

jimlerza (approved):
Unlike Copilot, I don't mind the repeated comment.

PR #9756, opened 2026-02-19
9521 geocode petitioner addresses to test 1771528333

zacharybrown204:
Dropping a FK that we initially thought to be valid. The petitioners contactIds line up well with the users in the users_on_case table, but there is a large gap between the users on that table and the user table. 

Dropping a where conditition in the backfill query that was limiting data too much.

### DIFF
--- a/scripts/geocoding/backfillUserGeocodes.ts
+++ b/scripts/geocoding/backfillUserGeocodes.ts
@@ -29,7 +29,6 @@ const userMissingGeocodesBaseQuery = async () =>
           .onRef(sql`petitioner->>'contactId'`, '=', 'uc.userId')
           .onRef('c.docketNumber', '=', 'uc.docketNumber'),
       )
-      .where('c.status', 'not in', ['Closed', 'Dismissed'])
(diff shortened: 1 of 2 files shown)


paulo-filho-ctr (approved):

devinganey613 (approved):

nechamakrigsman (approved):

scripts/geocoding/backfillUserGeocodes.ts
```diff
@@ -29,7 +29,6 @@ const userMissingGeocodesBaseQuery = async () =>
           .onRef(sql`petitioner->>'contactId'`, '=', 'uc.userId')
           .onRef('c.docketNumber', '=', 'uc.docketNumber'),
       )
-      .where('c.status', 'not in', ['Closed', 'Dismissed'])
```
jimlerza, line 32: Oh! I should've caught this the first time around, but "Dismissed" isn't a status. "Closed - Dismissed" is though. We have the `CLOSED_CASE_STATUSES` constant for this.

(Yes, I realize this code was removed, and that this PR has already been merged)

PR #9783, opened 2026-02-26
8821-refactor-file-selection-components-to-test-1772140792

paulo-filho-ctr:
### Summary
It is necessary to use the legacy uswds styles otherwise it will break the ui


zacharybrown204 (approved):

kampofo (approved):
So if I gather correctly? No tailwind on these pages/components until they're replaced by Dawson ui equivalents?

devinganey613 (approved):

Issue #9697, opened 2026-02-02
Clean up shared directory


TomElliottFlexion:
The shared directory contains _many_ modules that cannot, in fact, be shared by the client and the api. The idea behind this directory was that, combined with a well-maintained applicationContext, one could use these shared methods irrespective of context: any boundary-spanning calls would be replaced by a context-specific function. We have stopped maintaining and moved away from the applicationContext injection pattern, making this directory structure more confusing than beneficial. 

## Pre-Conditions
- None

## Acceptance Criteria
- All files in the shared directory can be called by both the client and the api
- All files specific to either of the above are moved to their respective directories

## Pain Avoided/Frustration Saved
- Less time spend validating functionality and searching for duplicative functions

## Breadth/Pervasiveness of Problem
- App-wide

## Complexity of the Problem (Low, Medium, High) and Why it's Complex
- Low

## Notes

-Orders, Petitions Service,

paulo-filho-ctr:
We removed the following because the files, methods and functions, were not being used. It would be good to test flows that references the them:

- Batch
- Scan
- Order
- documentServiceEmail
- petitionServiceEmail
- validatePdfHeader

paulo-filho-ctr:
## Test Notes
### Regression and workflow checklist
This PR is mostly boundary cleanup and module relocation, so QA should focus on behavior parity: same user-visible outcomes, no broken routes, no missing API wiring, and no role-based regressions.

- High-priority regression workflows (manual)
- Docket entry QC and edit flow
- Complete QC on a docket entry, then verify docket record reflects status change.
- Edit docket entry metadata and confirm updates persist after refresh.
- Seal, unseal, and strike a docket entry; verify visibility changes for internal vs public users.
- Filing workflows
- File external document (standard and at least one nonstandard scenario).
- Add/edit paper filing and confirm docket entry details and served parties behavior remain correct.
- Validate PDF upload flow (valid and invalid files) and confirm user-facing error handling still works.
- Petition workflows
- Create case from electronic petition.
- Create case from paper petition.
- Verify party/contact data mapping still behaves correctly for special petitioner types.
- Search workflows
- Case advanced search.
- Order advanced search.
- Opinion advanced search.
- Public case search and public order search.
- Trial session workflows
- Create trial session, add/remove case, update working copy notes.
- Set notices for calendared trial session.
- Generate trial calendar/minute-sheet related outputs.
- Practitioner and user workflows
- Practitioner lookup by bar number.
[...]

PR #9786, opened 2026-02-27
9745 - Amgen Support Large Cases

codyseibert:
- Large cases with many docket entries can exceed the API Gateway's 6MB response size limit on GET /cases/:docketNumber, causing failures when loading case detail
  - Splits docket entry loading into a separate paginated endpoint (GET /cases/:docketNumber/docket-entries) that returns entries in chunks of 1,000, capped at page 20
   (~20,000 entries max)
  - The existing case endpoint now excludes docket entries from its response (via excludeDocketEntries query param), keeping it well under the size limit              
  - getCaseAction transparently fetches all pages and merges them back, so all 50+ sequences that depend on it continue working with no changes


web-client/src/presenter/actions/getCaseAction.ts
```diff
@@ -18,11 +18,48 @@ export const getCaseAction = async ({
...
       docketNumber,
-    });
+    }),
+    applicationContext
+      .getUseCases()
+      .getCaseDocketEntriesInteractor(applicationContext, {
+        docketNumber,
+        page: 0,
+      }),
+  ]);
```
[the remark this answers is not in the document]
codyseibert: each page is 1,000 and the largest case we suspect is 8,000 total, so the "20" concurrent isn't a real issue

web-client/src/presenter/actions/getCaseAction.ts
```diff
@@ -18,11 +18,48 @@ export const getCaseAction = async ({
...
+        applicationContext
+          .getUseCases()
+          .getCaseDocketEntriesInteractor(applicationContext, {
+            docketNumber,
+            page,
+          }),
+      );
+    }
+    const results = await Promise.all(pagePromises);
+    for (const result of results) {
```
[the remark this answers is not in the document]
codyseibert: same as above, this will just cause the front end to take a lot longer to show the docket record, ignoring comment

shared/src/business/useCases/getCaseDocketEntriesInteractor.ts
```diff
@@ -0,0 +1,86 @@
...
+      ...docketEntry,
+      qcComplete: !!workItem?.completedAt,
+      qcViewed: !!workItem?.isRead,
+      workItemId: workItem?.workItemId,
+    };
+  });
+
+  const totalCount = allDocketEntries.length;
+  const start = page * PAGE_SIZE;
+  const docketEntries = allDocketEntries.slice(start, start + PAGE_SIZE);
```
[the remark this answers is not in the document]
codyseibert: we have to fetch all the docket entries so that the case will properly set it's computed values.  Getting away from this requires another large amount of refactoring.  This PR is already 245 file changes, so we opted to just do the bare minimum so all the existing functionality will work for the amgen case.

shared/src/proxies/getCaseProxy.ts
```diff
@@ -18,5 +18,6 @@ export const getCaseInteractor = (
   return get({
     applicationContext,
     endpoint: `/cases/${docketNumber}`,
+    params: { excludeDocketEntries: true },
   });
```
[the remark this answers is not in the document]
codyseibert: with the refactoring to the getCaseAction, this is expected

web-api/src/business/useCases/getCaseDocketEntriesInteractor.ts
```diff
@@ -0,0 +1,86 @@
+import { Case } from '@shared/business/entities/cases/Case';
+import { CaseFactory } from '@web-api/business/entities/cases/CaseFactory';
+import { NotFoundError, UnauthorizedError } from '@web-api/errors/errors';
+import {
+  UnknownAuthUser,
+  isAuthUser,
+} from '@shared/business/entities/authUser/AuthUser';
+import { getCaseByDocketNumber } from '@web-api/persistence/postgres/cases/getCaseByDocketNumber';
+import { getWorkItemsByDocketNumber } from '@web-api/persistence/postgres/workitems/getWorkItemsByDocketNumber';
```
jimlerza, line 9: This is a clean architecture violation, no? If we're importing from `web-api` shouldn't this interactor belong only in `web-api` and not `shared`?

web-client/src/views/DocketRecord/VirtualizedDocumentList.tsx
jimlerza, line 1: Can we add a Cypress test for this?

web-client/src/views/DocketRecord/DocketRecord.tsx
jimlerza, line 1: I think we need some Cypress coverage that asserts these new paginators.

web-client/src/views/DocketRecord/DocumentViewer.tsx
jimlerza, line 1: Could probably use a Cypress test covering this too.

jimlerza (changes requested):
I noticed a few small things, nothing major.

In addition to the file-specific comments, we need some swagger doc updates:

- `GET /cases/{docketNumber}/docket-entries`
- `GET /public-api/cases/{docketNumber}/docket-entries`
- `POST /public-api/cases/{docketNumber}/generate-docket-record`
- `GET /public-api/docket-record-status/{jobId}`
- the new `excludeDocketEntries` query parameter on `GET /cases/{docketNumber}`

web-api/src/business/useCases/getCaseDocketEntriesInteractor.ts
alexdziarn: moved it over to web-api, these files were made before #9697

jimlerza (approved):

paulo-filho-ctr (approved):

nechamakrigsman (approved):

web-api/src/persistence/postgres/utils/operation/pgInsertInto.ts
```diff
@@ -24,11 +26,15 @@ export const pgInsertInto = async <T extends keyof Database>({
       let query = writer.insertInto(table).values(values);
 
       if (onConflictColumns.length > 0) {
```
devinganey613, line 28: Should this also check if excludeFromUpdateColumns is populated?

web-api/src/lambdas/cases/updateQcCompleteForTrialLambda.ts
```diff
@@ -12,7 +12,7 @@ export const updateQcCompleteForTrialLambda = (
   event,
   authorizedUser: UnknownAuthUser,
 ) =>
-  genericHandler(event, async ({ applicationContext }): Promise<RawCase> => {
+  genericHandler(event, async ({ applicationContext }) => {
     return await updateQcCompleteForTrialInteractor(
```
devinganey613, line 16: Remove return here to match other lambdas

devinganey613 (approved):

Issue #9588, opened 2026-01-05
Order Response Workflow - Enhancements


cholly75:
As a Court, so that I am using clear language in generated orders for responses to motions, I need to adjust some of the generated text.

Assortment of feedback on the generated orders from the Order Response custom order workflow.

## Acceptance Criteria
* When the case for which the Order is being generated has more than one petitioner associated with it, where the generated text would be "Petitioner", it should generate instead "Petitioners"
* Change "(Document no. ###)" to "(doc. no. ###)"
* When a case is part of a consolidated group, change "Lead case" to "lead case"
* Add commas after all generated dates

## Test Cases
Test cases have been added to Test Rail

### Product Owner
 - [ ] Acceptance criteria have been met and validated on the Court's test environment.
 - [ ] Associated test cases defined in TestRail have been updated if necessary.
 - [ ] Successful test run is performed in TestRail.
 - [ ] User guides are updated if necessary.

### UX
 - [ ] All new functionality has been verified to work with keyboard navigation and screen reader software.
 - [ ] UI should be touch optimized and responsive for external users.

### Engineering
 - [ ] Automated test scripts have been written, including visual tests for newly added PDFs.
 - [ ] Successful test run is performed in TestRail.
 - [ ] New screens have been added to cypress accessibility axe.
 - [ ] Interactors should validate entities before calling persistence methods.
 - [ ] Types have been added to all added and updated functions.
 - [ ] Code refactored for clarity and to remove any known technical debt.
 - [ ] Acceptance criteria for the story has been met.
 - [ ] All schema changes are documented in the Data Dictionary (DD), and the Entity Relationship Diagram (ERD) is updated as appropriate.
 - [ ] If there are special deployment instructions, they have been added to the `CHANGES.md` file and the PR description.
 - [ ] Code that resides in the shared folder that only runs on the API or browser has been moved to either /web-client or /web-api.

Evelyn-Durbin:
Prerefinement: Demo please! :)

PR #9793, opened 2026-03-02
9588 order response to test 1772484510

devinganey613:
## [#9588](https://github.com/ustaxcourt/ef-cms/issues/9588) Updated text in order to response template:

- Changed "petitioner" to "petitioners" where multiple petitioners are on the case
- Changed "Lead case" to "lead case"
- Changed "Document no." to "doc. no."
- Added commas after all dates


AbelTaxCourt (approved):

jimlerza (approved):
TIL the word "movant"

PR #9796, opened 2026-03-03
9521-geocode-petitioner-addresses-to-test-1772574033

paulo-filho-ctr:
#9521 #9522

### Database
Adding a new table to hold user address geocode data.
- Table named dw_user_contact.
- Currently, it only holds PK and geocode data
- Each record corresponds to a petitioner in the dw_case petitioners column. So our PK is a docket number and a user ID.

### Geocoding
Added a script to geocode data.
- We ended up taking an on demand approach for geocoding user addresses as we need them, then storing them in our DB for future uses. If any addresses are updated they'll be removed from the table. Like invalidating records in a cache.
- We used an API provided by the US Census Beureu, and an NPM package to streamline the calls to this API.
- - https://geocoding.geo.census.gov/geocoder/
- - https://www.npmjs.com/package/us-census-geocoder
- - - If we want to avoid using this package then the alternative is to create CSV payloads and directly call the API ourselves.
- During manual testing it would take between 1 to 3 minutes to geocode 10,000 records. To geocode all of the addresses since DAWSONs release it will take around 18 calls to the geocoding API.
- Accepts parameters to geocode cases that were recieved within a date range.

### Report
Added a script that can call the geocoding script, and generates a CSV report with address data compelte with geocodes.


docs/postgres/schema/erd.mmd
```diff
@@ -107,6 +107,14 @@ erDiagram
         string userId FK "User ID"
     }
 
+    dwUserContact {
+        string userId PK,FK "Associated user ID"
```
devinganey613: This FK was dropped, so the erd should be updated

devinganey613 (approved):

kampofo (approved):

Issue #9622, opened 2026-01-09
BUG: Results Tables are not sorting by Page Number correctly


ttlenard:
**Describe the Bug**
This was reported to us by a Judge via the DAWSON Support helpdesk.  He reported page number sorting issues on the Today's Orders Page. The example he provided is that there is an Order that is 10 pages in length, multiple Orders that are 2 pages in length, and multiple orders that are 1 page in length. If you click on the page number header, the Orders sort by page number.  Notice that the 10 page order is below the 2 page orders in the screen grab below. (See screen grab of this below)

Note that this was also reported to via the helpdesk by a Tax blogger. We should probably fix this.

After additional review, it appears that other pages that display a page number value in a results table are also not sorting by the page number appropriately.

I found the same sorting issue on the following pages:
- Order Search page
- Opinion Search page

Note: The Docket Record appears to be sorting by page number appropriately

**Need to check the Today's Opinions page. Are there other pages that we display page numbers that is also doing this?**

**Business Impact/Reason for Severity**
low

**In which environment did you see this bug?**
Prod

**Who were you logged in as?**
Public user

**What were you doing when you discovered this bug? (Using the application, demoing, smoke tests, testing other functionality, etc.)**
This bug was reported to us by a Court user via the DAWSON support helpdesk.

**To Reproduce**
1. Go to the Today's Orders page
2. Make sure that there is a 10 page order along with some 2 page orders and 1 page orders (not sure if this bug is happening with other page numbers)
3. Click on the Pages header to sort the page
4. See that the Sorting is not working appropriately, the 10 page order is not displayed at the top when sorting by descending order.
5. Next go to the Order search page and the Opinion Search page (I was a public user, need to also check party and internal user views)
6. Add in a date and click search - sort the results by page number, and notice that the sorting is incorrect.

**Expected Behavior**
Sorting should be displaying the correct page number order when sorting by pages

**Actual Behavior**
A clear and concise description of what actually happened.
Sorting is not displaying correctly when sorting the Today's Orders page by number of pages

**Screenshots**

<img width="1483" height="1207" alt="Image" src="https://github.com/user-attachments/assets/3b01732f-a4ba-4b6f-aa7a-933efcae9ad7" />

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Cause of Bug, If Known**

**Process for Logging a Bug:**
* Complete the above information
[...]

PR #9800, opened 2026-03-06
9622 bug to test

akuny:
https://github.com/ustaxcourt/ef-cms/issues/9622

This PR addresses the following sorting issues on the public app:

- Today's Opinions: wire up sorting logic similar to that used by Today's Orders
- Today's Orders: handle page counts as numbers to sort them numerically
- Order and Opinion search: fix sorting by page count


devinganey613 (approved):

devinganey613 (approved):

web-client/src/presenter/computeds/Public/todaysOpinionsHelper.ts
```diff
@@ -24,5 +31,44 @@ export const todaysOpinionsHelper = (
...
+
+      const compare1 = tableSort.sortOrder === DESCENDING ? opinionB : opinionA;
+      const compare2 = tableSort.sortOrder === DESCENDING ? opinionA : opinionB;
+
+      if (tableSort.sortField === 'docketNumber') {
+        sortNumber = Case.docketNumberSort(
+          compare1.docketNumber,
+          compare2.docketNumber,
+        );
+      } else if (tableSort.sortField === 'numberOfPages') {
```
TomElliottFlexion, line 50: Do you think it's worth introducing constants for `docketNumber` and `numberOfPages` (and similar)? Genuinely not sure myself. Definitely wouldn't need to happen before test

TomElliottFlexion (approved):

web-client/src/presenter/computeds/Public/todaysOpinionsHelper.ts
akuny: I think it's fine as-is, but I'm not hard-bound to that opinion and could be persuaded otherwise.

PR #9815, opened 2026-03-13
9622 Bug

akuny:
https://github.com/ustaxcourt/ef-cms/issues/9622

This PR addresses the following sorting issues on the public app:

- Today's Opinions: wire up sorting logic similar to that used by Today's Orders
- Today's Orders: handle page counts as numbers to sort them numerically
- Order and Opinion search: fix sorting by page count


jimlerza (approved):

web-client/src/presenter/computeds/Public/todaysOpinionsHelper.test.ts
```diff
@@ -70,6 +79,12 @@ describe('todaysOpinionsHelper', () => {
     expect(result.formattedCurrentDate).toEqual(formattedCurrentDate);
   });
 
+  it('should return sortOptions', () => {
+    const result = runCompute(todaysOpinionsHelper, { state });
+    expect(Array.isArray(result.sortOptions)).toBeTruthy();
+    expect(result.sortOptions.length).toBeGreaterThan(0);
```
ChristieLaferriereTaxCourt: This is a great start here. Since `sortOptions` comes from a constants file, is there a way we could use that to make a more precise assertion here rather than just checking `toBeGreaterThan(0)`?

ChristieLaferriereTaxCourt (approved):

web-client/src/presenter/computeds/Public/todaysOpinionsHelper.test.ts
akuny: Good call, made that change.

Issue #9772, opened 2026-02-24
BUG: Leading space in Name field when adding Practitioner causes incorrect Bar Number


ttlenard:
**Describe the Bug**
I received a message from an admissions clerk letting me know that after admitting a new Practitioner to the Court, the Practitioner's Bar number was displaying incorrectly. What had happened was that the clerk was copying practitioner contact information from a document outside of DAWSON, and then pasting it into DAWSON. They happened to copy a leading space in the practitioner's Name, and then this caused the Bar number to generate with a blank space, rather than the letter.

Bar numbers are generated in DAWSON in the following manner:
BA26242
B = Practitioner's first letter of their last name
A = Practitioner's first letter of their first name
26 = This displays the current calendar year
242 = The sequential number that represents the number that the Practitioner was admitted this year (so this is the 242 practitioner to be admitted to the Court this year)

For this particular scenario, the BAR number that got generated was B 26242 (missing the first letter of their first name since there was a leading space when the clerk copied the name in)

After talking about this bug with the team, it was suggested that instead of ONLY removing the ability to include leading spaces for Practitioner names, we should look into removing leading spaces in many of our data fields in the app.

We would probably want to ensure that we keep the ability to have leading spaces for long text fields though, in case we needed to have paragraph indentations?

Any other gotchas if we were to implement this for more than just practitioner name fields?

**Business Impact/Reason for Severity**
low

**In which environment did you see this bug?**
Prod, Test

**Who were you logged in as?**
Admissions Clerk

**What were you doing when you discovered this bug? (Using the application, demoing, smoke tests, testing other functionality, etc.)**
This issue was reported to us by an Admissions clerk 

**To Reproduce**
1. Log in as an Admissions Clerk
2. Click on the advanced link
3. Click on the Practitioner Tab
4. Click on the Add New Practitioner button
5. Fill in the required fields, making sure to "accidentally" include a leading space in the First name field
6. Click Save
7. The Bar number will include a space
8. Repeat the steps above but this time leave a leading space in the last name field when creating a new practitioner - you should have the same result - an incorrectly formatted Bar number.
9. The only way to currently fix this is to go into Postgres and update the Bar Number for the Practitioner.
10. The problem with this is that OpenSearch does not reindex this w/o an edit to the practitioner record, so we could wind up with bad data in DAWSON.

**Expected Behavior**
Do not allow leading spaces in our data fields

**Actual Behavior**
A clear and concise description of what actually happened.
[...]

PR #9824, opened 2026-03-18
9772 Bug Fix to Test

akuny:
https://github.com/ustaxcourt/ef-cms/issues/9772

This PR ensures that practitioner first and last names are trimmed of any leading or trailing spaces before being used to generate a bar number. It also adds validation logic to ensure that the first name and last name passed to the server by the client start with a letter.


jimlerza (approved):

web-api/src/business/useCases/practitioner/createPractitionerUserInteractor.ts
```diff
@@ -20,14 +21,23 @@ export const createPractitionerUserInteractor = async (
...
-    user,
-  });
+  if (
+    !userDataCanGenerateValidBarNumber({
+      firstName: user.firstName,
+      lastName: user.lastName,
+    })
+  ) {
+    throw new InvalidRequest(
+      'Unable to generate a bar number: first and last names must start with a letter after trimming whitespace.',
```
devinganey613: Would this error be shown to the user on the frontend so they know what the issue is? Not sure how the error messages get passed around

devinganey613 (approved):

web-api/src/business/useCases/practitioner/createPractitionerUserInteractor.ts
akuny: Great question, here's what the user would see from their point of view. That's coming from `web-client/src/presenter/actions/createPractitionerUserAction.ts`. We could pass the error message along to the user if we like. @jimlerza what do you think?

user-pov

Issue #8847, opened 2024-10-04
Catching invalid entities after schema changes


zachrog:
## Acceptance Criteria
The below solutions are explored and implemented if they prove to be good options:
- (Proactive) Generate a fingerprint of validation rules. Notify devs when fingerprint changes.
- (Proactive) Utility to validate the whole database or a portion of the database.
- (Better Team Awareness) Creating Folders for persisted entities vs. non DB entities
- (Reactive) Logging when a persisted entity is invalid

## Pain Avoided/Frustration Saved
Making schema changes causes existing data in the database to become invalid.

## Complexity of Problem (Low, Medium, High) and Why it's Complex
  High

## Notes
This problem has been explored and voted on by the team(10/4/2024). The ideas to solve this problem can be found in this branch `devex-auto-detect-validation-errors` inside of this markdown file `docs/entities/entity-validation.md`.

The fingerprinting method has been explored and a partial solution exists. To run checkout `devex-auto-detect-validation-errors` branch and run `npx ts-node --transpile-only scripts/getEntityIdentifiers.ts`
  
These are experiments, not commitments. Let's see what works and eventually get rid of things that don't to reduce complexity.

8477 8527 8919

ben-eccles-gunnison:
Considering the major risk of breaking staging if a schema change without a migration makes it in, do we want to reconsider the priority on this one?

jimlerza:
# Testing Feedback

@kampofo and @zacharybrown204

- [ ] this needs a test in TestRail

Something like:

- make breaking change to joi validation, deploy -> expect fail
- make breaking schema change via migration, deploy -> expect fail
- validate obfuscated emails -> expect pass

PR #9843, opened 2026-03-23
8847 catching invalid entities

kampofo:
#8847 

## Entity Validation
Adding a script to run through our main entities, pull them from PG, and run them against our JOI validation rules.
Adding a circle CI job to run aforementioned script, and fail a build if the script finds any validation errors. This script hashes the JOI validation rules and stores those hashes in SSM. If any of those hashes are different then we run validation on those corresponding entities. Alternatively if there was a database migration then we run the whole validation step again.
If needed you can force a build to run validation on everything by setting the `entity-validation-required` parameter in SSM.

## Email Obfuscation
As part of this effort we realized that a few entities were throwing errors in test because the email addresses we obfuscate during a glue job can be much longer than they originally were. This leads to issues where some fields had a character limit that were exceeded when those email addresses got longer. To counteract this we revisited that obfuscation logic to use shorter hashes and a shorter domain. 
As we used shorter hashes, collisions were possible when hashing emails, so we added collision handling to chain hash the email until a free hash was found. This does make the set of obfuscated emails less deterministic as the order the emails are found in matter for the emails that do have collisions. 

The code for this was already merged to production, that half of the glue job is run, but it hasn't been merged to test or stage yet. 
#9858 

## Email Filtering
As a result of changing the domain used by the obfuscated emails, we wanted to ensure that no emails could leak out of the test environment. So we added a check to the email sending lambda function to check the To Address of emails, and if they were going to an email with a test domain, it would be filtered out and the lambda return success. 
We did also try using a IAM policy to restrict sending emails to those addresses, but that caused the email sending lambda to throw an error, the SQS message for the email would remain in queue and be retried until it timed out. 

This code for the email filtering was already merged into TEST before the last glue job. #9876


scripts/entity-validation/cleanupEntityValidation.ts
```diff
@@ -0,0 +1,27 @@
...
+}
+
+async function cleanupEntityValidation(): Promise<number> {
+  try {
+    await deleteSSMItem('entity-validation-required');
+  } catch (error) {
+    console.error('failed to delete ssm parameter for entity validation');
+    console.error(error);
+    process.exit(1);
+  }
```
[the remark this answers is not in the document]
kampofo: If no ssm item is there to be deleted then that's ok. We're more concerned with the step breaking.

scripts/entity-validation/createValidationIdentifier.ts
```diff
@@ -0,0 +1,38 @@
...
+import joi from 'joi';
+
+function stringifyEverything(input): string {
+  let bigOlString = '';
+  function stringifyStuff(obj) {
+    const theType = typeof obj;
+    if (theType === 'object' || Array.isArray(obj)) {
+      for (const key in obj) {
+        stringifyStuff(obj[key]);
+      }
```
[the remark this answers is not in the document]
kampofo: the only examples of big int values are in ttl columns in our table so we dont believe these are values we'd need to stringify

scripts/entity-validation/entityValidation.ts
```diff
@@ -0,0 +1,179 @@
...
+// TODO: Replace with dynamic entity discovery from shared/src/business/entities
+
+const ENTITIES_OF_CASE = [
+  'Case.ts',
+  'DocketEntry.ts',
+  'IrsPractitioner.ts',
+  'PrivatePractitioner.ts',
+  'Correspondence.ts',
+];
+
```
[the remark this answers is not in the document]
kampofo: We added a static member VALIDATION_RULES to DocketEntry, so now it's being properly picked up and finger printed by the script.

scripts/entity-validation/workers/entityValidationWorker.ts
```diff
@@ -0,0 +1,67 @@
...
+      if (validationErrors) {
+        if (entityName === 'Case') {
+          errors.push(
+            `Validation errors for ${entityName} ${record.docketNumber}: ${JSON.stringify(validationErrors)}`,
+          );
+        } else {
+          const idField = `${camelCase(entityName)}Id`;
+          errors.push(
+            `Validation errors for ${entityName} ${record[idField]}: ${JSON.stringify(validationErrors)}`,
+          );
```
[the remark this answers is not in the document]
kampofo: This is fine, this is purely for the console output, and even though the TrialSessionWorkingCopy records use trialSessionId its clearer what entity is failing with the name we gave it trialSessionWorkingCopyId.

.circleci/config.yml
```diff
@@ -281,6 +281,29 @@ jobs:
             cd ./scripts/postgres
             DB_HOST="$POSTGRES_HOST" ./create-rds-users.sh
 
+  validate-entities:
+    docker:
+      - image: *efcms-docker-image
+        aws_auth:
+          aws_access_key_id: $AWS_ACCESS_KEY_ID
+          aws_secret_access_key: $AWS_SECRET_ACCESS_KEY
+    resource_class: xlarge
```
jimlerza, line 288: Did you run some tests and determine that this step needs to be performed on an `xlarge` runner? Or is this copypasta?

.circleci/config.yml
```diff
@@ -281,6 +281,29 @@ jobs:
             cd ./scripts/postgres
             DB_HOST="$POSTGRES_HOST" ./create-rds-users.sh
 
+  validate-entities:
```
jimlerza, line 282: Why is this a separate job? Each job adds about 2 minutes of non-skippable environment setup to the workflow.

.circleci/config.yml
```diff
@@ -927,6 +954,10 @@ workflows:
           <<: *only-deployed-lower-environments
           requires:
             - deploy
+      - validate-entities:
+          <<: *only-deployed-lower-environments
+          requires:
+            - migrate
```
jimlerza, line 958: You added this to the workflow that deploys to lower environments, but not to the workflow that deploys to prod. Why?

jimlerza (changes requested):

.circleci/config.yml
jimlerza: I know! I know! So we can fix the data and re-run from failed!

.circleci/config.yml
zacharybrown204: Bit of a copy pasta, we did look at what the other jobs were running and figured the load would be similar to the migrate step. The worry was with pulling down large sets of data we would hit resource issues.

jimlerza (approved):

devinganey613 (approved):

PR #9864, opened 2026-03-31
9745: Fix pagination duplicates - add docketEntryId as secondary sort…

nechamakrigsman:
… key;


codyseibert:
will be merged as part of #9880

PR #9866, opened 2026-04-01
9805 add public copy link to working copy page to test 1775056646

devinganey613:


web-client/src/presenter/actions/gotoPrintPublicSessionCopyPreviewAction.ts
```diff
@@ -0,0 +1,27 @@
+import { state } from '@web-client/presenter/app.cerebral';
+
+/**
+ * sets the state.page which shows the docket record print preview (default, optional) and/or opens the document in a new tab (optional)
+ * @param {object} providers the providers object
+ * @param {object} providers.get the cerebral get helper function
+ * @param {object} providers.props the cerebral props object
+ * @param {object} providers.store the cerebral store used for setting the state.pdfPreviewUrl
+ */
```
TomElliottFlexion: I don't have a strong opinion myself, but a while ago the team resolved to remove jsDocs in favor of accurate typing and strict TS support. At this point half the team is adding jsDocs while the other half is removing them. We should probably all agree on an approach again

TomElliottFlexion (approved):
Nice job

web-client/src/presenter/actions/gotoPrintPublicSessionCopyPreviewAction.ts
paulo-filho-ctr: I think is a valid point. I replaced the JS Docs in favor of types.

zacharybrown204 (approved):

Issue #9811, opened 2026-03-12
Add Sortability to My Cases View


cholly75:
As a petitioner or practitioner, I want to sort my Open Cases and Closed Cases by any column in ascending or descending order, so that I can easily organize and review my cases based on what is most important to me.

For petitioners and practitioners, the Recent Filings screen was implemented with full spectrum sorting. We would like to extend this capability to the My Cases screen as well. This will enable petitioners and practitioners to sort the case lists in the Open Cases and Closed Cases views by any available column. Additionally, we want to add pagination to this screen for both the Open Cases and Closed Cases views. Sorting applies to the entire result set, not just the page currently displayed. 

## Pre-Conditions
#9810 

## Acceptance Criteria
* Users can sort the Open Cases view by any of the following columns:
  - Docket No.
  - Case Title
  - Filed Date
  - Filing Fee
  - Case Status

* Users can sort the Closed Cases view by any of the above columns (with the exception of Case Status - this column is not on the Close Cases list).
* Sorting can be applied in ascending or descending order for each column.
* When a user clicks a column header, the system updates the sort order and refreshes the list accordingly.
* Pagination is shown when the result set exceeds 100 cases.
* Changing pages retains the chosen sort order.
* Sorting reorders the entire result set, not just the currently visible page.
* The UI provides a clear indication of the active sort column and whether the sort is ascending or descending.

## Test Cases
Test Cases are in Test Rail

### Product Owner
 - [ ] Acceptance criteria have been met and validated on the Court's test environment.
 - [ ] Associated test cases defined in TestRail have been updated if necessary.
 - [ ] Successful test run is performed in TestRail.
 - [ ] User guides are updated if necessary.

### UX
 - [ ] All new functionality has been verified to work with keyboard navigation and screen reader software.
 - [ ] UI should be touch optimized and responsive for external users.

### Engineering
 - [ ] Automated test scripts have been written, including visual tests for newly added PDFs.
 - [ ] Successful test run is performed in TestRail.
 - [ ] New screens have been added to cypress accessibility axe.
 - [ ] Interactors should validate entities before calling persistence methods.
 - [ ] Types have been added to all added and updated functions.
 - [ ] Code refactored for clarity and to remove any known technical debt.
 - [ ] Acceptance criteria for the story has been met.
 - [ ] All schema changes are documented in the Data Dictionary (DD), and the Entity Relationship Diagram (ERD) is updated as appropriate.
[...]

Issue #9810, opened 2026-03-12
Add Case Status to My Cases/Recent Filings Views


cholly75:
As a petitioner or practitioner, I want to see the Case Status for each of my cases and access clear explanations of what each status means, so that I can understand the progress of my case and know what to expect next.

This feature displays Case Status values for all cases shown in My Cases and Recent Filings when a petitioner or practitioner is logged into DAWSON. We will also need to provide information explaining the meaning of any Case Status value, helping users better understand where their case stands in the Tax Court process and hopefully heading off potential support contacts.

## Acceptance Criteria
* When a logged‑in petitioner or practitioner views the My Cases screen, each listed case displays its current Case Status.
  - This only applies to the Open Cases view
* When a logged‑in petitioner or practitioner views the Recent Filings screen, each listed case displays its current Case Status.
* On the Recent Filings page, the Case Status column is eligible for sorting
* For every Case Status value displayed, users have a way to access information describing what that status means.
* The Case Status explanation is easily accessible from both My Cases and Recent Filings.
* The Case Status explanation includes clear guidance on Case Status values.
* The Case Status displayed in My Cases and Recent Filings matches current status of the case as seen by internal Court users.
  - **EXCEPTION**: When a Case Status is "Calendared", include both the date and location of the trial session for which the case is calendared as part of the Case Status value (example: see Messages or QC display for how this is done currently in the app)
* Case Status values appear consistently across both My Cases and Recent Filings.

## Notes
* Court to provide "decoder ring" for Case Status verbiage
* My Cases response currently pulls in Case Status; Recent Filings does not

## Test Cases
Test Cases are in Test Rail

### Product Owner
 - [ ] Acceptance criteria have been met and validated on the Court's test environment.
 - [ ] Associated test cases defined in TestRail have been updated if necessary.
 - [ ] Successful test run is performed in TestRail.
 - [ ] User guides are updated if necessary.

### UX
 - [ ] All new functionality has been verified to work with keyboard navigation and screen reader software.
 - [ ] UI should be touch optimized and responsive for external users.

### Engineering
 - [ ] Automated test scripts have been written, including visual tests for newly added PDFs.
 - [ ] Successful test run is performed in TestRail.
 - [ ] New screens have been added to cypress accessibility axe.
 - [ ] Interactors should validate entities before calling persistence methods.
 - [ ] Types have been added to all added and updated functions.
 - [ ] Code refactored for clarity and to remove any known technical debt.
[...]

PR #9873, opened 2026-04-03
9810 - Adding case status to my cases and recent filings tables

zacharybrown204:
https://github.com/ustaxcourt/ef-cms/issues/9810 https://github.com/ustaxcourt/ef-cms/issues/9811

Adds case status column to external users case table and recent filings table.
Adds sorting for Case Status on the recent filings table
Adds sorting to external users case table and changes the "load more" to pagination for handling larger sets of data.


jimlerza (approved):

devinganey613 (approved):
Added a suggestion for more test coverage, but otherwise looks great!

web-client/src/presenter/computeds/recentFilingsHelper.test.ts
```diff
@@ -20,6 +20,7 @@ const createFiling = (overrides: Partial<RecentFiling> = {}): RecentFiling => ({
   isFileAttached: true,
   eventCode: 'P',
   servedAt: '2024-01-15T10:00:00.000Z',
+  status: 'new',
```
devinganey613: Could add a test for sorting by status

zacharybrown204: Good call, there was only one test for sorting a field that wasn't the default sort. So I added tests for the other fields.

PR #9876, opened 2026-04-03
8847 restrict sending emails to test domains to test 1775245876

zacharybrown204:
As part of 8847 we are changing the domain used by obfuscated user emails in the test environment. To be sure we don't send any errant emails out to a domain that shouldn't exist we are adding filtering on the send-email function to restrict sending emails out to this domain.

### DIFF
--- a/web-api/src/dispatchers/ses/sendBulkTemplatedEmail.ts
+++ b/web-api/src/dispatchers/ses/sendBulkTemplatedEmail.ts
@@ -68,8 +68,25 @@ export const sendWithRetry = async ({
   params: SendBulkTemplatedEmailCommandInput;
   retryCount: number;
 }) => {
+  const testEmailDomains = ['@ustc.gov', '@example.com'];
   const sesClient: SESClient = applicationContext.getEmailClient();
   const { MAX_SES_RETRIES } = applicationContext.getConstants();
+
+  let sendingToTestEmail = false;
+  params.Destinations?.forEach(bulkDestination => {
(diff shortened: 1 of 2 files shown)


jimlerza (approved):

kampofo (approved):

web-api/src/dispatchers/ses/sendBulkTemplatedEmail.ts
```diff
@@ -68,9 +68,34 @@ export const sendWithRetry = async ({
...
+  if (validDestinations.length === 0) {
+    applicationContext.logger.info(
+      'No valid destinations found, exiting',
+      params,
+    );
+    return;
+  }
+  params.Destinations = validDestinations;
+
   applicationContext.logger.info('Bulk Email Params', params);
```
devinganey613: We might want to put this logging before the new logic. I know we use this to check in the logs if an email was sent while testing in test if we aren't using a real email. So when the function returns early we would no longer see this information.

akuny (approved):

paulo-filho-ctr (approved):

web-api/src/dispatchers/ses/sendBulkTemplatedEmail.ts
jimlerza: This is an excellent suggestion. Good eye!

zacharybrown204: Good catch, I moved that log statement above the filtering

devinganey613 (approved):

PR #9888, opened 2026-04-08
8847 catching invalid entities to test 1775676144

zacharybrown204:
#8847 

## Entity Validation
Adding a script to run through our main entities, pull them from PG, and run them against our JOI validation rules.
Adding a circle CI job to run aforementioned script, and fail a build if the script finds any validation errors. This script hashes the JOI validation rules and stores those hashes in SSM. If any of those hashes are different then we run validation on those corresponding entities. Alternatively if there was a database migration then we run the whole validation step again.
If needed you can force a build to run validation on everything by setting the `entity-validation-required` parameter in SSM.

## Email Obfuscation
As part of this effort we realized that a few entities were throwing errors in test because the email addresses we obfuscate during a glue job can be much longer than they originally were. This leads to issues where some fields had a character limit that were exceeded when those email addresses got longer. To counteract this we revisited that obfuscation logic to use shorter hashes and a shorter domain. 
As we used shorter hashes, collisions were possible when hashing emails, so we added collision handling to chain hash the email until a free hash was found. This does make the set of obfuscated emails less deterministic as the order the emails are found in matter for the emails that do have collisions. 

The code for this was already merged to production, that half of the glue job is run, but it hasn't been merged to test or stage yet. 
#9858 

## Email Filtering
As a result of changing the domain used by the obfuscated emails, we wanted to ensure that no emails could leak out of the test environment. So we added a check to the email sending lambda function to check the To Address of emails, and if they were going to an email with a test domain, it would be filtered out and the lambda return success. 
We did also try using a IAM policy to restrict sending emails to those addresses, but that caused the email sending lambda to throw an error, the SQS message for the email would remain in queue and be retried until it timed out. 

This code for the email filtering was already merged into TEST before the last glue job. #9876


.circleci/config.yml
```diff
@@ -281,6 +281,29 @@ jobs:
...
+  validate-entities:
+    docker:
+      - image: *efcms-docker-image
+        aws_auth:
+          aws_access_key_id: $AWS_ACCESS_KEY_ID
+          aws_secret_access_key: $AWS_SECRET_ACCESS_KEY
+    resource_class: xlarge
+    steps:
+      - setup_remote_docker:
+          version: docker23
```
devinganey613: Should this be docker29 now?

scripts/entity-validation/entityValidation.ts
```diff
@@ -0,0 +1,189 @@
...
+  const validationErrors = await validateEntitiesWithNewRules(changedEntities);
+
+  if (validationErrors.length > 0) {
+    console.log(validationErrors);
+    return validationErrors.length;
+  } else {
+    console.log(
+      'Entity validation fingerprints have changed. Writing new fingerprint to SSM.',
+    );
+    const ssmUpdateSucceeded = await putSSMItem(SSM_KEY, newFingerprint);
```
devinganey613, line 164: Should this update to the SSM still happen even if validation fails? Also seems like it will happen if it hasn't changed but entityValidationRequired is true, which might be fine.

devinganey613 (approved):
Some cool stuff in here!

scripts/helpers/consoleSpinner.ts
```diff
@@ -0,0 +1,43 @@
+import { getCurrentDateTimeInMillis } from '@shared/business/utilities/DateHandler';
+
+const SPINNER_FRAMES = ['⠋', '⠙', '⠹', '⠸', '⠼', '⠴', '⠦', '⠧', '⠇', '⠏'];
```
TomElliottFlexion, line 3: Nice

TomElliottFlexion (approved):

scripts/entity-validation/entityValidation.ts
kampofo: if validation fails, we return the validation errors length, otherwise we continue to update the value in SSM. With the idea being that we address any validation errors and only update SSM when validation has passed. Also the validation runs automatically if theres a new migration.

.circleci/config.yml
kampofo: good catch!

Issue #9458, opened 2025-12-04
Allow Editing of Trial Sessions After Start Date


cholly75:
As a Case Services Supervisor, so that I can maintain accurate trial session records for the Court, I need to be able to update specific information about that trial session after the session has begun.

Currently trial sessions cannot be updated after the start date has been reached. The Court has identified certain fields/conditions in which it would be useful for a CSS to be able to update information even after the trial session has begun. 

**UPDATE 4/14/26** While discussing this story during development, we found that the edit link on trial sessions that start TODAY is currently available to Trial clerks, Petitions Clerks, docket clerks, etc, and it shouldn't be.  The devs working on this story determined that they can fix this issue and roll the fix into this story.

## Acceptance Criteria
* Case Services Supervisor role can update the following fields on a Calendared Trial Session after the Start Date has been reached:
  - Estimated End Date
  - Trial Clerk (also Alternate trial clerk name if "other" is selected from the Trial clerk dropdown menu)
  - IRS Calendar Administrator fields (Search for IRS calendar administrator name, IRS calendar administrator name, Email, Phone mumber)
  - Court Reporter
* Editing and saving changes to any of these fields does NOT trigger any automated notices or service
* **UPDATE 4/14/26** - Remove edit link on Trial sessions that start TODAY from visibility for users that have the ability to edit trial sessions, with the exception of the Case Services Supervisor. CSS should still have the ability to edit Trial sessions that start TODAY, following the A/C above (only the 4 fields are available for editing of sessions that start TODAY)

## Test Cases
Available in TestRail

### Product Owner
 - [ ] Acceptance criteria have been met and validated on the Court's test environment.
 - [ ] Associated test cases defined in TestRail have been updated if necessary.
 - [ ] Successful test run is performed in TestRail.
 - [ ] User guides are updated if necessary.

### UX
 - [ ] All new functionality has been verified to work with keyboard navigation and screen reader software.
 - [ ] UI should be touch optimized and responsive for external users.

### Engineering
 - [ ] Automated test scripts have been written, including visual tests for newly added PDFs.
 - [ ] Successful test run is performed in TestRail.
 - [ ] New screens have been added to cypress accessibility axe.
 - [ ] Interactors should validate entities before calling persistence methods.
 - [ ] Types have been added to all added and updated functions.
 - [ ] Code refactored for clarity and to remove any known technical debt.
 - [ ] Acceptance criteria for the story has been met.
[...]

PR #9897, opened 2026-04-10
9458 edit trial session

alexdziarn:
[testrail](https://ustaxcourt.testrail.io/index.php?/runs/view/356&group_by=cases:section_id&group_order=asc&display=tree)
- moved validation rules for alternateTrialClerkName from NewTrialSession to TrialSession class.
- created new set of validation rules for a new class called EditTrialSession
- added seed data for trial sessions because were are unable to create trial sessions that start in the past through the UI
- refactored canEdit boolean in formattedTrialSessionDetails to a string that contains edit options of 'all', 'limited' or 'none'
- added estimatedEndDate values to trial session seed data since trial sessions can no longer be created without the estimatedEndDate
- added backend validation and refactored some existing code in updateTrialSessionInteractor


web-api/src/business/useCases/trialSessions/updateTrialSessionInteractor.ts
```diff
@@ -50,41 +55,100 @@ export const updateTrialSession = async (
...
+    'courtReporter',
+    'dismissedAlertForNott',
+    'estimatedEndDate',
+    'irsCalendarAdministrator',
+    'irsCalendarAdministratorInfo',
+    'term',
+    'termYear',
+    'trialClerk',
+    'trialClerkId',
+  ];
```
alexdziarn, line 107: I'd like to know if there are any more hidden fields that change. For example, 'dismissedAlertForNott', 'term', and 'termYear' change based on form values from the frontend.

paulo-filho-ctr (approved):

web-client/src/views/TrialSessions/SessionInformationForm.tsx
```diff
@@ -205,7 +227,7 @@ export const SessionInformationForm = connect<
                 }
                 id="estimated-end-date"
                 label="Estimated end date"
-                minDate={addTrialSessionInformationHelper.today}
+                minDate={addTrialSessionInformationHelper.tomorrow}
```
jimlerza, line 230: Shouldn't this be relative to the start date and not relative to real time?

Example - what if a session (that already started) ends early? Doesn't this make it that the soonest it can end is tomorrow?

We probably need some tests around this and other date boundaries.

cypress/local-only/tests/integration/trialSession/edit-a-trial-session-CSS.cy.ts
```diff
@@ -0,0 +1,439 @@
...
+      .type(formatDateString(tomorrowISO, FORMATS.MMDDYYYY));
+    cy.get('[data-testid="submit-edit-trial-session"]').click();
+    cy.get('[data-testid="success-alert"]').should('be.visible');
+    cy.get('[data-testid="trial-session-date"]').should(
+      'contain',
+      formatDateString(tomorrowISO, FORMATS.MMDDYY),
+    );
+  });
+
+  it('should not allow non case services supervisor roles to edit ongoing trial sessions', () => {
```
jimlerza, line 46: Do we need permutations for the other non-case services supervisor roles here? Or just a check to assert that trial clerks and docket clerks have the same trial session edit permissions?

web-client/src/views/TrialSessions/SessionInformationForm.tsx
jimlerza, line 304: This `<select>`probably needs to be disabled for folks that don't have the edit permission since this is not in `LIMITED_EDITABLE_FIELDS`.

cypress/local-only/tests/integration/trialSession/edit-a-trial-session-CSS.cy.ts
jimlerza, line 1: Should we add some checks to `cypress/local-only/tests/accessibility/trialSession/petitions-clerk.cy.ts` that assert that the disabled inputs don't trip up screen readers?

web-client/src/views/TrialSessions/SessionInformationForm.tsx
jimlerza: Update on this: Tenille says she found this in testing and isn't concerned about it since the date can still be typed manually.

cypress/local-only/tests/integration/trialSession/edit-a-trial-session-CSS.cy.ts
alexdziarn: We included this as a safety check because when we were building this we had an issue with permissions.

web-client/src/views/TrialSessions/SessionInformationForm.tsx
alexdziarn: I don't think this field is shown, but I added the disabled value anyways.

cypress/local-only/tests/integration/trialSession/edit-a-trial-session-CSS.cy.ts
alexdziarn: I'm not too sure how to test for this, but from what I am reading, since we are using `disabled` instead of `aria-disabled` we should have no problems with screen readers detecting dimmed inputs.

jimlerza: Yeah, I realized after I wrote this that the `checkA11y()` call in that test may be comprehensive enough to check for stuff like that anyway.

jimlerza (approved):

PR #9901, opened 2026-04-13
9458-edit-trial-session-to-test-1776105652

paulo-filho-ctr:
[testrail](https://ustaxcourt.testrail.io/index.php?/runs/view/344&group_by=cases:section_id&group_order=asc&display=tree)
- moved validation rule for alternateTrialClerkName from NewTrialSession to TrialSession class.
- created new set of validation rules for new class called EditTrialSession
- added seed data for ongoing trial session because were are unable to create trial sessions that start in the past through the UI


zacharybrown204 (approved):

web-client/src/presenter/computeds/formattedTrialSessionDetails.ts
```diff
@@ -32,21 +33,21 @@ export const formattedTrialSessionDetails = (
   let disableHybridFilter = false;
   let canDelete = false;
   let canEdit = false;
+  let canEditOngoingSession = false;
 
   let alertMessageForNOTT: string | undefined;
   let chambersPhoneNumber: string | undefined;
 
   const trialSession = get(state.trialSession);
+  if (!trialSession.trialSessionId) return {} as FormatTrialSessionHelperType;
```
TomElliottFlexion, line 42: Is type assertion the only option here?  If the helper doesn't guarantee a `trialSession`, I don't think that the type should assert that it does

web-client/src/views/TrialSessions/SessionInformationForm.tsx
```diff
@@ -100,6 +106,10 @@ export const SessionInformationForm = connect<
                   });
                   validateTrialSessionSequence();
                 }}
+                showDisabledDate={
+                  !addingTrialSession &&
+                  formattedTrialSessionDetails.canEditOngoingSession
+                }
```
TomElliottFlexion: Are `disabled` and `showDisabledDate` the same thing here?

TomElliottFlexion: Ah I think I see, we want to show the previous value but disable further inputs?

TomElliottFlexion (approved):

akuny (approved):
Looks good apart from `edit-ongoing-trial-session.cy.ts` failing locally and in CI.

web-client/src/views/TrialSessions/SessionInformationForm.tsx
alexdziarn: the DataSelector component would make the date field a blank string '' when disabled is true. We added showDisableDate to this component as a way to override this functionality and display the date. I noticed this is confusing and changed showDisabledDate to always true for this use case.

PR #9385, opened 2025-11-17
8477 s3 refactor

AbelTaxCourt:
- Added documentStorageId field to the Docket Entry entity instead of using docketEntryId as the S3 key so docket entries on unconsolidated cases can update their own S3 documents independently from the consolidated group
(Supports #8477)
- This PR includes the second migration that backfills the documentStorageId column on the Docket Entry table using docketEntryId
- This PR includes the functionality that uses the new documentStorageId field as the key when retrieving S3 documents
- See https://github.com/ustaxcourt/ef-cms/pull/9706

Issue #8919, opened 2025-02-10
Restrict Service of Simultaneous Documents In Consolidated Groups


cholly75:
As a Court, so that I do not inadvertently serve simultaneous documents, I need to prevent service of documents from a member case in the consolidated group.

NOTE: this is being fixed as part of the effort in #8477.

## Pre-Conditions
ustaxcourt/ef-cms#8477 

## Story Definition of Ready (updated on 12/23/22)
The following criteria must be met in order for the user story to be picked up by the Flexion development team.
The user story must: 
- [ ] Is framed in business/user need, the value has been addressed.
- [ ] Includes acceptance criteria
- [ ] Has been refined
- [ ] Pre conditions have been satisfied. 

**Process:** 
Flexion developers and designers will test if the story meets acceptance criteria and test cases in Flexion dev and staging environments (“standard testing”). If additional acceptance criteria or testing scenarios are discovered while the story is in progress, a new story should be created, added to the backlog and prioritized by the product owner. 

### Product Owner
 - [ ] Acceptance criteria have been met and validated on the Court's test environment.
 - [ ] Associated test cases defined in TestRail have been updated if necessary.
 - [ ] Successful test run is performed in TestRail.
 - [ ] User guides are updated if necessary.

### UX
 - [ ] All new functionality has been verified to work with keyboard navigation and screen reader software.
 - [ ] UI should be touch optimized and responsive for external users.

### Engineering
 - [ ] Automated test scripts have been written, including visual tests for newly added PDFs.
 - [ ] Successful test run is performed in TestRail.
 - [ ] New screens have been added to cypress accessibility axe.
 - [ ] Interactors should validate entities before calling persistence methods.
 - [ ] Types have been added to all added and updated functions.
 - [ ] Code refactored for clarity and to remove any known technical debt.
 - [ ] Acceptance criteria for the story has been met.
 - [ ] If there are special deployment instructions, they have been added to the `CHANGES.md` file and the PR description.
 - [ ] Code that resides in the shared folder that only runs on the API or browser has been moved to either /web-client or /web-api.

katiecissell:
UX Notes:
[Figma FIle](https://www.figma.com/design/Xn6a8dRy2DTA3P0o2MVTFm/USTC-%7C-10103%2C-10596-Consolidation%3A-Cover-Sheets-for-E-filed-Documents?node-id=84-653&t=CQg7WArzTZB89rR9-1)

[...]

Issue #8527, opened 2023-10-16
BUG: Multi-docketing UI appears when serving Simultaneous paper-filings in Lead case


ttlenard:
**Describe the Bug**
When a clerk goes to serve a paper-filed Simultaneous-type document in the Lead case of a consolidated group, they are presented with the option to multi-docket it across the consolidated group. Docket would need to serve each of the simultaneous documents individually, so this UI should not be activated in this situation.

Docket clerks may think they can/should multi-docket paper filed Simultaneous filings because they are getting the multi-docketing popup modal. When they select serve to all in group, the system actually doesn't serve to all, and instead just serves to the lead case, leading to confusion and the potential that clerks won't serve the simultaneous documents in the member cases. 

Note that this does not present as an issue with non-Simultaneous paper-filed documents.

**Business Impact/Reason for Severity**
medium

**In which environment did you see this bug?**
TEST, STG, PROD

**Who were you logged in as?**
Docket clerk

**What were you doing when you discovered this bug? (Using the application, demoing, smoke tests, testing other functionality, etc.)**
Testing other functionality

**To Reproduce**
1. Log in as a docket clerk and navigate to the lead case in a consolidated group
2. Click on the Create Menu and then select Paper filing
3. Select any "Simultaneous" type document from the document type picklist
4. Fill in the other required fields and then click Save Enty
5. Next, click on the document link for the simultaneous document that you just saved
5. Click on the Serve link from the document view
6. Notice that you will receive the multi-docketing pop-up window
7. Leave the default and save and serve to all in consolidated group
8. Once it is finished, you will see the green confirmation banner indicating that it was saved to the docket record (this is not the multi-docketing confirmation).
9. Notice that the document was filed in the lead case, and in no other member cases.
10. It is actually good that the document wasn't multi-docketed, the problem here is that they are getting the multi-docketing popup when they shouldn't be.
11. Refer to similar bug 10075 (this one was with eFiled simultaneous docs)

**Expected Behavior**
The paper filed simultaneous document should not get the multi-docketable pop-up modal when the docket clerk clicks the link to serve the document.  

**Actual Behavior**
A clear and concise description of what actually happened.
When a docket clerk is serving a paper filed simultaneous type document, they are getting the multi-docketable pop-up modal asking them if they want to serve it across the group. This modal shouldn't be showing up for these paper filings. Even if the clerk selects all to file it across the group, it is only filing it in the lead case (good).

**Screenshots**

[...]

PR #9903, opened 2026-04-13
8477 story to test

AbelTaxCourt:
- Overhauled Docket Entry QC, Edit, and Serve processes
- Added Work Item grouping and consolidated Work Item assignments in QC Inboxes and Outboxes
- Added counts to Individual QC Inbox, In Progress, and Outbox tables
- Updated counts to handle one Work Item row per consolidated group 
- Updated Initiate Service and Paper Service modals to support serving across multidocketed entries
- Grouped NODC Paper Service PDFs into a single PDF for easier printing
- Added multiDocketedOn to the Docket Entry Entity to track which cases the Docket Entry is currently filed across
- Added originallyFiledDocketNumber to the Docket Entry Entity to track the original filing case of the Docket Entry, which fixed a bug preventing access to the correct Previously-filed Documents when QCing or Editing from a case in which the document was not originally filed from
- Added documentStorageId to the Docket Entry entity (https://github.com/ustaxcourt/ef-cms/pull/9706 https://github.com/ustaxcourt/ef-cms/pull/9385) instead of using docketEntryId as the S3 key so docket entries on unconsolidated cases can update their own S3 documents independently from the consolidated group
#8477 
#8919 
#8527


web-client/src/presenter/actions/getDefaultAttachmentViewerDocumentToDisplayAction.ts
```diff
@@ -48,6 +54,6 @@ export const getDefaultAttachmentViewerDocumentToDisplayAction = ({
   }
 
   return {
-    messageViewerDocumentToDisplay,
+    messageViewerDocumentToDisplay: messageViewerDocumentToDisplay || {},
   };
```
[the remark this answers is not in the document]
AbelTaxCourt: messageViewerDocumentToDisplay is null checked in other actions

web-api/src/business/useCases/caseConsolidation/removeConsolidatedCasesInteractor.ts
```diff
@@ -110,6 +110,8 @@ const removeConsolidatedCases = async (
     docketNumbers: docketNumbersToRemove,
   });
 
+  const docketEnIdsToUpdate: Set<string> = new Set();
+  const docketEntriesToUpdate: Set<RawDocketEntry> = new Set();
```
alexdziarn: can these just be arrays?

alexdziarn: docketEnIdsToUpdate should be a set to remove duplicates, but docketEntriesToUodate can be an array

shared/src/business/utilities/getDocumentStorageId.ts
```diff
@@ -1,8 +1,17 @@
...
+}: {
+  caseDetail: RawCase;
+  docketEntryId: string;
+}): string => {
   const { docketEntries } = caseDetail;
   const docketEntry = docketEntries.find(de => {
     return de.docketEntryId === docketEntryId;
   });
 
+  if (!docketEntry)
```
akuny, line 13: Could be worth covering this conditional with a unit test.

web-api/src/business/useCaseHelper/docketEntry/noticeOfDocketChangeHelper.ts
```diff
@@ -0,0 +1,102 @@
...
+      relationship: DOCUMENT_RELATIONSHIPS.PRIMARY,
+    },
+    { authorizedUser, petitioners },
+  ).validate();
+
+export const needsNewCoversheet = ({
+  currentDocketEntry,
+  updatedDocketEntry,
+}: {
+  currentDocketEntry: any;
```
akuny: Could we type these parameters as DocketEntry instead of any? It looks like this function is used only in `web-api/src/business/useCases/docketEntry/completeDocketEntryQCInteractor.ts`, and I think it's only ever passed DocketEntry instances in that interactor.

web-api/src/business/useCaseHelper/service/processCaseForService.ts
```diff
@@ -0,0 +1,46 @@
...
+  electronicParties,
+  loadPdfDocument,
+  newPdfDoc,
+  PDFDocument,
+}: {
+  applicationContext: ServerApplicationContext;
+  caseEntity: Case;
+  docketEntryId: string;
+  electronicParties?: { email: string; name: string }[];
+  loadPdfDocument: () => Promise<any>;
```
akuny: Could this return type be `Promise<Uint8Array>`? That appears to be what the callback passed as `loadPdfDocument` in `web-api/src/business/useCaseHelper/serveDocumentAndGetPaperServicePdf.ts` would return.

web-api/src/business/useCaseHelper/service/processCaseForService.ts
```diff
@@ -0,0 +1,46 @@
...
+  newPdfDoc: any;
```
akuny: Could this be typed as `PDFDocument`?

web-api/src/business/useCases/docketEntry/updateDocketEntryMetaInteractor.ts
```diff
@@ -191,6 +197,59 @@ export const updateDocketEntryMeta = async (
...
+      docketNumbers: originalDocketEntry.multiDocketedOn,
+    });
+
+    const updatedDocketEntries = casesToUpdate
+      .map(caseRecord => {
+        const { docketNumber } = caseRecord;
+        if (docketNumber === caseToUpdate.docketNumber) {
+          return;
+        }
+        const consolidatedCaseEntity = new Case(caseRecord, { authorizedUser });
```
akuny, line 212: Lines 212-245 aren't covered by the existing unit test. May be worth having this under test if possible.

web-api/src/business/useCaseHelper/coverSheet/updateDocketEntriesWithPageCount.ts
```diff
@@ -0,0 +1,73 @@
...
+  docketEntryId,
+  docketNumber,
+  pageCount,
+}: {
+  authorizedUser: UnknownAuthUser;
+  caseEntity: Case;
+  consolidatedCases: Array<{
+    docketNumber: string;
+    documentNumber?: string;
+  }> | null;
```
alexdziarn: we should type this as OR undefined instead of null. This currently works because the conditional checks for a truthy or falsy value

web-client/src/presenter/actions/switchErrorActionFactory.ts
```diff
@@ -1,8 +1,17 @@
 export const switchErrorActionFactory =
   (errorMap: object) =>
   ({ path, props }) => {
+    let responseData = '';
+    try {
+      const data = props?.error?.originalError?.body;
+      responseData =
+        typeof data === 'string' ? data : JSON.stringify(data || '');
+    } catch (e) {
+      responseData = '';
```
akuny: If the intent behind this try/catch block is to silently swallow a `JSON.stringify` error, we don't need this assignment and can just leave the catch block empty.

web-client/src/presenter/computeds/confirmInitiateServiceModalHelper.ts
```diff
@@ -1,161 +1,144 @@
...
+  if (party.role === ROLES.privatePractitioner) {
+    return 'Petitioner Counsel';
+  } else if (party.role === ROLES.irsPractitioner) {
+    return 'Respondent Counsel';
+  } else {
+    return CONTACT_TYPE_TITLES[party.contactType];
+  }
+};
+
+export const shouldAllowMultiDocketing = ({ docketEntry, isLead }) => {
```
akuny: Feels like this should live in `@shared/business/utilities` since it's strictly business logic.

akuny (approved):
Just a few comments for consideration, no blockers as far as I can see.

web-api/src/business/useCaseHelper/docketEntry/noticeOfDocketChangeHelper.ts
```diff
@@ -0,0 +1,102 @@
...
+      });
+
+      if (parsedTitle) {
+        documentTitleForNotice = parsedTitle;
+      }
+
+      const { filedBy: parsedFiledBy } = parsedEditState;
+      if (parsedFiledBy) {
+        filedBy = parsedFiledBy;
+      }
```
alexdziarn:
```suggestion
      filedBy = parsedEditState.filedBy ?? filedBy;
```

paulo-filho-ctr (approved):

jimlerza (approved):
Amazing work!

web-api/src/business/useCaseHelper/docketEntry/noticeOfDocketChangeHelper.ts
```diff
@@ -0,0 +1,99 @@
...
+export const buildUpdatedDocketEntry = ({
+  authorizedUser,
+  docketEntry,
+  editableFields,
+  petitioners,
+}: {
+  authorizedUser: any;
+  docketEntry: any;
+  editableFields: Record<string, unknown>;
+  petitioners: any[];
```
jimlerza: I think we probably know these types! Can add a return type, too.

web-api/src/business/useCaseHelper/docketEntry/fileAndServeDocumentOnOneCase.ts
```diff
@@ -16,27 +16,24 @@ import { DocketEntry } from '@shared/business/entities/DocketEntry';
...
-  subjectCaseDocketNumber,
   user,
   caseHasDeadline = undefined,
 }: {
   caseEntity: any;
   docketEntryEntity: DocketEntry;
-  subjectCaseDocketNumber: any;
   user: any;
   caseHasDeadline?: boolean;
 }) => {
```
jimlerza: We know these types as well.

web-api/src/business/useCaseHelper/docketEntry/noticeOfDocketChangeHelper.ts
```diff
@@ -0,0 +1,99 @@
...
+      if (parsedTitle) {
+        documentTitleForNotice = parsedTitle;
+      }
+
+      filedBy = parsedEditState.filedBy ?? filedBy;
+    } catch (err) {
+      getDawsonLogger().error(
+        'Failed to parse docketEntry.editState for notice of docket change',
+      );
+    }
```
jimlerza: Should we still throw this? It looks like we don't have to, because we still have valid data to return, but I don't know if not having the parsed title is bad here.

jimlerza (approved):

Issue #9009, opened 2025-07-10
Add db transaction support (formerly flexion#10705)


Mwindo:
## Pre-Conditions
ustaxcourt/ef-cms#8828 and ustaxcourt/ef-cms#8978 

## Acceptance Criteria
We can do something like `await withTransaction(callback)`, and everything in callback either all makes it to the database and OpenSearch or none of it makes it to the database and OpenSearch.

## Pain Avoided/Frustration Saved
No more "shoot the process failed in between steps and now we have slightly corrupt data." One of the first things that prompted us to switch to Postgres was investigating how to support transactions.

## Breadth/Pervasiveness of Problem
Difficult to say, but big.

## Complexity of Problem (Low, Medium, High) and Why it's Complex
High. A really good solution will:

- Allow us to be fine-grained, i.e., wrap some things in transactions and not others
- Allow us to nest transactions
- Not prematurely write results to OpenSearch
- Work well with our current connection/locking setup

Mwindo:
I didn't add a point value to this because I have gone rogue: this is something that Jim has wanted for a long time, and I want to make it happen before we leave. No transaction support is crazy. I made this ticket because I think I have a fairly good solution to this, and I wanted to track it.

Mwindo:
The basic idea, which I want to discuss with a couple of other developers:

1. Top-level mutex to prevent two concurrent processes (e.g., Promise.all([process1, process2]) both trying to start a transaction on the same connection. (Only one transaction per connection is allowed with Postgres.) We can probably do this with kysely's [transaction API](https://kysely.dev/docs/category/transactions).
2. Within some `withTransaction` block, we want all children processes to operate on the same transaction/connection, but we don't want to try starting a _new_ transaction--we are already in one! So use [AsyncLocalStorage](https://docs.nestjs.com/recipes/async-local-storage) to pass in context to each child process so it knows it is already in a transaction.

Mwindo:
I think this is ready to go whenever we want to test it. I have added a lot of notes [here](https://github.com/ustaxcourt/ef-cms/pull/6349) for developer reference.

Mwindo:
My suspicion is that I will not be here to see this across the finish line. If I could, here is what I would do:

1. Merge the to-test PR to test, using and testing `withTransaction` in a low-stakes place.
2. Make a small update to the feature branch to use `withTransaction` in a less low-stakes place. Maybe consult with @jimlerza about where to use it. Create a to-test PR with this update and test.
[...]

PR #9905, opened 2026-04-14
10705 dxox to test 1776177548

devinganey613:
New PR for #9009. See this PR for an explanation of what this entails https://github.com/ustaxcourt/ef-cms/pull/6349

We also added more tests and error handing when running functions passed into onTransactionCommit.


scripts/user/cleanup-cypress-test-accounts.ts
```diff
@@ -120,8 +120,8 @@ async function deleteCognitoUser({
         UserPoolId: userPoolId,
         Username: email.toLowerCase(),
       });
-    } catch (err: any) {
-      const name = err?.name || err?.__type;
+    } catch (err: unknown) {
+      const name = (err as { name?: string; __type?: string })?.name || (err as { name?: string; __type?: string })?.__type;
```
paulo-filho-ctr, line 124: Nice! That is some wild typings!

web-api/src/business/useCases/saveCaseDetailInternalEditInteractor.ts
```diff
@@ -166,22 +170,26 @@ export const saveCaseDetailInternalEdit = async (
...
-
-    await upsertWorkItems({
-      workItems: [workItemEntity.validate().toRawObject()],
-    });
   }
 
-  await updateCaseAndAssociations({
-    authorizedUser,
-    caseToUpdate: caseEntity,
+  await withTransaction(async () => {
```
AbelTaxCourt, line 182:
```suggestion
    await withTransaction(async () => {
    if (workItemEntity) {
      await upsertWorkItems({
        workItems: [workItemEntity.validate().toRawObject()],
      });
    }

    return updateCaseAndAssociations({
      authorizedUser,
      caseToUpdate: caseEntity,
    });
  });
```

We may want to await instead of return updateCaseAndAssociations as it doesn't seem like we use the return value of it.

web-api/src/persistence/postgres/utils/transactions.ts
```diff
@@ -0,0 +1,55 @@
...
+    };
+
+    // Run the user-supplied function within the transaction context.
+    return ConnectionStore.run(transactionStore, () => fn());
+  });
+
+  // After the transaction completes successfully, run the onCommit callbacks.
+  if (transactionStore.onCommitCallbacks?.length) {
+    await settlePromises(transactionStore.onCommitCallbacks.map(cb => cb()));
+  }
```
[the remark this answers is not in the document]
akuny: I think that Copilot is onto something here.

akuny (changes requested):
This looks good apart from the issue that Copilot flagged in [web-api/src/persistence/postgres/utils/transactions.ts](https://github.com/ustaxcourt/ef-cms/pull/9905/files/52a9790485c9cee6b86b76eb759b6dcf07a4dd52#diff-f065be5ad47c75f91106966035b87297226d53b833c01d03ba7fbc15bf07b458): that seems worth addressing.

web-api/src/business/useCases/saveCaseDetailInternalEditInteractor.ts
devinganey613: Good point. It's tricky because on the staging branch we often are returning the value and using it, but on test due to 9745 we are not. Good to keep it consistent on test.

web-api/src/persistence/postgres/utils/transactions.ts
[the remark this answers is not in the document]
devinganey613: Didn't see your reply before I left mine. Will look more into this!

web-api/src/business/useCases/archiveDraftDocumentInteractor.ts
```diff
@@ -47,15 +48,17 @@ export const archiveDraftDocument = async (
...
+    if (workItem) {
+      await deleteWorkItem({
+        workItem,
+      });
+    }
 
-  await updateCaseAndAssociations({
-    authorizedUser,
-    caseToUpdate: caseEntity,
+    return updateCaseAndAssociations({
```
AbelTaxCourt: await instead of return here as well maybe

AbelTaxCourt (approved):

nechamakrigsman (approved):

akuny (approved):

web-api/src/persistence/postgres/databaseConnection.ts
```diff
@@ -0,0 +1,156 @@
...
+      await sleep(20 * 2 ** attempt);
+    }
+  }
+  dbInstance = null;
+  throw new DatabaseConnectionError(
+    `Failed to connect to database after 3 attempts`,
+  );
+}
+
+async function generateRDSAuthToken() {
```
jimlerza:
```suggestion
async function generateRDSAuthToken(): Promise<string> {
```

jimlerza (approved):

PR #9907, opened 2026-04-14
9697: remove unused files, moved certain files from shared to either …

alexdziarn:
### Intent
- Keep `shared/` limited to domain logic and types that both **web-client** and **web-api** actually need, and move browser-only HTTP proxy code and web-api–only implementation details beside their owning packages so imports reflect real boundaries.
### Scope
- **Remove dead code** (no non-test callers), including:
  - **Entities:** `Order`, `Scan`, `Batch` (and their specs).
  - **Document generators:** `documentServiceEmail`, `petitionServiceEmail` (and their specs).
  - **Utilities:** `validatePdfHeader`, `formatDollars`, `formatPositiveNumber`, `getClinicLetterKey`, `getFormattedPartiesNameAndTitle` (plus `validatePdfHeader`’s spec).
  - **Tools:** `pinkLog`, `generateCategories`.
- **Move** `shared/src/proxies/**` into **web-client** and other web-api–only internals into **web-api** where those are the sole callers (full list in the diff).
- Prefer **`@shared/...`** over long relative paths into `shared/`.
- Extend **TypeScript** on affected proxies, interactors, Lambdas/handlers as wired through `applicationContext`, and the web-client code that calls them, so request/response shapes stay consistent after the move.
### Testing / CI
- Fix regressions in unit tests, web-api tests, and Cypress where paths or mocks broke after relocation.
- **Shared workflow:** normalize paths from `split-tests-glob.ts` (strip leading `./`) so Jest 30’s `testPathPattern` matches all shard-selected files; previously some suites under paths such as `shared/.../dto/cases/` did not run, which showed up as missing branch coverage after merging shard `coverage-final.json` with `nyc`.


jdr81394 (approved):

shared/src/tools/generateCategories.ts
```diff
@@ -1,163 +0,0 @@
-/**
- * This is a convenience utility for converting an Event Code CSV from the court
- * to a JSON file that can be used by the application. It may never be used
- * again, but we've kept it in the project just in case.
```
devinganey613, line 4: Could this just be moved to the scripts folder? I have no context for if this is important or not, so deleting it might be fine too.

devinganey613 (approved):

shared/src/tools/generateCategories.ts
jimlerza: This was used to set up the categories (internal/external filing) of event codes that we currently have in EntityConstants. Even if the court were to decide to recategorize all of these event codes, the likelihood that they would supply it to us in this same format is probably pretty low. This file can go!

jimlerza (commented):
The PR description says

> Relocated files from the shared directory to client or web-api where they are exclusively used

But I don't see that in this diff. Are some files missing?

jimlerza:
@copilot Can you help me verify that the acceptance criteria has been met?

Read the original ticket for this:
```bash
GH_PAGER=cat gh issue view 9697 --comments
```

Understanding the objective, please analyze all code within the `shared/` directory and identify files that should be moved to either the `web-api` or `web-client` directories, or assert that all files can be utilized by either package and none need to be moved.

jimlerza (approved):

PR #9910, opened 2026-04-14
9697 cleanup shared directory to test 1776190649

paulo-filho-ctr:
https://github.com/ustaxcourt/ef-cms/issues/9697
- Removed unused headless files that were only referenced in their own tests
- Relocated files from the shared directory to client or web-api where they are exclusively used


AbelTaxCourt (approved):

shared/src/business/entities/Batch.ts
AbelTaxCourt, line 1: I did a global search for usages of `new Batch`and found it in getBatchClient.ts, which is also used in applicationContext, pollAWSBatchProgress.ts, and sendZipperBatchJob.ts. Are these files still used and are they also to be removed with this PR?

shared/src/business/entities/orders/Order.ts
AbelTaxCourt, line 1: I did a global search for `new Order` and found it used in validateOrderWithoutBodyInteractor.ts and validateOrderAdvancedSearchAction.ts. Are these files still used and are they also to be removed with this PR?

ChristieLaferriereTaxCourt (approved):

PR #9911, opened 2026-04-14
9745 amgen support huge cases to test one more time

codyseibert:
after the second seeding of the exhibits onto amgen, I accidently added around 9k instead of 8k, and now the public UI won't load either for the amgen cause due to the response payload being too large.  Also applying the exclude docket entry query string feature to the public api so that we can fetch all the data in pages.


nechamakrigsman (approved):

TomElliottFlexion (approved):

codyseibert:
@copilot resolve the merge conflicts in this pull request

PR #9932, opened 2026-04-21
9745 amgen support huge cases to test

codyseibert:


nechamakrigsman (approved):

web-api/src/lambdas/public-api/generatePublicDocketRecordPdfWorkerLambda.ts
```diff
@@ -0,0 +1,61 @@
...
+              statusCode: err?.statusCode || 500,
+            }),
+          ),
+          key: errorKey(jobId),
+          useTempBucket: true,
+        });
+        throw err;
+      }
+    },
+    { logResults: false },
```
TomElliottFlexion, line 60: Maybe we do want to log results for these?

codyseibert: I could set it to true, I don't think that endpoint has a return statement as well

web-client/integration-tests/docketClerkEditsDocketEntryNonstandardDMeta.test.ts
```diff
@@ -20,7 +21,10 @@ describe("Docket Clerk Edits a Docket Entry's Nonstandard D Metadata", () => {
     expect(caseDetail.docketNumber).toBeDefined();
 
     cerebralTest.docketNumber = caseDetail.docketNumber;
-    cerebralTest.previousDocumentId = caseDetail.docketEntries[0].docketEntryId;
+    const petitionDocketEntry = caseDetail.docketEntries.find(
+      entry => entry.eventCode === INITIAL_DOCUMENT_TYPES.petition.eventCode,
+    );
+    cerebralTest.previousDocumentId = petitionDocketEntry.docketEntryId;
```
TomElliottFlexion, line 27: I'd hope that the petition is always at [0]?

TomElliottFlexion (approved):

web-api/src/lambdas/public-api/generatePublicDocketRecordPdfWorkerLambda.ts
TomElliottFlexion: ...good point, I did not think this through

TomElliottFlexion: I guess the error would be returned and therefore logged?

PR #9939, opened 2026-04-23
9915 - initial commit to fix the coversheet bug

codyseibert:
The fundamental issue with this bug we invoke the coversheet endpoint from the UI in a two step process. This means if after the external document is filed and the user's internet stops or they close their browser, nothing ever invokes the coversheet endpoint.  

This refactor makes it so the backend api invokes the event to generate the coversheet, and we added it to the async worker queue so that it runs as a background task and doesn't block the user. The UI now polls until the docket entry is fully done processing using the status field, so nothing really changes much.


shared/src/proxies/documents/getDocketEntryProcessingStatusProxy.ts
```diff
@@ -0,0 +1,27 @@
+import { ClientApplicationContext } from '@web-client/applicationContext';
+import { getCurrentUserToken } from '../requests';
+
+const getDefaultHeaders = () => {
```
codyseibert: todo: i'm pretty sure we already have a standard way of attaching these tokens / headers.

web-api/src/business/useCaseHelper/coverSheet/enqueueAddCoversheet.ts
```diff
@@ -0,0 +1,35 @@
...
+    docketEntryId,
+    docketNumber,
+  }: {
+    authorizedUser: UnknownAuthUser;
+    docketEntryId: string;
+    docketNumber: string;
+  },
+): Promise<void> => {
+  // Mark the entry as pending so pollForCoversheetComplete won't see a stale
+  // COMPLETE status (e.g. carried over from a draft) and exit before the
```
codyseibert: todo: look into this

web-api/src/business/useCases/coversheet/addCoversheetWorker.ts
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  await addCoversheetInteractor(
+    applicationContext,
+    { docketEntryId, docketNumber },
+    authorizedUser,
+  );
+};
+
+export const addCoversheetWorker = withLocking(
```
codyseibert: todo: should this lock on the case as well? I'm concerned about race condition between the original external document being filed and this one kicking off.

web-api/src/business/useCases/coversheet/addCoversheetWorker.ts
```diff
@@ -0,0 +1,51 @@
...
+  if (
+    docketEntry.processingStatus === DOCUMENT_PROCESSING_STATUS_OPTIONS.COMPLETE
+  ) {
+    getDawsonLogger().info(
+      `addCoversheetWorker: docket entry ${docketEntryId} already complete — skipping duplicate SQS delivery`,
+    );
+    return;
+  }
+
+  await addCoversheetInteractor(
```
codyseibert: note: so this method will potentially affect multiple cases on a consolidated group.  it loops over the entire group and adds a coversheet to each document if the document was filed across the group.  This means that there might be potential race conditions were we are not locking all the cases for update... I do wonder if we need to actually put a lock on the entire case or not, as all this method does it update the docket entry records.

PR #9940, opened 2026-04-23
9458 edit trial session to test 1776967329

alexdziarn:
[testrail](https://ustaxcourt.testrail.io/index.php?/runs/view/356&group_by=cases:section_id&group_order=asc&display=tree)
- moved validation rules for alternateTrialClerkName from NewTrialSession to TrialSession class.
- created new set of validation rules for a new class called EditTrialSession
- added seed data for trial sessions because were are unable to create trial sessions that start in the past through the UI
- refactored canEdit boolean in formattedTrialSessionDetails to a string that contains edit options of 'all', 'limited' or 'none'
- added estimatedEndDate values to trial session seed data since trial sessions can no longer be created without the estimatedEndDate
- added backend validation and refactored some existing code in updateTrialSessionInteractor


devinganey613 (approved):

web-client/src/views/TrialSessions/InPersonProceedingForm.tsx
```diff
@@ -44,7 +44,7 @@ export const InPersonProceedingForm = connect<
             data-testid="trial-session-courthouse-name"
             disabled={
               !addingTrialSession &&
-              formattedTrialSessionDetails.canEditOngoingSession
+              formattedTrialSessionDetails.editPermissions === 'limited'
```
akuny, line 47: Could we store the result of `formattedTrialSessionDetails.editPermissions === 'limited'` as a property called `canEditOngoingSession` on `formattedTrialSessionDetails` instead of doing this boolean evaluation in .tsx files?

akuny (approved):
Looks good, just one suggestion to keep .tsx files a bit cleaner.

web-client/src/presenter/computeds/formattedTrialSessionDetails.ts
```diff
@@ -107,38 +107,26 @@ export const formattedTrialSessionDetails = (
...
+      editPermissions = docketClerkCanEditCheck(
+        formattedTrialSession.sessionType,
+      )
+        ? 'all'
+        : 'none';
+    }
 
-    if (user.role === USER_ROLES.docketClerk && canEdit) {
-      canEdit = docketClerkCanEditCheck(formattedTrialSession.sessionType);
+    if (user.role === USER_ROLES.caseServicesSupervisor && !trialDateInFuture) {
```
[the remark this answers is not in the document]
alexdziarn: after discussion with Tenille, we decided to leave out the check for whether a trial session is closed.

web-client/src/views/TrialSessions/InPersonProceedingForm.tsx
alexdziarn: We did try this initially, but it got too confusing because of the crisscross permissions between authorized users who can edit trial sessions and CSS users

PR #9951, opened 2026-04-24
9810 case status recent filings to test 1777062402

zacharybrown204:
#9810 #9811

Few more fixes for ux:
- Removing pagination div when we have only one page
- Adding a maximum width to the case title column to give other columns more room
- Adding a bit of space underneath the bottom paginator on mobile

### DIFF
--- a/web-client/src/views/CaseListRowExternal.tsx
+++ b/web-client/src/views/CaseListRowExternal.tsx
@@ -80,7 +80,7 @@ export const CaseListRowExternal = ({
                   key={consolidatedCase.docketNumber}
                   showFilingFee={showFilingFee}
                   showCaseStatusInfoSequence={showCaseStatusInfoSequence}
-                  showCaseStatus
+                  showCaseStatus={showCaseStatus}
                 />
               );
             })}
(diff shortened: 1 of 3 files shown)


TomElliottFlexion (approved):

ChristieLaferriereTaxCourt (approved):

web-client/src/views/CaseListRowExternal.tsx
```diff
@@ -49,7 +49,7 @@ export const CaseListRowExternal = ({
                 />
               </div>
             </td>
-            <td>{formattedCase.caseTitle}</td>
+            <td className="tw:max-w-[250px]">{formattedCase.caseTitle}</td>
```
jimlerza: I thought we were supposed to be using `vw`/`vh` units (`vw` in this case) instead of `px` for layouts?

(And yes, I can see there are already instances of `px` elsewhere in this diff...)

zacharybrown204: I had thought it was rem for layouts. I see only a few spots where we use vh, none where we use vw, and plenty more using rem. You're right though we should avoid px here so I did change it to rem, and I can change it to vw if we do want to use that.

PR #9952, opened 2026-04-27
9915 fix missing coversheet bug

codyseibert:
### Summary
Fixes a bug where docket entries were occasionally missing coversheets by moving coversheet generation from a separate client-triggered endpoint into the server-side interactors that create the documents.

### Suspected Root Cause
We believe the bug was caused by users getting disconnected while uploading an external document. The frontend was responsible for invoking the `addCoversheet` endpoint **only after a successful upload completed** — meaning there was a large window between the upload finishing and the follow-up coversheet request being made. If a user's internet dropped, the browser was closed, or the tab was navigated away during that window, the coversheet endpoint would never be called and the docket entry would be left without a coversheet.

By making coversheet generation part of the same server-side interactor that creates the docket entry, the coversheet is guaranteed to be generated atomically with the document — independent of client connectivity.

### Changes

**Server-side**
- Removed `addCoversheetLambda`, `addCoversheetProxy`, and the corresponding route registration in `web-api/src/app.ts`.
- Coversheet generation now happens within these interactors:
  - `createCaseInteractor`
  - `addPaperFilingInteractor`
  - `editPaperFilingInteractor`
  - `fileCourtIssuedDocketEntryInteractor`
  - `fileExternalDocumentInteractor`

**Client-side**
- Deleted `generateCoversheetAction` and `isCoversheetNeededAction`.
- Removed coversheet steps from `submitCourtIssuedDocketEntrySequence` and `serveDocumentCompleteSequence`.
- Simplified `saveAndSubmitCaseAction`, `uploadExternalDocumentsAction`, and the `submitCourtIssuedDocketEntry*` actions to no longer manage coversheet flow.

**Tests**
- Updated and added interactor tests covering the new server-side coversheet behavior.

### Why
Centralizing coversheet creation in the interactors eliminates the client-connectivity failure mode described above and makes coversheet generation an atomic part of the document operation itself.


jimlerza (changes requested):
Moving coversheet generation to the backend is definitely the right idea, but I’m not comfortable with the use of `Promise.all` here. That's a pattern we're trying to get away from.

This is multi-item state-changing work, not read-only parallelism. If one promise fails after others have already rewritten PDFs / updated docket entries, we’ve created a partial-success state with no rollback and no safe resume. That’s the kind of failure that turns into a big mess of manual data cleanup.

Can we perform the work from a queue with a retryable/idempotent job model? Or some other approach that allows idempotent retries without generating duplicate coversheets?

If that's all too much to ask, at very least we should generate coversheets sequentially instead of in a `Promise.all()`. That doesn't eliminate the manual cleanup in the event of an error, but at least that way we can more easily figure out where we "left off".

PR #9974, opened 2026-04-29
9697-cleanup-shared-directory-move-files

paulo-filho-ctr:


web-client/src/proxies/users/getUserPendingEmailStatusProxy.ts
alexdziarn, line 1: can we delete this and all associated routes and interactors?

web-client/src/proxies/getCurrentVersionProxy.ts
alexdziarn, line 1: can we remove

web-client/src/proxies/getPublicCaseProxy.ts
alexdziarn, line 1: can we remove

PR #9975, opened 2026-04-29
Dependency updates 04 27 2026

zacharybrown204:
## Dependencies Updates 4/27/2026
Simple set of updates this week.
- New docker image is `4.3.77` on exp7
- New terraform version `1.15.0`
- Couple of vulnerability fixes
  - A bad version of uuid was being used in a couple of our other dependencies, tagged uuid as an override so we use our pinned version.
  - Other vulerabilities that were fixed with `npm audit fix`. If we regenerate the package lock then we'll need to run this again.

[Exp7 deployment here](https://app.circleci.com/pipelines/github/ustaxcourt/ef-cms/21576/workflows/d56f25cc-0bc1-4692-a312-1ca856186289)


jimlerza (commented):
### Findings

**Medium**: `uuid` was major-bumped and globally overridden, but the rotation playbook/automation was not updated to explain that decision
`package.json:109` bumps `uuid` to 14.0.0
`package.json:274-276` adds a repo-wide override: `"uuid": "$uuid"`
But `scripts/npm/upgrade-npm-packages.ts:5-29` still lists `uuid` in the caveats/skip list
And `docs/dependency-updates.md` has no `uuid` caveat or rationale

**Why this matters**:
The branch makes `uuid` a deliberate manual exception, but the maintenance docs and upgrade automation still treat it as something to skip with no explanation.
Future rotations will have no source of truth for whether `uuid` is now safe to keep upgrading, why the override exists, or whether the caveat list should be removed.

**Recommendation**:
Either document the `uuid` decision in `docs/dependency-updates.md` and remove `uuid` from the caveat list in `scripts/npm/upgrade-npm-packages.ts`, or revert the major bump/override until that decision is captured.

Issue #8708, opened 2024-05-03
Consolidate `ENV` and `STAGE`


mmarcotte:
It appears that stage is a remnant of when we were using Serverless to deploy our infrastructure. And it seems to be causing confusion for no good reason. Do we still need it? Is it serving a purpose? Can we just consolidate behind one or the other? This is the ticket to find that out!

## Acceptance Criteria

- We no longer are using `process.env.STAGE` in the application
- OR we have identified a clear need for it, and we are able to document it such that it is no longer confusing in the least

## Pain Avoided/Frustration Saved

- 🧠 😩 

## Complexity of Problem (Low, Medium, High) and Why it's Complex
  
Low because we got this 💪 🧠

PR #10002, opened 2026-05-06
9392 prevent tabs from display on test

jdr81394:
This PR adds environment-based feature gating to the new Clerk of Court dashboard graph tabs (Pie Chart, Bar Graph, Line Graph) and the Dawson Library page.

### DIFF
--- a/web-client/src/views/Dashboards/DashboardClerkOfCourt.tsx
+++ b/web-client/src/views/Dashboards/DashboardClerkOfCourt.tsx
@@ -24,7 +24,7 @@ export const DashboardClerkOfCourt = connect(
   },
   function DashboardClerkOfCourt({ clerkOfCourtDashboard, user }) {
     const shouldDisplay =
-      process.env.USTC_ENV !== 'prod' && process.env.USTC_ENV !== 'test';
+      process.env.STAGE !== 'prod' && process.env.STAGE !== 'test';
 
     const {
       caseTypeBreakdownDatasets,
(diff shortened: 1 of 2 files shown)


web-client/src/views/Dashboards/DashboardClerkOfCourt.tsx
```diff
@@ -23,6 +23,9 @@ export const DashboardClerkOfCourt = connect(
     user: state.user,
   },
   function DashboardClerkOfCourt({ clerkOfCourtDashboard, user }) {
+    const shouldDisplay =
+      process.env.USTC_ENV !== 'prod' && process.env.USTC_ENV !== 'test';
```
jimlerza: This won't work on deployed environments. You need to use `process.env.STAGE` in deployed lambdas.

`STAGE` is a remnant from way way back when we used Serverless. There's a devex ticket to stop using it: https://github.com/ustaxcourt/ef-cms/issues/8708

jimlerza (changes requested):

web-client/src/views/Dashboards/DashboardClerkOfCourt.tsx
jdr81394: pushed up and changed

jimlerza (approved):
We probably need to clean up those "falsy" tabs before this goes to prod

AbelTaxCourt (approved):

Issue #9915, opened 2026-04-16
BUG: Party filed document is missing a coversheet


ttlenard:
**Describe the Bug**
We received a report from a Judge that a party filed document was missing a coversheet. The party was the IRS. The IRS reported that "their upload timed out/their system crashed" (explanation was unclear).

After reviewing the logs, I do not see any errors. We can see that the IRS initiated the upload and then 33 seconds later, Petitioners counsel also initiated an upload of a document on the case.  The logs show that Petitioners Counsel's upload was "moving" much faster. Petitioner Counsel's upload completed while the IRS doc was still uploading.  After Petitioner Counsel's doc finished, only then did the IRS's upload complete, but without a coversheet.

Good news is that Service for the document was still sent.

This occurred on case 17513-24, index 117.

These other Issues in Github may be related?  #8615, #8616, #9014?

**Business Impact/Reason for Severity**
med

**In which environment did you see this bug?**
Prod

**Who were you logged in as?**
This happened to an IRS practitioner

**What were you doing when you discovered this bug? (Using the application, demoing, smoke tests, testing other functionality, etc.)**
This issue was reported to us by a Judge

**To Reproduce**

**Expected Behavior**

Coversheets should always get added to documents that get added to the docket record

**Actual Behavior**
A clear and concise description of what actually happened.
A coversheet did not get added to the document that was uploaded

**Screenshots**
Notice that there is no coversheet response for the IRS in the screenshot below:

<img width="1608" height="548" alt="Image" src="https://github.com/user-attachments/assets/aebd6314-d909-45d4-906d-bacfb5cbb4b5" />

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Cause of Bug, If Known**

**Process for Logging a Bug:**
* Complete the above information
* Add a severity tag (Critical, High Severity, Medium Severity or Low Severity). See below for priority definition. 

**Severity Definition:**
* Critical Defect
Blocks entire system's or module’s functionality
No workarounds available
Testing cannot proceed further without bug being fixed.

* High-severity Defect
Affects key functionality of an application
There's a workaround, but not obvious or easy
App behaves in a way that is strongly different from the one stated in the requirements

* Medium-severity Defect
A minor function does not behave in a way stated in the requirements.
Workaround is available and easy

* Low-severity Defect
Mostly related to an application’s UI
Doesn't need a workaround, because it doesn't impact functionality

## Definition of Ready for Bugs (Created 2021-10-04)
[...]

PR #10003, opened 2026-05-06
9915 - Coversheet Bug

codyseibert:
# Fix missing coversheet bug (#9915)

## Summary

Fixes docket entries occasionally being created without coversheets by moving coversheet generation off the client request path and onto a server-side SQS worker, with the originating interactor enqueueing the job atomically with the document write. The client then polls a new processing-status endpoint until the worker finishes (or fails terminally).

## Suspected root cause

Coversheet generation used to run on a separate client-triggered `addCoversheet` HTTP call that fired *after* the upload interactor returned. If the user got disconnected, closed the tab, or navigated away in that window, the follow-up call never happened and the docket entry was left without a coversheet.

The fix removes that window: the same server-side interactor that creates the docket entry also enqueues the coversheet job, so coversheet generation is guaranteed to happen regardless of client connectivity.

### Server

- New `enqueueAddCoversheet` helper (`web-api/src/business/useCaseHelper/coverSheet/enqueueAddCoversheet.ts`) — flips the entry's `processingStatus` to `PENDING` synchronously, then queues an `ADD_COVERSHEET` worker message. If the queue write fails, the status is rolled forward to `ERROR_ADDING_COVERSHEET` so the client poll fails fast instead of timing out.
- New `addCoversheetWorker` (`web-api/src/business/useCases/coversheet/addCoversheetWorker.ts`) — SQS handler wrapped in `withLocking` keyed on `docket-entry|<id>`. On success the existing `addCoversheetInteractor` flips status to `COMPLETE`; on failure it writes `ERROR_ADDING_COVERSHEET` and rethrows so SQS records the failure and the message lands in the DLQ (current `maxReceiveCount=1`).
- Wired into the worker router as a new `MESSAGE_TYPES.ADD_COVERSHEET` route.
- New `getDocketEntryProcessingStatusInteractor` + lambda + proxy (`/case/:docketNumber/docket-entry/:docketEntryId/processing-status`) for the client poll. Authorization allows the original filer, internal users, and any user with case access.
- `addCoversheetInteractor` now has an idempotency gate: a `COMPLETE` entry returns early instead of stacking a second coversheet onto the S3 PDF — important on the queued path (duplicate SQS delivery) and for retry callers like `serveCaseToIrsInteractor`. Sync callers that intend to regenerate (e.g. `updateDocketEntryMetaInteractor` on a metadata edit) opt out via `bypassIdempotencyGate`.
- New `ERROR_ADDING_COVERSHEET` value on `DOCUMENT_PROCESSING_STATUS_OPTIONS`.
- New `retrySettled` utility (`web-api/src/utilities/retrySettled.ts`) — `Promise.allSettled` + retry-only-the-failures, used by `serveCaseToIrsInteractor` to retry partial coversheet failures without re-doing successful work.
- Coversheet enqueue calls added to: `createCaseInteractor`, `addPaperFilingInteractor`, `editPaperFilingInteractor`, `completeDocketEntryQCInteractor`, `fileCourtIssuedDocketEntryInteractor`, `fileExternalDocumentInteractor`, `serveExternallyFiledDocumentInteractor`, `updateDocketEntryMetaInteractor`.
- `addCoversheetLambda`, `addCoversheetProxy`, and the corresponding `app.ts` route are removed.

### Shared

- `CaseDTO` carries a transient `pendingCoversheetDocketEntryIds: string[]` field, populated by interactors that enqueued one or more workers. Captured before the `Case` rebuild so it survives `dataSecurityFilter`. Typed as an array (rather than a single id) so a future caller that fans out across multiple entries doesn't break the contract.

### Client

- New `awaitCoversheetCompleteAction` replaces the old `generateCoversheetAction`. It only observes — it does not initiate generation.
- New `pollForCoversheetComplete` utility — polls the new status endpoint with fast (2.5s) phase for ~25s then a 5s slow phase, both with up-to-500ms jitter to avoid thundering-herd. First iteration fires immediately so a worker that finished pre-poll is observed without dead time. Uses `Promise.allSettled` so a transient 5xx on one entry's check doesn't tear down the whole poll. Throws `CoversheetPollTimeoutError` after 300s or `CoversheetGenerationError` on a definitive `ERROR_ADDING_COVERSHEET`.
- `isCoversheetNeededAction` is now data-driven: it reads `pendingCoversheetDocketEntryIds` off `props` (set by the interactor response or the websocket payload) instead of re-deriving "does this need a coversheet" from event-code rules client-side.
- `submitCourtIssuedDocketEntrySequence` and `serveDocumentCompleteSequence` swap `generateCoversheetAction` → `awaitCoversheetCompleteAction`.
- `saveAndSubmitCaseAction`, `uploadExternalDocumentsAction`, and the `submitCourtIssuedDocketEntry*` actions are simplified — they no longer drive the coversheet flow.
- `generateCoversheetAction` is deleted.

## Tests

- New unit tests for `enqueueAddCoversheet`, `addCoversheetWorker`, `getDocketEntryProcessingStatusInteractor`, `awaitCoversheetCompleteAction`, `pollForCoversheetComplete`, and `retrySettled`.
- Updated interactor tests assert the enqueue is called from each filing path.
- One Cypress wait was bumped because the coversheet now appears asynchronously after the request returns.

## Why

Centralizing coversheet generation server-side and queueing it removes the client-connectivity failure mode entirely, while the worker + status-poll design keeps the UI experience close to the old synchronous flow (page count appears once the worker finishes) and gives us a terminal `ERROR_ADDING_COVERSHEET` state plus DLQ alarming when generation actually fails.


COVERSHEETS.md
jimlerza, line 1: If we're going to keep this file, it should probably be cleaned up, formatted a bit for better readability, and moved into `docs/`.

Or maybe just deleted?

jimlerza (approved):

cypress/local-only/tests/integration/documentQC/docket-clerk-edits-court-issued-coversheet-to-coversheet.cy.ts
```diff
@@ -0,0 +1,94 @@
...
+    externalUserCreatesElectronicCase().then(docketNumber => {
+      petitionsClerkServesPetition(docketNumber);
+      logout();
+
+      loginAsDocketClerk();
+      cy.visit(`/case-detail/${docketNumber}/upload-court-issued`);
+      cy.get('#upload-description').type('to petitioner');
+      attachFile({
+        filePath: '../../helpers/file/sample.pdf',
+        selector: 'input#primary-document-file',
```
jdr81394: Shouldnt we be using a data-testid for this?

cypress/local-only/tests/integration/documentQC/docket-clerk-adds-court-issued-doc-with-coversheet.cy.ts
```diff
@@ -0,0 +1,62 @@
...
+    externalUserCreatesElectronicCase().then(docketNumber => {
+      petitionsClerkServesPetition(docketNumber);
+      logout();
+
+      loginAsDocketClerk();
+      cy.visit(`/case-detail/${docketNumber}/upload-court-issued`);
+      cy.get('#upload-description').type('to petitioner');
+      attachFile({
+        filePath: '../../helpers/file/sample.pdf',
+        selector: 'input#primary-document-file',
```
jdr81394: A lot of the tests aren't using data-testid, just leaving this final comment as a blanket ccomment on this instead of pointing out each one.

cypress/local-only/tests/integration/documentQC/docket-clerk-edits-court-issued-coversheet-to-no-coversheet.cy.ts
```diff
@@ -0,0 +1,101 @@
...
+
+      loginAsDocketClerk();
+      cy.visit(`/case-detail/${docketNumber}/upload-court-issued`);
+      cy.get('#upload-description').type('to petitioner');
+      attachFile({
+        filePath: '../../helpers/file/sample.pdf',
+        selector: 'input#primary-document-file',
+        selectorToAwaitOnSuccess: '[data-testid^="upload-file-success"]',
+      });
+      cy.get('#save-uploaded-pdf-button').click();
```
jdr81394: data-testid?

web-client/src/presenter/actions/DocketEntry/awaitCoversheetCompleteAction.ts
```diff
@@ -0,0 +1,23 @@
...
+export const awaitCoversheetCompleteAction = async ({
+  applicationContext,
+  get,
+  props,
+}: ActionProps) => {
+  const docketNumber = get(state.caseDetail.docketNumber);
+  // isCoversheetNeededAction forwards the full pending list; the legacy
+  // single-id prop is kept as a fallback for websocket payloads that
+  // surface only docketEntryId.
+  const docketEntryIds: string[] = props.docketEntryIds ?? [props.docketEntryId];
```
jdr81394: ?? only falls back on null/undefined, so if docketEntryIds is absent and docketEntryId is also undefined, this resolves to [undefined] — a one-element array, not []. The type says string[] but at runtime it'd be (string | undefined)[]

Try something like this:
const docketEntryIds: string[] =
  props.docketEntryIds ?? (props.docketEntryId ? [props.docketEntryId] : []);

TomElliottFlexion (approved):
Good! There are a _lot_ of new cypress tests in here. This recurring bug may warrant them, but we definitely can't afford to be adding this many on a regular basis

jdr81394 (changes requested):

web-client/src/presenter/actions/DocketEntry/awaitCoversheetCompleteAction.ts
devinganey613: good catch!

jdr81394 (approved):

zacharybrown204 (approved):

PR #10044, opened 2026-05-14
9776 client notification bug

jdr81394:
## Summary

This PR adds a deployment timestamp handshake between the frontend and backend to detect version mismatches and prompt users to refresh. It also changes the infrastructure in order to invalidate Cloudfront cache. The bug was not able to be reproducible on our end, therefore after discussion and agreement we believe this will be the best possible way to fix this bug from occuring in the future.

---

### Client
- Stores the backend deployment timestamp in `localStorage` under `deploymentTimestamp`
- Sends it on every outgoing request via `X-Deployment-Timestamp`
- On a mismatch response:
  - **During bootstrap** → reloads so the app does not land on a blank screen
  - **After bootstrap** → opens the existing "app updated" modal

### Backend
- Compares the client-sent `X-Deployment-Timestamp` against the `DEPLOYMENT_TIMESTAMP` env var
- Returns `409` with `X-Manual-Refresh-Required: true` on mismatch
- Includes the current `X-Deployment-Timestamp` on every successful response so the client stays in sync

### Shared infrastructure
- Moved header constants into a new `shared/src/utils/headers.ts` so the frontend and backend share a single source of truth
- Added a `getHeaderValue` helper that handles both the `Headers.get()` API and plain object headers

### CloudFront / deployment
- Updated `deploy-ui.sh` and `deploy-public.sh` to serve `index.html` with `no-cache, no-store, must-revalidate` so browsers always fetch the latest entry point
- Fixed `switch-ui-colors.helpers.ts` to wait for each CloudFront distribution to reach `Deployed` status before updating Route 53, replacing the previous fire-and-forget `setTimeout` retry

---

## Why `localStorage` instead of Cerebral state

Cerebral manages runtime application state, but the deployment timestamp must survive full browser refreshes to be useful. `localStorage` is already used for this pattern in the codebase:

- `getItemInteractor` / `setItemInteractor` / `removeItemInteractor`
- Adapters under `web-client/src/persistence/localStorage/`

This follows the same convention rather than stretching ephemeral Cerebral state into a persistence role.

---

## Testing

```sh
npm run test:client:unit:file -- web-client/src/presenter/utilities/createForceRefreshCallback.test.ts
npm run test:client:unit:file -- web-client/src/providers/httpClient.test.ts
npm run test:client:unit:file -- web-client/src/presenter/actions/clearModalAction.test.ts
npm run test:shared:ci

### CODE REVIEW

File: web-client/src/providers/httpClient.ts [hunk 1]
```diff
@@ -13,17 +42,47 @@ export const getHttpClient = (
...
+        error.stack = stackError.stack;
+        throw error;
+      },
+    );
 
-    error.stack = stackError.stack;
-    throw error;
-  });
+    areInterceptorsRegistered = true;
+  }
```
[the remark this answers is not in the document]
jdr81394: addressed and resolved with commit [bf94c79](https://github.com/ustaxcourt/ef-cms/pull/10044/commits/bf94c7975377884edefc90a1c50efce30ac4bc44)

cypress/deployed-and-local/integration/caseDetail/docketRecord/public-docket-record.cy.ts
```diff
@@ -7,14 +7,21 @@ describe('Public Docket Record', () => {
       cy.get('[data-testid="account-menu-button"]').click();
       cy.get('[data-testid="logout-button-desktop"]').click();
 
-      cy.visit(getCypressEnv().publicSiteUrl);
-      cy.get('input#docket-number').type(docketNumber);
-      cy.get('button#docket-search-button').click();
-      cy.get('[data-testid="header-public-case-detail"]').contains(
-        `Docket Number: ${docketNumber}`,
+      const publicSiteUrl = getCypressEnv().publicSiteUrl;
```
akuny: I'm getting a lint error locally here. I think we could replace this line with `const { publicSiteUrl } = getCypressEnv();` and be in good shape. Same goes for the other two Cypress test suites in this PR.

jimlerza (approved):

akuny (approved):

devinganey613 (approved):

cypress/deployed-and-local/integration/caseDetail/docketRecord/public-docket-record.cy.ts
jdr81394: this is outdated

jdr81394: it is currently const { publicSiteUrl } = getCypressEnv();

jdr81394: Screenshot 2026-06-03 at 2 51 00 PM

PR #10067, opened 2026-05-20
9584: strip whitespace-only additional order clauses

paulo-filho-ctr:
## Summary
- **Normalize additional order text:** Drops array entries that are empty or only whitespace (after trim). PDF/rich text and **saved `draftOrderState`** only include substantive clauses.
- **First field always in the UI:** Form state keeps a mandatory first “Additional order text” row (`['']` when there’s nothing real). Extra rows stay optional; optional rows are not re-shown after preview/refresh if they were blank.
- **Persist clean payloads:** `submitCourtIssuedOrderAction` writes `additionalOrderTextArray` as normalized meaningful entries, or `[]` when everything was blank (no placeholder strings in metadata).
- **Helpers:** `normalizeAdditionalOrderTextArray` + `additionalOrderTextArrayWithRequiredFirstField`; used in prepare, edit, and submit paths. Joi again requires at least one slot when the array is present (`min(1)`).

### DIFF
--- a/web-client/src/business/entities/MotionOrderResponseForm.test.ts
+++ b/web-client/src/business/entities/MotionOrderResponseForm.test.ts
@@ -11,6 +11,21 @@ import {
 
 describe('MotionOrderResponseForm', () => {
   describe('valid cases', () => {
+    it('should be valid when no substantive additional order text is provided', () => {
+      const form = new MotionOrderResponseForm({
+        responseDate: getBusinessDateInFuture({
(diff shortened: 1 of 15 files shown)


alexdziarn (approved):

jdr81394 (approved):

devinganey613 (approved):
Updating the cypress tests to check these new scenarios would be good, but that can go into staging I think.

PR #10068, opened 2026-05-20
9776: removes timestamp header from s3 uploads

alexdziarn:
- code to not include timestamp headers for s3 upload requests


paulo-filho-ctr (approved):

kampofo (approved):
LGTM

PR #10069, opened 2026-05-20
9776 client notification bug to test 1779302639

jdr81394:
## Summary

This PR adds a deployment timestamp handshake between the frontend and backend to detect version mismatches and prompt users to refresh. It also updates the infrastructure to invalidate the CloudFront cache on deployment.

The bug was not reproducible on our end — after discussion and agreement, we believe this is the best approach to prevent it from occurring in the future.

### Client
- Stores the backend deployment timestamp in `localStorage` under `deploymentTimestamp`
- Sends it on every outgoing request via `X-Deployment-Timestamp`
- On a mismatch response:
  - During bootstrap → reloads so the app does not land on a blank screen
  - After bootstrap → opens the existing "app updated" modal

### Backend
- Compares the client-sent `X-Deployment-Timestamp` against the `DEPLOYMENT_TIMESTAMP` env var
- Returns `409` with `X-Manual-Refresh-Required: true` on mismatch
- Includes the current `X-Deployment-Timestamp` on every successful response so the client stays in sync

### Shared infrastructure
- Moved header constants into a new `shared/src/utils/headers.ts` so the frontend and backend share a single source of truth
- Added a `getHeaderValue` helper that handles both the `Headers.get()` API and plain object headers

### CloudFront / deployment
- Updated `deploy-ui.sh` and `deploy-public.sh` to serve `index.html` with `no-cache, no-store, must-revalidate` so browsers always fetch the latest entry point
- Fixed `switch-ui-colors.helpers.ts` to wait for each CloudFront distribution to reach `Deployed` status before updating Route 53, replacing the previous fire-and-forget `setTimeout` retry

## Why `localStorage` instead of Cerebral state

Cerebral manages runtime application state, but the deployment timestamp must survive full browser refreshes to be useful. `localStorage` is already used for this pattern in the codebase:

- `getItemInteractor` / `setItemInteractor` / `removeItemInteractor`
- Adapters under `web-client/src/persistence/localStorage/`

This follows the same convention rather than stretching ephemeral Cerebral state into a persistence role.

## Testing

```sh
npm run test:client:unit:file -- web-client/src/presenter/utilities/createForceRefreshCallback.test.ts
npm run test:client:unit:file -- web-client/src/providers/httpClient.test.ts
npm run test:client:unit:file -- web-client/src/presenter/actions/clearModalAction.test.ts
npm run test:shared:ci

### DIFF
--- a/shared/src/utils/headers.test.ts
+++ b/shared/src/utils/headers.test.ts
@@ -48,11 +48,11 @@ describe('headers', () => {
     });
 
     it('returns undefined when headers.get() returns a non-string value', () => {
-      const headers = {
+      const headers: Record<string, unknown> = {
         get: jest.fn().mockReturnValue(null),
         'X-Foo': 123,
       };
-      expect(getHeaderValue(headers as any, 'X-Foo')).toBeUndefined();
+      expect(getHeaderValue(headers, 'X-Foo')).toBeUndefined();
     });
 
     it('returns the value from plain object headers by exact key', () => {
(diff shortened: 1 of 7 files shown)


TomElliottFlexion (approved):
This'll fix it!

zacharybrown204 (approved):

PR #10072, opened 2026-05-21
9009: Add transaction support

nechamakrigsman:
Copied PR comment from the original [PR](https://github.com/ustaxcourt/ef-cms/pull/6283) 

### Details:

1. We say, `getConnection` and `establishConnection` ... but these are not getting a connection or establishing a connection. They are returning a kysely wrapper object around a client pool. Behind the scenes, kysely is creating/acquiring/releasing/removing connections when we call, e.g., `.execute`.
2. ~We have two connection pools, one for `getConnection` and one for `getLockingConnection`. For the latter especially, we abuse the pool: we say, "Make sure there is only ever one connection in your pool so that we don't accidentally create a lock on one connection and try to remove it with another, which won't work!" But this is not what a connection pool is for. Instead, we should have one connection pool and get an honest-to-goodness connection _from_ the pool and use _that_ throughout the lifetime of the lock.~ Nope. This is complicated.

2. ~We have _one_ connection pool. The equivalent of `getLockingConnection`, `acquireOneDbConnection`, now simply gets a particular connection from the existing pool and uses it throughout the lifetime of the locking code.~ Nope.

Implementation details:


paulo-filho-ctr (approved):

jimlerza (approved):

web-api/src/persistence/postgres/databaseConnection.md
```diff
@@ -0,0 +1,15 @@
+A few notes about database connections, which have proved vexing to get right:
+- Kysely generally prefers to work at the pool level with callbacks rather than allowing you to acquire specific connections. This can make dealing with per-connection requirements tricky.
+  - Example: getLockingConnection. We created a whole separate pool with 1 client to guarantee that a pg_advisory_lock is created and removed with the same connection. (Only the connection that acquired the lock can remove it.)
+  - Be careful! This might seem silly--"Re-creating a pool just for locking? That seems like a misuse of a pool!"--but it solves a lot of complicated issues:
```
TomElliottFlexion, line 4: ❤️

web-api/src/persistence/postgres/databaseConnection.md
```diff
@@ -0,0 +1,15 @@
...
+  - The query executor acquires this connection, runs the query, and releases the connection back into the pool.
+  - Subsequent queries (on the same Lambda invocation or on a warm Lambda) can acquire this previous connection without having to re-connect to the database via the RDS token.
+- A lot of the logic in databaseConnection.ts is for coordinating concurrent processes. For example, we don't want a Promise.all(someDBStuff, someDBStuff, someDBStuff) to all go out and get a token and re-create a pool; instead, we want one process to do this (whichever one gets to the relevant database connection code first) and all of them to wait on this result.
+- Remember that you are in a Lambda execution context. This matters for things like setTimeout/setInterval. The token refresh logic *will not work* if you try to do it with an interval because the Lambda execution environment will only respect that timeout/interval while it is actively running. As soon as the Lambda invocation ends, the execution environment is frozen, and your interval/timeout with it.
```
TomElliottFlexion, line 15: 💯

TomElliottFlexion (approved):

PR #10078, opened 2026-05-22
9776 client notification bug to test 1779473143

alexdziarn:


cypress/deployed-and-local/integration/myAccount/petitioner-updates-email.cy.ts
```diff
@@ -174,7 +178,7 @@ describe('Petitioner Updates e-mail', () => {
     }
   });
 
-  it('should show error alert and not update the petitioner email address when they enter the incorrect email confirmation code', () => {
+  it.only('should show error alert and not update the petitioner email address when they enter the incorrect email confirmation code', () => {
```
jimlerza:
```suggestion
  it('should show error alert and not update the petitioner email address when they enter the incorrect email confirmation code', () => {
```

cypress.config.ts
```diff
@@ -24,7 +24,7 @@ import { retry, setup } from '@cypress/puppeteer';
 import { toggleFeatureFlag } from './cypress/helpers/cypressTasks/postgres/featureFlagsCypress';
 
 export default defineConfig({
-  chromeWebSecurity: false,
+  chromeWebSecurity: true,
```
jimlerza, line 27: I think you'll need to add `cy.origin()` lines to `public-trial-session-details.cy.ts` (around line 37) as well.

jimlerza (approved):

cypress.config.ts
alexdziarn: no need to because that happens in the `local-only` directory

PR #10096, opened 2026-06-01
2026 06 01 hotfix to test 1780320334

jdr81394:
Bug: Restore indent-paragraph class in order PDF sanitizer

sanitizeUserHtml only kept ql-indent-N classes, stripping the indent-paragraph class used by Status Report Orders and Order Responses for first-line indents. Allowlist indent-paragraph so the text-indent CSS applies again. Allowlist stays fixed and safe.


alexdziarn (approved):

jimlerza (approved):
Should we add a visual test too? To hopefully catch this if it ever happens again in the future?

AbelTaxCourt (approved):

nechamakrigsman (approved):

paulo-filho-ctr (approved):

Issue #10098, opened 2026-06-01
BUG: No indentation on the auto-generated Status Report Order or Order Response documents


ttlenard:
**Describe the Bug**
We received an email over the weekend from a Judge reporting that the indentation was missing from a Status Report order. Upon further investigation, it appears that this is also affecting Order Response documents.

This does not seem to be impacting Auto-generated orders from Petitions

**Business Impact/Reason for Severity**
Med

**In which environment did you see this bug?**
Prod, Test

**Who were you logged in as?**
Judge, Chambers, ADC

**What were you doing when you discovered this bug? (Using the application, demoing, smoke tests, testing other functionality, etc.)**
This bug was reported to us by a Judge

**To Reproduce**
1. Log in as a Judge
2. Navigate to a Case that has a Status Report and a Motion filed on the docket record
3. Click on the Status Report link
4. Click on the Order Report link
5. Fill in some data and then click Preview PDF
6. Note that none of the sentences are indented.
7. Next, go back to the Docket record and click on the Motion document
8. Click on Order Response link
9. Fill in some data and then click Preview PDF
10. Note that none of the sentences are indented.

**Expected Behavior**
The sentences should have the standard indentation

**Actual Behavior**
A clear and concise description of what actually happened.
None of the sentences are indented anymore.

**Screenshots**

No indentation:
<img width="1185" height="482" alt="Image" src="https://github.com/user-attachments/assets/39b19951-fa31-495a-81da-5a7746ada4b8" />

With Indentation:

<img width="821" height="1063" alt="Image" src="https://github.com/user-attachments/assets/4320a797-0d68-472b-964f-c59551518968" />

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Cause of Bug, If Known**

**Process for Logging a Bug:**
* Complete the above information
* Add a severity tag (Critical, High Severity, Medium Severity or Low Severity). See below for priority definition. 

**Severity Definition:**
* Critical Defect
Blocks entire system's or module’s functionality
No workarounds available
Testing cannot proceed further without bug being fixed.

* High-severity Defect
Affects key functionality of an application
There's a workaround, but not obvious or easy
App behaves in a way that is strongly different from the one stated in the requirements

* Medium-severity Defect
A minor function does not behave in a way stated in the requirements.
Workaround is available and easy

* Low-severity Defect
Mostly related to an application’s UI
Doesn't need a workaround, because it doesn't impact functionality

[...]

PR #10101, opened 2026-06-01
10098 Bug: fix missing indents in generated order PDFs

jdr81394:
#10098

Bug: Restore indent-paragraph class in order PDF sanitizer

sanitizeUserHtml only kept ql-indent-N classes, stripping the indent-paragraph class used by Status Report Orders and Order Responses for first-line indents. Allowlist indent-paragraph so the text-indent CSS applies again. Allowlist stays fixed and safe.


devinganey613 (approved):
I still would like to see more testing for this. My main concern is that if something changes in the sanitizer or in the frontend html none of our current tests would catch it. It's a little roundabout, but maybe we could import the sanitizer function and run it in the `prepareStatusReportOrderAction.test.ts` (and the other affected order's) unit test to see if the html remains as expected. 

Approving because I know this is high priority, and we can follow up later with some automated tests.

nechamakrigsman (approved):

TomElliottFlexion (approved):

PR #10045, opened 2026-05-14
9584 add free text

alexdziarn:
### [TestRail Results](https://ustaxcourt.testrail.io/index.php?/runs/view/381&page_called_by=run_view&group_by=cases:section_id&group_order=asc&display=tree&group_id=129)
This PR adds the option to add more "Additional order text" textboxes to status report orders and motion order response forms.
What is included in this PR:

- refactored status report order and motion order forms to use additionalOrderTextArray instead of additionalOrderText. see MotionOrderResponseForm.ts and StatusReportOrderForm.ts
- migrated the additionalOrderText field to additionalOrderTextArray
- added expand and contract migration files for additionalOrderText and additionalOrderTextArray
- added/edited various unit tests
- created structure around the documentMetadata in `submitCourtIssuedOrderAction.ts‎`

PR #10112, opened 2026-06-03
9584 add free text to test 1780498196

alexdziarn:
- refactor for the motion order response pdf `<p>` text
- migrate additionalOrderText to additionalORderTextArray


jimlerza (approved):

devinganey613 (approved):

cypress/local-only/tests/integration/statusReportOrder/file-status-report-order.cy.ts
```diff
@@ -242,7 +303,7 @@ describe('file status report order', () => {
         cy.get('[data-testid="save-draft-button"]').click();
 
         cy.wait('@courtIssuedOrder').then(({ request: req }) => {
-          expect(req.body.addedDocketNumbers).to.have.length(0);
+          expect(req.body.addedDocketNumbers ?? []).to.deep.equal([]);
```
TomElliottFlexion: Can we not guarantee that this will be an array?

shared/src/business/entities/DocketEntry.ts
```diff
@@ -156,7 +156,7 @@ export class DocketEntry extends JoiValidationEntity {
   public servedParties?: any[];
   public signedAt?: string;
   public draftOrderState?: {
-    additionalOrderText?: string; // deprecated - use additionalOrderTextArray instead
+    additionalOrderText?: string; // deprecated - use additionalOrderTextArray instead - see PR #10045
```
TomElliottFlexion: Forgive me if I'm missing it, but is there more info in PR #10045? Is there a plan to remove this field?

TomElliottFlexion: In a vacuum, I'd recommend migrating this field rather than leaving two sources of truth

alexdziarn: I'd also agree that we should do this. Shall we make a part 2 to this story?

cypress/local-only/tests/integration/statusReportOrder/file-status-report-order.cy.ts
alexdziarn: We ready deployed it into test, I believe it is in setEditStatusReportOrderAction.ts. We convert the string or array into an array in the state.form

shared/src/business/entities/DocketEntry.ts
TomElliottFlexion: That would work, is there a reason not to do it as part of this PR?

cypress/local-only/tests/integration/statusReportOrder/file-status-report-order.cy.ts
TomElliottFlexion: Shouldn't this test fail if `addedDocketNumbers` isn't an array?

alexdziarn: not sure why this change is in here in the first place, addedDocketNumbers should always be an array.

shared/src/business/entities/DocketEntry.ts
alexdziarn: just now implemented is in this PR

TomElliottFlexion: love it!

web-api/src/persistence/postgres/utils/migrate/migrations/2026-06-04T00_00_00Z-migrate-additional-order-text-to-array.ts
```diff
@@ -0,0 +1,25 @@
...
+  await sql`
+    UPDATE dw_docket_entry
+    SET draft_order_state = (draft_order_state - 'additionalOrderText')
+      || jsonb_build_object('additionalOrderTextArray',
+        jsonb_build_array(draft_order_state ->> 'additionalOrderText'))
+    WHERE draft_order_state IS NOT NULL
+      AND draft_order_state ? 'additionalOrderText'
+      AND NOT (draft_order_state ? 'additionalOrderTextArray')
+  `.execute(db);
+}
```
TomElliottFlexion, line 13: This will lock all affected records for the duration of the transaction. Not sure how long that will be for this set, but it's worth testing during the `test` migration

TomElliottFlexion (approved):

web-api/src/persistence/postgres/utils/migrate/migrations/2026-06-04T00_00_00Z-migrate-additional-order-text-to-array.ts
alexdziarn: took 32 seconds

TomElliottFlexion: Sweet!

Issue #9725, opened 2026-02-10
Search By Name Improvements


cholly75:
As a case search user, so that I can quickly find the case I am looking for, I need the Search by Name function to return results with a look and feel consistent with our other Search functions.

We have made improvements to our Order and Opinion searches by implementing a bigger result set limit, pagination, and column sorting. We would like to extend these improvements to the Search by Name function for Case Search for all users.

## Acceptance Criteria
* Search by Name results no longer limit displayed results to 100 items
* Maximum number of results returned is 5000
* Search by Name results include pagination mechanism (100/p) seen across the app when appropriate
* Search Results table can be sorted by the following columns:
  - Petitioner(s)
  - Docket No.
  - Filed Date
  - Case Title
  - State
* Counter is displayed showing the total number of search results returned
* All relevant results from a query are returned
* Existing Search by Name functionality/criteria remains unaffected
* Search by Name functionality by role remains unaffected
* Search performance does not become unacceptable

## Notes
* Business logic for case search should be added to story

**Logged in Private Practitioner/IRS Practitioner**
- All served cases with keyword display in search results
- Sealed cases that the Practitioner is associated with are displayed in results
     - Lock icon does not display for sealed case
- Sealed cases that the Practitioner is NOT associated with are NOT displayed in results

**Public User/Public Terminal User**
- All served cases with keyword display in search results
- Sealed cases do not display in the results

**Court User**
- All served cases with keyword display in search results
- Sealed cases are displayed in results
     - Lock icon does not display for sealed case

### Product Owner
 - [ ] Acceptance criteria have been met and validated on the Court's test environment.
 - [ ] Associated test cases defined in TestRail have been updated if necessary.
 - [ ] Successful test run is performed in TestRail.
 - [ ] User guides are updated if necessary.

### UX
 - [ ] All new functionality has been verified to work with keyboard navigation and screen reader software.
 - [ ] UI should be touch optimized and responsive for external users.

### Engineering
 - [ ] Automated test scripts have been written, including visual tests for newly added PDFs.
 - [ ] Successful test run is performed in TestRail.
 - [ ] New screens have been added to cypress accessibility axe.
 - [ ] Interactors should validate entities before calling persistence methods.
 - [ ] Types have been added to all added and updated functions.
 - [ ] Code refactored for clarity and to remove any known technical debt.
 - [ ] Acceptance criteria for the story has been met.
[...]

PR #10116, opened 2026-06-03
9725 Search By Name Improvements

AbelTaxCourt:
# 9725: Search By Name Improvements

- #9725

## Overview

This PR updates Case Search by Name to use the same result-handling patterns as the improved Order and Opinion searches: a 5,000-accessible-result cap, 100-result pagination, sortable columns, a visible result count, and mobile-friendly sorting. It preserves backend relevance as the default `No.` order and keeps cases without State values at the bottom when sorting by State.

## Changes

- Increases Case Search by Name results to a 5,000 accessible-case cap with `search_after` pagination, preventing sealed cases from consuming the limit.
- Adds sortable columns (`No.`, `Petitioner(s)`, `Docket No.`, `Filed Date`, `Case Title`, `State`), a result count, pagination at 100/page, and mobile sort controls.
- Converts State/Territory abbreviations to full names, blank values sort to the bottom in both State sort directions.
- Hides the limited-information and sealed-case bullets for internal users on the Case Search tab, matching Order and Opinion.
- Extracts case search formatting and display logic into `advancedCaseSearchHelper` and reuses `updateAdvancedSearchResultsSortSequence` across all search tabs.
- Advanced search files cleaned up

## Reviewer Notes

**Default sort order**: The default `No.` order preserves backend relevance. OpenSearch results are sorted by relevance score and then docket number. The docket number provides stable ordering when multiple cases have the same relevance score.

**Pagination and access filtering**: Authorization filtering still occurs in the interactor. If inaccessible cases are removed, the search continues from the last raw OpenSearch result until 5,000 accessible cases are collected or no additional matches remain. The `search_after` cursor uses the last raw result rather than the last accessible result so inaccessible records are not repeatedly fetched.

**State sort**: Per ticket discussion, cases without State values are sorted to the bottom in both State sort directions.

**Duplicate results fix**: Case search sorts by relevance score (`_score`) as its primary cursor key. Unlike a stored field value, `_score` is a float recomputed per data copy on every request. Order and Opinion search sort by filing date and docket entry id, both stored values that never change, so they are not affected by this. In Case search, a boundary document can score slightly differently between two `search_after` pages if those pages are served by different copies of the data. If the score drifts below the cursor value on the next page, OpenSearch returns that case again, producing a duplicate row. Passing a per-search UUID as the `preference` parameter tells OpenSearch to always use the same copy of the data for all pages of a single search session, keeping scores consistent and preventing duplicates.

## Verification
- [TestRail Run](https://ustaxcourt.testrail.io/index.php?/runs/view/411&group_by=cases:section_id&group_order=asc&display=tree)
- Added and updated automated tests

## Pull Request Checklist

  - [x] Locally


web-client/src/presenter/computeds/AdvancedSearch/advancedSearchHelper.ts
```diff
@@ -1,98 +1,170 @@
...
-  result.caseTitle = applicationContext.getCaseTitle(result.caseCaption || '');
-
-  return result;
-};
-
 export const advancedSearchHelper = (
   get: Get,
   applicationContext: ClientApplicationContext,
-): any => {
+) => {
```
jdr81394: It'd be nice to have the return type of the result here so we know what the shape of the object is going to look like if we call the function

web-client/src/presenter/computeds/AdvancedSearch/advancedSearchHelper.ts
```diff
@@ -1,98 +1,170 @@
...
+    const bValue = caseSearchSortValue(b, sortColumn);
+
+    return aValue.localeCompare(bValue) * direction;
+  });
+};
+
+const formatSearchResultRecord = (
+  result,
+  { applicationContext }: { applicationContext: ClientApplicationContext },
+) => {
```
jdr81394: Same for this

jdr81394 (approved):

TomElliottFlexion (approved):

jimlerza (approved):

PR #10121, opened 2026-06-03
10015 paginate todays orders to test 1780519576

kampofo:
This pull request implements client-side pagination and sorting for the "Today's Orders" public page. The changes ensure that all order data is fetched up front, and pagination/sorting are handled entirely in the client, resulting in a more responsive UI and simplified state management. The update also introduces a dedicated pagination state and refactors related Cypress tests and actions for clarity and correctness.

**Key changes include:**

### Client-side Pagination & Sorting

- Refactored the `getTodaysOrdersAction` to fetch all pages of today's orders in a single operation, storing the full result set in state for client-side pagination and sorting. This eliminates the need for server-side pagination logic in the UI.
- Updated the `TodaysOrders` React component to use a new `todaysOrdersCurrentPaginationPage` state value, and implemented client-side slicing of results for pagination. Added a paginator UI component and logic to reset the page if the number of results changes. [[1]](diffhunk://#diff-792db000773e3a1e48577f5ea1d3c73a5f503a7c7945c2aaa568a1df0378d3b6L8-R8) [[2]](diffhunk://#diff-792db000773e3a1e48577f5ea1d3c73a5f503a7c7945c2aaa568a1df0378d3b6R18-R51) [[3]](diffhunk://#diff-792db000773e3a1e48577f5ea1d3c73a5f503a7c7945c2aaa568a1df0378d3b6R72-R90)

### State Management & Actions

- Introduced a new state field `todaysOrdersCurrentPaginationPage` and corresponding action/sequence (`setTodaysOrdersCurrentPaginationPageAction` and sequence) to manage the current page in client-side pagination. [[1]](diffhunk://#diff-ca231992fcca1e99c92fa9f2ec337e4108cf8514faa722fa9ab811ec3d83beaaR99) [[2]](diffhunk://#diff-86d9c3a213c85960440a1cf2d29fe8b0ba692225c963de9f219893096e0c8de1R1-R13) [[3]](diffhunk://#diff-3ed9dfdc5b0876d9e9980e5b8f320f3eddd516daed59448d25acbb71b9794167R1-R5) [[4]](diffhunk://#diff-b6134eb6425601352750a3b441c4aee34a51efafaf1a2ba6244ecede51b1ca93R45) [[5]](diffhunk://#diff-b6134eb6425601352750a3b441c4aee34a51efafaf1a2ba6244ecede51b1ca93R139) [[6]](diffhunk://#diff-1a678aecd9b3e8bbaad88915bf35f465bf85647a6ae2efb7602637b2b74cd944R1-R17)
- Simplified `setTodaysOrdersAction` to always replace the results in state with the full set of orders from props, and set the total count, removing previous logic for incremental page loading.

### Test Updates

- Updated and refactored tests for `setTodaysOrdersAction` to reflect the new logic, ensuring that state is correctly replaced rather than appended, and that pagination state is managed correctly. [[1]](diffhunk://#diff-4e6ab375b05d4cdaae1672f01ee86cb77bd73d3e560bd9d268a8903c0d604027L6-R7) [[2]](diffhunk://#diff-4e6ab375b05d4cdaae1672f01ee86cb77bd73d3e560bd9d268a8903c0d604027L32-R65)
- Added new Cypress integration tests to cover sorting and pagination behavior for the "Today's Orders" page, including mock data for multi-page results and various sorting scenarios.

These changes collectively move all pagination and sorting logic for today's orders to the client, improving UI responsiveness and simplifying the codebase.


cypress/local-only/tests/integration/public/todays-orders.cy.ts
```diff
@@ -0,0 +1,251 @@
...
+    cy.get('[data-testid="0-filingDate-header-button"]')
+      .find('.sortActive')
+      .should('not.exist');
+  });
+});
+
+// 103 orders → 2 pages (100 on page 1, 3 on page 2).
+// i=0 is newest so that filingDate DESC (the default) puts it first.
+const TOTAL_PAGINATION_ORDERS = 103;
+const paginationMockOrders = Array.from(
```
TomElliottFlexion, line 138: Just some thoughts, not requesting any changes!

We generally avoid using mock data in cypress tests, because it causes the tests to hide issues caused by data/schema/infrastructure changes and requires the mock data be updated if the schema/entities change. Having said that, the only other options I see here are to reduce the page size temporarily and/or use seeded data which also must be updated and still requires wonky manipulation to fudge the date. Also this test is very limited in scope, so not as big a deal.

I think this is a good new(ish) pattern to introduce, but we should be wary of using too liberally

web-client/src/presenter/actions/Public/getTodaysOrdersAction.ts
```diff
@@ -11,14 +11,27 @@ export const getTodaysOrdersAction = async ({
...
 
-  return { todaysOrders: results, totalCount };
+  // Fetch all pages so the full result set is available for client-side sort & pagination
+  do {
+    const { results, totalCount: fetchedTotalCount } = await applicationContext
+      .getUseCases()
+      .getTodaysOrdersInteractor(applicationContext, {
+        page,
+        todaysOrdersSort,
+      });
```
TomElliottFlexion, line 29: Out of curiosity, are we still splitting this up into multiple requests for efficiency? I know response size is an ongoing issue, but I would have assumed that single day's orders wouldn't push up against the limit

alexdziarn: It’s true that we could retrieve all of the data in a single API call as long as a days orders never exceed the limit. However, we don't want the solution to depend on that assumption. By keeping the calls separate, we won't need to revisit this in the future.

cypress/local-only/tests/integration/public/todays-orders.cy.ts
alexdziarn: We did this to keep the test focused on pagination and sorting. It also prevents the test from becoming flaky if other tests in the same Cypress run create orders that could affect the expected data set.

devinganey613 (approved):

scripts/archived/run-once/create-exhibits/insert-exhibits-via-sql.ts
```diff
@@ -211,7 +211,6 @@ const confirmAction = async (message: string): Promise<boolean> => {
             isDraft: false,
             isFileAttached: true,
             isOnDocketRecord: true,
-            multiDocketedOn: JSON.stringify([]),
```
AbelTaxCourt, line 214: strange that this was a duplicate, good catch and removal

TomElliottFlexion (approved):

AbelTaxCourt (approved):

jimlerza (approved):

PR #10141, opened 2026-06-09
9726 client side logging redo

jdr81394:
## Summary

Adds client-side error handling and reporting so unhandled UI errors no longer
blank the page silently and are now captured in CloudWatch RUM. Three pieces:
a React error boundary, RUM error reporting, and readable production stack traces.

## Changes

**Error boundary**
- New `ErrorBoundary` component (`getDerivedStateFromError` + `componentDidCatch`)
  that renders a graceful fallback instead of unmounting the whole React tree
  (which previously left users on a blank white page).
- Wired into both entry points — `app.tsx` (`AppComponent`) and
  `appPublic.tsx` (`AppComponentPublic`) — inside the Cerebral `Container`.

**CloudWatch RUM error reporting**
- `realUserMonitoring.ts`: exposed a safe `recordError()` (no-op when RUM is not
  initialized) and call it from `ErrorBoundary.componentDidCatch`.
- Enabled the `'errors'` and `'http'` RUM telemetries so uncaught JS errors and
  failed HTTP requests (4xx/5xx and network failures, `recordAllRequests: false`)
  are captured automatically.

**Readable production stack traces**
- Added `keepNames: true` to the shared esbuild config so minified prod bundles
  preserve original function/class names — RUM traces now show `ErrorBoundary` /
  `AppComponent` instead of single-letter identifiers.

## Testing

- Unit tests for `ErrorBoundary` (error-state transition, default/custom fallback,
  `recordError` invocation) and `realUserMonitoring` (`recordError` no-op vs.
  forwarding, http telemetry config).
- **Verified end-to-end on exp6**: confirmed in the RUM console that JS errors,
  HTTP errors (incl. a real `status: 400`), and readable `keepNames` stack traces
  are all captured and ingested.


jimlerza:
I looked over the new documentation, and I think you missed a step. After we determine the user's DAWSON `userId`, we can filter errors for that `userId` like so:

Screenshot 2026-07-23 at 9 42 57 AM

jimlerza (approved):

web-client/src/router.ts
```diff
@@ -97,14 +101,31 @@ const ifHasAccess = (
...
     setPageTitle('U.S. Tax Court');
     // expose route function on window for use with cypress
 
     (window as Window & { __cy_route?: (path: string) => void }).__cy_route =
       path => route(path || '/');
     const { ROLE_PERMISSIONS } = app.getState('constants');
 
+    // Wrap every route registration to fire a RUM page view event on
+    // navigation, mirroring the trackedRoute pattern in routerPublic.ts.
+    const registerRoute = (
```
zacharybrown204, line 114: ❤️

web-api/terraform/modules/rum/rum.tf
```diff
@@ -65,3 +65,160 @@ resource "aws_rum_app_monitor" "app_monitor" {
...
+  }
+}
+
+resource "aws_rum_app_monitor" "public_app_monitor" {
+  name   = "${var.environment}_dawson_public_rum_app_monitor"
+  domain = "*.${var.domain}"
+  app_monitor_configuration {
+    allow_cookies       = true
+    session_sample_rate = var.sample_rate
+    telemetries         = ["performance"]
```
zacharybrown204: Should this be set to performance, errors, and http?

I thought that if these weren't sent then RUM wouldn't accept events of those types, but looking at the test env this is not the case. I tried looking for some documentation on what exactly these values do on when set on AWS side but wasn't able to find any.

web-api/terraform/modules/rum/rum.tf
zacharybrown204: Same as below, should this be performance, errors, and http?

zacharybrown204 (approved):

paulo-filho-ctr (approved):

web-api/terraform/modules/rum/rum.tf
kampofo: the telemetries being used in rum are being configured and used in realUserMonitoring.ts where it specifies http, errors, and performance so those events can still be tracked. I looked into this a bit further and what I found was that AWS's RUM normally offers a getting started snippet in the AWS console and this setting here basically generates the snippet to configure the rum telemetries with only performance turned on.

So its separate from our rum config thats actually running this is a good catch though and it's a tiny change to get this in parity with the actually live configuration i'll add that change now!

web-api/terraform/modules/rum/rum.tf
kampofo: addressed in above comment.

PR #10142, opened 2026-06-09
9726 client side logging redo to test 1781032549

jdr81394:
## Summary

Adds client-side error handling and reporting so unhandled UI errors no longer
blank the page silently and are now captured in CloudWatch RUM. Three pieces:
a React error boundary, RUM error reporting, and readable production stack traces.

## Changes

**Error boundary**
- New `ErrorBoundary` component (`getDerivedStateFromError` + `componentDidCatch`)
  that renders a graceful fallback instead of unmounting the whole React tree
  (which previously left users on a blank white page).
- Wired into both entry points — `app.tsx` (`AppComponent`) and
  `appPublic.tsx` (`AppComponentPublic`) — inside the Cerebral `Container`.

**CloudWatch RUM error reporting**
- `realUserMonitoring.ts`: exposed a safe `recordError()` (no-op when RUM is not
  initialized) and call it from `ErrorBoundary.componentDidCatch`.
- Enabled the `'errors'` and `'http'` RUM telemetries so uncaught JS errors and
  failed HTTP requests (4xx/5xx and network failures, `recordAllRequests: false`)
  are captured automatically.

**Readable production stack traces**
- Added `keepNames: true` to the shared esbuild config so minified prod bundles
  preserve original function/class names — RUM traces now show `ErrorBoundary` /
  `AppComponent` instead of single-letter identifiers.

## Testing

- Unit tests for `ErrorBoundary` (error-state transition, default/custom fallback,
  `recordError` invocation) and `realUserMonitoring` (`recordError` no-op vs.
  forwarding, http telemetry config).
- **Verified end-to-end on exp6**: confirmed in the RUM console that JS errors,
  HTTP errors (incl. a real `status: 400`), and readable `keepNames` stack traces
  are all captured and ingested.

### DIFF
--- a/web-client/src/app.tsx
+++ b/web-client/src/app.tsx
@@ -5,6 +5,7 @@ import '../../node_modules/@fortawesome/fontawesome-svg-core/styles.css';
 import { AppComponent } from './views/AppComponent';
 import { AppInstanceManager } from './AppInstanceManager';
 import { Container } from '@cerebral/react';
+import { ErrorBoundary } from './views/ErrorBoundary';
 import { GlobalModalWrapper } from './views/GlobalModalWrapper';
 import { IdleActivityMonitor } from './views/IdleActivityMonitor';
 import { createForceRefreshCallback } from '@web-client/presenter/utilities/createForceRefreshCallback';
@@ -296,7 +297,9 @@ const app = {
             <AppInstanceManager />
             <GlobalModalWrapper />
           </>
-          <AppComponent />
+          <ErrorBoundary>
+            <AppComponent />
+          </ErrorBoundary>
 
           {process.env.CI && <div id="ci-environment">CI Test Environment</div>}
         </Container>,
--- a/web-client/src/providers/realUserMonitoring.ts
+++ b/web-client/src/providers/realUserMonitoring.ts
@@ -2,15 +2,25 @@ import { AwsRum, AwsRumConfig } from 'aws-rum-web';
 
 let awsRum: AwsRum | undefined;
 
-export const initializeRealUserMonitoring = () => {
+export const initializeRealUserMonitoring = (): void => {
   if (process.env.ENV === 'local') return;
   try {
+    const identityPoolId = process.env.RUM_IDENTITY_POOL_ID;
+    const sampleRateStr = process.env.RUM_SAMPLE_RATE;
+    const appMonitorId = process.env.RUM_APP_MONITOR_ID;
+
+    if (!identityPoolId || !sampleRateStr || !appMonitorId) {
+      throw new Error(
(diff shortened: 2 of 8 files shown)


web-client/src/providers/realUserMonitoring.ts
```diff
@@ -1,16 +1,25 @@
...
       allowCookies: true,
       enableXRay: false,
       endpoint: 'https://dataplane.rum.us-east-1.amazonaws.com',
       identityPoolId: process.env.RUM_IDENTITY_POOL_ID,
       sessionSampleRate: Number(process.env.RUM_SAMPLE_RATE!),
-      telemetries: ['performance'],
+      telemetries: [
+        'performance',
+        'errors',
+        // Record failed HTTP requests (4xx/5xx and network errors). axios runs
```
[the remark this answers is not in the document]
jimlerza: I agree here. Should be easy enough to default sample rate to 0 and app id to an empty string if they are not set.

jimlerza (changes requested):
I agree with copilot's type suggestions. Looks good otherwise.

web-client/src/providers/realUserMonitoring.ts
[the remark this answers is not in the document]
jdr81394: addressed

jimlerza (approved):

akuny (approved):

web-client/src/providers/realUserMonitoring.ts
```diff
@@ -29,3 +48,16 @@ export const initializeRealUserMonitoring = () => {
     console.log('Error initializing real user monitoring: ', error);
   }
 };
+
+/**
+ * Records an error to CloudWatch RUM. Safe to call when RUM has not been
+ * initialized (e.g. local env or failed initialization) - it is a no-op.
+ */
+export const recordError = (error: unknown): void => {
+  if (!awsRum) return;
```
TomElliottFlexion, line 57: Not sure if it's necessary, but we could consider either adding a retry of  `initializeRealUserMonitoring` here, or an error/console.log. I do realize that an error would potentially be annoying and obviously wouldn't log to RUM, but a silent failure makes me a _little_ nervous

TomElliottFlexion (approved):

web-client/src/providers/realUserMonitoring.ts
jdr81394: Not a bad idea at all - I agree. On the to staging branch I'll add one of the two.

web-client/src/app.tsx
```diff
@@ -296,7 +297,9 @@ const app = {
             <AppInstanceManager />
             <GlobalModalWrapper />
```
nechamakrigsman, line 298: Does `GlobalModalWrapper` also need to be wrapped in the `ErrorBoundary` - like how its done in `appPublic.tsx`

PR #10156, opened 2026-06-12
10076 notice setting case trial to test 1781277389

paulo-filho-ctr:
[Test Rail Results](https://ustaxcourt.testrail.io/index.php?/runs/view/412&group_by=cases:section_id&group_order=asc&display=tree)
## Summary
Updates the **Notice Setting Case For Trial (NTD)** PDF for both **remote** and **in-person** trial sessions so it matches the Court's standardized notice formatting used on other order-style documents.
### Header & caption
- Replaces the plain-text `PrimaryHeader` / `DocketHeader` with the **order-style court header** (`OrderPrimaryHeader`: seal, decorative court name, Washington, DC address).
- Adds a dedicated **`NoticeSettingCaseForTrialDocketHeader`** with:
  - Uppercase petitioner names and a trailing comma
  - Uppercase `COMMISSIONER OF INTERNAL REVENUE,` respondent line
  - Uppercase underlined document title: `NOTICE SETTING CASE FOR TRIAL`
### Judge info box
- Extracts judge details into **`TrialNoticeJudgeInfoBox`**, showing:
  - Judge last name only (no `"Judge"` prefix)
  - Chambers address (`400 Second St., NW` / `Washington, DC 20217`)
  - `Chambers Phone: {chambersPhoneNumber}` from the trial session
### Body & signature
- Applies **Times New Roman** styling to the notice body via `notice-of-trial-issued.scss`.
- Adds **`DocumentParagraphIndent`** (shared `INDENT` constant) to indent the first line of body paragraphs.
- Right-aligns the Clerk of the Court signature block using the existing `ClerkOfTheCourtSignature` component.
### Data / generation
- **`generateNoticeOfTrialIssuedInteractor`** now sets `formattedJudge` from `trialSession.judge.name` (last name) instead of the titled judge string from `getJudgeWithTitle`.
- Applies to all NTD generation paths (Set Calendar flow). **Paper-service address label/cover sheet generation is unchanged.**


alexdziarn:
Just wanted to confirm the approach for this story. Are we planning to create a new component in `NoticeSettingCaseForTrialDocketHeader.tsx`?

I had thought the plan was to update `DocketHeader.tsx` so the change would be applied consistently across most or all document types.

shared/src/business/utilities/pdfGenerator/components/DocumentParagraphIndent.tsx
```diff
@@ -0,0 +1,4 @@
+import { INDENT } from '@shared/business/entities/EntityConstants';
+import React from 'react';
+
+export const DocumentParagraphIndent = () => <>{INDENT}</>;
```
devinganey613, line 4: I really like this

devinganey613 (commented):
I think just updating DocketHeader.tsx makes sense. The two files are really similar and they probably want the consistent styling across documents.

jimlerza (approved):

akuny (approved):

TomElliottFlexion (approved):

Issue #9561, opened 2025-12-30
CotC Dashboard - Petitions


cholly75:
As a Clerk of the Court, so that I can easily see up to date information about petition volume, I need current year petition data to be shown on my dashboard.

This story will address the Petitions data required as part of the epic #9112 . See that story for overall design.

## Pre-Conditions
#9392 

## Acceptance Criteria
* CotC dashboard contains a new tab labeled "Petitions"
* CotC can toggle all displayed data between YTD (1/1 - 12/31) and FYTD (10/1 - 9/30)
* Petitions tab contains the following data:
  - Pie chart showing overall number of petitions filed F/YTD w/ percentage of Paper and Electronic
  - Pie chart showing overall number of petitions filed F/YTD w/ percentage of Pro Se and Represented
  - Bar graph displaying totals by month broken down by Paper/Electronic

## Notes
Relevant mock from #9112 : 

<img width="1122" height="1626" alt="Image" src="https://github.com/user-attachments/assets/8bfefe21-5bb1-4355-9488-58a509f71878" />

### Product Owner
 - [ ] Acceptance criteria have been met and validated on the Court's test environment.
 - [ ] Associated test cases defined in TestRail have been updated if necessary.
 - [ ] Successful test run is performed in TestRail.
 - [ ] User guides are updated if necessary.

### UX
 - [ ] All new functionality has been verified to work with keyboard navigation and screen reader software.
 - [ ] UI should be touch optimized and responsive for external users.

### Engineering
 - [ ] Automated test scripts have been written, including visual tests for newly added PDFs.
 - [ ] Successful test run is performed in TestRail.
 - [ ] New screens have been added to cypress accessibility axe.
 - [ ] Interactors should validate entities before calling persistence methods.
 - [ ] Types have been added to all added and updated functions.
 - [ ] Code refactored for clarity and to remove any known technical debt.
 - [ ] Acceptance criteria for the story has been met.
 - [ ] All schema changes are documented in the Data Dictionary (DD), and the Entity Relationship Diagram (ERD) is updated as appropriate.
 - [ ] If there are special deployment instructions, they have been added to the `CHANGES.md` file and the PR description.
 - [ ] Code that resides in the shared folder that only runs on the API or browser has been moved to either /web-client or /web-api.

Evelyn-Durbin:
This ticket relies on #9392 Library Graph to be completed

Evelyn-Durbin:
Delivery expectations for this ticket:
[...]

Issue #9392, opened 2025-11-19
Library: Graph


Evelyn-Durbin:
**To update the library, please follow these guidelines.**
Figma:
https://www.figma.com/design/S53FiPR3UEgA7AVUVymVy6/DAWSON-Library?node-id=0-1&t=yEwiNPKaNCvDGE3W-1

**Design Acceptance Criteria**

- Hover and Click interaction
- Mobile vs Desktop
- Text font and size
- Color
- Margins and padding
- Navigate via screen reader
- Tab through the site

**Deliver:**
- link to components in Dawson Library, viewable different variants (primary, secondary, etc) + interactions (hover, focus, etc) + screen reader functionality.
- single implementation in Dawson (for 2+ variants with links + context).

**Process to handle Dawson Library Tickets:**
DEVELOPMENT
Create PR to Staging
TEST IN LOCAL AS GOING
PUSH TO EXP9
Move to UX Review and post in slack for Evelyn Review (provide links + context) + Comments in issue.
Evelyn DENIAL
_ Evelyn moves back to in-progress, rinse + repeat.
Evelyn APPROVAL
_ Create PR to Test + Merge with any 1 approval
_ PUSH TEST
_ Move to UX Review and post in slack for Evelyn Review (provide links + context) + Comments in issue.
_ Evelyn DENIAL
___ Moved back to in-progress, rinse + repeat.
_ Evelyn APPROVAL
___ Move to Review for Niusha/Tenille + Evelyn notifies Niusha/Tenille
_____ Niusha/Tenille DENIAL
_______ Moved back to in-progress, rinse + repeat.
_____ Niusha/Tenille APPROVAL
_______ 🚨Open PR to staging, make sure there are NO CHANGES. If changes, we need to redeploy to test.🚨
_______ Jim Merges Staging PR, all test pass, and has min 1 USTC + 1 Gunnison + 1 Flexion Reviews

Evelyn-Durbin:
Will need access to Chart.js (Ask Jake or Evelyn or Tenille)

Evelyn-Durbin:
https://www.figma.com/design/Xjbwwm1umYzUJ2gZuTOYHU/9112-%7C-Clerk-of-Court-Dashboard?node-id=144-3521&t=Bwc6dfVKMxXSu6sp-1

Evelyn-Durbin:
EXP9 UX Review
**Bar Graph:**

[...]

PR #10166, opened 2026-06-15
9561 cotc petitions to test

devinganey613:
Implements #9561, pulling petitions data required for first tab of the clerk of the court dashboard and displaying it using the new graph library. This PR is also the first time this graph library (#9392) is going into the test environment.


web-client/src/views/Public/DawsonLibrary/PieGraph.tsx
```diff
@@ -0,0 +1,248 @@
...
+    const rows = data
+      .map(d => {
+        const pct = total > 0 ? ((d.value / total) * 100).toFixed(1) : '0.0';
+        return `<tr><td>${d.name}</td><td>${d.value}</td><td>${pct}%</td></tr>`;
+      })
+      .join('');
+    const html = `<html><body><h2>${title}</h2><table border="1" cellpadding="4" cellspacing="0"><thead><tr><th>Name</th><th>Value</th><th>Percentage</th></tr></thead><tbody>${rows}</tbody></table></body></html>`;
+    const blob = new Blob([html], { type: 'text/html' });
+    window.open(URL.createObjectURL(blob), '_blank');
+  };
```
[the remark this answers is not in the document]
devinganey613: We just included updating window.open options to include noopener and noreferrer. Given that this data does not come from a user source, we did not think the rest was necessary. But we are open to the rest of the team's opinion.

web-api/src/business/useCases/reports/getClerkDashboardStatsInteractor.ts
```diff
@@ -0,0 +1,154 @@
...
+  const petitionData = await petitionsDataByYear(yearStart, yearEnd);
+
+  const petitonPaperMonths = petitionData.filter(
+    pet => pet.isPaper && pet.month,
+  );
+  const petitonElectronicMonths = petitionData.filter(
+    pet => !pet.isPaper && pet.isPaper != null && pet.month,
+  );
+
+  const petitionFullPaperMonths = Array.from({ length: 12 }, (_, i) => {
```
paulo-filho-ctr, line 73: Could we use an asceding sort order  `fiscalMonthPriority` instead of here?

paulo-filho-ctr (approved):

web-api/src/business/useCases/reports/getClerkDashboardStatsInteractor.ts
zacharybrown204: We're iterating over a range here because we want to have objects for each month even if we don't have data for that specific month. Later down stream the bar graphs line data and labels up using array position, so to ensure that everything lines up we're stepping through each "month" and checking if we have data for it.

jimlerza (approved):

AbelTaxCourt (approved):

Issue #10190, opened 2026-06-23
BUG: Some Court Users Unable To Edit Case Note(s) After the Seal Of Petitioner(s) Address


ddonaldson-ustax:
**Describe the Bug**
This bug currently lives in PROD/TEST/STG - Multiple user roles lose the functionality to add or edit existing 'Case Notes(s)' after the sealing of any Petitioner Addresses. These users can still DELETE case notes though.

This issue was reported to DAWSON support by an ADC user.  Adding/Editing of case notes works as per normal for other cases, it appears to only be an issue if a Petitioner address is sealed.

After investigating this issue in the TEST environment, any role that can view sealed petitioner addresses can still add/edit case notes on cases that have a sealed P address. Any role that cannot view sealed petitioner addresses cannot add/edit case notes on cases that have a sealed P address.

**OUTSTANDING QUESTION FOR DEVS** - Is there other functionality that is currently being impacted if a P's address is sealed?  Can you please be sure to review the code while also solving this specific issue? Thank you.

**Business Impact/Reason for Severity**
Medium-Severity - Impacted Internal Users Workflow

**In which environment did you see this bug?**
STG/PROD/TEST

**Who were you logged in as?**
ADC, docket, CSS, Petitions clerk, Admissions clerk, etc.

**What were you doing when you discovered this bug? (Using the application, demoing, smoke tests, testing other functionality, etc.)**
This bug was reported to us by a Court user via DAWSON Support helpdesk ticket

**To Reproduce**
1. Sign in as ADC 
2. Select a case & use that role to add or edit the Case Notes
3. Then logout; Sign in as Case Services Supervisor, navigate to the same case as above & Seal the Petitioner(s) addresses
4. Then Logout again; Sign back in as ADC
5. Go back to your original case try to Edit Case Notes
6. Note that the save button does not click, it will highlight in blue, but the save does not take affect. No errors in the logs, no errors in the console display
7. Next, log back in as the CSS (or any other role that can view sealed petitioner addresses), and navigate to that same case
8. Click on the Notes tab, and click Edit note
9. Make an edit to the case note and click save
10. The edited note saves successfully

Note: ADC and other users can still DELETE case notes when P's address is sealed.

**Expected Behavior**
ADC and any Court users - Should be able to ADD or EDIT the Case Note(s) Even though the Petitioner(s) Addresses are sealed

**Actual Behavior**
[...]

PR #10196, opened 2026-06-23
10190 ADC Unable to edit sealed petitioner case 1782246181

zacharybrown204:
# 10190: BUG Some Court Users Unable To Edit Case Note(s) After the Seal Of Petitioner(s) Address

#10190

## Overview

Certain internal users (ex ADC) were unable to add or edit case notes if a petitioners contact information is sealed. The issue was with our JOI validation on the petitioner. When a user tried to add a note we would try validating the case, but because the petitioner was sealed fields that are required were missing from the case and it was considered invalid. Because the case was invalid we could not update it. The case notes tab doesn't expect to receive any validation errors and so it did not display the validation errors when they were received, appearing to do nothing when we attempt to save the case notes.

## Changes

Updated the validation rules in the petitioner entity for the sealed fields. Now for those fields we check if the petitioner is sealedAndUnavailable then those fields are forbidden, otherwise they are required along with any other previous validation rules.

isAddressSealed and sealedAndUnavailable are two different fields, isAddressSealed should always be true if an address is sealed. However sealedAndUnavailable should only be true if the petitioner was filtered for users who are not able to see the sealed fields. So sealedAndUnavailable should be appropriate for the condition in the JOI rules.

## Verification

Manually sealing a peitioner and editing the case with both users who can and cannot see the sealed values. 
Added new cypress test to cover this scenario.

## Pull Request Checklist

_PRs that do not meet these criteria may be closed without review._

   - [x] Locally
   - [ ] Experimental Environment (if necessary)


shared/src/business/entities/contacts/Petitioner.ts
```diff
@@ -74,6 +74,36 @@ export class Petitioner extends JoiValidationEntity {
 
   static VALIDATION_RULES = {
     ...User.USER_CONTACT_VALIDATION_RULES,
+    address1: joi.when('sealedAndUnavailable', {
```
TomElliottFlexion, line 77: Not a problem for this workflow; my only concern with relaxing these validation rules is the possibility that it allows us to accidentally overwrite these fields in the database if a user without sealed permissions updates the case (petitioners included)

akuny (approved):

shared/src/business/entities/contacts/Petitioner.ts
```diff
@@ -74,6 +74,36 @@ export class Petitioner extends JoiValidationEntity {
...
+      is: true,
```
TomElliottFlexion, line 78: Unit test at some point

TomElliottFlexion (approved):

jimlerza (approved):
Excellent discussion on Petitioner.ts! Great points all around.

nechamakrigsman (approved):

Issue #10076, opened 2026-05-22
Generated PDF Formatting Changes - Notice Setting Case For Trial (NTD)


cholly75:
As a court, I want the generated Notice Setting Case for Trial PDF to follow standardized formatting rules, so that the document meets court presentation expectations and appears consistent and professional.

Part of the Court's project to standardize our documentation, the story focuses on changes to the generated Notice Setting Case For Trial document created when setting a calendar.

## Acceptance Criteria
* Replace plain text header with Gothic header/address version
* ~~Ensure font is Times New Roman across remainder of the notice~~
* The caption includes a comma immediately following the party names for both petitioners and respondent.
* The party names in the caption (for both petitioners and respondent) are displayed in all capital letters.
* The document title is displayed in all capital letters.
* In the "Judge" box:
  - Remove the word "Judge" from before the judge's last name
  - Add the following information underneath the judge's last name:
    - 400 Second St., NW
      Washington, DC 20217
      Chambers Phone: [TRIAL JUDGE CHAMBERS PHONE]
* The first line of text in each paragraph in the body of the document is fully indented relative to the left margin.
* The Clerk of the Court's signature block is aligned with the right margin as per other notices
* These formatting changes are present whenever an NTD is generated.
* When generating this notice for paper service there are NO changes to the address label/page generated.

## Notes
* Let's also change the text in the generated Notice Setting Case for Trial PDF to 14-point font and see how it looks/generates before deciding whether to incorporate that change.

<img width="839" height="851" alt="Image" src="https://github.com/user-attachments/assets/193a982c-47f1-4054-ad39-dd0d538a1c86" />

### Product Owner
 - [ ] Acceptance criteria have been met and validated on the Court's test environment.
 - [ ] Associated test cases defined in TestRail have been updated if necessary.
 - [ ] Successful test run is performed in TestRail.
 - [ ] User guides are updated if necessary.

### UX
 - [ ] All new functionality has been verified to work with keyboard navigation and screen reader software.
 - [ ] UI should be touch optimized and responsive for external users.

### Engineering
 - [x] Automated test scripts have been written, including visual tests for newly added PDFs.
 - [x] Successful test run is performed in TestRail.
 - [x] New screens have been added to cypress accessibility axe.
 - [x] Interactors should validate entities before calling persistence methods.
 - [x] Types have been added to all added and updated functions.
 - [x] Code refactored for clarity and to remove any known technical debt.
 - [x] Acceptance criteria for the story has been met.
[...]

PR #10158, opened 2026-06-12
10076-notice-setting-case-trial

paulo-filho-ctr:
# 10076: Generated PDF Formatting Changes – Notice Setting Case For Trial (NTD)

#10076

## Overview

This PR standardizes the formatting of the generated **Notice Setting Case For Trial** PDF (both the remote/Zoomgov and in-person variants) so the document meets the Court's presentation expectations and matches the approved Figma mocks. It swaps the legacy plain-text header for the Gothic court header/address block, restructures the caption and judge boxes, applies 14pt body text with fully indented paragraphs, right-aligns the Clerk of the Court signature block, and adds the judge's chambers address and phone number.

Per UX direction on the ticket, font changes to Times New Roman were explicitly dropped from scope (see the crossed-out acceptance criterion and the UX comment: "Let's not worry about fonts on this ticket").

## Changes

**Document templates** (`shared/src/business/utilities/pdfGenerator/documentTemplates/`)
- `NoticeOfTrialIssued.tsx` (remote) and `NoticeOfTrialIssuedInPerson.tsx` now use the shared `OrderPrimaryHeader` (Gothic court header) instead of the legacy `PrimaryHeader`, and the new `NoticeSettingCaseForTrialDocketHeader` component instead of the generic `DocketHeader`.
- Body paragraphs use a consistent `notice-body-indent` class (first-line indent relative to the left margin), and bolding is done via `.text-bold` spans rather than `<b>` tags.
- The remote notice restructures the "Access Remote Proceeding" section, placing **Meeting ID** and **Passcode** on a single line with added horizontal spacing between them (testing feedback), and styles the `www.zoomgov.com` link.
- The Clerk of the Court signature block is wrapped in `#notice-clerk-signature` so it aligns to the right margin.

**New shared PDF components** (`shared/src/business/utilities/pdfGenerator/components/`)
- `NoticeSettingCaseForTrialDocketHeader.tsx` – renders the caption with party names in ALL CAPS, a comma after the party names, "COMMISSIONER OF INTERNAL REVENUE, Respondent", the docket number, and the ALL-CAPS "NOTICE SETTING CASE FOR TRIAL" title.
- `TrialNoticeJudgeInfoBox.tsx` – renders the judge box with the word "Judge" removed from before the last name, plus the chambers address (`400 Second St., NW / Washington, DC 20217`) and, when present, `Chambers Phone: …`.

[...]

PR #10202, opened 2026-06-24
10076-notice-setting-case-trial-to-test-1782342750

paulo-filho-ctr:
# 10076: Standardize Notice Setting Case for Trial (NTD) PDF formatting

#10076

## Overview

This PR updates the generated **Notice Setting Case for Trial** PDF (remote and in-person variants) to match the Court's standardized document formatting rules. The notice now uses the Gothic court header/address block, Times New Roman typography throughout the body, standardized caption and judge-box layout, 56pt paragraph indents, and 14pt body text. UX feedback from the issue comments (header weight, seal sizing, spacing, and font consistency) was incorporated across two follow-up commits.

Paper-service address labels/pages are **unchanged** — only the notice PDF template and its generation pipeline were modified.

### PDF templates & layout
- **`NoticeOfTrialIssued.tsx`** (remote) and **`NoticeOfTrialIssuedInPerson.tsx`** (in-person):
  - Use shared `OrderPrimaryHeader` (Gothic seal + "United States Tax Court" + Washington, DC address) instead of a plain-text header.
  - Replace `<DocumentParagraphIndent />` with CSS-driven `.notice-body-indent` paragraphs (56pt first-line indent).
  - Replace `<b>` tags with `.text-bold` spans for consistent Times New Roman rendering.
  - Remote notice: restructure body sections (`.notice-access-remote-section`, `.notice-remote-closing`) to improve spacing and keep content on one page.

### Caption & judge box
- **`NoticeSettingCaseForTrialDocketHeader.tsx`**:
  - Caption party names uppercased with trailing comma after petitioners (`JOHN SMITH AND JANE SMITH,`).
  - Caption extension and "RESPONDENT" uppercased.
  - Document title remains `NOTICE SETTING CASE FOR TRIAL` (all caps).
- **`TrialNoticeJudgeInfoBox.tsx`**:
  - Displays judge last name only (no "Judge" prefix).
  - Adds chambers address (`400 Second St., NW`, `Washington, DC 20217`) and `Chambers Phone:` line beneath the name.
  - Prevents chambers phone from wrapping mid-number (`.info-box-chambers-phone`).

### Styling (`notice-of-trial-issued.scss`)
- Court header: reduced title weight (`font-weight: normal`), adjusted seal dimensions, standardized address sizing.
- Body text: 14pt font with 15pt line height; paragraph indents at 56pt.
- Info boxes: 12pt content with consistent line height.
- Remote-specific overrides (`.notice-of-trial-remote`) for tighter vertical spacing on the single-page remote notice.
- Clerk signature block retains right-margin alignment via existing `#notice-clerk-signature` rules.

### Generation pipeline
- **`generateNoticeOfTrialIssuedInteractor.ts`**:
  - Formats `chambersPhoneNumber` and `joinPhoneNumber` via `formatPhoneNumber` before passing to the PDF template.
  - Validates assigned judge exists in persistence via `getJudgeWithTitle` when a judge name is present; throws if the judge cannot be found.
  - Skips judge lookup when no judge is assigned (`"Not assigned"`).

### Tests & snapshots
- Updated visual PDF expected images:
  - `Notice_Trial_Issued.pdf.1.png`
  - `Notice_Trial_Issued_In_Person.pdf.1.png`
- Updated document generator test fixtures (simpler case caption, formatted phone numbers).
- Added/extended interactor tests for phone formatting, missing-judge handling, and judge-not-found error path.

## Verification

- [x] `npm run test:api:file -- web-api/src/business/useCases/trialSessions/generateNoticeOfTrialIssuedInteractor.test.ts` — **8/8 passed**
- [x] `npm run test:document-generation:file -- shared/src/business/utilities/documentGenerators/noticeOfTrialIssued.test.ts shared/src/business/utilities/documentGenerators/noticeOfTrialIssuedInPerson.test.ts` — visual diff tests updated with new expected PNGs (re-run in CI to confirm pixel match)
- [ ] Court test environment — PO validation of generated notice against acceptance criteria
- [x] TestRail test run

### Acceptance criteria mapping

| Criterion | Status |
|-----------|--------|
| Gothic header/address version | ✅ `OrderPrimaryHeader` |
| Times New Roman for remainder of notice | ✅ SCSS + `.text-bold` / `.text-italic` spans |
| Caption comma after party names | ✅ `{caseTitle.toUpperCase()},` |
| Party names in all caps | ✅ Petitioners, extension, and respondent |
| Document title in all caps | ✅ `NOTICE SETTING CASE FOR TRIAL` |
| Judge box: no "Judge" prefix; chambers address & phone | ✅ `TrialNoticeJudgeInfoBox` |
| 56px body paragraph indents | ✅ `.notice-body-indent { text-indent: 56pt; }` |
| Clerk signature right-aligned | ✅ Existing signature block styles |
| Applies to all NTD generations (remote + in-person) | ✅ Both templates updated |
| No changes to paper-service address label/page | ✅ Out of scope; no related files modified |
| 14pt body font (per issue Notes) | ✅ Applied to notice body paragraphs |

### UX feedback addressed
Per [issue comment](https://github.com/ustaxcourt/ef-cms/issues/10076#issuecomment) from @Evelyn-Durbin:
- All Nimbus fonts replaced with Times New Roman (court title retains Gothic styling via existing header component).
- "NOTICE SETTING CASE FOR TRIAL" header and body text under trial/judge boxes set to 14pt.
- Court title weight reduced; seal sizing adjusted.
- Body indents set to 56pt.


web-api/src/business/useCases/trialSessions/generateNoticeOfTrialIssuedInteractor.ts
```diff
@@ -67,10 +68,12 @@ export const generateNoticeOfTrialIssuedInteractor = async (
     title: string;
   };
   const trialInfo: FormattedTrialInfoType = {
+    ...trialSession,
+    chambersPhoneNumber: formatPhoneNumber(trialSession.chambersPhoneNumber),
```
TomElliottFlexion: I ask only because of the old seed data in the document generator tests: do we need to support phone extensions?

shared/src/business/utilities/pdfGenerator/documentTemplates/NoticeOfTrialIssuedInPerson.tsx
```diff
@@ -50,40 +49,37 @@ export const NoticeOfTrialIssuedInPerson = ({
       </div>
 
       <div id="notice-body">
-        <p>
-          <DocumentParagraphIndent />
+        <p className="notice-body-indent">
```
devinganey613, line 52: The paragraphs in this notice seem a little closer together than in the remote notice, and particularly there's less of a gap between the info box and the start of the notice text. I think it might some CSS in the class `notice-of-trial-remote` that isn't applied here accounting for the difference. (Sorry if this is intended, I didn't have access to the figma to check myself)

paulo-filho-ctr: You're right that the remote notice looks tighter — that's intentional.

The remote template uses the `.notice-of-trial-remote` class, which applies tighter margins and line-height so the longer remote copy (Meeting ID, passcode, join instructions, etc.) fits on **one page**. That constraint is explicit in the ticket; the in-person notice has shorter copy and doesn't need the same compression.

We only have a Figma spec for the **in-person** variant. There is no remote Figma, so we can't confirm whether remote should match in-person spacing 1:1. The in-person layout follows the spec; the remote layout is tuned for the one-page requirement.

Happy to revisit if UX provides a remote spec or if PO wants the variants visually aligned at the cost of a second page.

devinganey613 (approved):

shared/src/business/utilities/htmlGenerator/documents/notice-of-trial-issued.scss
```diff
@@ -1,23 +1,93 @@
 #notice-of-trial-pdf {
-  font-family: 'Times New Roman', Times, serif;
```
jimlerza, line 2: I am sure you've tested this locally, but unless I'm missing a new class in the document template, I think this changes the font on these documents.

I do see the new class on line 100 of this file is used but that doesn't set this font either.

jimlerza (approved):
Approving with a shout to @Evelyn-Durbin to check the font on this when it hits UX Review.

shared/src/business/utilities/htmlGenerator/documents/notice-of-trial-issued.scss
paulo-filho-ctr: Good catch on the diff — yes, we're removing `font-family: 'Times New Roman'` from `#notice-of-trial-pdf`, and that lines up with what we found while testing.

We tried several ways to satisfy the Times New Roman acceptance criteria on this notice, including scoping it to `#notice-body`. The blocker wasn't just how wide the selector was — it's how PDFs are built:

1. All document SCSS compiles into one stylesheet that gets injected into every PDF (`generateHTMLTemplateForPDF` → a single `<style>` block from `index.scss`).
2. The **SERVED** footer is rendered separately through Puppeteer's `footerTemplate` (`DateServedFooter` → `generatePdfFromHtmlHelper`), not as part of the notice markup.

Whenever we added Times New Roman — even with what looked like tight scoping — the notice body rendered correctly, but the **footer came out serif instead of sans-serif**. The footer wrapper does set `font-family: sans-serif` inline, but in practice the footer still picked up the serif context from the page stylesheet. We saw the same thing on `test` via #10158, which is why this PR backs that out.

For now the notice body uses the default PDF serif (`nimbus_roman` from `shared.scss`), consistent with most court documents. Everything else in this PR — 14pt body, 56pt indents, Gothic header, caption/judge box, `.text-bold` spans — is unchanged.

If UX confirms Times New Roman is still required (@Evelyn-Durbin), my read is we'd need a small follow-up outside this PR — something like an explicit `font-family: sans-serif` on `DateServedFooter`, or per-document CSS — rather than another selector tweak on the notice alone. Happy to spike that once UX weighs in.

nechamakrigsman (approved):

web-api/src/business/useCases/trialSessions/generateNoticeOfTrialIssuedInteractor.ts
paulo-filho-ctr: Good question — we don't need new extension handling for this PR.

`formatPhoneNumber` only adds hyphens when the value is a bare 10-digit string (e.g. `2025213339` → `202-521-3339`). Anything already formatted — including extensions like `1-721-740-9885 x4239` — is returned unchanged. So if production data includes an extension, it still renders on the notice; we just don't parse or reformat it.

That's the same behavior used elsewhere (standing pretrial orders, change-of-proceeding notices, etc.). Trial session phone fields are validated as free-form strings (max 100 chars), not strictly 10-digit numbers.

The old document-generator fixture (`1-721-740-9885 x4239`) was legacy seed data. This PR updates those fixtures to `(202) 521-3339` to match how chambers/join numbers are typically stored today. The interactor test covers the 10-digit → hyphenated path, which is the main gap we were closing.

If the Court wants a standardized extension display format, that would be a separate utility change across all documents that use `formatPhoneNumber` — out of scope for #10076.

TomElliottFlexion: Thanks for the explanation!

TomElliottFlexion (approved):

devinganey613 (approved):

PR #10204, opened 2026-06-25
9725 search by name improvements to test 1782420057

AbelTaxCourt:
# 9725: Search By Name Improvements

- #9725

- [Passing test run on TestRail
](https://ustaxcourt.testrail.io/index.php?/runs/view/411&group_by=cases:section_id&group_order=asc&display=tree)

## Overview

This PR updates Case Search by Name to use the same result-handling patterns as the improved Order and Opinion searches: a 5,000-accessible-result cap, 100-result pagination, sortable columns, a visible result count, and mobile-friendly sorting. It preserves backend relevance as the default `No.` order and keeps cases without State values at the bottom when sorting by State.

## Changes

- Increases the maximum Case Search by Name results to 5,000 accessible cases.
- Uses OpenSearch `search_after` pagination to continue retrieving cases when sealed or otherwise inaccessible cases are filtered from a result batch.
- Prevents inaccessible cases from consuming the 5,000-result limit and causing fewer accessible results to be returned when additional matches exist.
- Adds a docket-number tie-breaker to the OpenSearch relevance sort, preventing equal-scoring cases from being duplicated or skipped between `search_after` requests.
- Preserves the existing exact-match search behavior and falls back to the non-exact query only when the initial exact search has no results.
- Adds case search pagination at 100 results per page.
- Adds sortable Case Search by Name columns:
  - `No.`
  - `Petitioner(s)`
  - `Docket No.`
  - `Filed Date`
  - `Case Title`
  - `State`
- Adds mobile sorting controls for case search results.
- Adds a result count and limited-results indicator when the 5,000-result cap is reached.
- Extracts case-specific formatting, sorting, pagination, and display metadata into `advancedCaseSearchHelper`.
- Adds explicit case search sort and pagination state.
- Reuses `updateAdvancedSearchResultsSortSequence` across case, order, and opinion search tabs.
- Handles missing `Filed Date` and `State` values so blank values sort after populated values.
- Converts State and Territory abbreviations to full names while preserving blank and unknown values.
- Hides the limited-information and sealed-case boilerplate bullets for internal users on the Case Search tab, matching existing Order and Opinion behavior.
- Adds test coverage for accessible-result pagination, deterministic OpenSearch sorting, exact-to-non-exact fallback, result limits, and State/Territory formatting.
- Reorganizes and cleans up related Advanced Search code and files.

## Reviewer Notes

- The default `No.` order preserves backend relevance.
- OpenSearch results are sorted by relevance score and then docket number. The docket number provides stable ordering when multiple cases have the same relevance score.
- Authorization filtering still occurs in the interactor. If inaccessible cases are removed, the search continues from the last raw OpenSearch result until 5,000 accessible cases are collected or no additional matches remain. The `search_after` cursor uses the last raw result rather than the last accessible result so inaccessible records are not repeatedly fetched.
- Non-exact matching is used only when the initial exact search returns no matches.
- Per ticket discussion, cases without State values are sorted to the bottom in both State sort directions.

## Manual Deployment Steps

None.

## Pull Request Checklist

  - [x] Locally
- [ ] I assert that all DoD criteria for this issue have been met.


web-api/src/business/useCases/caseAdvancedSearchInteractor.ts
```diff
@@ -14,11 +14,17 @@ import {
   ProcedureType,
   MAX_CASE_SEARCH_RESULTS,
   US_STATES,
+  US_STATES_OTHER,
 } from '@shared/business/entities/EntityConstants';
 import {
   isAuthorized,
   ROLE_PERMISSIONS,
 } from '@shared/authorization/authorizationClientService';
+import type { Common } from 'node_modules/@opensearch-project/opensearch/api/_types';
```
akuny: Others may disagree, but for me importing an OpenSearch-specific abstraction into the business layer is leaky. You could define a `SearchAfter` type (or whatever you want to call it) in `caseAdvancedSearch` and re-export it from there.

web-api/src/business/useCases/caseAdvancedSearchInteractor.ts
```diff
@@ -79,32 +85,63 @@ export const caseAdvancedSearchInteractor = async (
...
+  };
+  const accessibleCases: CaseAdvancedSearchResult[] = [];
+  let searchAfter: Common.SortResults | undefined;
+  let useNonExactQuery = false;
 
-  const filteredCases = filterCaseSearchResultsNotAccessibleToUser(
-    foundCases,
-    authorizedUser,
-  ).slice(0, MAX_CASE_SEARCH_RESULTS);
+  while (accessibleCases.length < MAX_CASE_SEARCH_RESULTS) {
```
akuny, line 92: This loop makes repeated persistence calls because `filterCaseSearchResultsNotAccessibleToUser` may discard results, which can cause many round trips in the worst case. Maybe that filtering can be pushed into the query instead?

web-client/src/presenter/actions/AdvancedSearch/setAdvancedSearchResultsSortAction.ts
```diff
@@ -0,0 +1,34 @@
...
+  store,
+}: ActionProps<{
+  sortColumn: string;
+  sortDirection: typeof ASCENDING | typeof DESCENDING;
+}>) => {
+  const { sortColumn, sortDirection } = props;
+  const advancedSearchTab = get(state.advancedSearchTab) as string;
+  let stateKey;
+
+  if (advancedSearchTab === ADVANCED_SEARCH_TABS.CASE) {
```
akuny, line 25: The CASE branch sets state directly and returns early, while the others assign a stateKey, which makes this if block a bit difficult to reason about. Consider refactoring the CASE branch to assign a stateKey like the others, if the downstream logic allows it.

web-client/src/views/AdvancedSearch/SearchResults.tsx
```diff
@@ -1,123 +1,442 @@
...
+
+      updateAdvancedSearchResultsSortSequence({
+        sortColumn,
+        sortDirection,
+      });
+      setCurrentPaginationPageSequence({
+        advancedSearchTab: 'case',
+        currentPaginationPage: 0,
+      });
+    };
```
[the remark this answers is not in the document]
jimlerza: I agree with this.

web-client/src/views/Public/SearchBoilerplateText.tsx
```diff
@@ -28,7 +28,9 @@ export function SearchBoilerplateText({
 
   const showBullets =
     isPractitionerUser ||
-    (formTypeText !== 'an order' && formTypeText !== 'an opinion') ||
+    (formTypeText !== 'a case' &&
+      formTypeText !== 'an order' &&
+      formTypeText !== 'an opinion') ||
```
jimlerza, line 33: We should be able to assert this explicitly in a cypress test.

jimlerza (dismissed):
Everything here looks good, but I think we need some cypress tests to assert the new view logic and pagination behavior.

akuny (approved):
A few comments for consideration, particularly for `web-api/src/business/useCases/caseAdvancedSearchInteractor.ts`, but no blockers as far as I can tell.

web-client/src/views/AdvancedSearch/SearchResults.tsx
```diff
@@ -1,123 +1,442 @@
...
+                      iconAriaLabel={`Search is limited to ${MAX_CASE_SEARCH_RESULTS.toLocaleString()} results.`}
+                      iconClass="fa-icon-blue icon-spacing-4"
+                      title={`Search is limited to ${MAX_CASE_SEARCH_RESULTS.toLocaleString()} results.`}
+                    />
+                  )}
+                  <span
+                    aria-label={`Search is limited to ${MAX_CASE_SEARCH_RESULTS.toLocaleString()} results.`}
+                    className="cursor-default"
+                    title={`Search is limited to ${MAX_CASE_SEARCH_RESULTS.toLocaleString()} results.`}
+                  >
```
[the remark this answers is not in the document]
AbelTaxCourt: AC expects the message to show on hover even if the results don't hit the limit

web-client/src/views/Public/SearchBoilerplateText.tsx
AbelTaxCourt: Agreed, updated the case search cypress test to cover the boilerplate text. Also used Today's Orders patterns to cover new view logic and pagination behavior for case search.

web-api/src/business/useCases/caseAdvancedSearchInteractor.ts
AbelTaxCourt: Agreed, defined SearchAfter in `caseAdvancedSearch` instead.

web-api/src/business/useCases/caseAdvancedSearchInteractor.ts
AbelTaxCourt: Leaving the sealed-case filtering in the business layer for now.

The interactor already pages through OpenSearch results in batches of 5000, so the extra round-trip risk should be limited unless a search happens to match thousands of inaccessible sealed cases. Pushing this into OpenSearch would also mean duplicating access logic that currently lives in `filterCaseSearchResultsNotAccessibleToUser` / `isAssociatedUser`.

Can revisit this if we see evidence this path is causing latency.

web-client/src/presenter/actions/AdvancedSearch/setAdvancedSearchResultsSortAction.ts
AbelTaxCourt: Agreed, refactored the CASE branch to assign a stateKey like the others.

jimlerza (approved):
🚀

paulo-filho-ctr (approved):

paulo-filho-ctr (changes requested):
Can we have a link with passing test rail results?

jtdevos (approved):
Approve, but +1 to @paulo-filho-ctr's request re: wanting link to testrail results

ChristieLaferriereTaxCourt:
@paulo-filho-ctr @jtdevos Retesting test cases on Test Rail after implementing PR feedback began yesterday. We will continue to completion and link the successful test run then.

paulo-filho-ctr (approved):

PR #10284, opened 2026-07-21
10144 part 2 to test

TomElliottFlexion:
# 10144: BUG: Fix Entity Info — Restrict Public API Response Data (Part 2)

#10144

## Overview

This PR implements a systematic, allowlist-based mechanism for enforcing that public-facing API endpoints return only data belonging to approved public entity schemas. If a response contains any field not permitted by those schemas, the request fails with a 500 error rather than leaking data, making over-exposure a loud, detectable failure rather than a silent one.
This is **Part 2** of the fix. Part 1 (branch `10144-public-data`) introduced the foundational `PublicCase` and related entities; this branch builds on that by wiring in enforcement at the API Gateway middleware layer and auditing every public-facing interactor and Lambda response.

### New Enforcement Layer

- **`web-api/src/middleware/publicDataSanitizer.ts`** — Detects public-site requests by inspecting the `host` header (`public-api.*`) or URL path (`/public-api/`). Delegates to the shared `findUnauthorizedPublicFields` utility.
- **`web-api/src/middleware/apiGatewayHelper.ts`** — Adds `validatePublicResponse()`, called before returning any response on a public-site request. Throws `UnsanitizedEntityError` if unauthorized fields are detected, producing a 500 rather than silently leaking data.
- **`web-api/src/genericHandler.ts`** — Wired to participate in public response validation.

### Shared Validation Utility

- **`shared/src/business/utilities/publicDataValidation.ts`** — New utility that traverses a response object and checks every nested entity/object against a registry of known public entity factories. Detects both: (1) fields not present on any public entity, and (2) entities that were not constructed through a validated public entity class (`not_validated`). URL-exempt paths can be configured to bypass checking for specific non-entity routes.

### New/Updated Public Response Entities (DTOs)

All public API responses now flow through validated entity classes that explicitly declare their allowed fields:

| New Class | Replaces / Notes |
|---|---|
| `PublicCaseResponse` | Renamed from `PublicCaseDTO` |
| `RestrictedCaseResponse` | Replaced `RestrictedCaseDTO` |
| `PublicTrialSessionInfo` | Renamed from `PublicTrialSessionInfoDTO` |
| `FeatureFlagResponseDTO` | New — validates feature flag responses |
| `HealthCheckResponse` | New — validates health check responses |
| `PublicDocumentDownloadUrl` | New — validates download URL responses |
| `PublicDocketRecordPdfJobResponse` | New — validates docket record PDF job status |

### Interactors & Lambdas Updated to Use Response Entities

Public-facing interactors now wrap their return values in the appropriate response entity class before returning, ensuring the validated field set is enforced end-to-end:

- `getPublicCaseInteractor.ts`
- `getPublicDownloadPolicyUrlInteractor.ts`
- `getPublicTrialSessionDetailsInteractor.ts`
- `getPublicTrialSessionsInteractor.ts`
- `getAllFeatureFlagsInteractor.ts`
- `getHealthCheckInteractor.ts`
- `getCaseInteractor.ts`
- `getPublicDocketRecordStatusLambda.ts`
- `generatePublicDocketRecordPdfLambda.ts`

### Removed: Public Users-in-Section Route

- **`web-client/src/proxies/users/getPublicUsersInSectionProxy.ts`** — Deleted; this route exposed user data that has no legitimate public use case.
- **`web-api/src/app-public.ts`** — Route registration for the above removed.
- Related client actions/computeds updated to remove dead references.

### Cypress Task Updated

- **`cypress/helpers/cypressTasks/network/assertCorrectNetworkData.ts`** — Refactored to delegate to the shared `findUnauthorizedPublicFields` utility, ensuring Cypress tests and the production middleware use identical validation logic.

### Infrastructure

- **`web-api/src/business/entities/cases/CaseFactory.ts`** / **`getEntityByName.ts`** — Updated to register the new response entity classes.

## Verification

- All modified files pass unit tests: `npm run test:api`, `npm run test:shared`.
- Linting passes: `npm run lint`.
- The Cypress `assertCorrectNetworkData` task was updated to use the same validation code as the middleware, so existing Cypress runs that intercept public API responses will automatically enforce the new allowlists.
- Manual verification: loading a public case page, public trial session, and health check endpoint no longer returns fields outside the permitted set; attempting to return an unvalidated entity produces a 500.

## Notes for Reviewers

- **Allowlist vs. blocklist**: The new approach is explicitly allowlist-based — a response field is only permitted if it is a known property of the matching public entity class. Any new endpoint that returns a new shape of data to the public site will require a corresponding public entity (or expansion of an existing one) to avoid a runtime `UnsanitizedEntityError`. This is by design.
- **Entity name detection**: The scanner relies on the `entityName` field being present in response objects to look up the correct allowlist. Objects without `entityName` emit an error.
- **Public-site detection**: Detection is based on the `host` request header (`public-api.` prefix) or the URL path prefix `/public-api/`. This covers both local dev (path-based routing) and deployed environments (subdomain-based routing).
- The `validatePublicResponse` check runs **after** the interactor returns but **before** the Lambda writes bytes to API Gateway. It is intentionally a hard failure — returning a 500 is preferable to silently leaking data. This means any future interactor that touches a public route and returns an unvalidated object will fail loudly in tests and locally before reaching production.
- The `PUBLIC_PATH_REGEX` in `publicDataSanitizer.ts` covers both the local dev path prefix (`/public-api/`) and the deployed subdomain convention (`public-api.*`). Both detection paths are tested in `apiGatewayHelper.test.ts`.

## Manual Deployment Steps

_None required._

- [ ] I have conducted thorough manual testing:
   - [x] Locally
   - [ ] Experimental Environment (if necessary)
- [ ] If this PR is for a user story, or tech debt (devex, opex, design debt, etc.) that is user-facing, I have created test case(s) in TestRail and executed them.


shared/src/business/utilities/publicDataValidation.ts
```diff
@@ -0,0 +1,365 @@
...
+      const fieldPath = args.path ? `${args.path}.${key}` : key;
+
+      if (Array.isArray(value) || isRecord(value)) {
+        nestedFindings.push(
+          ...findUnauthorizedPublicFieldsRecursive(value, fieldPath, args.url),
+        );
+      }
+    }
+
+    return nestedFindings;
```
jimlerza, line 257: **Nested scalar fields bypass validation**: For untagged nested objects, the validator only recurses into objects/arrays and ignores scalar keys. For example, `HealthCheckResponseDTO.s3.secret` produces no finding. Top-level numeric-only objects are also accepted (`:240`).

We should be able to add tests to assert validation of all nested keys that will show these limitations.

web-api/src/middleware/apiGatewayHelper.ts
jimlerza, line 67: This path never invokes `findUnauthorizedPublicFields()`. The comment on the preceding line assumes the lambda is "more advanced" - does this also mean "trusted to not leak non-public information" too?

web-api/src/business/useCases/caseAdvancedSearchInteractor.ts
```diff
@@ -137,7 +137,7 @@ export const caseAdvancedSearchInteractor = async (
       docketNumberWithSuffix: filteredCase.docketNumberWithSuffix,
       petitionerNames: filteredCase.petitioners?.map(p => p.name),
       petitionerStateNames: filteredCase.petitioners
-        ?.map(p =>
+        .map(p =>
           p.state
```
jimlerza, line 141: If we need the `?` on 138, we probably need it here too.

jimlerza (commented):
I asked Copilot to analyze our public routes for regressions and it came up with this suggestion:

Public `/sections/:section/users` responses use `User` entities, but `User` is absent from `PUBLIC_ENTITY_FACTORIES`; the sanitizer will reject every response with HTTP 500. Add an appropriate public DTO/entity or adjust the response shape.

shared/src/business/utilities/publicDataValidation.ts
TomElliottFlexion: Very very good point! New approach: remove all (non-entity) DTOs from the public site and ensure that every entity passed to the public site has had `.validate()` called on it. This shifts the burden of ensuring that any shape of nested response _must_ have been validated according to the shape defined by the allowlist of public-safe entities

web-api/src/middleware/apiGatewayHelper.ts
TomElliottFlexion: After looking at the history, this block was meant to ensure that lambdas that have constructed their own full response (rather than returning plain data) aren't subject to the pre-existing sanitization. As near as I can tell, this should only apply to responses that are intercepted in `genericHandler` to return a `X_MANUAL_REFRESH_REQUIRED` header. I've added validation to this branch with an exclusion for the refresh response, but obviously I'll need to watch closely and test to make sure this isn't brittle

web-api/src/business/useCases/caseAdvancedSearchInteractor.ts
TomElliottFlexion: Ah thank you, merge error. It should not be required

TomElliottFlexion:
> I asked Copilot to analyze our public routes for regressions and it came up with this suggestion:
> 
> Public `/sections/:section/users` responses use `User` entities, but `User` is absent from `PUBLIC_ENTITY_FACTORIES`; the sanitizer will reject every response with HTTP 500. Add an appropriate public DTO/entity or adjust the response shape.

Good news, this route is never used by the public site. `/public-api/judges` is, but returns a `PublicUser`. I've removed this route, proxy, etc from the public site

web-api/src/business/useCases/trialSessions/getPublicTrialSessionsInteractor.ts
```diff
@@ -1,14 +1,14 @@
...
 export const getPublicTrialSessionsInteractor = async (): Promise<
-  PublicTrialSessionInfoDTO[]
+  PublicTrialSessionInfo[]
 > => {
   const trialSessions = await getTrialSessions();
 
   return trialSessions
     .map(t => new TrialSession(t).toRawObject())
-    .map(trialSession => new PublicTrialSessionInfoDTO(trialSession))
+    .map(trialSession => new PublicTrialSessionInfo(trialSession))
```
jimlerza: Does this need to be validated before it is returned?

jimlerza (approved):

web-api/src/business/useCases/trialSessions/getPublicTrialSessionsInteractor.ts
TomElliottFlexion: It sure does, and that should have failed. I've updated and simplified the handling

web-api/src/business/getEntityByName.ts
```diff
@@ -12,8 +12,12 @@ import { User } from '@shared/business/entities/User';
...
-import { PublicCaseDTO } from '@shared/business/dto/cases/PublicCaseDTO';
-import { RestrictedCaseDTO } from '@shared/business/dto/cases/RestrictedCaseDTO';
+import { PublicCaseResponse } from '@shared/business/dto/cases/PublicCaseResponse';
+import { RestrictedCaseResponse } from '@shared/business/dto/cases/RestrictedCaseResponse';
+import { HealthCheckResponse } from '@shared/business/dto/public/HealthCheckResponse';
+import { PublicDocketRecordPdfJobResponse } from '@shared/business/dto/public/PublicDocketRecordPdfJobResponse';
+import { PublicDocumentDownloadUrl } from '@shared/business/dto/public/PublicDocumentDownloadUrl';
+import { PublicTrialSessionInfo } from '@shared/business/dto/trialSessions/PublicTrialSessionInfo';
 
 export const getEntityByName = (name: string) => {
```
jtdevos, line 32: Several DTOs returned from public interactors aren't registered here — FeatureFlagResponseDTO, PublicUser, PublicDocketEntry, PublicTrialSessionDetails, PublicCaseSearchResult, PublicContact, PublicDocumentSearchResult. When getEntityByName returns undefined, dataSecurityFilter silently passes the response through unfiltered. 

Should the aforementioned be added?

shared/src/business/dto/cases/PublicCaseResponse.ts
```diff
@@ -58,4 +71,36 @@ export class PublicCaseDTO {
...
+    irsPractitioners: joi.array().optional(),
+    isPaper: joi.boolean().optional(),
+    isSealed: CASE_IS_SEALED_RULE,
+    leadDocketNumber: CASE_LEAD_DOCKET_NUMBER_RULE,
+    partyType: JoiValidationConstants.STRING.valid(
+      ...Object.values(PARTY_TYPES),
+    ).optional(),
+    petitioners: joi.array().optional(),
+    privatePractitioners: joi.array().optional(),
+    receivedAt: JoiValidationConstants.ISO_DATE.required(),
```
jtdevos: `PublicCase`  treats `caseCaption` & `receivedAt` as optional, whereas they are required here.  Is this intentional, or should they agree (and if so, should it be optional or required)?

shared/src/business/dto/cases/PublicCaseResponse.ts
```diff
@@ -58,4 +71,36 @@ export class PublicCaseDTO {
     this.privatePractitioners = rawPublicCase.privatePractitioners;
     this.consolidatedCases = rawPublicCase.consolidatedCases;
   }
+
+  static VALIDATION_RULES = {
+    canAllowDocumentService: joi.boolean().optional(),
+    canAllowPrintableDocketRecord: joi.boolean().optional(),
+    canDojPractitionersRepresentParty: joi.boolean().optional(),
+    caseCaption: CASE_CAPTION_RULE,
```
jtdevos: `PublicCase`  treats `caseCaption` & `receivedAt` as optional, whereas they are required here.  Is this intentional, or should they agree (and if so, should it be optional or required)?

Currently it seems like you could pass a _valid_ PublicCase to construct a PublicCaseResponse that is _invalid_.

jtdevos (commented):
Added a couple questions.  I'm unfamiliar w/ the codebase so not sure if these are real issues or not.

shared/src/business/dto/public/HealthCheckResponse.ts
```diff
@@ -0,0 +1,69 @@
+import { JoiValidationEntity } from '@shared/business/entities/JoiValidationEntity';
+import joi from 'joi';
+
+export type S3BucketsStatusDTO = {
```
devinganey613, line 4: Do the files in this folder need tests?

web-api/src/business/getEntityByName.ts
TomElliottFlexion: Thank you, this is a good point and opened up a can of worms! I've added the new entities. Turns out, we've let this get pretty stale: CalendaredCase, CaseWorksheet, ConsolidatedCaseSummary, Contact, DocketEntryWorksheet, EligibleCase, Note, Petitioner, PractitionerDocument, RestrictedCase, TrialSessionInfoDTO, and WorkItem are also all missing. Obviously, I'd like just to fix those here, but I'm worried about the blast radius of regression testing all of that (some fields do seem to be stripped out on the front end). I'll add a devex ticket.

shared/src/business/dto/cases/PublicCaseResponse.ts
TomElliottFlexion: Good point. These rules were taken from Case, which doesn't take into account the sealed response variants. I've pulled all the relevant rules from PublicCase

shared/src/business/dto/public/HealthCheckResponse.ts
TomElliottFlexion: It certainly couldn't hurt!

devinganey613 (approved):

web-api/src/business/useCases/trialSessions/getPublicTrialSessionsInteractor.ts
```diff
@@ -1,14 +1,14 @@
...
 export const getPublicTrialSessionsInteractor = async (): Promise<
-  PublicTrialSessionInfoDTO[]
+  PublicTrialSessionInfo[]
 > => {
   const trialSessions = await getTrialSessions();
 
   return trialSessions
     .map(t => new TrialSession(t).toRawObject())
-    .map(trialSession => new PublicTrialSessionInfoDTO(trialSession))
+    .map(trialSession => new PublicTrialSessionInfo(trialSession).validate())
```
nechamakrigsman: Should the validate happen after the filter?

web-client/src/presenter/computeds/Public/publicTrialSessionDetailsHelper.ts
```diff
@@ -100,7 +100,7 @@ export const publicTrialSessionDetailsHelper = (
 
 const isRestrictedCaseDTO = (
   calendaredCase: CalendaredPublicCase,
-): calendaredCase is RestrictedCaseDTO => {
+): calendaredCase is RestrictedCaseResponse => {
   return calendaredCase.entityName === 'RestrictedCaseDTO';
```
nechamakrigsman: This would always return false since the entity name was updated to `RestrictedCaseResponse`.

web-api/src/middleware/apiGatewayHelper.ts
```diff
@@ -50,12 +77,8 @@ export const handle = async (event, fun) => {
         statusCode: 200,
       };
     } else {
-      const privateKeys = ['pk', 'sk', 'gsi1pk'];
-      (Array.isArray(response) ? response : [response]).forEach(item => {
-        if (item && Object.keys(item).some(key => privateKeys.includes(key))) {
-          throw new UnsanitizedEntityError();
-        }
-      });
+      validatePublicResponse(event, response);
```
nechamakrigsman, line 80: `validatePublicResponse` only runs for public route requests, so this check won't run for authenticated routes. Is that intentional?

shared/src/business/dto/cases/RestrictedCaseResponse.ts
```diff
@@ -0,0 +1,47 @@
...
+import { JoiValidationConstants } from '@shared/business/entities/JoiValidationConstants';
+import { JoiValidationEntity } from '@shared/business/entities/JoiValidationEntity';
+import joi from 'joi';
+
+export class RestrictedCaseResponse extends JoiValidationEntity {
+  public docketNumber: string;
+  public docketNumberSuffix?: string;
+  public docketNumberWithSuffix?: string;
+  public isPaper?: boolean;
+  public isSealed?: string;
```
nechamakrigsman: Nit - is this a boolean?

nechamakrigsman (commented):
Nice work!

web-api/src/business/useCases/trialSessions/getPublicTrialSessionsInteractor.ts
TomElliottFlexion: That _does_ make more sense, ty

web-client/src/presenter/computeds/Public/publicTrialSessionDetailsHelper.ts
TomElliottFlexion: Oops! Good catch

web-api/src/middleware/apiGatewayHelper.ts
TomElliottFlexion: It is intentional!

shared/src/business/dto/cases/RestrictedCaseResponse.ts
TomElliottFlexion: It is! Another inherited type mismatch

jtdevos (approved):

Issue #6551, opened 2018-11-13
Pay.gov Integration - E-filing Petition Flow (Mock Transactions)


JessicaMarine:
As a petitioner or practitioner, I want to be able to pay my filing fee directly from DAWSON, so that I can securely complete my petition filing and clearly understand whether my payment was successful, pending, or failed.

In order for a case to proceed, a Petitioner must pay the Court’s $60.00 filing fee.  Currently this process is completely disconnected from the e-filing submission flow, and the process for documenting paid fees is manual work for both the petitioner and the Court.

This feature will enable petitioners and practitioners to initiate their filing fee payment directly in DAWSON. DAWSON sends required payment details to the USTC payment portal, which uses that information to route the user to Pay.gov’s Hosted Collections Page to enter payment information. After payment is attempted, users return to DAWSON and can view the most up‑to‑date payment status within the case.

This story will address **invoking the payment portal for a single petition at the end of the petition e-filing flow**. As the HCP for pay.gov has not yet been built, we want to start by being able to invoke the portal and mock transactions using the payment portal test server.

## Pre-Conditions:
ustaxcourt/ef-cms#8488

## Acceptance Criteria:
* Payment Workflow Initiation
  - Users filing a petition are presented on the last page of the e-filing flow with an option to pay the required filing fee 
  - When the user selects the payment option, DAWSON sends the necessary payment and case information to the USTC payment portal.
  - The payment portal uses the information received from DAWSON to establish a connection to Pay.gov’s Hosted Collections Page.
  - The user can send mock transactions to the payment portal.

* Successful Payment Handling
  - If the payment portal reports that the payment was successful, DAWSON displays a message that the filing fee has been paid

* Pending Payment Handling
  - If the payment portal reports a “Pending” status, DAWSON displays that the payment is pending.

* Failed Payment Handling
  - If the payment portal returns a failed or declined payment, DAWSON displays that the filing fee has not been paid.
  - DAWSON presents a clear, user‑friendly explanation that the payment attempt failed.
  - If the user closes the Hosted Collections Page or abandons the workflow without completing payment, DAWSON shows that the filing fee remains unpaid.
  - If the payment portal cannot process the payment (e.g., Hosted Collections Page error), DAWSON displays that the payment failed and the fee is unpaid.

* Error Handling for Status Retrieval
[...]

PR #10311, opened 2026-07-30
6551 payment portal integration e filing petition flow to test 1785444142

devinganey613:
#6551 
Updated cerebral actions and sequences for initiating and processing the filing fee payment to fix an issue where the DAWSON error page flashed when a transaction failed, and to keep the loading spinner showing during the full process navigating to the hosted collections page and back to the dashboard. 

Created a new error message for if the process-payment endpoint fails to tell the user to contact support. Includes an update to the `ErrorNotification` component to support inserting the support email with styling and a mailto link


web-client/src/presenter/sequences/initFilingFeePaymentSequence.ts
```diff
@@ -7,5 +7,8 @@ export const initFilingFeePaymentSequence = [
   clearErrorAlertsAction,
   setWaitingForResponseAction,
   initFilingFeePaymentAction,
-  unsetWaitingForResponseAction,
+  {
+    success: [],
+    error: [unsetWaitingForResponseAction],
+  },
```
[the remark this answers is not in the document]
devinganey613: I tested this, and when you leave the app to go to the hosted collections page, the state is reset when you return to the app. This was deliberately removed so the loading spinner shows until you redirect to the HCP.

paulo-filho-ctr (approved):

web-client/src/presenter/actions/FilingFee/setFilingFeeAlertsAction.ts
```diff
@@ -27,5 +27,13 @@ export const setFilingFeeAlertsAction = ({ get, store }: ActionProps) => {
         overwritable: true,
       });
       break;
+    case 'unknown':
```
nechamakrigsman, line 30: Do we need to add a test for this scenario?

web-client/src/presenter/actions/FilingFee/setFilingFeeAlertsAction.ts
```diff
@@ -27,5 +27,13 @@ export const setFilingFeeAlertsAction = ({ get, store }: ActionProps) => {
...
+      store.set(state.alertError, {
+        message: 'Unable to verify payment status.',
+        title: 'Filing Fee Status Unknown',
```
nechamakrigsman: Nit: Wondering if the title's casing should be consistent with the other titles?

nechamakrigsman (approved):

web-client/src/presenter/actions/FilingFee/setFilingFeeAlertsAction.ts
devinganey613: I can check with Evelyn about this

web-client/src/presenter/actions/FilingFee/setFilingFeeAlertsAction.ts
devinganey613: Good catch, updated.

web-client/src/presenter/actions/FilingFee/setFilingFeeAlertsAction.ts
devinganey613: This has been updated

web-client/src/presenter/actions/FilingFee/processFilingFeePaymentAction.ts
```diff
@@ -10,20 +9,18 @@ export const processFilingFeePaymentAction = async ({
...
-          'Something went wrong when paying the filing fee. Please try again.',
-        overwritable: true,
-        title: 'Filing fee payment failed',
+    return {
+      processPaymentStatus: {
+        docketNumber: props.docketNumber,
+        paymentStatus: 'unknown',
       },
-    });
+    };
```
jimlerza, line 24: This is a wide net for all failure cases. Do we need more specific error handling for different types of payment failure?

web-client/src/views/ErrorNotification.tsx
jimlerza, line 1: In this diff, you've added "contact support" rendering for single and multiple messages, but the Cypress test only exercises the single-message path, and only for petitioners. We probably need coverage asserting the multiple-messages branch in this file.

web-client/src/presenter/actions/FilingFee/processFilingFeePaymentAction.ts
devinganey613: This path is for if our process-payment endpoint fails and we can't tell what happened without checking (e.g. the api connection to the payment portal app breaks, a database write fails). If the payment fails on pay.gov's side, the endpoint will return that status and we will take the original "failed" path in this action.

web-client/src/views/ErrorNotification.tsx
devinganey613: I don't think the multiple messages + contact support scenario currently happens anywhere in our app, I just added it as an option for later use. I'll either remove it or find a way to unit test it.

jimlerza (approved):

Issue #10164, opened 2026-06-15
BUG: State Selection Dropdown Contains Incorrect Sorting For States


ddonaldson-ustax:
**Describe the Bug**
Currently in DAWSON - Several State Selection Dropdown Fields have incorrect sorting applied & are not correctly being sorted in alphabetical order. 

The values appear partially ordered, but do not follow a proper full-string alphabetical sort. For example, items that should be ordered as:

Maine → Maryland → Massachusetts

May instead appear in an incorrect sequence such as:

Massachusetts → Maryland → Maine

This Bug can currently be found affecting multiple State Selection Dropdowns in various places within DAWSON

**`Current State Dropdown Selector(s) Affected`**

- Advanced Search for Case select United States [State Selection Dropdown]
- Petition workflow [State Selection Dropdown]
- Petition workflow Legal Residence [State Selection Dropdown]
- Petition workflow Petitioner + Spouse fields [State Selection Dropdown]
- Add Practitioner State Field + Original Bar State Field [State Selection Dropdown]
- Edit Practitioner State Field + Original Bar State Field [State Selection Dropdown]
- Add Trial Sessions Location Information [State Selection Dropdown]
- Edit Trial Sessions Location Information [State Selection Dropdown]
- Edit Petitioner Information (As Docket Clerk) [State Selection Dropdown]
- Petitioner Edits their OWN contact information [State Selection Dropdown]
- Practitioner Edits Their Own Contact Information [State Selection Dropdown]

**Business Impact/Reason for Severity**
Low Severity

**In which environment did you see this bug?**
Test / Staging

**Who were you logged in as?**
Petitioner / Trial Clerk / Docket Clerk / Admissions Clerk

**What were you doing when you discovered this bug? (Using the application, demoing, smoke tests, testing other functionality, etc.)**
Ad Hoc Testing

**Expected Behavior**
All State Selection Dropdown values should be sorted using a consistent, standard alphabetical (lexicographical) order across the full list.

**Actual Behavior**
Dropdown values appear inconsistently sorted for State Selection Dropdown Fields, with items out of expected alphabetical order.

**Screenshots**

<img width="1031" height="736" alt="Image" src="https://github.com/user-attachments/assets/69652c30-bc62-4307-b566-3555f1a248aa" />

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Cause of Bug, If Known**

**Process for Logging a Bug:**
* Complete the above information
[...]

PR #10325, opened 2026-08-06
10164-state-selection-dropdown

paulo-filho-ctr:
# 10164: Sort state selection dropdowns alphabetically by full name

#10164

## Overview

State selection dropdowns across DAWSON were listing U.S. states and territories in abbreviation-key order (e.g., Massachusetts before Maryland before Maine; Puerto Rico before Palau) instead of alphabetical order by full name. This branch introduces shared `US_STATES_SORTED` and `US_STATES_OTHER_SORTED` constants—arrays of abbreviations ordered by full name—and uses them wherever state/territory dropdown options are rendered, so users see a consistent A–Z list in every affected selector.

## Changes

- Added `US_STATES_SORTED` and `US_STATES_OTHER_SORTED` to `EntityConstants`, derived from `US_STATES` and `US_STATES_OTHER` via `localeCompare` on full names. The original lookup maps are unchanged so validation lists and abbreviation lookups are unaffected.
- Exposed both sorted constants through private client constants (`getConstants.ts`) and the public app (`applicationContextPublic.ts`).
- Updated `getGroupedStateOptions` to build the "States" and "Other" optgroups from the sorted key arrays instead of `Object.entries`.
- Switched state/territory dropdown rendering to the sorted constants in:
  - Advanced Search — Case Search by Name (petitioner/respondent state filters)
  - Start Case — `StateSelect` and `PlaceOfLegalResidenceSelect` (petition workflow, petitioner/spouse address fields, and other flows that reuse these components such as edit-petitioner and contact-info updates)
  - Practitioners — Add/Edit Practitioner form and practitioner document form (state and original bar state fields)
  - Trial Sessions — In Person Proceeding location state field
- Added unit tests asserting both sorted constants contain every abbreviation in full-name order and that `getGroupedStateOptions` returns alphabetically sorted state and territory labels.

## Verification

- Unit tests added/updated:
  - `npm run test:shared:file -- shared/src/business/entities/EntityConstants.test.ts`
  - `npm run test:client:unit:file -- web-client/src/business/utilities/groupStatesOptions.test.ts`
- No manual deployment steps required; no schema, API, or migration changes.
- No Cypress or accessibility spec changes in this branch.
- No merged PR to `test` for this branch at time of writing.

## Notes

- In `CaseSearchByName.tsx`, the respondent "Other" optgroup renders `{usStatesOtherSorted[abbrev]}` while the petitioner optgroup uses `{usStatesOther[abbrev]}`. Because `US_STATES_OTHER_SORTED` is a key array (not a lookup map), reviewers may want to confirm the respondent labels render as intended.

Using the contributions gracefully sent by @arpitjain099

Using the contributions gracefully sent by @arpitjain099


web-client/src/views/AdvancedSearch/CaseSearchByName.tsx
```diff
@@ -38,6 +39,7 @@ export const CaseSearchByName: React.FC<CaseSearchByNameProps> = connect(
       sequences.updateCaseAdvancedSearchByNameFormValueSequence,
     usStates: state.constants.US_STATES,
     usStatesOther: state.constants.US_STATES_OTHER,
+    usStatesSorted: state.constants.US_STATES_SORTED,
```
jdr81394, line 43: I'm pretty sure that state.constants.US_STATES_SORTED will be undefined here when this renders on the public site. this component is shared via PublicSearch.tsx > CaseSearchForm.tsx > CaseSearchByName.tsx, but US_STATES_SORTED was only added to web-client/src/getConstants.ts, not applicationContextPublic.ts's frozenConstants. usStatesSorted.map(...) below (lines 333/369) will throw on mount for public users. Needs adding to applicationContextPublic.ts too.

jdr81394 (changes requested):

web-client/src/views/AdvancedSearch/CaseSearchByName.tsx
paulo-filho-ctr: You are correct, great catch! We new import and make use of the sorted US States and US Territories in the Public client.

arpitjain099:
@paulo-filho-ctr LGTM.

The important property is that `US_STATES` itself is untouched, so every lookup and validation list keyed by abbreviation keeps working and only the presentation order changes. Deriving `US_STATES_SORTED` from `Object.keys(US_STATES)` rather than maintaining a second hand-written list is the right call too, since the two can't drift apart.

The tests are the part I'd have asked for if they weren't there. Asserting set-equality against `Object.keys(US_STATES)` alongside the ordering check means a dropped or duplicated state fails the test rather than silently shipping, which is the real risk with a derived constant. Good catch adding `US_STATES_OTHER_SORTED` for the territories as well.

One question, not a blocker. `StateSelect` takes `useFullStateName`, and when it's `false` the options render as abbreviations while still being ordered by full name. So that variant shows `ME, MD, MA` (Maine, Maryland, Massachusetts), which reads as unsorted to someone seeing only the abbreviations. If any caller uses the abbreviation form on a long list, it may be worth sorting that case by abbreviation instead. If the abbreviation form is only used in short or contextual lists, this is fine as is.

One small ask, and no problem at all if it isn't appropriate. The original description carried a line crediting the fork I sent, and it looks like it came out in a later edit. Would you mind restoring it? I work on open-source security and contributions like this are part of the public record I point to, so an attribution line in the PR is genuinely useful to me. Entirely your call, and thanks either way for picking the change up and carrying it through the remaining dropdowns.

paulo-filho-ctr:
@arpitjain099 Thanks so much for the contribution and review — we tagged you on [#10325](https://github.com/ustaxcourt/ef-cms/pull/10325) because your fork was the right foundation, and we’d love to see you keep contributing. Since your comment, we’ve carried the sorted constants through the remaining dropdowns (including territories via `US_STATES_OTHER_SORTED`), wired both into the public client after review feedback, and added the unit test coverage you called out. On `useFullStateName`: every current `StateSelect` caller passes it as `true`, so users see full names today — if abbrev-only ever lands on a long list, sorting by abbreviation there would be a good follow-up. Attribution is back in the PR description too — appreciate you flagging that.

jimlerza (approved):

jdr81394 (approved):

Issue #10329, opened 2026-08-07
BUG: Validation error triggered incorrectly when editing Submitted/CAV or Pending Motions on the Judge's Dashboard


ttlenard:
**Describe the Bug**

A judge reported this issue to us.  The Judge wanted to edit the Status of Matter field when editing data on the Submitted CAV or Pending Motions section of the Judge's Dashboard. When he attempted to Save the data, he received a validation error in the Final Brief Due Date field that stated "Enter a Valid Date".

The date that was in the field was in the correct format.

We found a work around for the time being, all they have to do is click on the date picker and select the same date, and the validation error goes away.  For some reason the existing date that is in this field is throwing the validation error. We need to fix this.

**Business Impact/Reason for Severity**
medium

**In which environment did you see this bug?**
Test, Prod

**Who were you logged in as?**
Judge/Chambers

**What were you doing when you discovered this bug? (Using the application, demoing, smoke tests, testing other functionality, etc.)**
This issue was reported to us by a Judge

**To Reproduce**
1. Log in as a Judge or chambers
2. Click on the Submitted/CAV tab or the Pending Motions tab
3. Find data that has an existing date in the Final Brief Due Date field (this issue does not happen with freshly added dates - it's only existing dates in this field)
4. Click on the Edit link
5. Update the Status of matter
6. Click Save
7. Receive Validation error to "Enter a Valid Date"
8. See Console Error
9. We also reviewed the error in RUM (see below for specific error message)

**Expected Behavior**
Users should not be receiving a validation error when there is indeed a valid date in the Final Brief Due Date Field

**Actual Behavior**
A clear and concise description of what actually happened.
Users are getting a validation error when attempting to update the data on the Submitted CAV or Pending Motions from their dashboard, if there was an existing date in the Final Brief Due Date Field.

**Screenshots**

The error message and console error in the UI:

<img width="1615" height="966" alt="Image" src="https://github.com/user-attachments/assets/3884ae08-6ac2-4fe3-b2f2-c356ed0d44ae" />

The information from RUM:
` "event_details": {
    "version": "1.0.0",
    "type": "TypeError",
    "message": "Cannot read properties of undefined (reading 'name')",
    "filename": null,
    "lineno": null,
    "colno": null,
    "stack": "TypeError: Cannot read properties of undefined (reading 'name')\n    at callback (https://app.test.ef-cms.ustaxcourt.gov/index.PPRSZUQB.js:25:61832)",
[...]

PR #10332, opened 2026-08-07
10329 Normalize finalBriefDueDate before seeding the worksheet edit forms

arpitjain099:
#10329

## Overview

A judge editing the Status of Matter on the Submitted/CAV or Pending Motions tab gets "Enter a valid date" on the Final Brief Due Date field, on a date they never touched. Reselecting the same date in the picker clears it, which is the workaround in the ticket.

## Cause

The postgres mapper stores `finalBriefDueDate` as a `Date` and returns it with `.toISOString()`, so the API hands the client a full timestamp. Both worksheet entities validate that field with `JoiValidationConstants.DATE`, which is `joi.date().iso().format(['YYYY-MM-DD'])`. The two modal state actions copy the worksheet straight into `state.form`, so the form starts out holding a value the entity will reject:

```
new CaseWorksheet({ finalBriefDueDate: '2026-08-07T04:00:00.000Z' })
  -> { finalBriefDueDate: 'Enter a valid due date' }
new CaseWorksheet({ finalBriefDueDate: '2026-08-07' })
  -> null
```

Touching the date picker replaces the value through `formatDateFromDatePickerAction`, which writes it back as `YYYY-MM-DD`. That is why the workaround works, and it is the same reason the error only appears when the field is left alone.

## Change

Format `finalBriefDueDate` once when seeding `state.form`, in `setAddEditCaseWorksheetModalStateAction` and `setAddEditDocketEntryWorksheetModalStateAction`, so the form starts with the shape the entity validates. The API contract is unchanged, and nothing else that reads the timestamp is affected.

`formatDateString` resolves in `America/New_York`, which is the zone `calculateDate` used when the value was stored, so the date survives the round trip. I checked both sides of daylight saving:

```
2026-08-07T04:00:00.000Z (EDT) -> 2026-08-07
2026-01-15T05:00:00.000Z (EST) -> 2026-01-15
2026-08-07                     -> 2026-08-07   (already date-only, unchanged)
```

## Testing

Unit tests added to both action test files: one asserting the ISO timestamp is normalized and the resulting form passes entity validation, one asserting an already-correct value is left alone.

Reverting only the source change and rerunning makes the new case-worksheet test fail on the assertion, so it is covering the fix rather than passing incidentally.

Ran with node 24.18.1:

```
web-client/src/presenter/actions/CaseWorksheet/setAddEditCaseWorksheetModalStateAction.test.ts   4 passed
web-client/src/presenter/actions/PendingMotion/setAddEditDocketEntryWorksheetModalStateAction.test.ts   2 passed
```

eslint and tsc are clean on the four changed files.

I did not reproduce this against a running environment, since I do not have one, so the judge-facing symptom is inferred from the entity behaviour rather than observed. Worth a manual check on Test with an existing worksheet that has a due date set, editing only Status of Matter.

### DIFF
--- a/web-client/src/presenter/actions/CaseWorksheet/setAddEditCaseWorksheetModalStateAction.ts
+++ b/web-client/src/presenter/actions/CaseWorksheet/setAddEditCaseWorksheetModalStateAction.ts
@@ -1,3 +1,8 @@
+import {
+  FORMATS,
+  formatDateString,
+} from '@shared/business/utilities/DateHandler';
+import { RawCaseWorksheet } from '@shared/business/entities/caseWorksheet/CaseWorksheet';
 import { state } from '@web-client/presenter/app.cerebral';
 
 export const setAddEditCaseWorksheetModalStateAction = ({
@@ -11,9 +16,20 @@ export const setAddEditCaseWorksheetModalStateAction = ({
 
   const { submittedAndCavCasesByJudge } = get(state.submittedAndCavCases);
 
-  const caseWorksheet = submittedAndCavCasesByJudge.find(
-    ws => ws.docketNumber === docketNumber,
-  )?.caseWorksheet || { docketNumber };
+  const caseWorksheet: Partial<RawCaseWorksheet> & { docketNumber: string } =
+    submittedAndCavCasesByJudge.find(ws => ws.docketNumber === docketNumber)
+      ?.caseWorksheet || { docketNumber };
 
-  store.set(state.form, { ...caseWorksheet });
+  // The API returns finalBriefDueDate as a full ISO timestamp, while CaseWorksheet
+  // validates it as YYYY-MM-DD, so saving without touching the date picker fails
+  // validation on a date the judge never edited.
+  store.set(state.form, {
+    ...caseWorksheet,
+    ...(caseWorksheet.finalBriefDueDate && {
+      finalBriefDueDate: formatDateString(
+        caseWorksheet.finalBriefDueDate,
+        FORMATS.YYYYMMDD,
+      ),
+    }),
+  });
 };
--- a/web-client/src/presenter/actions/CaseWorksheet/setAddEditCaseWorksheetModalStateAction.test.ts
+++ b/web-client/src/presenter/actions/CaseWorksheet/setAddEditCaseWorksheetModalStateAction.test.ts
@@ -1,5 +1,8 @@
 import { MOCK_CASE } from '@shared/test/mockCase';
-import { RawCaseWorksheet } from '@shared/business/entities/caseWorksheet/CaseWorksheet';
+import {
+  CaseWorksheet,
+  RawCaseWorksheet,
+} from '@shared/business/entities/caseWorksheet/CaseWorksheet';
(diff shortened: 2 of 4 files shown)


paulo-filho-ctr:
@arpitjain099  Thanks so much for the contribution on [#10332](https://github.com/ustaxcourt/ef-cms/pull/10332) — the root-cause write-up and tests are really helpful, and we’d love to see you keep contributing.

One process note: external PRs should target **`test` first** (for Court review and PO verification), then **`staging`** after sign-off — not straight to `staging`. The full workflow is in [`docs/external-contributions.md`](https://github.com/ustaxcourt/ef-cms/blob/staging/docs/external-contributions.md); the short version is branch from `staging`, run [Pre-PR Validation](https://github.com/ustaxcourt/ef-cms/blob/staging/docs/external-contributions.md#pre-pr-validation), then open a PR to `test` with `npm run git:pr-to test` (see [Handling Merge Conflicts](https://github.com/ustaxcourt/ef-cms/blob/staging/docs/external-contributions.md#handling-merge-conflicts) if needed). Happy to help if you want to retarget this one — thanks again for picking it up.

web-client/src/presenter/actions/CaseWorksheet/setAddEditCaseWorksheetModalStateAction.ts
```diff
@@ -11,9 +16,20 @@ export const setAddEditCaseWorksheetModalStateAction = ({
...
-    ws => ws.docketNumber === docketNumber,
-  )?.caseWorksheet || { docketNumber };
+  const caseWorksheet: Partial<RawCaseWorksheet> & { docketNumber: string } =
+    submittedAndCavCasesByJudge.find(ws => ws.docketNumber === docketNumber)
+      ?.caseWorksheet || { docketNumber };
 
-  store.set(state.form, { ...caseWorksheet });
+  // The API returns finalBriefDueDate as a full ISO timestamp, while CaseWorksheet
+  // validates it as YYYY-MM-DD, so saving without touching the date picker fails
+  // validation on a date the judge never edited.
```
paulo-filho-ctr, line 25: While helpful, perhaps we can do without comments for cleaner code?

arpitjain099:
Thanks @paulo-filho-ctr, and apologies for the wrong target. I had the flow backwards in my head, and re-reading `docs/external-contributions.md` it is unambiguous: branch from `staging`, PR to `test` first for review and PO verification, then to `staging` once it passes.

Retargeted as #10334. I did not flip the base on this PR, since `test` and `staging` have diverged and the intermediary-branch step in Handling Merge Conflicts exists precisely to keep untested `test` changes out of a feature branch. The commit applied to `test` cleanly with no conflicts, so #10334 is the same four files and the same tests. Closing this one to keep your queue clean.

One thing worth flagging rather than leaving for you to find. Three cypress e2e jobs went red here. The only failure log I could read reports `Expected to find element: .docket-number-header a`, which is a docket-number header link and has nothing to do with the field this change touches, and cypress is green on #10325. I am not calling it pre-existing on that basis alone, so I will watch the run on #10334 and dig in if anything fails that touches the worksheet.

Thanks for the pointer to the docs, and for the review on the write-up.

PR #10334, opened 2026-08-07
10329 Normalize finalBriefDueDate before seeding the worksheet edit forms

arpitjain099:
#10329

Retarget of #10332 to `test`, per the external contribution flow. Same commit, applied onto `test` rather than `staging`. #10332 will be closed.

## Overview

A judge editing the Status of Matter on the Submitted/CAV or Pending Motions tab gets "Enter a valid date" on the Final Brief Due Date field, on a date they never touched. Reselecting the same date in the picker clears it, which is the workaround in the ticket.

## Cause

The postgres mapper stores `finalBriefDueDate` as a `Date` and returns it with `.toISOString()`, so the API hands the client a full timestamp. Both worksheet entities validate that field with `JoiValidationConstants.DATE`, which is `joi.date().iso().format(['YYYY-MM-DD'])`. The two modal state actions copy the worksheet straight into `state.form`, so the form starts out holding a value the entity will reject:

```
new CaseWorksheet({ finalBriefDueDate: '2026-08-07T04:00:00.000Z' })
  -> { finalBriefDueDate: 'Enter a valid due date' }
new CaseWorksheet({ finalBriefDueDate: '2026-08-07' })
  -> null
```

Touching the date picker replaces the value through `formatDateFromDatePickerAction`, which writes it back as `YYYY-MM-DD`. That is why the workaround works, and why the error only appears when the field is left alone.

## Change

Format `finalBriefDueDate` once when seeding `state.form`, in `setAddEditCaseWorksheetModalStateAction` and `setAddEditDocketEntryWorksheetModalStateAction`. The API contract is unchanged and nothing else that reads the timestamp is affected.

`formatDateString` resolves in `America/New_York`, the zone `calculateDate` used when the value was stored, so the date survives the round trip. Checked on both sides of daylight saving:

```
2026-08-07T04:00:00.000Z (EDT) -> 2026-08-07
2026-01-15T05:00:00.000Z (EST) -> 2026-01-15
2026-08-07                     -> 2026-08-07   (already date-only, unchanged)
```

## Testing

Unit tests added to both action test files: one asserting the ISO timestamp is normalized and the resulting form passes entity validation, one asserting an already-correct value is left alone.

Reverting only the source change and rerunning makes the new case-worksheet test fail on the assertion, so it covers the fix rather than passing incidentally.

Ran with node 24.18.1:

```
setAddEditCaseWorksheetModalStateAction.test.ts          4 passed
setAddEditDocketEntryWorksheetModalStateAction.test.ts   2 passed
```

eslint and tsc are clean on the four changed files.

I do not have a running environment, so the judge-facing symptom is inferred from the entity behaviour rather than observed. Worth a manual check in the test environment with an existing worksheet that has a due date set, editing only Status of Matter.

On the three cypress jobs that failed on #10332: the one failure log I could read is `Expected to find element: .docket-number-header a`, which is a docket-number header link and unrelated to this field. I will watch this run and dig in if anything fails that touches the worksheet.

### DIFF
--- a/web-client/src/presenter/actions/CaseWorksheet/setAddEditCaseWorksheetModalStateAction.test.ts
+++ b/web-client/src/presenter/actions/CaseWorksheet/setAddEditCaseWorksheetModalStateAction.test.ts
@@ -1,5 +1,8 @@
 import { MOCK_CASE } from '@shared/test/mockCase';
-import { RawCaseWorksheet } from '@shared/business/entities/caseWorksheet/CaseWorksheet';
+import {
+  CaseWorksheet,
+  RawCaseWorksheet,
+} from '@shared/business/entities/caseWorksheet/CaseWorksheet';
 import { applicationContextForClient as applicationContext } from '@web-client/test/createClientTestApplicationContext';
 import { judgeColvin } from '@shared/test/mockUsers';
 import { presenter } from '@web-client/presenter/presenter-mock';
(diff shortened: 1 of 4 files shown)


arpitjain099:
Comments removed in `030b349`, both files. Thanks for the review.

Two notes on CI, flagged rather than asserted since I cannot see the runner.

The earlier red shard on `9b3212e` (`e2e_cypress (16, 8)`) never executed a test. Cypress could not launch the browser: 62 rounds of "Still waiting to connect to Edge", then `connect ECONNREFUSED 127.0.0.1:35477` and "Could not find Cypress test run results". Zero assertions ran.

On `030b349` that shard is green and a different job is red, `e2e_cypress_smoketests_local (3, 1)`. That one is a genuine assertion:

```
1) Striking removes pending items
     "before each" hook for "Should not remove the case block when striking one of multiple pending items."
   AssertionError: Timed out retrying after 120000ms:
   expected 'http://localhost:1234/file-a-petition/step-1' to match /\/case-detail\/[^/]+/
   at getDocketNumberAfterPaperPetitionSubmit (cypress/helpers/fileAPetition/create-and-serve-paper-petition.ts:16:41)
```

I do not think it is this branch, for two reasons. `030b349` deletes five comment lines and changes nothing else, so it cannot alter behaviour, and the smoketests job did not run on `9b3212e` so there is no baseline to compare against. Within the same spec the first test passed in 25s using the same `createAndServePaperPetition` helper, and the failure is the second call to it timing out waiting for the paper petition to redirect to case detail. That path does not touch `finalBriefDueDate` or either worksheet modal.

Reads like flake or a shared-environment timeout to me, but you know the smoketests better than I do. Happy to re-run or dig further if you would rather be sure.

devinganey613:
@arpitjain099 thanks for contributing again! We decided to go in a different direction for this fix, instead updating the entities themselves along with the date pickers on the forms to use the ISO date format. See #10351 for more details. Closing this PR.

PR #10391, opened 2026-08-28
10389 story to test 1787932342

devinganey613:
#10389

### DIFF
--- a/shared/src/business/utilities/canPractitionerFileEntryOfAppearance.ts
+++ b/shared/src/business/utilities/canPractitionerFileEntryOfAppearance.ts
@@ -3,17 +3,11 @@ import {
   PRACTICE_TYPE,
   ROLES,
 } from '@shared/business/entities/EntityConstants';
-import { RawIrsPractitioner } from '@shared/business/entities/IrsPractitioner';
 import { RawPractitioner } from '@shared/business/entities/Practitioner';
 import { RawUser } from '@shared/business/entities/User';
 
-const isUserADojPractitioner = (
-  user: RawUser | RawPractitioner | RawIrsPractitioner,
-): boolean => {
-  if (user.role !== ROLES.irsPractitioner) return false;
-  const irsPractitioner: RawPractitioner = user as RawPractitioner;
-  if (irsPractitioner.practiceType !== PRACTICE_TYPE.DOJ) return false;
-  return true;
+const isUserADojPractitioner = (user: RawPractitioner): boolean => {
+  return user.practiceType === PRACTICE_TYPE.DOJ;
 };
 
 export const canPractitionerFileEntryOfAppearance = ({
@@ -34,7 +28,7 @@ export const canPractitionerFileEntryOfAppearance = ({
   if (user.role === ROLES.irsPractitioner) {
     const caseHasRespondent = !!caseDetail.irsPractitioners?.length;
     const isCaseSealed = isSealedCase(caseDetail);
-    const isDojPractitioner = isUserADojPractitioner(user);
+    const isDojPractitioner = isUserADojPractitioner(user as RawPractitioner);
     return (
       caseHasRespondent &&
       !isCaseSealed &&
--- a/shared/src/business/utilities/canPractitionerFileEntryOfAppearance.ts
+++ b/shared/src/business/utilities/canPractitionerFileEntryOfAppearance.ts
@@ -35,7 +35,6 @@ export const canPractitionerFileEntryOfAppearance = ({
     const caseHasRespondent = !!caseDetail.irsPractitioners?.length;
     const isCaseSealed = isSealedCase(caseDetail);
     const isDojPractitioner = isUserADojPractitioner(user);
-    console.log(caseHasRespondent, isCaseSealed, isDojPractitioner, caseDetail);
     return (
       caseHasRespondent &&
       !isCaseSealed &&
--- a/shared/src/business/utilities/canPractitionerFileEntryOfAppearance.ts
+++ b/shared/src/business/utilities/canPractitionerFileEntryOfAppearance.ts
@@ -1,0 +1,47 @@
+import { isSealedCase } from '@shared/business/entities/cases/Case';
+import {
+  PRACTICE_TYPE,
+  ROLES,
+} from '@shared/business/entities/EntityConstants';
+import { RawIrsPractitioner } from '@shared/business/entities/IrsPractitioner';
+import { RawPractitioner } from '@shared/business/entities/Practitioner';
+import { RawUser } from '@shared/business/entities/User';
+
+const isUserADojPractitioner = (
+  user: RawUser | RawPractitioner | RawIrsPractitioner,
+): boolean => {
+  if (user.role !== ROLES.irsPractitioner) return false;
+  const irsPractitioner: RawPractitioner = user as RawPractitioner;
+  if (irsPractitioner.practiceType !== PRACTICE_TYPE.DOJ) return false;
+  return true;
+};
+
+export const canPractitionerFileEntryOfAppearance = ({
+  user,
+  caseDetail,
+  canDojPractitionersRepresentParty,
+  hasPendingAssociation,
+}: {
+  user: RawUser;
+  caseDetail: RawCase;
+  canDojPractitionersRepresentParty?: boolean;
+  hasPendingAssociation?: boolean;
+}): boolean => {
+  if (user.role === ROLES.privatePractitioner) {
+    const isCaseSealed = isSealedCase(caseDetail);
+    return !hasPendingAssociation && !isCaseSealed;
+  }
+  if (user.role === ROLES.irsPractitioner) {
+    const caseHasRespondent = !!caseDetail.irsPractitioners?.length;
+    const isCaseSealed = isSealedCase(caseDetail);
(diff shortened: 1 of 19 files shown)


web-api/src/business/useCases/externalDocument/fileExternalDocumentInteractor.ts
jimlerza, line 181: This code trusts `consolidatedCasesToFileAcross` to not have been tampered with.

We probably need to recompute this list.

jimlerza, line 104: I mentioned this in a previous comment (later down in the file) - I think `consolidatedCasesToFileAcross` should probably be recomputed here. We should validate that the requester actually belongs to those cases.

shared/src/business/utilities/canUserFileFirstIrsFiling.ts
```diff
@@ -0,0 +1,23 @@
...
+  user: RawUser,
+  caseDetail: RawCase,
+): boolean => {
+  if (user.role === ROLES.irsPractitioner) {
+    const isCaseSealed = isSealedCase(caseDetail);
+    const isDojPractitioner =
+      (user as RawPractitioner).practiceType === PRACTICE_TYPE.DOJ;
+    const caseHasRespondent = !!caseDetail.irsPractitioners?.length;
+
+    return !caseHasRespondent && !isCaseSealed && !isDojPractitioner;
```
jimlerza, line 19: [blocking] This helper grants the server-side exception to every valid external filing whenever the case has no IRS practitioner and is unsealed. The client filters the first filing list with `canBeFirstIrsDocument`, but a caller can bypass that client-only filter and submit an arbitrary document directly to the API. Enforce the first-IRS-document allowlist (and any event/document pairing) in the interactor before allowing this exception.

web-api/src/business/useCases/externalDocument/fileExternalDocumentInteractor.ts
```diff
@@ -54,6 +60,25 @@ export const fileExternalDocument = async (
 
   const currentCaseEntity = new Case(currentCase, { authorizedUser });
 
+  const isFilingEntryOfAppearance =
+    user.role === ROLES.privatePractitioner &&
+    PRACTITIONER_ASSOCIATION_DOCUMENT_TYPES_MAP.some(
+      event => event.eventCode === documentMetadata.eventCode,
```
jimlerza, line 83: [blocking] This exception is keyed only by the client-supplied `eventCode`; it does not verify that `documentType` is the corresponding association document. `DocketEntry` separately validates both fields against the global allowlists, so an unauthenticated private practitioner can claim `eventCode: 'EA'` while submitting another valid document type (for example, `Answer`) and bypass the association check. Validate the event-code/document-type pair and restrict the exception to the intended association filing.

jimlerza (changes requested):
Request changes: the server-side authorization still has two additional bypasses, called out inline. First-IRS eligibility is not constrained to the allowed first-IRS document set, and the private-practitioner exception trusts an event code without validating its matching document type. The existing Copilot consolidated-target authorization finding is also unresolved and remains blocking. The other existing Copilot comments (reset shared mocks, remove the RawPractitioner assertion, and fix the Cypress typo) are non-blocking cleanup items; no replies or fixes were present.

shared/src/business/utilities/canUserFileFirstIrsFiling.ts
jimlerza: Eh, clanker got confused here. I don't think this is true given that "the caller" has already checked the event code against `PRACTITIONER_ASSOCIATION_DOCUMENT_TYPES_MAP` by this point.

web-api/src/business/useCases/externalDocument/fileExternalDocumentInteractor.ts
jimlerza: Hmm. Not sure I agree that an "unauthenticated private practitioner" can POST, so I'm not sure what that means.

web-api/src/business/useCases/externalDocument/fileExternalDocumentInteractor.ts
jdr81394: I agree with Jim^

jdr81394 (changes requested):

web-api/src/business/useCases/externalDocument/fileExternalDocumentInteractor.ts
devinganey613: Yeah I don't think this is necessary, I think we use the event code as the main key for the document type everywhere right? So if someone made a bad request with the wrong documentType, the app would probably still treat the upload as the eventCode that was passed.

web-api/src/business/useCases/externalDocument/fileExternalDocumentInteractor.ts
devinganey613: This has been fixed

jimlerza (approved):

jdr81394 (approved):

shared/src/business/utilities/canPractitionerFileEntryOfAppearance.ts
```diff
@@ -0,0 +1,46 @@
...
+import { RawUser } from '@shared/business/entities/User';
+
+const isUserADojPractitioner = (
+  user: RawUser | RawPractitioner | RawIrsPractitioner,
+): boolean => {
+  if (user.role !== ROLES.irsPractitioner) return false;
+  const irsPractitioner: RawPractitioner = user as RawPractitioner;
+  if (irsPractitioner.practiceType !== PRACTICE_TYPE.DOJ) return false;
+  return true;
+};
```
alexdziarn, line 11: why not just check if user.role === ROLES.irsPractititoner and user.practitionerType === PRACTICE_TYPE.DOJ

shared/src/business/utilities/canPractitionerFileEntryOfAppearance.ts
```diff
@@ -0,0 +1,46 @@
...
+  hasPendingAssociation?: boolean;
+}): boolean => {
+  if (user.role === ROLES.privatePractitioner) {
+    const isCaseSealed = isSealedCase(caseDetail);
+    return !hasPendingAssociation && !isCaseSealed;
+  }
+  if (user.role === ROLES.irsPractitioner) {
+    const caseHasRespondent = !!caseDetail.irsPractitioners?.length;
+    const isCaseSealed = isSealedCase(caseDetail);
+    const isDojPractitioner = isUserADojPractitioner(user);
```
alexdziarn: We do the same check again for if user.role === ROLES.irsPractitioner

alexdziarn (approved):
just some code cleanup comments

web-client/src/presenter/actions/isUserDirectlyAssociatedToCaseAction.test.ts
```diff
@@ -0,0 +1,150 @@
...
+        caseDetail: {},
+      },
+      state: {
+        user: {
+          role: ROLES.petitioner,
+        },
+      },
+    });
+
+    expect(pathNoStub).toHaveBeenCalled();
```
ChristieLaferriereTaxCourt: Great start here! Assertions here can be made stricter. Each test should use toHaveBeenCalledTimes(1) on the expected stub and not.toHaveBeenCalled() on the opposing stub, to verify exactly one path was taken. What are your thoughts?

web-client/src/presenter/actions/isUserDirectlyAssociatedToCaseAction.test.ts
```diff
@@ -0,0 +1,150 @@
...
+        },
+      },
+      state: {
+        user: {
+          role: ROLES.petitioner,
+        },
+      },
+    });
+
+    expect(pathNoStub).toHaveBeenCalled();
```
ChristieLaferriereTaxCourt: Same as above

web-client/src/presenter/actions/isUserDirectlyAssociatedToCaseAction.test.ts
```diff
@@ -0,0 +1,150 @@
...
+      },
+      state: {
+        user: {
+          role: ROLES.irsPractitioner,
+          practiceType: PRACTICE_TYPE.IRS,
+        },
+      },
+    });
+
+    expect(pathYesStub).toHaveBeenCalled();
```
ChristieLaferriereTaxCourt: Same here and the assertions below in this file

ChristieLaferriereTaxCourt (approved):

shared/src/business/utilities/canPractitionerFileEntryOfAppearance.ts
zacharybrown204: Good call, since this function is only being called once after we do the IRS check below, we removed it from this function. Since all that was blocking was the user casting we also moved that to the function below, then yeah we just return the result of the comparison.

web-client/src/presenter/actions/isUserDirectlyAssociatedToCaseAction.test.ts
zacharybrown204: Can do, and it has been done for all the assertions in this test!

paulo-filho-ctr (approved):

Issue #9460, opened 2025-12-05
SSO POC


cholly75:
As a Court User, so that I can login without needing my password, I need a single sign on option with the Court's Microsoft Entra ID tenancy.

This story is a proof of concept as outlined in #4826 .

## Acceptance Criteria
* Proof of concept that the strategy outlined in 4826 will work and should be pursued
* Working model for SSO integration with DAWSON

## Notes
* Can we do this and still retain the DAWSON-native login UI?
* Be aware of any competing efforts involving Cognito work/refactoring
* Zendesk automations creating/provisioning Court users?
* Some court users have multiple user logins

### Product Owner
 - [ ] Acceptance criteria have been met and validated on the Court's test environment.
 - [ ] Associated test cases defined in TestRail have been updated if necessary.
 - [ ] Successful test run is performed in TestRail.
 - [ ] User guides are updated if necessary.

### UX
 - [ ] All new functionality has been verified to work with keyboard navigation and screen reader software.
 - [ ] UI should be touch optimized and responsive for external users.

### Engineering
 - [ ] Automated test scripts have been written, including visual tests for newly added PDFs.
 - [ ] Successful test run is performed in TestRail.
 - [ ] New screens have been added to cypress accessibility axe.
 - [ ] Interactors should validate entities before calling persistence methods.
 - [ ] Types have been added to all added and updated functions.
 - [ ] Code refactored for clarity and to remove any known technical debt.
 - [ ] Acceptance criteria for the story has been met.
 - [ ] All schema changes are documented in the Data Dictionary (DD), and the Entity Relationship Diagram (ERD) is updated as appropriate.
 - [ ] If there are special deployment instructions, they have been added to the `CHANGES.md` file and the PR description.
 - [ ] Code that resides in the shared folder that only runs on the API or browser has been moved to either /web-client or /web-api.

Evelyn-Durbin:
https://github.com/ustaxcourt/ef-cms/issues/9429

PR #10410, opened 2026-09-02
9460 sso poc to test 1788374442

devinganey613:
# 9460: Single Sign-on Proof of Concept

#9460 

## Overview

This PR includes infrastructure changes, UI updates, and a new API endpoint required for court users to sign in to DAWSON using Microsoft Entra. These changes are hidden behind both an infrastructure flag, where the updates to cognito are not included if `IDP_NAME` is not set in secrets, and by a feature flag, `allow-idp-login`, that will hide the SSO login button if not set to `true`.

### Terraform updates
All of the following only occur if the proper secrets are set in AWS

- User pool client is updated to allow oauth flows
- Variables are set in client to handle OIDC login
- A new OIDC cognito identity provider is created to allow users authenticated in Microsoft Entra to log into our existing Cionito user pool
- A managed login page domain and default styling. A managed login page is required for this to work, however users will never see it; the native loging page using the api to authenticate will still be available, and SSO sign in bypasses the managed login page to authenticate directly with Microsoft Entra
- A pre sign-up lambda is created that will link incoming authenticated Microsoft users to their exisiting user in Cognito
- An inbound federation lambda is created that will remap the user's Microsoft email to keep it lowercase in cognito (Our taxcourt emails are formatted as First.Last@ustaxcourt.gov, and without this step the casing of our email is updated in Cognito, breaking the normal password login)
  - One todo is to see if the work done in the pre sign-up lambda can be done in this lambda
  - Terraform does not yet support setting the inbound federation trigger for a user pool client, so for the time being this is being set in our deployment script. Ideally terraform will be updated before we use this in production. See here https://github.com/hashicorp/terraform-provider-aws/pull/47954
- Roles and policies for these new lambda functions

### UI updates
- A new button is added to the login page that will redirect users to be authenticated with Microsoft, then redirected back to the app to login fully. This first step will generate a code after authentication, which will then be exchanged for tokens in a new api endpoint. 

### API updates
- A new endpoint, `/auth/code`, where the code generated after authenticating with Microsoft Entra is exchanged with Cognito for tokens. These tokens are sent back to the UI in the same format as the native login endpoint, including putting the refresh token in a cookie. 
- The `createUserInteractor` used by zendesk to create new users will now link a new internal user to their Microsoft account. This is the same work that happens in the pre sign-up lambda. Here it is done for new users, and in the latter it is done for existing users signing in using Microsoft for the first time.

## Verification

This is deployed and working in `exp4`

## Manual Deployment Steps

If SSO is needed in an environment, first set up a new application in Microsoft Entra for this environment, then set the following secrets. Those prefixed with `OIDC_` will be found on the Microsoft side

```bash
. ./scripts/env/set-env.zsh {YOUR_ENV}
./scripts/secrets/update-secret.ts -k IDP_NAME -v ustc_entra
./scripts/secrets/update-secret.ts -k OIDC_ISSUER_URL -v https://login.microsoftonline.com/{application id}/v2.0
./scripts/secrets/update-secret.ts -k OIDC_CLIENT_ID -v {app client id}
./scripts/secrets/update-secret.ts -k OIDC_CLIENT_SECRET -v {app client secret}
```
An account specific deployment will then be required for new roles that will be created
```bash
npm run deploy:account-specific
```

Once everything is deployed, set the feature flag to show the SSO button on the login page
```bash
./scripts/postgres/featureFlags/setup-idp-login.ts
```

If this feature is not needed, no updates are required. The terraform updates will not take place if `IDP_NAME` is not set. 

## Pull Request Checklist

_PRs that do not meet these criteria may be closed without review._

   - [x] Locally
- [ ] If this PR is for a user story, or tech debt (devex, opex, design debt, etc.) that is user-facing, I have created test case(s) in TestRail and executed them.
- [ ] All user-facing changes have been verified to be free of accessibility issues via manual testing and automated accessibility tests.


web-api/terraform/modules/everything-else-deprecated/cognito.tf
```diff
@@ -201,6 +205,185 @@ resource "aws_cognito_user_pool_client" "client" {
...
+  source_arn    = aws_cognito_user_pool.pool.arn
+}
+
+resource "aws_lambda_permission" "cognito_inbound_federation_lambda_invoke" {
+  statement_id  = "AllowCognitoInvoke"
+  action        = "lambda:InvokeFunction"
+  function_name = module.cognito_inbound_federation_lambda.function_name
+  principal     = "cognito-idp.amazonaws.com"
+  source_arn    = aws_cognito_user_pool.pool.arn
+}
```
paulo-filho-ctr, line 287: Super minor: maybe format using `terraform fmt`?

web-api/terraform/bin/deploy-app-all-colors.sh
```diff
@@ -136,9 +148,18 @@ terraform init -upgrade -backend=true \
...
+
+  # this is temporary until terraform supports setting inbound federation trigger
+  # we also have to re-set the presignup lambda trigger or it will be overwritten with the default and removed
+  if [[ -n "$IDP_NAME" ]]; then
+    USER_POOL_ID=$(terraform output -raw aws_cognito_user_pool_id)
+    aws cognito-idp update-user-pool \
+      --user-pool-id "$USER_POOL_ID" \
+      --lambda-config PreSignUp=arn:aws:lambda:us-east-1:"$LOWER_ENV_ACCOUNT_IDS":function:cognito_pre_signup_lambda_"$ENV",InboundFederation=\{LambdaVersion=V1_0,LambdaArn=arn:aws:lambda:us-east-1:"$LOWER_ENV_ACCOUNT_IDS":function:cognito_inbound_federation_lambda_"$ENV"\}
+  fi
+else
```
andrewcameronsims, line 166: Beware: I'm sure you guys know, but this API is a full replace API so it might also overwrite other configuration you have associated with the pool. You have to provide every attribute (not just every attribute in the lambda configuration).

https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-pool-updating.html#cognito-user-pool-updating-api-cli

devinganey613: Good point, I think I misinterpreted this.

web-api/terraform/modules/everything-else-deprecated/cognito.tf
devinganey613: Done!

web-api/terraform/modules/everything-else-deprecated/cognito.tf
```diff
@@ -157,6 +157,10 @@ resource "aws_cognito_user_pool" "pool" {
   lifecycle {
     prevent_destroy = true
   }
+
+  lambda_config {
+    pre_sign_up = module.cognito_pre_signup_lambda.arn
+  }
```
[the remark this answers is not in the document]
paulo-filho-ctr: Is Copilot comment addressed on this bullet in the description or is it something else?

  - Terraform does not yet support setting the inbound federation trigger for a user pool client, so for the time being this is being set in our deployment script. Ideally terraform will be updated before we use this in production. See here https://github.com/hashicorp/terraform-provider-aws/pull/47954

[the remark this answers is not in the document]
devinganey613: Oh no this actually is valid, I need to do add the trigger only if IDP_NAME is populated, like I did with other resources. Something like `pre_sign_up = IDP_NAME ? module.cognito_pre_signup_lambda.arn : null`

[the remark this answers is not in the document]
devinganey613: I missed the conditional add for most of the things related to this lambda, but all the other lines mentioned in this comment are actually already correct.

shared/admin-tools/user/admin.ts
```diff
@@ -128,15 +128,34 @@ export async function createOrUpdateUser(
...
+        role: rawUser.role,
+        sendWelcomeEmail: true,
+        temporaryPassword: password,
+        userId,
+      });
+    if (
+      cognitoUser &&
+      rawUser.entityName === 'User' &&
+      rawUser.role !== ROLES.petitioner
+    )
```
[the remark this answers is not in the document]
devinganey613: Fixed this using the feature flag instead

web-client/src/presenter/computeds/Login/LoginHelper.ts
```diff
@@ -0,0 +1,30 @@
...
+  idpLoginUrl: string;
+};
+
+export const loginHelper = (get: Get): LoginHelperType => {
+  const showIdpLoginButton =
+    !!process.env.IDP_NAME &&
+    !!get(state.featureFlags[ALLOWLIST_FEATURE_FLAGS.ALLOW_IDP_LOGIN.key]);
+
+  const idpLoginParams = new URLSearchParams();
+  idpLoginParams.append('identity_provider', process.env.IDP_NAME || '');
```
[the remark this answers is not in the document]
zacharybrown204: Added PKCE Challenges

web-client/src/presenter/computeds/Login/LoginHelper.ts
```diff
@@ -0,0 +1,30 @@
...
+  idpLoginParams.append(
+    'redirect_uri',
+    `https://app.${process.env.EFCMS_DOMAIN}/auth-code`,
+  );
+  idpLoginParams.append('response_type', 'CODE');
```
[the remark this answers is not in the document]
zacharybrown204: We are no longer setting this ourselves, as part of adding the PKCE challenges we added a package to handle creating the oath authorize request.

web-client/src/presenter/computeds/Login/LoginHelper.ts
```diff
@@ -0,0 +1,30 @@
...
+
+type LoginHelperType = {
+  showIdpLoginButton: boolean;
+  idpLoginUrl: string;
+};
+
+export const loginHelper = (get: Get): LoginHelperType => {
+  const showIdpLoginButton =
+    !!process.env.IDP_NAME &&
+    !!get(state.featureFlags[ALLOWLIST_FEATURE_FLAGS.ALLOW_IDP_LOGIN.key]);
```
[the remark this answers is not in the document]
zacharybrown204: Added test

web-api/src/app.ts
```diff
@@ -1222,6 +1223,7 @@ app.post(
     .delete(lambdaWrapper(deleteAuthCookieLambda))
     .post(lambdaWrapper(loginLambda));
   app.post('/auth/refresh', lambdaWrapper(renewIdTokenLambda));
+  app.post('/auth/code', lambdaWrapper(exchangeAuthCodeLambda));
```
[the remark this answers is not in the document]
zacharybrown204: Added to swagger spec

web-client/src/views/Login/Login.tsx
```diff
@@ -162,6 +164,21 @@ export const Login = connect(
                         Create your account now.
                       </Button>
                     </div>
+                    {loginHelper.showIdpLoginButton && (
+                      <div className="tw:flex tw:w-full">
+                        <Button
+                          aria-label="Sign in with Microsoft"
+                          variant="primaryTertiary"
```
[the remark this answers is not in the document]
zacharybrown204: Added the flags needed so the accessibility test can see the new IDP login button.

web-api/terraform/modules/everything-else-deprecated/cognito.tf
```diff
@@ -201,6 +205,193 @@ resource "aws_cognito_user_pool_client" "client" {
     "website",
     "zoneinfo",
   ]
+
+  callback_urls                = var.idp_name != "" ? ["https://app.${var.environment}.ef-cms.ustaxcourt.gov/auth-code"] : null
```
jimlerza: This URL happens to work for lower environments, but it will be wrong on prod. Prod's URL does not have "prod" in it.

Use `${var.dns_domain}` instead.

```suggestion
  callback_urls                = var.idp_name != "" ? ["https://app.${var.dns_domain}/auth-code"] : null
```

web-api/terraform/bin/deploy-app-all-colors.sh
```diff
@@ -136,9 +148,21 @@ terraform init -upgrade -backend=true \
...
+  # this is temporary until terraform supports setting inbound federation trigger
+  # you have to re-set everything in the user pool or it will be overwritten by defaults
+  if [[ -n "$IDP_NAME" ]]; then
+    USER_POOL_ID=$(terraform output -raw aws_cognito_user_pool_id)
+    SES_IDENTITY_ARN=$(terraform output -raw aws_ses_email_identity_arn)
+    sed \
+      "s|REPLACE_MAIN_POOL_ID|${USER_POOL_ID}|g;s|REPLACE_ENVIRONMENT|${ENV}|g;s|REPLACE_ACCOUNT_ID|${LOWER_ENV_ACCOUNT_IDS}|g;s|REPLACE_DNS_DOMAIN|${DNS_DOMAIN}|g;s|REPLACE_SES_IDENTITY_ARN|${SES_IDENTITY_ARN}|g" \
+      ../../bin/cognito-user-pool-update.yml > ../../bin/output.yml
+    aws cognito-idp update-user-pool --cli-input-yaml file://../../bin/output.yml
+    rm ../../bin/output.yml
```
jimlerza, line 164: The `aws cognito-idp update-user-pool` command replaces the pool, but this YAML input omits existing settings such as username attributes and schema attributes. This will result in our custom attributes (`userId` and `role`) to be lost.

web-api/src/lambdas/auth/exchangeAuthCodeLambda.ts
```diff
@@ -0,0 +1,32 @@
+import { createCookieString } from '@web-api/utilities/cookieFormatting';
+import { genericHandler } from '../../genericHandler';
+import { exchangeAuthCodeInteractor } from '@web-api/business/useCases/auth/exchangeAuthCodeInteractor';
+
+export const exchangeAuthCodeLambda = event =>
+  genericHandler(
```
jimlerza, line 6: `genericHandler` will log the full content of this request, which includes the `authCode` and PKCE verifier. We need to update its log redaction logic to redact these values.

web-api/terraform/modules/everything-else-deprecated/cognito.tf
```diff
@@ -201,6 +205,193 @@ resource "aws_cognito_user_pool_client" "client" {
...
+  user_pool_id  = aws_cognito_user_pool.pool.id
+  provider_name = var.idp_name
+  provider_type = "OIDC"
+  provider_details = {
+    authorize_scopes          = "openid"
+    client_id                 = var.oidc_client_id
+    client_secret             = var.oidc_client_secret
+    oidc_issuer               = var.oidc_issuer_url
+    attributes_request_method = "POST"
+  }
```
jimlerza, line 241: The OIDC provider maps `email` and `name`, and the pre-signup Lambda requires `email`, but Microsoft is only asked for `openid`. Optional claims may be absent, causing federated sign-in or `.toLowerCase()` to fail.

Request `openid email profile` from the upstream provider.

```suggestion
  provider_details = {
    authorize_scopes          = "openid email profile"
    client_id                 = var.oidc_client_id
    client_secret             = var.oidc_client_secret
    oidc_issuer               = var.oidc_issuer_url
    attributes_request_method = "POST"
  }
```

web-api/src/business/useCases/auth/exchangeAuthCodeInteractor.ts
```diff
@@ -0,0 +1,51 @@
...
+      .getHttpClient()
+      .post(`${process.env.MANAGED_LOGIN_DOMAIN}/oauth2/token`, params, {
+        headers: {
+          'Content-Type': 'application/x-www-form-urlencoded',
+        },
+      });
+
+    const expiresAt = applicationContext.getUtilities().calculateISODate({
+      dateString: applicationContext.getUtilities().createISODateString(),
+      howMuch: response.data.expires_in,
```
jimlerza: Cognito’s OAuth `expires_in` is the access-token lifetime. This cookie therefore expires after one hour, while Terraform configures refresh tokens for one day.

This could cause users to lose refresh capability prematurely. We should use the refresh-token lifetime, consistent with `loginLambda`, when calculating the cookie expiry here.

shared/admin-tools/user/admin.ts
```diff
@@ -128,15 +131,39 @@ export async function createOrUpdateUser(
...
+        SourceUser: {
+          ProviderName: process.env.IDP_NAME,
+          ProviderAttributeName: 'email',
+          ProviderAttributeValue: rawUser.email,
+        },
+        DestinationUser: {
+          ProviderName: 'Cognito',
+          ProviderAttributeValue: cognitoUser.Username,
+        },
+      });
```
jimlerza, line 167: If the feature flag is enabled while `IDP_NAME` is unset, a newly created internal user is linked with `ProviderName: undefined`, causing provisioning to fail after the Cognito user has already been created. We should verify that we have a complete IDP configuration before linking here.

web-api/src/business/useCases/auth/exchangeAuthCodeInteractor.ts
```diff
@@ -0,0 +1,51 @@
...
+    return {
+      accessToken: response.data.access_token,
+      idToken: response.data.id_token,
+      refreshToken: response.data.refresh_token,
+      expiresAt,
+    };
+  } catch (err: any) {
+    if (err.name === 'NotAuthorizedException') {
+      throw new UnauthorizedError('Invalid refresh token');
+    }
```
jimlerza: The token exchange uses Axios, whose invalid-code errors are not named `NotAuthorizedException`. Expired or reused codes therefore escape as generic errors and produce a 500 response instead of an authentication/client error. I think we can detect Axios `invalid_grant` responses here so that they also get mapped to an `UnauthorizedError`.

web-api/terraform/modules/everything-else-deprecated/cognito.tf
```diff
@@ -201,6 +205,193 @@ resource "aws_cognito_user_pool_client" "client" {
...
+  user_pool_id          = aws_cognito_user_pool.pool.id
+  managed_login_version = 2
+}
+
+resource "aws_cognito_managed_login_branding" "login_branding" {
+  count                       = var.idp_name != "" ? 1 : 0
+  client_id                   = aws_cognito_user_pool_client.client.id
+  user_pool_id                = aws_cognito_user_pool.pool.id
+  use_cognito_provided_values = true
+}
```
jimlerza, line 228: Branding depends on the managed-login domain existing, but Terraform has no dependency between these resources. Initial provisioning may call branding before the domain is ready.

```suggestion
resource "aws_cognito_user_pool_domain" "domain" {
  count                 = var.idp_name != "" ? 1 : 0
  domain                = "ef-cms-${var.environment}"
  user_pool_id          = aws_cognito_user_pool.pool.id
  managed_login_version = 2
}

resource "aws_cognito_managed_login_branding" "login_branding" {
  count                       = var.idp_name != "" ? 1 : 0
  client_id                   = aws_cognito_user_pool_client.client.id
  user_pool_id                = aws_cognito_user_pool.pool.id
  use_cognito_provided_values = true
  depends_on                  = [aws_cognito_user_pool_domain.domain]
}
```

web-api/swagger.json
```diff
@@ -8165,6 +8165,50 @@
         }
       }
     },
+    "/auth/code": {
```
jimlerza, line 8158: The endpoint accepts `authCode` and `code_verifier`, but Swagger documents no request body or `Set-Cookie` response header and incorrectly presents it as requiring Cognito authentication before login.
Please document the request schema, cookie response, error responses, and unauthenticated callback semantics.

jimlerza (changes requested):
I found a few things worth your attention.

Plus these, which can't be neatly tied to existing lines of code:

- Need a Lambda unit test for the new cookie and error behavior
- Need a Cypress test for exercising the SSO redirect/callback flow
- Add the manual deployment steps to `CHANGES.md`

web-client/src/presenter/actions/Login/idpLoginAction.ts
```diff
@@ -0,0 +1,45 @@
...
+    clientId,
+  );
+
+  const code_verifier = randomPKCECodeVerifier();
+  const code_challenge = await calculatePKCECodeChallenge(code_verifier);
+
+  applicationContext.getPersistenceGateway().setItem({
+    key: 'code_verifier',
+    value: code_verifier,
+  });
```
jimlerza, line 35: My copilot review included this:

> **Add OAuth state generation and validation** — `idpLoginAction.ts:27-42` uses PKCE but no `state`; `router.ts:1272-1278` and the callback action never validate one. PKCE does not replace OAuth state protection against login CSRF.

I know some of those words! But not if it's worth requesting here. Can you investigate this and determine if we need "state protection" here?

web-api/terraform/bin/deploy-app-all-colors.sh
devinganey613: The custom attributes are set by a different cli command, `add-custom-attributes`. Also verified we still have these attributes in `exp4` after deploying this code. In addition, the modifications to the standard attributes `name` and `email` can only be done on creation of the user pool, so the update command does not update these settings either.

web-api/src/business/useCases/auth/exchangeAuthCodeInteractor.ts
devinganey613: Good point didn't realize what token this expires_in variable referred to. Updated.

web-api/terraform/modules/everything-else-deprecated/cognito.tf
devinganey613: Updated

shared/admin-tools/user/admin.ts
devinganey613: Updated

web-api/terraform/modules/everything-else-deprecated/cognito.tf
devinganey613: Updated

web-client/src/presenter/actions/Login/idpLoginAction.ts
zacharybrown204: A state check has been added! I had thought the state check was a backup for a PKCE check, but they fill two different roles. A state check is for the client to help ensure that the request wasn't intercepted along the way, and a PKCE check is for the API to verify that the same client that made the auth request is making the token request.

web-api/src/lambdas/cognitoPreSignup/cognitoPreSignupLambda.test.ts
```diff
@@ -0,0 +1,106 @@
...
+
+const getCognito = jest.mocked(getCognitoMock);
+
+const adminGetUser = jest.fn();
+const adminLinkProviderForUser = jest.fn();
+
+getCognito.mockReturnValue({
+  adminGetUser,
+  adminLinkProviderForUser,
+} as unknown as ReturnType<typeof getCognitoMock>);
```
alexdziarn:
```suggestion
import { CognitoIdentityProvider } from '@aws-sdk/client-cognito-identity-provider';

const getCognito = jest.mocked(getCognitoMock);

const adminGetUser = jest.fn();
const adminLinkProviderForUser = jest.fn();

getCognito.mockReturnValue({
  adminGetUser,
  adminLinkProviderForUser,
} as Partial<CognitoIdentityProvider> as CognitoIdentityProvider);
```
some better typing rather than using unknown to force it

alexdziarn (changes requested):

web-api/src/lambdas/cognitoPreSignup/cognitoPreSignupLambda.test.ts
devinganey613: Nice thanks I like it

web-api/src/lambdas/auth/exchangeAuthCodeLambda.ts
zacharybrown204: Added redaction of authCode and the code_verifier next to the password redaction

web-api/src/business/useCases/auth/exchangeAuthCodeInteractor.ts
zacharybrown204: Updated error handling here. Instead of looking for an error name I'm checking for a response status of 403, otherwise we just throw the error.

web-api/swagger.json
zacharybrown204: Swagger has been updated. Added the body of the request, the set-cookie header, and removed the security block defining cognito auth (also did this for the other login endpoints)

jimlerza (approved):

web-api/terraform/applyables/blue/blue.tf
```diff
@@ -62,7 +62,9 @@ resource "terraform_data" "locals" {
     EMAIL_SERVED_PETITION_TEMPLATE     = "petition_served_${var.environment}"
     EMAIL_SOURCE                       = "U.S. Tax Court <noreply@${var.dns_domain}>"
     IRS_SUPERUSER_EMAIL                = var.irs_superuser_email
+    IDP_NAME                           = var.idp_name
     LOG_LEVEL                          = "info"
+    MANAGED_LOGIN_DOMAIN               = "https://ef-cms-${var.environment}.auth.us-east-1.amazoncognito.com"
```
jimlerza, line 67: Does this URL format work on prod too?

devinganey613: It will, since the formatting of the domain name right now always includes the environment. We can leave it out for prod though to be consistent if that's better. This is currently how it is set up in cognito.tf.

```
resource "aws_cognito_user_pool_domain" "domain" {
  count                 = var.idp_name != "" ? 1 : 0
  domain                = "ef-cms-${var.environment}"
  user_pool_id          = aws_cognito_user_pool.pool.id
  managed_login_version = 2
}
```

web-api/src/business/useCases/auth/exchangeAuthCodeInteractor.test.ts
```diff
@@ -0,0 +1,93 @@
...
+          };
+        },
+      } as any,
+      { authCode: '1234abcd', code_verifier: '1234' },
+    );
+    await expect(callPromise).rejects.toThrow(
+      new UnauthorizedError('Code exchange failed: Unauthorized'),
+    );
+  });
+});
```
alexdziarn, line 139:
```suggestion
import { exchangeAuthCodeInteractor } from '@web-api/business/useCases/auth/exchangeAuthCodeInteractor';
import { UnauthorizedError } from '@web-api/errors/errors';
import { applicationContext } from '@shared/business/test/createTestApplicationContext';
import { calculateISODate } from '@shared/business/utilities/DateHandler';
import { AxiosError } from 'node_modules/axios/index.cjs';
import { ServerApplicationContext } from '@web-api/applicationContext';

describe('exchangeAuthCodeInteractor', () => {
  it('should successfully exchange auth code', async () => {
    const mockNow = '2026-09-01T00:00:00.000Z';
    const createISODateString = jest.fn().mockReturnValue(mockNow);

    const mockResult = {
      data: {
        expires_in: 200,
        refresh_token: '1234',
        id_token: '5678gefd',
        access_token: '12341234',
      },
    };
    const result = await exchangeAuthCodeInteractor(
      {
        ...applicationContext,
        getHttpClient: () => {
          return {
            CancelToken: {
              source: () => ({
                cancel: () => null,
              }),
            },
            post: () => Promise.resolve(mockResult),
          };
        },
        getUtilities: () => {
          return {
            createISODateString,
            calculateISODate,
          };
        },
      } as ServerApplicationContext,
      { authCode: '1234abcd', code_verifier: '1234' },
    );
    expect(result).toEqual({
      accessToken: '12341234',
      idToken: '5678gefd',
      refreshToken: '1234',
      expiresAt: '2026-09-02T00:00:00.000Z',
    });
  });

  it('should throw an error if auth is unsuccessful', async () => {
    const callPromise = exchangeAuthCodeInteractor(
      {
        ...applicationContext,
        getHttpClient: () => {
          return {
            CancelToken: {
              source: () => ({
                cancel: () => null,
              }),
            },
            post: () => Promise.reject(new Error('Bad Request')),
          };
        },
      } as ServerApplicationContext,
      { authCode: '1234abcd', code_verifier: '1234' },
    );
    await expect(callPromise).rejects.toThrow(new Error('Bad Request'));
  });

  it('should throw an unauthorized error if auth is unsuccessful', async () => {
    const callPromise = exchangeAuthCodeInteractor(
      {
        ...applicationContext,
        getHttpClient: () => {
          return {
            CancelToken: {
              source: () => ({
                cancel: () => null,
              }),
            },
            post: () =>
              Promise.reject({
                name: 'NotAuthorizedException',
                response: { status: 403 },
              } as AxiosError),
          };
        },
      } as ServerApplicationContext,
      { authCode: '1234abcd', code_verifier: '1234' },
    );
    await expect(callPromise).rejects.toThrow(
      new UnauthorizedError('Code exchange failed: Unauthorized'),
    );
  });
});
   post: () => Promise.reject(new Error('Bad Request')),
          };
        },
      } as any,
      { authCode: '1234abcd', code_verifier: '1234' },
    );
    await expect(callPromise).rejects.toThrow(new Error('Bad Request'));
  });

  it('should throw an unauthorized error if auth is unsuccessful', async () => {
    const callPromise = exchangeAuthCodeInteractor(
      {
        getHttpClient: () => {
          return {
            CancelToken: {
              source: () => ({
                cancel: () => null,
              }),
            },
            post: () =>
              Promise.reject({
                name: 'NotAuthorizedException',
                response: { status: 403 },
              } as AxiosError),
          };
        },
      } as any,
      { authCode: '1234abcd', code_verifier: '1234' },
    );
    await expect(callPromise).rejects.toThrow(
      new UnauthorizedError('Code exchange failed: Unauthorized'),
    );
  });
});
```
this sets the type for the parameter instead of using any

web-api/src/business/useCases/auth/exchangeAuthCodeInteractor.ts
```diff
@@ -0,0 +1,51 @@
...
+    });
+
+    return {
+      accessToken: response.data.access_token,
+      idToken: response.data.id_token,
+      refreshToken: response.data.refresh_token,
+      expiresAt,
+    };
+  } catch (err: any) {
+    if (err?.response?.status === 403) {
```
alexdziarn:
```suggestion
  } catch (err) {
    if (axios.isAxiosError(err) && err.response?.status === 403) {
```

web-api/src/gateways/user/createUser.ts
```diff
@@ -68,5 +69,9 @@ export async function createUser(
     Username: email.toLowerCase(),
   };
 
-  await applicationContext.getCognito().adminCreateUser(createUserArgs);
+  const result = await applicationContext
+    .getCognito()
+    .adminCreateUser(createUserArgs);
+
```
alexdziarn, line 75: should we throw an error if result.User is undefined? Does this mean that a user was not created?

alexdziarn (changes requested):

web-api/src/business/useCases/auth/exchangeAuthCodeInteractor.test.ts
devinganey613: We refactored this test a bunch while you were reviewing, but I think I captured the changes you wanted here. Let me know if not.

web-api/src/gateways/user/createUser.ts
devinganey613: I couldn't really find a reason why this function would return success but not include the user data, but I don't think checking hurts. Added a check here.

alexdziarn (approved):

paulo-filho-ctr (approved):

PR #10415, opened 2026-09-03
10387 contact updates to test 1788464440

devinganey613:
#10387

### DIFF
--- a/shared/src/business/utilities/canUserUpdatePetitionerContact.ts
+++ b/shared/src/business/utilities/canUserUpdatePetitionerContact.ts
@@ -1,0 +1,45 @@
+import {
+  isAuthorized,
+  ROLE_PERMISSIONS,
+} from '@shared/authorization/authorizationClientService';
+import { AuthUser } from '@shared/business/entities/authUser/AuthUser';
+import {
+  canAllowDocumentServiceForCase,
+  getPractitionersRepresenting,
+} from '@shared/business/entities/cases/Case';
+import { ROLES } from '@shared/business/entities/EntityConstants';
+
+export const canUserUpdatePetitionerContact = ({
+  petitionerCaseRaw,
+  updatedPetitionerData,
+  user,
+}: {
+  petitionerCaseRaw: RawCase;
+  updatedPetitionerData: any;
+  user: AuthUser;
+}) => {
+  if (!canAllowDocumentServiceForCase(petitionerCaseRaw)) return false;
+
+  let isRepresentingCounsel = false;
+  if (user.role === ROLES.privatePractitioner) {
+    const practitioners = getPractitionersRepresenting(
+      petitionerCaseRaw,
+      updatedPetitionerData?.contactId,
+    );
+
+    isRepresentingCounsel = practitioners?.find(
+      practitioner => practitioner.userId === user.userId,
+    );
+  }
+
+  let isCurrentPetitioner = false;
+  if (user.role === ROLES.petitioner) {
+    isCurrentPetitioner = updatedPetitionerData?.contactId === user.userId;
+  }
+
+  return (
+    isRepresentingCounsel ||
+    isCurrentPetitioner ||
+    isAuthorized(user, ROLE_PERMISSIONS.EDIT_PETITIONER_INFO)
+  );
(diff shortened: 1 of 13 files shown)


web-client/src/presenter/sequences/gotoContactEditSequence.ts
```diff
@@ -13,6 +15,16 @@ export const gotoContactEditSequence =
     clearFormAction,
     clearScreenMetadataAction,
     getCaseAction,
-    setupContactFormAction,
-    setupCurrentPageAction('ContactEdit'),
+    canUserUpdatePetitionerAction,
+    {
+      yes: [setupContactFormAction, setupCurrentPageAction('ContactEdit')],
```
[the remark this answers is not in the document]
zacharybrown204: Tests have been added

web-api/src/business/useCases/user/updatePetitionerInformationInteractor.ts
```diff
@@ -28,35 +19,7 @@ import { getUserById } from '@web-api/persistence/postgres/users/getUserById';
...
-    isCurrentPetitioner = updatedPetitionerData?.contactId === user.userId;
-  }
-
-  return (
-    isRepresentingCounsel ||
-    isCurrentPetitioner ||
-    isAuthorized(user, ROLE_PERMISSIONS.EDIT_PETITIONER_INFO)
-  );
-};
+import { canUserUpdatePetitionerContact } from '@shared/business/utilities/canUserUpdatePetitionerContact';
```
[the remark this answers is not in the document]
zacharybrown204: Bad import removed

shared/src/business/utilities/canUserUpdatePetitionerContact.ts
```diff
@@ -0,0 +1,45 @@
...
+  canAllowDocumentServiceForCase,
+  getPractitionersRepresenting,
+} from '@shared/business/entities/cases/Case';
+import { ROLES } from '@shared/business/entities/EntityConstants';
+
+export const canUserUpdatePetitionerContact = ({
+  petitionerCaseRaw,
+  updatedPetitionerData,
+  user,
+}: {
```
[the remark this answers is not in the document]
zacharybrown204: Tests added

paulo-filho-ctr (approved):

jimlerza (approved):

shared/src/business/utilities/canUserUpdatePetitionerContact.ts
```diff
@@ -0,0 +1,45 @@
...
+  petitionerCaseRaw: RawCase;
+  updatedPetitionerData: any;
```
alexdziarn: we should write out the types for this parameter

web-client/src/presenter/actions/canUserUpdatePetitionerAction.ts
```diff
@@ -0,0 +1,21 @@
+import { canUserUpdatePetitionerContact } from '@shared/business/utilities/canUserUpdatePetitionerContact';
+import { state } from '@web-client/presenter/app.cerebral';
+
+export const canUserUpdatePetitionerAction = ({
+  props,
+  path,
+  get,
+}: ActionProps) => {
+  const user = get(state.user);
+  const { caseDetail, contactId } = props;
```
alexdziarn, line 10: can go with or without it, but I'd like to have some error handling if these were to exist or not

alexdziarn (approved):

shared/src/business/utilities/canUserUpdatePetitionerContact.ts
zacharybrown204: Since we only need the contact id for this I figured it'd be better just to pass in the contact id instead, and let anyone calling this be responsible for grabbing it out of the petitioner objects instead.

web-client/src/presenter/actions/canUserUpdatePetitionerAction.ts
zacharybrown204: Added a quick existence check for those three variables, and if it fails then return the no route.

PR #10418, opened 2026-09-06
10398 Keep the docket record filter out of the pending motions list

### LINKED ISSUE #10398: BUG: Filtering the docket record filters the Pending Motions tab in Tracked Items
**Describe the Bug**
This issue was reported to us by a Court User.  If a case has pending motions, and the docket record is filtered by Orders or exhibits, then when viewing the pending motions tab in Tracked items, no pending motions display.

We need to ensure that if you filter the docket record, it does NOT filter the Pending motions tab.

**Business Impact/Reason for Severity**
low

**In which environment did you see this bug?**
Prod/Test

**Who were you logged in as?**
Court user

**What were you doing when you discovered this bug? (Using the application, demoing, smoke tests, testing other functionality, etc.)**
Using the application

**To Reproduce**
1. Log in as a court user
2. Navigate to a case that has pending motions
3. Click on the Tracked items tab and take note of the pending motions that display 
4. Click on the Docket Record tab
5. Filter the docket record by Orders
6. Next, Click on Tracked items tab, and be sure to view the Pending motions tab
7. Notice that it displays "There are no pending motions"
8. Go back to the docket record and either filter by motions or no filters to show all documents on the dockte record.
9. Next, Click on Tracked items tab, and be sure to view the Pending motions tab
10. Notice that all of the pending motions display again.
 
**Expected Behavior**
If the docket record is filtered by Orders or Exhibits, it should not affect the pending motions tab on the case

**Actual Behavior**
A clear and concise description of what actually happened.
Filtering the Docket record by orders or Exhibits will also filter the Pending motions tab in Tracked items

**Screenshots**

Filter the docket record by Orders or Exhibits:

Image

Then View Pending motions tab and none display, even if there are Pending motions

Image

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Cause of Bug, If Known**

**Process for Logging a Bug:**
* Complete the above information
* Add a severity tag (Critical, High Severity, Medium Severity or Low Severity). See below for priority definition. 

**Severity Definition:**
* Critical Defect
Blocks entire system's or module’s functionality
No workarounds available
Testing cannot proceed further without bug being fixed.

* High-severity Defect
Affects key functionality of an application
There's a workaround, but not obvious or easy
App behaves in a way that is strongly different from the one stated in the requirements

* Medium-severity Defect
A minor function does not behave in a way stated in the requirements.
Workaround is available and easy

* Low-severity Defect
Mostly related to an application’s UI
Doesn't need a workaround, because it doesn't impact functionality

## Definition of Ready for Bugs (Created 2021-10-04)
Definition used: A failure or flaw in the system which produces an incorrect or undesired result that deviates from the expected result or behavior. (Note: Expected results are use cases that have been documented in past user stories as acceptance criteria and test cases, and do not include strange behavior unrelated to use cases.)

The following criteria must be met in order for the development team to begin work on the bug.

The bug must:
 - Be focused on solving a user problem
 - Contain data for all fields in the bug template, so the team can pick it up and begin working immediately

Process: If the unexpected results are new use cases that have been identified, but not yet built, new acceptance criteria and test cases should be captured in a new user story and prioritized by the product owner. 

If the Court is not able to reproduce the bug, add the “Unable to reproduce” tag. This will provide visibility into the type of support that may be needed by the Court. In the event that the Court cannot reproduce the bug, the Court will work with Flexion to communicate what type of troubleshooting help may be needed.

### Product Owner
 - [ ] Bug fix has been validated in the Court's test environment.
 - [ ] Associated test cases defined in TestRail have been updated if necessary.
 - [ ] Successful test run is performed in TestRail.

### UX
 - [ ] Verify that language for docket record for internal users and external users is identical.
 - [ ] All new functionality has been verified to work with keyboard navigation and screen reader software.
 - [ ] UI should be touch optimized and responsive for external users.

### Engineering
 - [ ] Automated test scripts have been written.
 - [ ] Successful test run is performed in TestRail.
 - [ ] New screens have been added to cypress accessibility axe.
 - [ ] Interactors should validate entities before calling persistence methods.
 - [ ] Types have been added to all added and updated functions.
 - [ ] Code refactored for clarity and to remove any known technical debt.
 - [ ] All schema changes are documented in the Data Dictionary (DD), and the Entity Relationship Diagram (ERD) is updated as appropriate.
 - [ ] If there are special deployment instructions, they have been added to the `CHANGES.md` file and the PR description.

arpitjain099:
The `formattedDocketEntries` computed applies `docketRecordFilter` once, at the top, to the entries it then formats. `formattedPendingDocketEntriesOnDocketRecord` is derived from that same filtered set, and `CaseDetailPendingReportList` renders it, so filtering the docket record by Orders or Exhibits empties the pending motions tab under Tracked Items. Filtering by Motions happens to leave it looking correct, which is probably why it went unnoticed for a while.

The per-entry formatting and sort move into a local `formatDocketEntries`, and the pending list is built from the unfiltered entries. `getDocketEntriesByFilter` returns the array it was handed when the filter is All documents or unset, so the identity check means nothing is formatted twice in the common case; the second pass only happens while a filter is actually selected.

Three parameterized cases in `formattedDocketEntries.test.ts` put a pending motion on a case alongside an exhibit, an order and a motion, and assert the pending list still holds exactly that motion under each filter. Against the current code two of the three fail, matching the report:

    ✕ should keep the pending list intact when the docket record is filtered by exhibits
    ✕ should keep the pending list intact when the docket record is filtered by orders
    ✓ should keep the pending list intact when the docket record is filtered by motions

    - Array [ ... ]
    + Array []

Run in a node:24 container, since the repo pins that and my host is on 26. `formattedDocketEntries.test.ts` is 43 passed, and the whole `web-client/src/presenter/computeds` tree is 182 suites and 2069 tests passed. Prettier and eslint are clean on both files through your lint-staged hook.

One unrelated thing I ran into: `.husky/pre-commit` ends with `[ "$HAS_CIRCLECI" -eq 1 ] && circleci config validate`. Without the circleci CLI installed that test is false, so the script's exit status is 1 and the commit is rejected even though gitleaks and lint-staged both passed. I committed past it here. Happy to send a one-line fix separately if that is worth having.

Fixes #10398

### DIFF
--- a/web-client/src/presenter/computeds/formattedDocketEntries.test.ts
+++ b/web-client/src/presenter/computeds/formattedDocketEntries.test.ts
@@ -808,6 +808,50 @@ describe('formattedDocketEntries', () => {
       ]);
     });
 
+    it.each([
+      ['exhibits', DOCKET_RECORD_FILTER_OPTIONS.exhibits],
+      ['orders', DOCKET_RECORD_FILTER_OPTIONS.orders],
+      ['motions', DOCKET_RECORD_FILTER_OPTIONS.motions],
+    ])(
+      'should keep the pending list intact when the docket record is filtered by %s',
+      (_label, docketRecordFilter) => {
+        const pendingMotion = {
+          ...mockDocketEntry,
+          docketEntryId: 'b8dd9fa0-46bd-4a2a-a1d1-7bcbc2c78d0d',
+          documentTitle: 'Motion for Continuance',
+          documentType: 'Motion for Continuance',
+          eventCode: 'M006',
+          index: 4,
+          isFileAttached: true,
+          isOnDocketRecord: true,
+          isStricken: false,
+          pending: true,
+          servedAt: '2020-09-18T17:38:32.418Z',
+          servedParties: [
+            { email: 'petitioner@example.com', name: 'Mona Schultz' },
+          ],
+        };
+        const caseDetail = {
+          ...MOCK_CASE,
+          docketEntries: [...mockDocketEntries, pendingMotion],
+        };
+
+        const result = runCompute(formattedDocketEntries, {
+          state: {
+            ...getBaseState(petitionsClerkUser),
+            caseDetail,
+            sessionMetadata: { docketRecordFilter },
+          },
+        });
+
+        expect(
+          result.formattedPendingDocketEntriesOnDocketRecord.map(
+            entry => entry.eventCode,
+          ),
+        ).toEqual(['M006']);
+      },
+    );
+
     it('should ONLY show exhibit docket entries when "Exhibits" has been selected as the filter', () => {
       const caseDetail = {
         ...MOCK_CASE,
--- a/web-client/src/presenter/computeds/formattedDocketEntries.ts
+++ b/web-client/src/presenter/computeds/formattedDocketEntries.ts
@@ -448,35 +448,36 @@ export const formattedDocketEntries = (
       docketRecordFilter,
     });
 
-  let docketEntriesFormatted = preformattedDocketEntries
-    .map(entry =>
-      getFormattedDocketEntry({
-        applicationContext,
-        docketNumber,
-        entry: { ...entry } as FormattedCaseDetailDocketEntry,
-        get,
-        permissions,
-        rawCase: caseDetail,
-        user,
-        visibilityPolicyDateFormatted,
-      }),
-    )
-    .map(docketEntry => {
-      return {
-        ...docketEntry,
-        isDocumentSelected: documentsSelectedForDownload.some(
-          docEntry => docEntry.docketEntryId === docketEntry.docketEntryId,
-        ),
-        isSelectableForDownload: !!isSelectableForDownload(docketEntry),
-        signatory: '',
-      };
-    });
+  const formatDocketEntries = (entries: RawDocketEntry[]) =>
+    sortDocketEntryTable(
+      entries
+        .map(entry =>
+          getFormattedDocketEntry({
+            applicationContext,
+            docketNumber,
+            entry: { ...entry } as FormattedCaseDetailDocketEntry,
+            get,
+            permissions,
+            rawCase: caseDetail,
+            user,
+            visibilityPolicyDateFormatted,
+          }),
+        )
+        .map(docketEntry => {
+          return {
+            ...docketEntry,
+            isDocumentSelected: documentsSelectedForDownload.some(
+              docEntry => docEntry.docketEntryId === docketEntry.docketEntryId,
+            ),
+            isSelectableForDownload: !!isSelectableForDownload(docketEntry),
+            signatory: '',
+          };
+        }),
+      docketRecordSortField,
+      docketRecordSortOrder,
+    );
 
-  docketEntriesFormatted = sortDocketEntryTable(
-    docketEntriesFormatted,
-    docketRecordSortField,
-    docketRecordSortOrder,
-  );
+  const docketEntriesFormatted = formatDocketEntries(preformattedDocketEntries);
 
   const selectableDocumentsCount = docketEntriesFormatted.filter(entry =>
     isSelectableForDownload(entry),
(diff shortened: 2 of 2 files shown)


web-client/src/presenter/computeds/formattedDocketEntries.test.ts
```diff
@@ -808,6 +808,50 @@ describe('formattedDocketEntries', () => {
...
+        });
+
+        expect(
+          result.formattedPendingDocketEntriesOnDocketRecord.map(
+            entry => entry.eventCode,
+          ),
+        ).toEqual(['M006']);
+      },
+    );
+
```
paulo-filho-ctr: While great for compactness, the `it.each` pattern is not a convention of our codebase. Plus, when a test fails it is far easier to pin point the exact location and the error reason if we have individual test case. I recommend to split each value in inside the `it.each` array into its own `it` case statement.

web-client/src/presenter/computeds/formattedDocketEntries.ts
```diff
@@ -503,8 +504,20 @@ export const formattedDocketEntries = (
       docketEntryId: docEntry.docketEntryId,
     }));
 
-  const formattedPendingDocketEntriesOnDocketRecord =
-    formattedDocketEntriesOnDocketRecord.filter(docketEntry =>
+  // The pending motions tab under Tracked Items is a separate view of the case,
```
paulo-filho-ctr: The convention is to avoid comments to explain code. A good set of unit tests, along with the change description on the PR is sufficient. If it's ok, I recommend removing the comments for a cleaner codebase.

paulo-filho-ctr:
Thank you again for your contribution @arpitjain099, keep them coming! And, also kudos for remembering our contribution process by opening a PR to `test` first so our QA and PO can test it. If you would like, you can also open a PR to `staging` set to `Draft` and put a `Do Not Merge` label on it. That way once this PR is merged we can include the feature branch PR to `staging` in our release package.

web-client/src/presenter/computeds/formattedDocketEntries.ts
```diff
@@ -503,8 +504,20 @@ export const formattedDocketEntries = (
...
+  // so the docket record's filter must not reach it. Everything above is derived
+  // from the filtered entries, which left the tab empty whenever the docket
+  // record was filtered by orders or exhibits (#10398). getDocketEntriesByFilter
+  // returns the array it was given when no filter applies, so the reformat only
+  // happens when a filter is actually set.
```
paulo-filho-ctr, line 512: The convention is to avoid comments to explain code. A good set of unit tests, along with the change description on the PR is sufficient. If it's ok, I recommend removing the comments for a cleaner codebase.

web-client/src/presenter/computeds/formattedDocketEntries.test.ts
```diff
@@ -808,6 +808,50 @@ describe('formattedDocketEntries', () => {
       ]);
     });
 
+    it.each([
```
paulo-filho-ctr, line 811: Thank you for your contribution and improving the codebase making it more robust. While great for compactness, the `it.each` parameterization of test case pattern is not a convention of our codebase. Plus, when a test fails it is far easier to pin point the exact location and the error reason if we have individual test case. I recommend to split each value in inside the `it.each` array into its own `it` case statement.

web-client/src/presenter/computeds/formattedDocketEntries.test.ts
paulo-filho-ctr, line 1: Nice coverage of the bug. Optional polish: in each filter case, also show the docket record is still filtered so we know the pending tab is truly decoupled, not just accidentally correct. Happy with whatever assertion style fits the surrounding tests

paulo-filho-ctr (commented):
Looks great, just a few clean ups.

web-client/src/presenter/computeds/formattedDocketEntries.test.ts
jimlerza: I think I disagree with this, we use `it.each()` in a bunch of places.

paulo-filho-ctr: Oh nevermind then, resolving.

alexdziarn (approved):

PR #10435, opened 2026-09-15
devex: Add checklist for dependency PRs

devinganey613:
Added a new checklist for the dependency rotation pull requests.


paulo-filho-ctr (approved):

ChristieLaferriereTaxCourt (approved):

jimlerza:
:thinking: What if we add a new section to the template?

> ## Dependencies Updated
>
> ### Runtime dependencies
>
> | Package | Version | Purpose | Used in | Possible areas of testing |
> |---|---:|---|---|----|
> | | | | | |
>
> ### Development dependencies
>
> | Package | Version | Purpose | Used in |
> |---|---:|---|---|
> | | | | | |

I would envision this could fit in between the existing **Verification** and **Manual Deployment Steps** sections. We already let folks delete the Manual Deployment Steps section altogether on PRs that don't have any deployment steps, so they could do the same if their PR doesn't contain any dependency updates.

Plus, this way we've actually provided a template for rendering the dependency updates changes in tables.

devinganey613:
> 🤔 What if we add a new section to the template?
> 
> > ## Dependencies Updated
> > ### Runtime dependencies
> > 
> > 
> > 
> > Package
> > Version
> > Purpose
> > Used in
> > Possible areas of testing
> > 
> > 
> > 
> > 
> > 
> > 
> > 
> > 
> > 
> > 
> > 
> > 
> > ### Development dependencies
> > 
> > 
> > 
> > Package
> > Version
> > Purpose
> > Used in
> 
> I would envision this could fit in between the existing **Verification** and **Manual Deployment Steps** sections. We already let folks delete the Manual Deployment Steps section altogether on PRs that don't have any deployment steps, so they could do the same if their PR doesn't contain any dependency updates.
> 
> Plus, this way we've actually provided a template for rendering the dependency updates changes in tables.

I like that idea, it's less disruptive than having a whole second template. Updated this to include the new table and nested the checklist under this section

paulo-filho-ctr (approved):

jimlerza:
> I like that idea, it's less disruptive than having a whole second template. Updated this to include the new table and nested the checklist under this section

I don't think we need the checklist at all. It's not going to force folks who don't do these things to do them. The other checklists have not achieved this either, to be quite honest.

